PR #87, opened 2011-03-21
First pass at a ICCTransform

malcolmhumphreys:
Had sometime on the plane so did a first pass at a ICCTransform, which will be needed for 'iv color profiles/LUTs' on the https://github.com/OpenImageIO/oiio/wiki/GSoC-Idea-Page.

Would like some feedback on where the filepath resolution should be? It's easy for me to add this in the Op but most other things have this in the Transform, is this where this logic should sit?

I have not implemented the setInputMem(const void \* inputPtr, int size) functions yet, but have added these to the transform interface, these would be used when loading icc profile embedded in file headers.


hodoulp:
Now Refer to #541

scoopxyz:
Closing in favor of #541 : Will continue to use as another reference implementation

PR #304, opened 2013-01-17
CSP file read failing fix

andrewbunday:
Edited CSP read function to use pystring to split, strip and format the lines being checked.

* Previously the (line != "CSPLUTV100") would error when reading a CSP.
  Writing was fine.
* I thought it would be better to match the pattern matching in other
  fileFormat source files.

Signed-off-by: Andrew Bunday <andrew.bunday@gmail.com>


malcolmhumphreys:
I had to change the code a bit as the existing unit test didn't work. 0dae004fc006284620424e0db7472b969a6e6f1b

Merged into master

PR #328, opened 2013-09-16
Nuke: OCIODisplay now has an implemented GPU path

slooper:
Modified CMake to only include this in the Linux builds. Contemplating implementing the GPU path on the other plugins, as well.

### DIFF
--- a/src/nuke/CMakeLists.txt
+++ b/src/nuke/CMakeLists.txt
@@ -2,6 +2,10 @@ if(APPLE)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -undefined dynamic_lookup")
 endif()
 
+if(UNIX)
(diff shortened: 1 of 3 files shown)


KevinJW:
I'd like to suggest that the cube size should probably be variable somehow, 32^3 isn't always the right answer. We've seen cases where this introduces more banding than necessary (in ociodisplay for instance), though really the problem relates to the lack of shaper, being able to brute force it by increased sampling is simpler.

Kevin

michdolan:
@slooper any interest in updating this PR for OCIO v2? We are still discussing the best path forward for the Nuke nodes in the repo since Foundry has been maintaining their own version, but it would be cool for those to support the OCIO GPU path. @mjtitchener-fn has that been discussed at all for Nuke? It could also be interesting to adapt this work to the in-development OFX nodes.

michdolan:
I'm going to close this for now. Please reopen if you would like to explore it further.

PR #346, opened 2014-02-17
Cdl file format support

KevinJW:
One of our clients wanted to send us CDL values in a Color Decision List rather than a Color Correction Collection, here is an option to allow parsing the cdl files as well as to select specific cccid values when baking luts.


malcolmhumphreys:
Merged d9f19da1417bb19a99d3bda01db1348742f6842e

I will also look at the "Update test to detect incorrect base usage in ApplyL" which was hiding in that branch as well.

Issue #344, opened 2013-12-16
Config "description:" key for looks


dbr:
Minor thing: would be nice if looks could contain a `description:` key without raising a warning, e.g:

``` yaml
looks:
  - !<Look>
    name: example
    process_space: log
    transform: !<FileTransform> {src: "example.cc"}
    description: |
      Does such and such
```

Currently causes:

```
Unknown key in Look: 'description'. (line 40, column 4)
```

PR #352, opened 2014-04-29
Add description to Look object

rminsk:
Add a description field to look objects.  This addresses issue #344.
On the python binding the description was added as the last field to not break existing positional arguments.


malcolmhumphreys:
Merge e625903d64164356d5a9bc4a971b01be4f93f3b6

PR #354, opened 2014-05-01
Expand context variables in search path before checking if absolute path

rminsk:
Paths on the search path where being check if they were absolute before expanding context variables.  If I had a search_path like "${SHOT_PATH}/color/ocio/luts" OCIO would see this as a relative path since it does not begin with a slash.  This patch expands the context variables before checking if it an absolute path.


malcolmhumphreys:
Merged 8d75efba3719bd266de5aba15c2fd8c56d39a559

PR #357, opened 2014-05-05
Add description to Display/View combination

rminsk:
To be consistent with other objects a description field was added to a display/view combination.  The description field could be used in an UI for pop up help.

It follows in the style of the current API with a new method getDisplayDescription(const char *display, const char *view) method added to the Config object.  A second overloaded addDisplay method was added with a description parameter instead of making a description parameter with a default value.  This should ensure binary compatibility with the current library.


malcolmhumphreys:
Do we really need another Config::addDisplay I think just break the ABI and add the description argument to the existing one.

doug-walker:
Closing this as it is no longer relevant after the merge of PR #1038, which added the same functionality.

PR #359, opened 2014-05-30
cmake support for OSX universal binaries

rminsk:
Added support to specify multiple architectures with CMAKE_OSX_ARCHITECTURES. Needed to pass architecture flags to ExternalProject_add.

Also implemented USE_EXTERNAL_LCMS which was in the main cmake file but not used in the ociobakelut cmake file.


scoopxyz:
Closing due to age, and not supporting PowerPC MacOS. Apple devs please tell me otherwise.

PR #361, opened 2014-06-03
Get all displays from the config file

rminsk:
getNumDisplays and getDisplay filter the display list by the active display list. A method was needed to be able to get all the displays from the config. This also corrects the problem of reading a config file and writing it back out with all the displays intacted. To ensure backward compatibility the original methods were not changed.


doug-walker:
Closing this since it is no longer relevant after the merge of PR #1038, which added the same functionality.

PR #363, opened 2014-06-12
ClampTransform

rminsk:
Adding a ClampTransform to be able to limit the range of a channel.  The inverse of a ClampTransform is a noop.  If min>max then the channel is pass through.  The ClampTransform will be used clamp the values before apply the power operator in a CDLTransform.

### DIFF
--- a/docs/userguide/config_syntax.rst
+++ b/docs/userguide/config_syntax.rst
@@ -844,6 +844,19 @@ Keys:
 * ``direction``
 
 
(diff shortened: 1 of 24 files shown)


hodoulp:
@rminsk  In the context of the OCIO v2 project, I need to implement a range op which is what I understand from your PR. Could I use your work as a starting point?

hodoulp:
Refer Range Op in 556

PR #366, opened 2014-07-24
Add ExpressionTransform op

Shootfast:
I've created an ExpressionTransform op that will allow users to describe transformations in terms of a mathematical expression string. I was unable to find a library that both supported expression parsing and shader generation, so I have written and included my own.

I have tested the expression transformation with various different equations, and it has been used in production for a few months:

Here is an example of an Alexa Log C transform:

```
  - !<ColorSpace>
    name: Alexa
    family: log
    bitdepth: 32f
    description: |
      Alexa Log C
    allocation: uniform
    allocationvars: [-0.125, 1.125]
    to_reference: !<ExpressionTransform> {rgb: "v > 0.1496582 ? (pow(10.0, (v - 0.385537) / 0.2471896) - 0.052272) / 5.555556 : (v -0.092809) / 5.367655 "}
```


mfe:
This is awesome !
What the status of this PR ? 

I've just tested it through the python binding, it works nicely :)

However, I have some troubles with the nuke nodes. Chan / pixel values does not look like they are written or read in the right order.
See [this screenshot](https://www.dropbox.com/s/v215p2qxzdukxq0/ocio_expr.jpg). 
For this test, I'm using an identity expr : `{rgb: "v"}`.
Other transform types work well.

(Built on a Centos 5.4 Linux using Nuke8v7 API with an old gcc 4.1.2, which is - I know -  not ideal for optim ;p)

andrewhwood:
adding on to the pile to try to keep this PR alive!  Would be handy especially if r, g, and b can have separate expressions for reversible, but not 1-d functions (like a conversion to L*A*B and back again)

hodoulp:
Very interesting work.

I would like to read that third-party library documentation to 1) double the licensing scheme and 2) have an overview of the capabilities and limitations. Where I can find the project web site or project repo?

michdolan:
@Shootfast have you explored this op further? Any interest in continuing with it? Or do you think the OpenColorMath work would replace it?

Shootfast:
This Op as it stands, no. There are some kinks in the expression parser here that were never fixed, and the LLVM API has shifted dramatically since this was written.

I am hoping the functionality can be duplicated in OpenColorMath when I get the chance to develop it futher.

PR #377, opened 2014-09-12
Fix define check for GCC symbol visibility

pinotree:
Define the ``__attribute__`` symbol visibility (for GCC) by checking for GCC itself, instead of different OSes (which might also have different compilers).

Clang should define ``__GNUC__`` as well for GCC compatibility, so this should cause no regressions (while fixing the symbols visibility on any OS which uses GCC).


malcolmhumphreys:
Merged de9491e67825687918a35d49a72daca698f112a7

PR #378, opened 2014-09-12
Fix getcwd to not rely on MAXPATHLEN

pinotree:
Dynamically allocate the buffer for getcwd, instead of assuming it is at most some arbitrarly-fixed size. Also, optionally use PATH_MAX (which is optional in POSIX) as starting size for the buffer, instead of MAXPATHLEN (which is not even POSIX).


malcolmhumphreys:
Merged 21e55af1f0845541aa7103b0a2abee5005476f5a

PR #380, opened 2014-09-21
external include dirs should always be last

devernay:
- compiles even if an older version of OCIO is installed


malcolmhumphreys:
Merged 5fde5614fd33249d325d3d766adfbfc9830e3fd5

PR #389, opened 2015-01-30
Write full precision of float to config file

KevinJW:
As this depends on the patch to the yaml code, consider this proof of concept, rather than fully mergable (although it should work).

This avoids issues when (re)writing config files and drifting floating point numbers due to the required precision to exactly store floating point numbers.

For details see http://www2.open-std.org/JTC1/SC22/WG21/docs/papers/2005/n1822.pdf

I have reported the yaml-cpp bug upstream (https://code.google.com/p/yaml-cpp/issues/detail?id=272)

### DIFF
--- a/ext/yaml-cpp-0.5.1.patch
+++ b/ext/yaml-cpp-0.5.1.patch
@@ -10,3 +10,24 @@ diff -Naur yaml-cpp-0.5.1.orig/CMakeLists.txt yaml-cpp-0.5.1/CMakeLists.txt
  	#
  	add_custom_target(debuggable $(MAKE) clean
  		COMMAND ${CMAKE_COMMAND} -DCMAKE_BUILD_TYPE=Debug ${CMAKE_SOURCE_DIR}
+
+--- src/emitterstate.cpp-orig	2015-01-30 13:53:26.468688052 +0000
++++ src/emitterstate.cpp	2015-01-30 13:53:33.507624388 +0000
+@@ -267,7 +267,7 @@
+ 
+     bool EmitterState::SetFloatPrecision(int value, FMT_SCOPE scope)
+     {
+-        if(value < 0 || value > std::numeric_limits<float>::digits10)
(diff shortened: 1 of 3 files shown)


KevinJW:
looks like upstream moved to github.... and the issue might be lost but there is a pull request targeting a similar fix

https://github.com/jbeder/yaml-cpp/pull/470

Kevin

hodoulp:
@KevinJW Please double-check with master because changes were recently done where some Ops now hold double values (instead of float values) i.e. #640 for Matrix, etc.

doug-walker:
Kevin, no need to recheck this quite yet.  As Patrick wrote, the underlying MatrixOpData is now able to hold doubles and we read double precision from CLF files (and have a unit test for that).  HOWEVER, the YAML load/save is still float based.  So it's very likely that the issue you wrote this PR to address is still present.  Please stay tuned and we'll get back to you on this.

hodoulp:
Even if the MatrixOp values moved to double precision, the YAML read/write was not updated.
A pull request will soon pop to complete the MatrixTransform transition to doubles.

hodoulp:
The pull request #724 fixes the Matrix precision.

KevinJW:
Note that the real bug in the yaml library limits the correct output of float or double - this aspect relates to #690 if we bundle a broken library we'll always be limited to its output capabilities. yes switching to doubles 'fixes' the float precision issues, but from my reading of the code doubles will still break.

For reference the fix to the yaml-cpp https://github.com/jbeder/yaml-cpp/pull/649 is from December 2018, so either we bump the bundled version and deal with any fall out, and we document that older versions will have a slight chance of this occurring.

Kevin

hodoulp:
Agree. As you mention, it should be part of the `ext` library refactoring needed for the ASWF transition (i.e. update all 'core' third-party libraries to their latest version).

michdolan:
OCIO v2 is now using the newest yaml-cpp release by default. Does that solve the issue this PR was intending to fix @KevinJW ?

hodoulp:
@KevinJW What is the status of the pull request?

PR #390, opened 2015-02-11
Adding an interface for defining custom transform without changes to the core

lukasstockner:
This patch allows developers to use self-defined transforms inside their OpenColorIO pipeline without maintaining a modified version of the library by adding the CustomTransform, a class that can be inherited to implement arbitrary transforms and Ops.

### DIFF
--- a/export/OpenColorIO/OpenColorTransforms.h
+++ b/export/OpenColorIO/OpenColorTransforms.h
@@ -274,6 +274,38 @@ OCIO_NAMESPACE_ENTER
     
     
     //!rst:: //////////////////////////////////////////////////////////////////
+
+    class Op;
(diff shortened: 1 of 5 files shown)


Shootfast:
Hi,

Just chiming in to say this hasn't gone unnoticed. I'm on the fence about this particular feature. 
On the one hand, I know how handy it would be if you are just using OCIO as an engine (without a config). But ideally we are creating a common pipeline for all applications to share, and encouraging developers to make their own transformations outside of that pipeline seems in violation of that goal. 
In addition to that, we are also investigating a more formalised plugin arcitecture which would make this feature superfluous. So on those grounds I would be opposed to this pull request.

I spoke to a few other OCIO devs who seemed to be in agreement with me, but does anyone else have an opinion on the matter?

Cheers,
Mark

PR #395, opened 2015-06-13
Fix build when OCIO_BUILD_DOCS=ON and OCIO_BUILD_PYGLUE=OFF

sambler:
Add a conditional to define which libs get added to the DEPENDS list for the doc target.


malcolmhumphreys:
Merged 8eedc2bc96002ce0784327263ce39b8e78e7eb77

PR #400, opened 2015-11-11
Changes explanation for 'reference' role

mplec:
Per discussion in https://groups.google.com/d/topic/ocio-users/V3oZmjkcfTU/discussion
the existing explanation is not accurate, this role seems to most commonly be used to
get the reference space against which others are defined.

Also makes the other items in the list consistent in initial capitalization and ending punctuation.


malcolmhumphreys:
Merged 4cdb0a73b4bf025fd35b2f7ff3e3961e06d40645

PR #404, opened 2016-04-13
Make it possible to build OpenColorIO for Windows on Linux

boudewijnrempt:
With this patch, it's possible to use the mingw cross-compile toolchain to build opencolorio on Linux for Windows: the #include <Windows.h> becomes #include <windows.h>, all the places where if (WIN32) was used but if (MSVC) was meant are fixed. And on Windows itself, use myptch.exe as a patch command instead of the blocked patch.exe

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -310,7 +310,7 @@ endif()
 messageonce("SSE Optimizations: ${OCIO_USE_SSE}")
 
 if(OCIO_USE_SSE)
-    if(WIN32)
+    if(MSVC)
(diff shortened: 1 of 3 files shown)


hodoulp:
@boudewijnrempt  With all the changes in master (starting with #520 and later), it should work.

hodoulp:
@boudewijnrempt The pull request is now outdated, and some recent changes should have fixed the issue. Would you like to double-check ?
Without any answer, I will close it.

boudewijnrempt:
Oh, gosh, that's so old... Just close it, because it's no longer relevant. These days I don't cross-compile anything anymore.

PR #405, opened 2016-04-15
AE plugin update

fnordware:
Update AE plug-in build files for the latest OCIO, including new YAML. Also fixed one bug and made a few small tweaks.


fnordware:
Let's do this as part of #476 instead.

PR #411, opened 2016-09-16
Fix py3 compatibility for building docs

sambler:
replace xrange with range and file with open
this change works with py2.7 and 3.4, 3.5


malcolmhumphreys:
Merged 61bcbeaa167423bfd409cbd8193287280f627b49

PR #413, opened 2016-09-22
Added configurable delimiters support to Config object

malcolmhumphreys:
Added configurable delimiters support which is a modified version of #381

@mike1158, @shootfast, @mplec and @lgritz can you take a look also
- What do you think colorspace name should be returned from this string "somefile-Gamma22-ACEScg.exr”? Currently we return the longest right most colorspace name e.g. Gamma22 not ACEScg. This is what the old code looked like it did but maybe now with delimiters it should just be the first delimited colorspace found (ACEScg in this example).
- I have set the default delimiters to be [-, _, /, \, .] what are your thoughts?
- Note. To get the old behaviour you just need to add ‘’ to the delimiters list.

-- CHANGES
- added getNumDelimiters()
- added getDelimiterByIndex()
- added addDelimiters()
- added clearDelimiters()
- added containsToken()
- Set default delimiters  [-, _, /, \, .]
- Updated parseColorSpaceFromString to use containsToken()
- Added unit tests for containsToken() and parseColorSpaceFromString()
- Updated pyglue and jinglue bindings
- Added missing EnvironmentMode.java file

### DIFF
--- a/export/OpenColorIO/OpenColorIO.h
+++ b/export/OpenColorIO/OpenColorIO.h
@@ -341,6 +341,26 @@ OCIO_NAMESPACE_ENTER
         //!cpp:function::
         void setStrictParsingEnabled(bool enabled);
         
+        //!cpp:function:: Returns the number of string delimiters that are
+        // defined
+        int getNumDelimiters() const;
+        //!cpp:function:: Returns the string delimiter at index
+        // .. note::
+        //    This will return null if the specified index is not found.
+        const char * getDelimiterByIndex(int index) const;
+        //!cpp:function:: Add a string delimiter to the config
+        void addDelimiters(const char * str);
+        //!cpp:function:: Remove all the defined delimiters
+        void clearDelimiters();
+        
+        //!cpp:function:: Returns true if str contains the token which is
+        // surrounded by one of the defined delimiters. This function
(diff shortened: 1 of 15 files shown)


scoopxyz:
Makes sense to me. There probably aren't too many cases where that was an
issue, but it's definitely better to be open to different file naming
conventions.

On Thu, Sep 22, 2016 at 1:44 AM, Malcolm Humphreys <notifications@github.com

michdolan:
Echoing my comments from #381 :

> Curious if we feel like the new OCIO v2 file rules section solves this adequately. It's a different solution, but since it supports wildcards and regex it is more flexible. @hodoulp @doug-walker do you think it would be useful to support a special ${name} variable or similar in file rules regex strings to imply the placement of a color space name, rather than needing to have a pattern that accounts for every color space naming convention?

doug-walker:
Michael, the FileRules does already allow a special rule that applies the parseColorSpaceFromString type matching.  An optional parameter could be added to that rule that would specify the delimiters to use.  I think it would be cleaner and clearer to put it there rather than add a separate "delimiters" attribute in the config.  That would also be a lot easier to implement than allowing a ${name} parameter in the other rule types.

PR #415, opened 2016-10-01
Update Travis-CI build file

dbr:
The old config was set up before Travis-CI properly supported C++ projects

New one also tests against clang and GCC on both OS X and Linux

To get working requires signing in to the "imageworks" TravisCI account (it authenticates with Github) and enabling the OpenColorIO project


scoopxyz:
I would like to pull this in, any objections?

Issue #399, opened 2015-10-14
FileTransform not clamping low values when applying inverse of 1D LUT


hpd:
In the case where values below the lower bound of the output range of a 1D LUT are run through the inverse of that LUT, the FileTransform does not behave as expected. Specifically, the FileTransform does not clamp values below the lower bound of the LUT output before applying the inverse LUT transform.

An example .spi1d 1D LUT is as follows
Version 1
From -0.1 2.0
Length 2
Components 1
{
         0.0
         1.0
}

A very simple Nuke script demonstrating the problem is copied below. The Red channel only covers the 0.0 to 1.0 range and inverts as expected. The green channel goes beyond the 0.0 to 1.0 range and produces large negative values for values below 0.0. Values above 1.0 are clamped properly before the inverse LUT is applied.

set cut_paste_input [stack 0]
version 9.0 v7
Constant {
 inputs 0
 channels rgb
 format "512 512 0 0 512 512 1 square_512"
 name Constant1
 selected true
 xpos -155
 ypos -111
}
Expression {
 expr0 x/(width-1)
 expr1 y/(width-1)*3-1.5
 name Expression3
 selected true
 xpos -155
 ypos -11
}
OCIOFileTransform {
 file ocio_lut_issue_spi1d
 direction inverse
 working_space linear
 name OCIOFileTransform2
 selected true
 xpos -155
 ypos 48
}

KevinJW:
Relates to issues #391 #326 and possibly #289 #196

KevinJW:
This may explain the problems when using ACEScc transforms in image filtering see https://www.researchgate.net/publication/305278559_A_retrospective_on_the_adoption_of_the_ACES_technology_at_Framestore

sobotka:
Bump on this now that OCIO is back to life.

malcolmhumphreys:
Any chance someone can do a change and a pull request for this? Should be
straightforward to add a clamp.

Longer term we could add a BOUNDS option to support both BOUNDS_EXTRAPOLATE
and BOUNDS_CLAMP.

scoopxyz:
Not the most elegant solution, but I have submitted a hotfix #418 for this particular issue, longer term the BOUNDS parameters suggested by @malcolmhumphreys seems the best solution

PR #418, opened 2016-11-16
hotfix Lut1DOp negative value clamp

scoopxyz:
address issue #399

### DIFF
--- a/src/core/Lut1DOp.cpp
+++ b/src/core/Lut1DOp.cpp
@@ -404,7 +404,7 @@ OCIO_NAMESPACE_ENTER
                 delta = (v - *lowbound) / (*highbound - *lowbound);
             }
             
-            return ((float)(lowbound - start) + delta) * invMaxIndex;
+            return std::max(((float)(lowbound - start) + delta) * invMaxIndex, 0.0f);
         }
         
         void Lut1D_LinearInverse(float* rgbaBuffer, long numPixels, const Lut1D & lut)


KevinJW:
If I am reading the patch right ... I'm not sure it should clamp at zero, but at the LUT minimum - depending on the direction this would be the first value in the LUT  (simple_lut[indexLow] ?) or the lower bound in the range (*start ?), - on the assumption of monotonicity.

As a note at one point I did try just adding extra entries to the LUTs at the bottom end and clamping them to a minimum on he assumption that extrapolation of flat values would also effectively clamp - but it didn't.

Kevin

scoopxyz:
This implementation provides the above behavior via it's encapsulating function "Lut1D_LinearInverse" which stores the values via a _**mx+b**_ format, where the clamped function returns 0.0 as _**x**_ when the value becomes negative. Since _b == lut.from_min_ it will return the expected _from_min_ value. This has been the behavior in the tests I have run.

I've removed the initial clamp from the forward transform as I don't believe that will result in desired behavior.

scoopxyz:
Would we be able to get some extra eyes on this @malcolmhumphreys @Shootfast @KevinJW as this is directly impacting imagery in current OCIO ACES implementations?

If my solution is too simplistic I'd be happy to work on a more comprehensive solution in due time.

scoopxyz:
Conversation with @Dithermaster reveals that "Jeremy said it was a bug - the intent was for clamping (not extrapolation)"

After a successful Travis build I intend to merge the above.

Dithermaster:
Digging up an old email, Jeremy had this to say: "I consider the correct behavior to be clamping values outside the defined range" and "I submitted an issue report on it when discovered earlier this year: https://github.com/imageworks/OpenColorIO/issues/326"

Dithermaster:
This fix looks like it would fix the errant extrapolation on the bottom end. I agree that the `lut.from_min` is handled in the caller, not this function.

PR #444, opened 2017-04-18
Fix black pixels returned by Lut1D_Linear in certain cases

Nazg-Gul:
It was possible to have linear lookup returning black pixel for values
which are far beyond lookup table range. This was caused by precision
issues with floating point. Basically, if index is really high, then
it'll make delta really high and for such high values delta will be
equal to (1.0f - delta) because of the way how machine math works.
This will then produce 0 as the output value.

This patch performs early output check to avoid any possible floating
point precision issues in such cases.

### DIFF
--- a/src/core/Lut1DOp.cpp
+++ b/src/core/Lut1DOp.cpp
@@ -296,6 +296,11 @@ OCIO_NAMESPACE_ENTER
         {
             int indexLow = clamp(std::floor(index), 0.0f, maxIndex);
             int indexHigh = clamp(std::ceil(index), 0.0f, maxIndex);
+            if (indexLow == indexHigh) {
+                // Happens when index outside of the LUT range is requested. We do early
+                // output here to avoid possible precision issues during interpolation.
+                return simple_lut[indexLow];
+            }
             float delta = index - (float)indexLow;
             return (1.0f-delta) * simple_lut[indexLow] + delta * simple_lut[indexHigh];
         }


Dithermaster:
That's the same math that was causing instability in the result values for large excursions past the bounds. I had suggested that Kevin's numeric stability fix (which was a gist, not a MR) should also be applied to other interpolation that likely have the same instability, in particular, *Lut3D_Linear* and *Lut3D_Tetrahedral*.

Here's his gist: https://github.com/imageworks/OpenColorIO/issues/420

Can we merge these efforts and apply them in all of the necessary places?

Nazg-Gul:
Dithermaster, ah, didn't know there was an existing patch for that. There is actually a PR #441. Good thing about that patch is that it includes a unit test.

I was thinking that 3D LUT interpolation also needs a fix, but so far i did not run into the case where it fails, so can't really verify any changes there. Maybe with unit tests it'll be easier to test.

I'm also all up for merging forces on getting this issues solved. How do we move forward from this point?

Dithermaster:
Someone posted your PR on the OCIO Slack and we had some discussion there. I don't know the next steps; I think Sean is able to take PRs so hopefully he can take #441 soon after it's been reviewed (I suggested that the 3D look be adjusted too since it uses similar math).

scoopxyz:
Thanks for the PR @Nazg-Gul I am going to work on reviewing this and #441 this week and testing them on my end

Dithermaster:
Now that https://github.com/imageworks/OpenColorIO/pull/441 has been merged, this change is not needed.

PR #450, opened 2017-06-22
update README

scoopxyz:


dbr:
You'll need to update docs/index.rst too as it directly include the file `README`

PR #462, opened 2017-08-11
Update to TinyXML2 & avoid string->c-string->string conversions on error.

marsupial:


KevinJW:
This looks like it should be split into separate parts, separating out the TinyXML changes, CMake and exception bits. on the XML side should cross check the impact on #389.

Kevin

KevinJW:
On the exception specification that is potentially a bigger can of worms (or might turn out to be a search and replace, but that would depend on which versions of C++ the library needs to support.

Kevin

marsupial:
The TinyXML update is dependent on the additional Exception constructor parts.
I could add a separate PR for just that, but GitHub doesn't manage dependencies like that very well.

michdolan:
Thanks for the contribution @marsupial . Recent OCIO development has replaced TinyXML with Expat in preparation for adding CLF support, so I'm closing this PR.

PR #468, opened 2017-10-10
Fix the documentation build under Windows & nmake (e.g. html, tex & pdf)

hodoulp:
This branch is fixing the doc generation on the windows platform but using nmake. 
Unfortunately using a Visual Studio project adds a new set of issues.

In order to have the html, tex & pdf documentation one should install extra tools such as [MiKTex](https://miktex.org/) and [GnuWin32](http://gnuwin32.sourceforge.net/) tools.


lgritz:
LGTM.

PR #469, opened 2017-10-11
Add the pdf doc generation on linux platform only

hodoulp:
The idea is to have at least one platform generating all the potential documentation outputs to ensure that the CMake 'mechanism' is working. A side effect is also to document (through the travis file) how to produce the pdf documentation.


lgritz:
LGTM.

PR #470, opened 2017-10-12
Fix Windows warnings

hodoulp:
The goal is to fix most of the warnings in order to have a build break to any new warnings. Most of the warnings were on Windows around some code security concerns.

### DIFF
--- a/src/core/Platform.cpp
+++ b/src/core/Platform.cpp
@@ -50,12 +50,14 @@ OCIO_NAMESPACE_ENTER
             const errno_t err = ::_dupenv_s(&value, &len, name);
             if(err!=0 || len==0 || !value || !*value)
             {
+                if(value) free(value);
                 return 0x0;
             }
             else
             {
                 str.resize(len+1);
-                ::strncpy_s(&str[0], len, value, len);
+                ::snprintf(&str[0], len, "%s", value);
+                if(value) free(value);
             }
 
             return const_cast<char*>(str.c_str());
@@ -98,6 +100,16 @@ OIIO_ADD_TEST(Platform, getenv)
         char* env = OCIO::Platform::getenv("NotExistingEnvVariable");
         env = OCIO::Platform::getenv("PATH");
         OIIO_CHECK_ASSERT(env && *env);
+    }
+}
+
+OIIO_ADD_TEST(Platform, putenv)
+{
+    {
+        ::putenv("MY_DUMMY_ENV=SomeValue");
+        const char* env = OCIO::Platform::getenv("MY_DUMMY_ENV");
+        OIIO_CHECK_ASSERT(env && *env);
+        OIIO_CHECK_ASSERT(0==strcmp("SomeValue", env));
     }
 }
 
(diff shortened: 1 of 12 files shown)


src/core/Platform.cpp
```diff
@@ -0,0 +1,116 @@
...
+                return 0x0;
+            }
+            else
+            {
+                str.resize(len+1);
+                ::snprintf(&str[0], len, "%s", value);
+                if(value) free(value);
+            }
+
+            return const_cast<char*>(str.c_str());
```
lgritz: The use of this static is not thread safe, it will be a problem if there's any chance that Platform::getenv() is called simultaneously by two threads.

I'm a little confused about why this function is needed. I just use getenv even on Windows, for OIIO/OSL. Am I doing something dangerous?

This function should contain a comment with some discussion about exactly why _dupenv_s should be preferred over getenv (it wasn't immediately apparent to me even looking at the MSDN pages for both functions).

hodoulp:
I was not aware of the gitHub diff tool limitation. Next time, I will be careful. 
Concerning pystring.cpp, here are the warnings I wanted to fix:
> pystring.cpp
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(458): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(468): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(483): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(500): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(516): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(532): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(548): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(565): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)
> D:\OpenSource\ocio_rw_1\src\core\pystring\pystring.cpp(608): warning C4800: 'int': forcing value to bool 'true' or 'false' (performance warning)

hodoulp:
@lgritz  The ::getenv method is using an internal static buffer so the proposed method was not adding extra limitations. But with the _dupenv_s usage on Windows platform, that internal buffer could be removed. Unfortuantely, it could not be done for *nix platforms.

scoopxyz:
I'm not familiar with the needs of the Platform specific memory fix, though I am familiar with the errors they raised, they look fine by my eyes. @lgritz how strongly do you feel about the formatting PR?

lgritz:
I feel strongly about segregating formatting changes to their own PRs... in the future. There's no reason to screw up an existing PR.

lgritz:
I'm fine with this as it stands. LGTM. I will merge.

PR #471, opened 2017-10-13
Fix the CSP file format issue

hodoulp:
The branch is fixing the CSP format issue where the preLut was wrongly computed. A commented unit test was already highlighting that problem. Using the baker, the right computation for the preLUT is to define a lut to convert the colors from 'input' to 'shaper' color spaces. 

A major refactoring by commit 45caa7716c5203fc11c14f0b60599fca04141b66 introduced the problem and the unit test was disabled by commit 1cfa00324e8e435d27d771357e6bb19ff2a3f7b4

The metadata bug is from 0dae004fc006284620424e0db7472b969a6e6f1b

Refer to [CineSpace File Format](http://mjbcolor.com/cineSpaceManual.pdf)

### DIFF
--- a/src/core/FileFormatCSP.cpp
+++ b/src/core/FileFormatCSP.cpp
@@ -425,7 +425,7 @@ OCIO_NAMESPACE_ENTER
                        !istream)
                 {
                     nextline (istream, line);
-                    if (startswithU(line, "END METADATA"))
+                    if (!startswithU(line, "END METADATA"))
                         metadata += line + "\n";
                 }
                 lineUpdateNeeded = true;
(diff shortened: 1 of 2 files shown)


src/core/FileFormatCSP.cpp
```diff
@@ -686,18 +686,20 @@ OCIO_NAMESPACE_ENTER
                 GenerateIdentityLut1D(&shaperOutData[0], shaperSize, 3);
                 GenerateIdentityLut1D(&shaperInData[0], shaperSize, 3);
                 
-                ConstProcessorRcPtr shaperToInput = config->getProcessor(baker.getShaperSpace(), baker.getInputSpace());
-                if(shaperToInput->hasChannelCrosstalk())
+                ConstProcessorRcPtr inputToShader
+                    = config->getProcessor(baker.getInputSpace(), baker.getShaperSpace());
```
hodoulp: Here is the fix around the preLut computation bug.

src/core/FileFormatCSP.cpp
```diff
@@ -425,7 +425,7 @@ OCIO_NAMESPACE_ENTER
                        !istream)
                 {
                     nextline (istream, line);
-                    if (startswithU(line, "END METADATA"))
+                    if (!startswithU(line, "END METADATA"))
```
hodoulp: Here is the metadata fix.

hodoulp:
Closed the pull request without merging it in master because I have questionning about the shaper to input computation which could have been done to have a 0.0f to 1.0f uniform range to the next Op (1D or 3D lut). Unfortunately, no comments explain the reason of the inverse computation...  Have to do my home work to have the best fix.

PR #475, opened 2017-10-17
Fixed the crash with missing role's color space

hodoulp:
Fixed the crash when the color space associated to a role is not defined:
1. Fixed the problem by checking if the color space exists before trying to serialize anything.
2. Added a unit test to validate that it does not crash and trigger an OCIO exception.


lgritz:
LGTM

PR #476, opened 2017-10-27
After Effects update, Photoshop plug-in

fnordware:
Includes some updates to the AE plug-in, as well as a new Photoshop plug-in.


scoopxyz:
I doubt there are other resident AE or Photoshop experts around, but if so please take a look. I'll look to accepting this within the week since they are relatively self contained.

PR #477, opened 2017-11-06
Color space transform now accepts env. variable

hodoulp:
The Color space transform is now accepting env. variable for the src and dst color space names.

1. Accept env. variables for the color space names
2. Add unit tests
3. Fix a typo and improve a unit test macro


scoopxyz:
Well, I attempted to resolve the merge conflict but that apparently broke the build. It looks like you had tried that before as well?

hodoulp:
Refer to pull request #479 to have the right/final code change.

PR #478, opened 2017-11-08
Enable all apps on windows

hodoulp:
The goal is to have all apps working on the Windows platform. Obviously, the OIIO library (and all its third-party libraries) compilation must have been done prior to OCIO.
In order to demonstrate the success, I've added the family classification in the color space menu of the ociodisplay apps.

### DIFF
--- a/docs/configurations/index.rst
+++ b/docs/configurations/index.rst
@@ -19,10 +19,6 @@ Public Configs
     spi_vfx
     spi_anim
     nuke_default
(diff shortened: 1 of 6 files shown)


scoopxyz:
What was the reasoning behind removing the other supported ACES configs in the documentation?

hodoulp:
The corresponding aces_xxx.rst files are missing so it was generating some errors.

PR #480, opened 2017-11-13
Fix AllocationTransform to use a proper enum.

### LINKED ISSUE #473: Typo in Python AllocationTransform Example
Line 20 of https://github.com/imageworks/OpenColorIO/blob/master/docs/developers/bindings/PythonTransforms.rst
 
The constant ALLOCATION isn't present, as defined by the constants.

Should probably be a viable constant?

       # Allocation
        self.assertEqual(OCIO.Constants.ALLOCATION_UNKNOWN, "unknown")
        self.assertEqual(OCIO.Constants.ALLOCATION_UNIFORM, "uniform")
        self.assertEqual(OCIO.Constants.ALLOCATION_LG2, "lg2")

sobotka:
Fixes #473

### DIFF
--- a/docs/developers/bindings/PythonTransforms.rst
+++ b/docs/developers/bindings/PythonTransforms.rst
@@ -17,7 +17,7 @@ AllocationTransform
     
     import PyOpenColorIO as OCIO
     transform = OCIO.AllocationTransform()
-    transform.setAllocation(OCIO.Constants.ALLOCATION)
+    transform.setAllocation(OCIO.Constants.ALLOCATION_LG2)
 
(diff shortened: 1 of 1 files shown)


hodoulp:
It could better to add a reference (if it's possible with the rst format) to the Allocation enum instead of hardcoding the enum values.

dbr:
It is possible to link to a reference (used all over the docs).. although on first glance it seems like the constants aren't listed in the Python section of the docs?

scoopxyz:
Seeing as how this is just example code, I dont think it needs to be exhaustive at this point. Documentation generation needs some love as it is definitely not exhaustive. Moving @dbr comment to a new issue #493

PR #482, opened 2017-11-25
Fix MacOS >= 10.12 Deprecation

scoopxyz:
MacOS >= 10.12 (Sierra) break OCIO builds via deprecated mutex functions. For example:
`error: 'OSSpinLock' is deprecated: first deprecated in macOS 10.12 - Use os_unfair_lock() from <os/lock.h> instead` 
and 
`error: 'OSSpinLockLock' is deprecated: first deprecated in macOS 10.12 - Use os_unfair_lock_lock() from <os/lock.h> instead`

I've swapped out the required functions, added the MacOS version logic, and added the build criteria to TravisCI.

@hodoulp and @lgritz could you take a look? Don't trust myself with mutex-y things just yet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,24 @@
 # travis-ci.org build file
 # https://docs.travis-ci.com/user/languages/cpp
 
-language: cpp
-
-compiler:
-  - clang
-  - gcc
-
-os:
-  - linux
-  - osx
+matrix:
+  include:
+    - os: linux
+      language: cpp
+      compiler: clang
+    - os: linux
+      language: cpp
+      compiler: gcc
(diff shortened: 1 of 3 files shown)


lgritz:
Is OCIO using C++11 minimum yet? If so, is there a good reason why we shouldn't just use std::mutex on all platforms?

If not, I think it should be.

hodoulp:
**SpinLock** is never used so you can remove the class.

scoopxyz:
Updated, worth keeping the additional MacOS CI build since it was a known failure before?

hodoulp:
Yes, it is. Anything solidifying the build is welcome...

Issue #486, opened 2017-11-28
ociodisplay namespace error


scoopxyz:
Line [133](https://github.com/imageworks/OpenColorIO/blob/master/src/apps/ociodisplay/main.cpp#L133) of ociodisplay/main.cpp produces an error on Linux build.

`f->read_image(OIIO::TypeFloat, &img[0]);` needs to change to `f->read_image(OIIO::TypeDesc::TypeFloat, &img[0]);`

@hodoulp did this resolve correctly on your Windows builds?

Also raises the point that not all apps are built on CI, _ociodisplay_ has the large dependency on OIIO so may not be worth the build times.

hodoulp:
Yes, it was a build break on Windows. All the builds from master are green. So what is the difference?

Concerning ociodisplay app, as OIIO is not present that app should not be built even if OCIO_BUILD_APPS is ON.

scoopxyz:
I was building _ociodisplay_ on a Linux local machine just wanted to run through all the paces.

> Yes, it was a build break on Windows

With or without the `::TypeDesc` ?

hodoulp:
My bad, ociodisplay is not compiled on Travis... So revert it and I will check on my Windows machine.
Sorry for that noise.

hodoulp:
You can check OpenImageIO/typedesc.h to understand why the two are possible. Based on my understanding, the former one TypeDesc::FLOAT is deprecated. What OIIO version are you using ?

hodoulp:
I use the 1.9.0 on from master.

scoopxyz:
I was pulling from 1.0.7 since the higher versions require C++11 as a minimum, just wanted to keep consistent

scoopxyz:
closed by #487

PR #487, opened 2017-11-29
Adjust ociodisplay build depending of the OIIO version used

hodoulp:
During my tests for a previous branch, I used the most recent OIIO and update the code accordingly. 
* Add an ifdef depending of the OIIO version to use the right type.


hodoulp:
#486

PR #488, opened 2017-12-01
Create GPU framework for unit tests

hodoulp:
As the GPU engine is more and more critical for applications using OCIO, it seems important to have unit tests validating the shader programs generated by the library. To have an easy way to write GPU unit tests, the framework is built on top of the CPU engine. The unit test defines the OCIO::Transform to be validated, and the framework automatically ensures consistency between CPU and GPU engine. So, the CPU implementation becomes the reference which means it should be fully validated using CPU unit tests. Using that approach GPU unit test are easy to write.

To demonstrate the framework and provide some examples, several unit tests were added: 1) some around MatrixOp and 2) some around Lut3DOp. It covers the two extreme cases.

The framework works well on Win7 & VS15, RH 6.8 & gcc 4.4.7. In addition, the branch was compiled on ubuntu 16.4 & gcc 5.4.0 (using docker).


hodoulp:
Unfortunately, Travis & appVeyor could not execute GPU unit tests by default because they exclusively use headless machines. I tried to use Xvfb (Linux utility to create a virtual framebuffer) but a freeglut bug is crashing the app.

scoopxyz:
On a quick test build, I'm seeing a couple warnings and errors on the GPU Unit Test build possibly related to the OpenGL version used, I haven't looked into it too much yet.

`warning: 'glutCreateWindow' is deprecated: first deprecated in macOS 10.9 [-Wdeprecated-declarations] `

and 

`error: use of undeclared identifier GL_CLAMP_VERTEX_COLOR'`

I'm also seeing that the GPU Unit Tests aren't built on TravisCI because GLUT isn't installed, we should probably modify the configuration to at least build the tests, even if they can't run.

src/core_gpu_tests/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+    set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR} )
+    # For multi-config builds (e.g. msvc)
+    foreach( OUTPUTCONFIG ${CMAKE_CONFIGURATION_TYPES} )
+        string( TOUPPER ${OUTPUTCONFIG} OUTPUTCONFIG )
+        set( CMAKE_RUNTIME_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${CMAKE_CURRENT_BINARY_DIR} )
+        set( CMAKE_LIBRARY_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${CMAKE_CURRENT_BINARY_DIR} )
+        set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${CMAKE_CURRENT_BINARY_DIR} )
+    endforeach( OUTPUTCONFIG CMAKE_CONFIGURATION_TYPES )
+
+    # Mute a design issue where the Exception public class inherits
```
scoopxyz: This should be removed after the OCIO Exception is updated

src/core_gpu_tests/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+        string( TOUPPER ${OUTPUTCONFIG} OUTPUTCONFIG )
+        set( CMAKE_RUNTIME_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${CMAKE_CURRENT_BINARY_DIR} )
+        set( CMAKE_LIBRARY_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${CMAKE_CURRENT_BINARY_DIR} )
+        set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${CMAKE_CURRENT_BINARY_DIR} )
+    endforeach( OUTPUTCONFIG CMAKE_CONFIGURATION_TYPES )
+
+    # Mute a design issue where the Exception public class inherits 
+    # from a STL Exception. STL classes are never supposed to
+    # be exported among different dynamic libraries.
+    set(EXTERNAL_COMPILE_FLAGS "${EXTERNAL_COMPILE_FLAGS} /wd4275")
```
scoopxyz: This should be updated after the OCIO Exception is modified

scoopxyz: This should be updated after the OCIO Exception is modified

hodoulp: Unfortunately, there are two issues around STL classes in public API: 1) the `std::string` and 2) the `std::exception`. The previous PR removed the `std::string` (easy change as it's a private member, and code is fully backward compatible) but did nothing for the `std::exception` (now a `std::runtime_error`). The later change is much more problematic because some existing code could stop working properly (if the inheritance is removed).

So these lines are still needed to mute some warnings.

PR #498, opened 2018-01-09
Fix Linux compilation for ubuntu gcc 5.4.0

hodoulp:
Fix gcc 5.4.0 build breaks
Fix the temp filename for Linux


hodoulp:
That PR could be a candidate for RB-1.1 if needed.

PR #514, opened 2018-02-05
Bump version numbers for After Effects and Photoshop plug-ins

fnordware:
The plug-ins are not built with CMake, so the version numbers are hard-coded, unfortunately.


fnordware:
I'm not sure why this pull request includes a bunch of extra commits. The only ones that should matter are 79bbfbb8357cc549b4f5e27bef4bf5bd8a07db29, c2794bed6abb9973cb83b7452e324f37b3894c69, and 508336d577152b21c241718abf854ed4a3c5511e.

scoopxyz:
Thanks for catching that!

Yeah looks like there was some get-foo in there, that's alright, everything will be squashed into one commit on merge.

PR #518, opened 2018-02-20
Fix build of Python bindings with GCC 8

AdamWill:
GCC 8 appears to be rather stricter about various issues to do
with type conversions and casts than GCC 7 was. This affects
OpenColorIO's Python bindings quite heavily, producing a large
number of warnings (which are converted to errors by `-Werror`)
and outright errors.

The changes here are almost all one of three basic types:

1. Many functions that become Python methods with no arguments
(using the METH_NOARGS flag) did not include the expected second
parameter in their signatures at all. METH_NOARGS does not
prevent this second parameter being passed *at all*, it only
ensures that it will always be NULL. It's still not technically
correct to leave it out of the function signature; as a comment
from 'yak' on  https://stackoverflow.com/questions/10256315
points out, there are situations where this could cause a crash.
I've added the second parameter (with no name, per convention)
to every one of these cases.

2. In several cases, classes specified a custom destructor, with
a cast to the `destructor` type, which only takes a single
parameter. However, the signatures for these destructor functions
included two parameters, assuming that they'd get an 'args'
parameter (they do not). I've corrected all these cases.

3. In several cases, classes specified custom str or repr
methods. However, in the `PyTypeObject` structures for these
classes, these methods were not cast to the `reprfunc` type, as
they ought to be. I've added these casts.

There are two warnings I just can't get rid of with my limited
C++ knowledge. The `Config` class (in PyConfig.cpp) defines a
couple of methods that take kwargs as well as args. This is
done by setting the `METH_KEYWORDS` flag, which ultimately seems
to result in a cast from type `PyCFunctionWithKeywords` to
`PyCFunction` happening somewhere behind the scenes. There's some
discussion of this at https://stackoverflow.com/questions/9496753

GCC 8 does not like this cast - it causes a 'cast-function-type'
warning. I've messed around a bit with `reinterpret_cast` and
stuff, but didn't really understand precisely what I was doing
and didn't manage to find anything that got rid of the warnings.
So I just added `-Wno-error=cast-function-type` to the compiler
flags instead, which turns these back into warnings and allows
the compilation to succeed.

I've tested at least that the compilation succeeds with only
those two warnings, and I can import the Python module and
instantiate a few classes and examine their docstrings and stuff
with no apparent errors or crashes.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/src/pyglue/PyAllocationTransform.cpp
+++ b/src/pyglue/PyAllocationTransform.cpp
@@ -53,10 +53,10 @@ OCIO_NAMESPACE_ENTER
         ///
         
         int PyOCIO_AllocationTransform_init(PyOCIO_Transform * self, PyObject * args, PyObject * kwds);
-        PyObject * PyOCIO_AllocationTransform_getAllocation(PyObject * self);
+        PyObject * PyOCIO_AllocationTransform_getAllocation(PyObject * self,  PyObject *);
         PyObject * PyOCIO_AllocationTransform_setAllocation(PyObject * self,  PyObject * args);
-        PyObject * PyOCIO_AllocationTransform_getNumVars(PyObject * self);
-        PyObject * PyOCIO_AllocationTransform_getVars(PyObject * self);
+        PyObject * PyOCIO_AllocationTransform_getNumVars(PyObject * self,  PyObject *);
+        PyObject * PyOCIO_AllocationTransform_getVars(PyObject * self,  PyObject *);
         PyObject * PyOCIO_AllocationTransform_setVars(PyObject * self,  PyObject * args);
         
         ///////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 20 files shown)


ignatenkobrain:
That's why people SHOULD NOT use `-Werror` for production builds[⁰](https://mail.gnome.org/archives/desktop-devel-list/2012-July/msg00100.html).

ignatenkobrain:
@AdamWill just make it to not use -Werror when doing production builds, so you won't need `-Wno-error`

AdamWill:
It's not my project, so not my decision.

AdamWill:
Stole an idea from [openembedded](https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-bsp/grub/files/fix.build.with.gcc-7.patch) to use GCC pragmas...this should work better I hope.

AdamWill:
Note: upstream https://gcc.gnu.org/ml/gcc-patches/2018-02/msg01315.html / https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84531 should stop the remaining two warnings being emitted. Not sure what GCC version they'll land in. Thanks @davidmalcolm for the reference.

PR #520, opened 2018-03-04
Repository Restructure and CMake refactor

Shootfast:
Hi,

This isn't quite 100% ready yet (I need to rebase with some latest patches and unit tests), but I'm putting it here for public review / criticism on what I've done so far.

* Rearranged directory structure - libs / binaries / bindings are in their own directories. Headers are in include.
* CMake rewritten. Uses target properties and exports where possible (still some work on ext libs to be done)
* Windows builds out-of-the-box (using bundled yamlcpp and tinyxml)
* Removed Boost dependency

What's not done yet:
* unit tests (cpu and GPU)
* make external packages correctly export their targets via CMake patches
* misc installers I haven't added back in yet.

I'm hoping to tick of the remaining issues over the next week or so!


CMakeLists.txt
```diff
@@ -1,677 +1,87 @@
-project(OpenColorIO)
-set(OCIO_VERSION_MAJOR 1)
-set(OCIO_VERSION_MINOR 1)
-set(OCIO_VERSION_PATCH 0)
+cmake_minimum_required(VERSION 3.1)
+project(OpenColorIO VERSION 1.2.0)
+
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
hodoulp: Based on [this page](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/) it seems that `set(CMAKE_CXX_EXTENSIONS OFF)` must be added to avoid platform specific C++11 extensions.

share/cmake/FindLCMS2.cmake
```diff
@@ -0,0 +1,69 @@
...
+#  LCMS2_VERSION, the version of the library
+#
+# By default, the dynamic libraries of lcms2 will be found. To find the static ones instead,
+# you must set the LCMS2_STATIC_LIBRARY variable to TRUE before calling find_package(LCMS2 ...).
+#
+include(FindPkgConfig)
+
+# attempt to find static library first if this is set
+if(LCMS2_STATIC_LIBRARY)
+	set(LCMS2_STATIC liblcms2.a)
```
hodoulp: That's a linux specific name.
For Windows, the names of the 'external' libraries are: lcms2.lib, libyamll-cppmd.lib and tinyxml.lib.

dbr:
Looks like a great improvement!

A few minor comments on first glance:

1. I prefer the original `src/core/` to `src/OpenColorIO` (otherwise it kind of sounds like all the other code isn't part of OCIO)
2. Confusing having `src/pystring/` - seems like it should be under util (same as argparse etc), or I guess ideally under `ext/` since it's a thirdparty project too.
3. I'd expect the `tests` to be under `src/tests/` rather than a separate top-level folder
4. Lots of the files have been moved in a way which they lose their history, which is unfortunate (git blame etc can be very useful when tracking down bugs)

Might be out-of-scope for this ticket, but since it relates to tidying up the repo:

1. There's a separate changelog in docs which is empty and untouched for 6 years, could be removed
2. The rv integration code in `thirdparty/rv/` is very outdated and should be deleted. Probably same with Mari one
3. `ocioconvert` and maybe `ociodisplay` feel a bit like example code rather than practical utilities (compared to ociocheck/ociobakelut/etc). Not sure if these should be handled differently, somehow. Particularly given the somewhat circular dependency where OIIO depends on OCIO (and these parts of OCIO then depend on OIIO)

scoopxyz:
> I prefer the original src/core/ to src/OpenColorIO (otherwise it kind of sounds like all the other code isn't part of OCIO)

I think its just to highlight that they aren't a part of the core library, which is true. You have executables that link against it, and language bindings that are built on top of it. I'd say they _use_ OpenColorIO but aren't in and of themselves "OpenColorIO" the library, but are a part of "OpenColorIO" the project.

> There's a separate changelog in docs which is empty and untouched for 6 years, could be removed
The rv integration code in thirdparty/rv/ is very outdated and should be deleted. Probably same with Mari one
ocioconvert and maybe ociodisplay feel a bit like example code rather than practical utilities (compared to ociocheck/ociobakelut/etc). Not sure if these should be handled differently, somehow. Particularly given the somewhat circular dependency where OIIO depends on OCIO (and these parts of OCIO then depend on OIIO)

I agree on the Changelog, just needs to be deleted. RV and Mari could also disappear since they're natively supported now. I'd say that _ocioconvert_ should go, OIIO has supported OCIO for some time now, so it's redundant and frankly not useful.

 _ociodisplay_ is a little different. I think it's useful to provide a good reference implementation of OCIO, but I don't think _ociodisplay_ satisfies that completely. It could be moved to its own repository so as to remove the OIIO dependency from OCIO completely.

scoopxyz:
ah, _ociolutimage_ also depends on OIIO

dekekincaid:
Why would you remove ocioconvert?  For automated tests it shouldn’t have an
external dependency just to test itself.  If you have a problem how do you
quickly know if the issue is oiio or ocio?

scoopxyz:
*ocioconvert* _already_ has a dependency on OIIO, and there isn't anything that *ocioconvert* does that couldn't be done by `Processor::apply`

dbr:
>> I prefer the original src/core/ to src/OpenColorIO (otherwise it kind of sounds like all the other code isn't part of OCIO)
>
> I think its just to highlight that they aren't a part of the core library

Indeed, but seems like having `src/core/` highlights this distinction as clearly as anything?

> ociodisplay is a little different. I think it's useful to provide a good reference implementation of OCIO

ociodisplay is a definitely valuable piece of example code to have around (same with ocioconvert). I was thinking they could be organised a bit more like, say, the Qt "demos" example code - an clearly separate (optional) component when you build Qt (still be built via CI of course), creates a separate executable, and structured distinctly from the main source - e.g https://github.com/qt/qtbase/tree/5.11/examples/

Wouldn't be a big change, just a slightly clearer distinction between the "core" applications like ociocheck (might just be a case of moving them into `examples/` instead of `src/`)


> ah, ociolutimage also depends on OIIO

Hmm oh yeh - that is a bit unfortunate. I wonder if instead of OIIO it could use a few smaller libraries which we could bundle into the thirdparty modules (maybe [TinyTIFF](https://github.com/jkriege2/TinyTIFF) and [TinyEXR](https://github.com/syoyo/tinyexr)). Much less flexible than OIIO, but that isn't necessary a bad thing (e.g would mean people couldn't accidentally try and extract a LUT from a compressed JPEG)

Probably worth splitting this into a separate ticket

hodoulp:
1) Adding to the discussion: I would like to mention that at some point, we will have to add the half lut and perhaps the half bit depth for the apply(). Both require the half library part of the OpenEXR. So openEXR could be a good candidate for OCIO tools. 

2) ocioDisplay is definitely a valuable part of the OCIO as example and as test tool.

3) CPU unit tests are 'white box' ones which means that they access private OCIO classes/methods. It explains why they are 'embedded' in core files (using OIIO macros). On the other hand, GPU unit tests are 'black box' ones (i.e. only using OCIO public API) so they could be in a separate directory.   This distinction is not mandatory. So I will not argue on changes but I think that these two approaches are good to have.

hodoulp:
@Shootfast  Following the agreement explained in #556, this work is now critical for the next steps of OCIO v2.    So, I took few hours to investigate the branch. Only few minor changes are needed to make it compile on Windows (except the unit tests which are missing). 

[ocio_mark_patch.diff.txt](https://github.com/imageworks/OpenColorIO/files/2384738/ocio_mark_patch.diff.txt)

[FindIlmBase.cmake.txt](https://github.com/imageworks/OpenColorIO/files/2384742/FindIlmBase.cmake.txt)

Not beeing a cmake expert please only use these files as a way to highlight where are the remaining problems.

scoopxyz:
@Shootfast, noticed there we updated commits but I'm not certain if you're done.

For those following at home, here's the proposed structure from the latest commit on this fork:

```
├── docs
│   ├── configurations
│   │   └── images
│   │       └── spi-vfx
│   ├── developers
│   │   ├── api
│   │   └── bindings
│   ├── ociotheme
│   │   └── static
│   └── userguide
│       └── images
│           └── ps_icc
├── ext
│   ├── oiio
│   │   └── src
│   │       └── include
│   └── sampleicc
│       └── src
│           └── include
├── include
│   └── OpenColorIO
├── share
│   ├── cmake
│   ├── docker
│   ├── nuke
│   │   ├── examples
│   │   └── ocionuke
│   ├── ocio
│   └── sphinx
├── src
│   ├── java
│   │   ├── org
│   │   │   └── OpenColorIO
│   │   └── tests
│   │       └── org
│   │           └── OpenColorIO
│   ├── ociobakelut
│   ├── ociocheck
│   ├── ocioconvert
│   ├── ociodisplay
│   ├── ociolutimage
│   ├── OpenColorIO
│   │   ├── md5
│   │   └── pkgconfig
│   ├── openglbuilder
│   ├── pystring
│   ├── python
│   │   ├── DocStrings
│   │   ├── PyOpenColorIO
│   │   └── tests
│   └── util
├── testdata
├── tests
│   ├── cpu
│   └── gpu
└── thirdparty
    ├── aftereffects
    │   ├── mac
    │   ├── vc
    │   │   └── vc9
    │   │       ├── aftereffects
    │   │       └── ext
    │   ├── win
    │   └── xcode
    │       ├── aftereffects
    │       │   └── OpenColorIO_AE.xcodeproj
    │       ├── ext
    │       │   ├── lcms.xcodeproj
    │       │   ├── tinyxml.xcodeproj
    │       │   └── yaml.xcodeproj
    │       └── OpenColorIO.xcodeproj
    ├── mari
    │   ├── 1.4v1
    │   └── prototype
    ├── nuke
    │   ├── OCIOCDLTransform
    │   ├── OCIOColorSpace
    │   ├── OCIODisplay
    │   ├── OCIOFileTransform
    │   ├── OCIOLogConvert
    │   └── OCIOLookTransform
    ├── photoshop
    │   ├── mac
    │   ├── vc
    │   │   └── vc9
    │   ├── win
    │   └── xcode
    │       └── OpenColorIO_PS.xcodeproj
    └── rv
        ├── Mu
        ├── Packages
        └── Python

92 directories
```

I do think it is much more clear and straight forward, though I'm still not 100% on the overall structure. In particular `ext` and `thirdparty` are ambiguous to me. Isn't it pretty standard to put third party libraries in `thirdparty`? I also preferred having the apps in a separate directory, even if only for visual clarity.

So anyway, I took a stab at what it would like to me as a additional commit on top of your cmake work (which is much much better) and I didn't touch it much apart from dir changes. Found here:
[https://github.com/scoopxyz/OpenColorIO/tree/shootfast-cmake_refactor](scoopxyz/shootfast-cmake_refactor)

Directory layout here:
```
├── doc
│   ├── configurations
│   │   └── images
│   │       └── spi-vfx
│   ├── developers
│   │   ├── api
│   │   └── bindings
│   ├── ociotheme
│   │   └── static
│   └── userguide
│       └── images
│           └── ps_icc
├── include
│   └── OpenColorIO
├── share
│   ├── cmake
│   ├── docker
│   ├── nuke
│   │   ├── examples
│   │   └── ocionuke
│   ├── ocio
│   └── sphinx
├── src
│   ├── apps
│   │   ├── ociobakelut
│   │   ├── ociocheck
│   │   ├── ocioconvert
│   │   ├── ociodisplay
│   │   ├── ociolutimage
│   │   └── util
│   ├── bindings
│   │   ├── java
│   │   │   ├── org
│   │   │   │   └── OpenColorIO
│   │   │   └── tests
│   │   │       └── org
│   │   │           └── OpenColorIO
│   │   └── python
│   │       ├── DocStrings
│   │       ├── PyOpenColorIO
│   │       └── tests
│   ├── OpenColorIO
│   │   ├── md5
│   │   └── pkgconfig
│   └── util
│       ├── openglbuilder
│       └── pystring
├── test
│   ├── cpu
│   ├── data
│   └── gpu
├── thirdparty
│   ├── oiio
│   │   └── src
│   │       └── include
│   └── sampleicc
│       └── src
│           └── include
└── vendor
    ├── aftereffects
    │   ├── mac
    │   ├── vc
    │   │   └── vc9
    │   │       ├── aftereffects
    │   │       └── ext
    │   ├── win
    │   └── xcode
    │       ├── aftereffects
    │       │   └── OpenColorIO_AE.xcodeproj
    │       ├── ext
    │       │   ├── lcms.xcodeproj
    │       │   ├── tinyxml.xcodeproj
    │       │   └── yaml.xcodeproj
    │       └── OpenColorIO.xcodeproj
    ├── mari
    │   ├── 1.4v1
    │   └── prototype
    ├── nuke
    │   ├── OCIOCDLTransform
    │   ├── OCIOColorSpace
    │   ├── OCIODisplay
    │   ├── OCIOFileTransform
    │   ├── OCIOLogConvert
    │   └── OCIOLookTransform
    ├── photoshop
    │   ├── mac
    │   ├── vc
    │   │   └── vc9
    │   ├── win
    │   └── xcode
    │       └── OpenColorIO_PS.xcodeproj
    └── rv
        ├── Mu
        ├── Packages
        └── Python

95 directories
```

scoopxyz:
Regardless of the final decision it should try to happen swiftly, so as to unblock the ADSK contributions...

Also your Python bindings weren't building correctly due to a missing definition:
`add_definitions(-DPYOCIO_NAME=PyOpenColorIO)`

Shootfast:
> @Shootfast, noticed there we updated commits but I'm not certain if you're done.

Yeah, I'm almost finished. Just chasing the last few build bugs.

>  In particular ext and thirdparty are ambiguous to me.

`ext` are our own (OpenColorIO) external dependencies.
`thirdparty` are the third-party plugins we provide (that link against OpenColorIO)

I still have to shuffle some tests around (python and java binding tests in particular), and I have some linker errors on the windows side that I've not 100% found the cause of yet.

Shootfast:
I like your changes though! I might pull those across / make the same changes to my branch. I haven't yet included putting the ops, fileformats, and transforms into their own subdirectories yet. Do you think that would be a worthwhile change to make here at the same time?

Shootfast:
> In addition to Python3 being pulled in on MacOS and breaking

Python3 is not supported yet right? It's the build bot configuration that is wrong, not me right? :innocent:

scoopxyz:
For Python stuff, I think we need to manually force it to Python2.7 if Python3 doesn't work out of the box... at least until its fixed and working.

Seems like you'll specify by:
```
find_package(PythonLibs 2.7 REQUIRED)
find_package(PythonInterp 2.7 REQUIRED)
``` 
In both the python binding and docs CMakeLists.txt

scoopxyz:
Submitted a PR to your branch if you wanted to pull my changes in.

scoopxyz:
That did indeed fix the MacOS issue, but the Windows builds are horribly maimed...

hodoulp:
Hoping it could help, here are my findings for the Windows platform:
1. *jom ociodisplay*, apps need an access to 'ilmbase'  (Refer to my previous comment)
2. *jom test_gpu_exec* needs an access to 'glew'
3. *jom test_cpu_exec*  symbol visibility issue, (will investigate this later morning)

Shootfast:
> 1. jom ociodisplay, apps need an access to 'ilmbase' (Refer to my previous comment)

Nothing in OCIO currently depends on ilmbase, and there don't seem to be any includes for any of its components. This might be a red-herring

> 2. jom test_gpu_exec needs an access to 'glew'

Thanks for this one, I'll add it to the CMake for it (though I've disabled the GPU tests on the CI hosts)

> 3. jom test_cpu_exec symbol visibility issue, (will investigate this later morning)

Yeah any help here would be great. I'm fairly sure it's me building sub-components with the wrong visibility settings when they pull in the public api. Might be the unit tests too as they pull the same files (but don't link against the OpenColorIO dll)

hodoulp:
BTW, I like @scoopxyz proposal for the new file structure.

hodoulp:
> > jom test_cpu_exec symbol visibility issue, (will investigate this later morning)

> Yeah any help here would be great. I'm fairly sure it's me building sub-components with the wrong visibility settings when they pull in the public api. Might be the unit tests too as they pull the same files (but don't link against the OpenColorIO dll)

That's a visibility issue introduced by the OpenColorABI.h.in. By default, the include imposes to export the symbols (and does not offer any explicit way to 'import' symbols). To demo my finding, I used the legacy '-DOpenColorIO_STATIC' in the test/cpu/CMakeLists.txt.

I would recommand to have the 'import' as the default behavior, and let src/OpenColorIO/CMakeLists.h adding a define to mention that it exports symbols (i.e. it's the only one having to export symbols). Hence, all apps/tests... would import symbols without anything to add at compile time.

All my investigations were done on Windows with Visual C++.

hodoulp:
The former **test_verbose** option would good to preserve.
Note: could be done in a follow-up PR.

hodoulp:
Add these lines to the file tests\CMakeLists.txt seems to do the job:
`if(OCIO_BUILD_TESTS OR OCIO_BUILD_GPU_TESTS)
add_custom_target(test_verbose
                  COMMAND ctest -VV
                  DEPENDS ${TEST_DEPENDS}
                  COMMENT "Running ctest with verbose output")
endif()
`

scoopxyz:
Just looping back after yesterdays comments?

hodoulp:
The Windows build failure could be to replace the src\utils\CMakeLists.txt content by:
```
find_package(OpenGL QUIET)
find_package(GLEW QUIET)
if(OpenGL_FOUND AND GLEW_FOUND)
    add_subdirectory(openglbuilder)
else()
    message(WARNING "OpenGL or GLEW not found. Skipping build for openglbuilder")
endif()

add_subdirectory(pystring)
```

Shootfast:
Unfortunately that wouldn't work because the GPU test target will then depend on something that doesn't exist. It's very annoying that the find_package calls aren't conditionally tied to a target being a dependency. I've added the correct exclusions now.

hodoulp:
Not enough because serval CMakeList.txt need to change. But I still continue my investigations, I should have something better soon.

hodoulp:
Here is a patch tosuccessfully build on Windows:
[ocio_mark_patch.diff.txt](https://github.com/imageworks/OpenColorIO/files/2416406/ocio_mark_patch.diff.txt)

Only Python unit tests are still failing.

scoopxyz:
Also, a summary of the changes included:
- Jump to C++11
- Dropping Boost dependency
- CMake Minimum 3.10
- Restructured/Modernized CMake
- Restructured repo directories
- Bumping `master` version to 2.0.0
- Build Shared _or_ Static libs not both
- Explicit linking to Python2.7 libraries (for now)

@Shootfast correct me if wrong/missing things

Shootfast:
The only side note is that the java bindings are not working at the moment, but that is already the case on the master branch, and I will try to fix them up soon. I do also want to make it so that the apps aren't built by default (as this leads to the circular dependency issue), but again I can make those changes after this patch is merged

hodoulp:
**[Minor]** It would be nice to have a target like **jom test_verbose** to run test in verbose mode (i.e. like ctest -V)

hodoulp:
Using the target **Visual Studio 14 2015 Win64** with Vc14 in release mode, the command GPU & Python uni tests fail:
```
1/3 Test #1: test_cpu .........................   Passed    0.56 sec
test 2
    Start 2: test_gpu

2: Test command: D:\OpenSource\ocio_mark\build_rls\tests\gpu\Release\test_gpu_exec.exe
2: Test timeout computed to be: 10000000
2/3 Test #2: test_gpu .........................***Exception: Exit code 0xc0000135
  0.02 sec
test 3
    Start 3: test_python

3: Test command: D:\Tools\cygwin64\bin\python2.7.exe "D:/OpenSource/ocio_mark/tests/python/OpenColorIOTestSuite.py" "D:/OpenSource/ocio_mark/build_rls" "Release"
3: Test timeout computed to be: 10000000
3: Traceback (most recent call last):
3:   File "D:/OpenSource/ocio_mark/tests/python/OpenColorIOTestSuite.py", line 18, in <module>
3:     import PyOpenColorIO as OCIO
3: ImportError: No module named PyOpenColorIO
3/3 Test #3: test_python ......................***Failed    1.77 sec

33% tests passed, 2 tests failed out of 3

Total Test time (real) =   2.40 sec

The following tests FAILED:
          2 - test_gpu (Exit code 0xc0000135
)
          3 - test_python (Failed)
Errors while running CTest
```

hodoulp:
**[Minor]** The compilation variable **OCIO_WARNING_AS_ERROR** is not used by the build
```-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    OCIO_WARNING_AS_ERROR
```

src/OpenColorIO/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+	transforms/GroupTransform.cpp
+	transforms/LogTransform.cpp
+	transforms/LookTransform.cpp
+	transforms/MatrixTransform.cpp
+	transforms/TruelightTransform.cpp
+	UnitTest.cpp
+)
+
+add_library(OpenColorIO ${SOURCES})
+set_property(TARGET OpenColorIO PROPERTY CXX_STANDARD 11)
```
hodoulp: The C++ version selection is already done in the CMakeLists.txt line 5. Why is that needed here too?
If oneday the C++ version is changed, only one location should be changed.

src/OpenColorIO/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+endif()
+
+if(OCIO_USE_SSE)
+	target_compile_definitions(OpenColorIO
+		PRIVATE
+			USE_SSE
+	)
+endif()
+if(MSVC)
+	# Because our Exceptions derive from std::runtime_error we can safely disable this warning
```
hodoulp: That's an unfortunate side-effect of deriving the OCIO exception from STL exceptions. The STL library classes were never designed to be used as base classes and exported. At the same time, it's extremely an useful design to ease the code writing.

src/OpenColorIO/Op.cpp
```diff
@@ -149,10 +149,10 @@ OCIO_NAMESPACE_EXIT
...
 
 namespace OCIO = OCIO_NAMESPACE;
-#include "UnitTest.h"
-#include "MatrixOps.h"
-#include "LogOps.h"
-#include "NoOps.h"
+#include "unittest.h"
+#include "ops/Matrix/MatrixOps.h"
+#include "ops/Log/LogOps.h"
+#include "ops/NoOp/NoOps.h"
```
hodoulp: **[Minor]** NoOp does not deserve to have its own directory. The original location was good at my point of view.

src/apps/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+if(OCIO_BUILD_APPS)
+	add_subdirectory(apputils)
+	add_subdirectory(ociobakelut)
+	add_subdirectory(ociocheck)
+
+	if(TARGET OpenImageIO)
```
hodoulp: Sometime some external libraries are checked (i.e. OpenImageIO) and the build is 'adjusted'. Sometimes some external libraries (GLUT, GLEW) are required, and the build fails if they are not present. On the Windows platform, the two cases are strictly equivalent as none of these libraries are installed by default.

I would recommend following the same rule (i.e. the OpenImageIO one) for all external libraries.

src/bindings/java/CMakeLists.txt
```diff
@@ -0,0 +1,161 @@
+find_package(Java REQUIRED)
+find_package(JNI REQUIRED)
+
+set(SOURCES
```
hodoulp: What not using a 'glob' to find all the sources?
On long-term, hard-coding the file names introduces unnecessary extra maintenance cost.

tests/cpu/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+	endif(PRIVATE_INCLUDES)
+	add_test(${TEST_NAME} ${TEST_BINARY})
+endfunction(add_ocio_test)
+
+
+
+# Eventually we will factor out each test into it's own executable
+# but for now, we will maintain the status quo and copy all from the
+# OpenColorIO target
+set(SOURCES
```
hodoulp: Same comment as above about hard-coded file names vs. using a glob

tests/gpu/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+find_package(GLEW REQUIRED)
+find_package(GLUT REQUIRED)
+
```
hodoulp: The code should add the OpenGL dependency.

hodoulp:
Compiling (i.e. Vc14 in Release mode) with **-DBUILD_SHARED_LIBS=OFF** fails because of symbol issues.
...
```
PyConstants.cpp.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) enum OpenColorIO::v2_0::BitDepth __cdecl OpenColorIO::v2_0::BitDepthFromString(char const *)" (__imp_?BitDepthFromString@v2_0@OpenColorIO@@YA?AW4BitDepth@12@PEBD@Z
)
PyColorSpace.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: static class std::shared_ptr<class OpenColorIO::v2_0::ColorSpace> __cdecl OpenColorIO::v2_0::ColorSpace::Create(void)" (__imp_?Create@ColorSpace@v2_0@Open
ColorIO@@SA?AV?$shared_ptr@VColorSpace@v2_0@OpenColorIO@@@std@@XZ) referenced in function "int __cdecl OpenColorIO::v2_0::`anonymous namespace'::PyOCIO_ColorSpace_init(struct OpenColorIO::v2_0::PyOCIOObject<class std::shared_ptr<class OpenColorIO::v2
_0::ColorSpace const >,class std::shared_ptr<class OpenColorIO::v2_0::ColorSpace> > *,struct _object *,struct _object *)" (?PyOCIO_ColorSpace_init@?A0x454721e1@v2_0@OpenColorIO@@YAHPEAU?$PyOCIOObject@V?$shared_ptr@$$CBVColorSpace@v2_0@OpenColorIO@@@s
td@@V?$shared_ptr@VColorSpace@v2_0@OpenColorIO@@@2@@23@PEAU_object@@1@Z)
jom: D:\OpenSource\ocio_mark\build_rls\src\bindings\python\CMakeFiles\PyOpenColorIO.dir\build.make [src\bindings\python\PyOpenColorIO.pyd] Error 2
```

**Note:** That case should be added to the CI builds.

**Note:** If it's problematic to fix, it could be done in a follow-up pull request i.e. not blocking

ext/CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+######################################################################
+# TinyXML
+add_library(tinyxml UNKNOWN IMPORTED GLOBAL)
+if(USE_EXTERNAL_TINYXML)
+	find_package(TinyXML 2.6.1 REQUIRED)
+else()
+	set(TINYXML_INSTALL_DIR "${CMAKE_BINARY_DIR}/ext/dist")
+	set(TINYXML_CMAKE_ARGS
+		-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+		-DBUILD_SHARED_LIBS:BOOL=OFF
```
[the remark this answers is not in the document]
Shootfast: The -DBUILD_SHARED_LIBS:BOOL=OFF already forces the dependent libs to build statically

ext/CMakeLists.txt
```diff
@@ -0,0 +1,154 @@
...
+	)
+	if(CMAKE_TOOLCHAIN_FILE)
+		set(TINYXML_CMAKE_ARGS 
+			${TINYXML_CMAKE_ARGS}
+			-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+		)
+	endif()
+
+	if(NOT BUILD_SHARED_LIBS)
+		#TODO: Find a way to merge in the static libs when built with internal yamlcpp
```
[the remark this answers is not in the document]
Shootfast: Fixed, and added the same warning to lcms2!

ext/CMakeLists.txt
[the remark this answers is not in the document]
hodoulp: Too much code to review :) The code is doing the right thing.

src/OpenColorIO/CMakeLists.txt
Shootfast: Correct, This was leftover from an earlier test. I've removed the additional check.

src/OpenColorIO/CMakeLists.txt
Shootfast: That's an extremely common use case, some would say it's what std::runtime_error was made for. It's kinda silly that the MSVC STL has such flimsy binary compatability that this would warrant a warning.
We can only hope that this is mitigated by developers only building this library in the same configuration as the app it will end up embedded in.

src/OpenColorIO/Op.cpp
Shootfast: I debated that too, but I think it might as well live with the other ops :P

src/bindings/java/CMakeLists.txt
Shootfast: Globbing in build systems is evil for a number of reasons. This stackoverflow post lists some:

https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak/1060061#1060061

Another big one is that it ruins the ability to create reproducible builds:
https://reproducible-builds.org/docs/stable-inputs/

tests/cpu/CMakeLists.txt
Shootfast: as above, globbing is evil

tests/gpu/CMakeLists.txt
Shootfast: Good catch! I'll add that

src/OpenColorIO/CMakeLists.txt
hodoulp: However, it's a quite risky assumption as VS could do lot of things in your back. When there are some memory model mismatches between libraries, it could change it without asking i.e. only a command line warning.

tests/gpu/CMakeLists.txt
Shootfast: Fixed all instances of this

src/apps/CMakeLists.txt
Shootfast: This is a tricky one. The presence of OIIO is the common requirement for those additional apps, hence me using it to drive their inclusion. In the past we considered OIIO to be a "non blocking" requirement, as you can still build some tools without it. I would be tempted to remove the special check for OIIO and make the entire OCIO_BUILD_APPS require OIIO, as this might also help the people with the circular dependency problem.
This would treat OIIO, GLUT and GLEW as hard dependencies for all apps, and we could just rely on the traditional find_package(OIIO) calls.

Shootfast:
> **[Minor]** It would be nice to have a target like **jom test_verbose** to run test in verbose mode (i.e. like ctest -V)

I am not a fan of additional targets that are created solely for non-build related tasks. If you want verbose tests, run ctest -V!

I've made the verbose output be the default for the CI jobs.

Shootfast:
>  After some investigations, I found that VS does not create the output results at the specified location, but rather at the location plus "/Release". The GPU unit tests are using the OpenColorIO dynamic library (black box testing) and failed to find it.
>
> A solution is to update the execution environment.

The current Python test runner is a bit of a hack. I might try and get cmake to pass the built location of OpenColorIO and PyOpenColorIO explicitly, rather than relying on my guesswork.

src/bindings/java/CMakeLists.txt
hodoulp: Really interesting discussion.

src/apps/CMakeLists.txt
hodoulp: On the windows platform, the makefile generation will fail right away as none of these libraries are present by default. I would prefer to let Windows devs be able to generate the makefiles and compile successfully even if some apps are missing (and GPU unit tests are missing too).
Another point is that the libraries are part of two different categories, one to read/write a file (OIIO) and some to have GPU. For example, GPU unit tests only need GPU related libraries while ociodisplay needs both categories.
Note: The find_package with some ' potential path' will never work on Windows.

My proposal is to only have soft dependencies with clear messages explaining the 'adjustment'.

With the long-term maintenance perpective in mind, we should help devs (*nix or Windows platforms) successfully compile the library itself whatever is the platform.

hodoulp:
After a discussion with @Shootfast and the fact that the work is blocking for any development on the repo, we agreed to merge the work as-is. The modernization of the depot (refactor/simplify of cmake files, and file source code retructuration) is now successfully completed.

Here is a summary of missing parts:
1) Compile in 'static library' mode has some symbol issues (dynamic library is working fine)
2) Discussion around hard vs. soft requirement for external libraries (i.e. OIIO, Glew...) is still pending
3) Python unit test failure on Windows
4) Python debug issue on Windows when linking PyOpenColorIO
5) Java binding & unit tests not done

@Shootfast please correct me if needed.

PR #521, opened 2018-03-15
After Effects and Photoshop Windows fixes

fnordware:
Some small fixes to the After Effects and Photoshop plug-ins on Windows. Mainly changing how I get the $OCIO environment variable on Windows. There was also a bad menu label.


src/aftereffects/OpenColorIO_AE.cpp
```diff
@@ -634,14 +634,14 @@ static PF_Err DoRender(
...
                 {
-				#ifdef MAC_ENV
-					char *envFile = std::getenv("OCIO");
-					const bool gotOCIO = (envFile != NULL);
-				#else
-					char envFile[32767] = { '\0' };
-					const DWORD envResult = GetEnvironmentVariable("OCIO", envFile, 32767);
-					const bool gotOCIO = (envResult > 0);
-				#endif
+                #ifdef MAC_ENV
```
hodoulp: Could you use/enhance the getenv() from platform.h/cpp ?

fnordware: Hmmm, I never saw that. One problem is OCIO::Platform::getenv() is not part of the OCIO API, but used by the library internally. Maybe that's why ociobakelut and other programs we include don't use it.

Seems like there should be an API call to get the environment config. Or at least a constant with "OCIO".

fnordware: So I figured if I couldn't use OCIO::Platform::getenv() I should at least make my own rather than have all those #ifdefs floating around. Couple more commits made.

src/photoshop/OpenColorIO_PS_Context.cpp
```diff
@@ -283,3 +283,19 @@ OpenColorIO_PS_Context::getDefaultTransform(const std::string &device) const
     return _config->getDefaultView( device.c_str() );
 }
 
+
+void
+OpenColorIO_PS_Context::getenv(const char *name, std::string &value)
+{
+#ifdef __APPLE__
```
hodoulp: What about the linux support ?

src/photoshop/OpenColorIO_PS_Context.cpp
```diff
@@ -283,3 +283,19 @@ OpenColorIO_PS_Context::getDefaultTransform(const std::string &device) const
...
+    char *env = std::getenv(name);
```
hodoulp: A test on the name could be done to ensure that it is not null or empty.

src/aftereffects/OpenColorIO_AE.cpp
hodoulp: Good idea.

src/photoshop/OpenColorIO_PS_Context.cpp
fnordware: Photoshop and After Effects are Mac and Windows only.

scoopxyz:
Hi @fnordware thanks again for your photoshop and ae support. We've done a refactor of the master branch and now place this code in vendor/aftereffects and vendor/photoshop. If you don't have the time to refactor I can submit a PR with your changes

fnordware:
A lot has happened since I submitted this! Now superseded by #865 for OCIO v2.

Issue #524, opened 2018-03-20
GPU Unit Tests - Add Lut 1D tests


hodoulp:

hodoulp:
Mostly fixed by #640

Issue #523, opened 2018-03-20
GPU Unit Tests - Add Exponent tests


hodoulp:

hodoulp:
Fixed.

PR #530, opened 2018-04-11
Unit tests for ops

BernardLefebvre:
Unit tests for operators: review existing tests and add no tests.
Add macro to test the content of an Exception being thrown.
Tests for AllocationOps, ExponentOps, LogOps, Lut1DOps, Lut3DOps, MatrixOps, NoOps.
Fix GPU tests disabling.
Remove an hardcoded dependency.
Fix a gcc 4.4.x build break. The program was only used by gcc 8.x.x build so hide it for lower gcc versions


hodoulp:
Address part of #523 and #524

PR #539, opened 2018-06-12
Add the new GPU API support

hodoulp:
Here is the first version of the generic shader engine for OCIO v2. The change implements the new public GPU API to support either the legacy shader (i.e. baking all luts in one single 3D lut), or the generic shader (i.e preserving all the luts).
The legacy shader should perfectly work. But the generic still does not support the lut inverses. Lot of new GPU unit tests are now validating the legacy and generic shaders against the CPU color processing. 

To ease the review, one could start with the ociodisplay/main.cpp.

The current work does not finalize the inverse lut support because of some missing functionalities we are currently adding for the clf support. It explains why the legacy shader is still the recommended shader to be used.


talonrenaud:
We are trying to get Autodesk to incorporate those changes in RV since this is a major issue for us. I believe the RV team is still using OCIO 1.0.9, would they get this patch if they were to update to a more recent version or is it a OCIO 2.0+ fix only ?
Thank you.

doug-walker:
It is v2 only as it is a major change to the API.

PR #541, opened 2018-06-27
Adsk contrib/add icc support

BernardLefebvre:
Add support to allow reading icc profiles.
Extract code for reading from SampleICC.
Add tests.


hodoulp:
An ICC profile could now be added to a configuration file.

PR #550, opened 2018-07-09
Match code convention on casing of SatNode to clear up ambiguity

tanant:
We had some issues at work the other week where some code was doing up CCs for export using <SATNode> rather than <SatNode> tokens - I originally thought this was a legacy issue but it might've been due to the fact that the easiest-available example the developer copied from was the user documentation here in the OCIO repo.

It's not a _major_ issue, but wanted to suggest a documentation update to match the test code and the XML element being created in core/CDLTransform.cpp


tanant:
I'll see if anyone at the office has a more recent copy of the spec (I'm not sure if one exists, but hey, can't hurt to dig) tomorrow and let you know - same logic with this PR though, I'd be more inclined to crowbar the samples to match the spec than anything else.

For the use of both SatNode and SATNode, I'm a bit on the fence - makes life easier but for anyone using the spec the latter is technically not valid (unless I'm going nuts, XML tags are case sensitive, aren't they?) so my tech hat says it's probably a 'feature' (bug) to allow both forms.

tanant:
So, noone had anything handy - next stop is to send out an email to the relevant bodies to see if they have something :|

PR #551, opened 2018-07-11
Add CI build option

scooperly:
Seeing as how we test very few build combinations through CI, I feel its best not to completely halt compilation for someone else that is trying to build OCIO for themselves. Even one minor version up in GCC produces warnings.

So submitting this to limit -Werror purely to contributors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,10 @@ env:
 # Install missing packages
 before_install:
   # Linux ones
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then 
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       sudo apt-get update -qq;
       sudo apt-get install -y -qq texlive-full;
-      sudo apt-get install -y -qq freeglut3-dev libglew-dev libxmu-dev libxi-dev; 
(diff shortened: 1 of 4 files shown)


scoopxyz:
Any opinions on this? I'm not 100% on the option name "OCIO_CI" maybe "WERROR" makes more sense.

But in practice, I still don't think its a good idea to halt compilation for other folks. Unless someone knows some easy CMake-foo to override this for your average CMake user.

hodoulp:
There are two parts to your concern:
1. 'Warnings as Errors' is a powerful trick to impose some thinking on warnings. Targeting several platforms and compilers it could greatly help ‘discovering’ bugs. I think that OCIO should enforce the ‘Warnings as Errors’ to all parts of the code (including Python bindings, apps…). But I would not disagree to have a flag to disable that behavior on demand.   So, I would prefer something like ‘OCIO_WARNING_AS_ERROR’ with true by default.
2. The CI build do not cover all possible platforms/compilers/versions which is normal. But it could be certainly improved by adding more compiler/platform versions.

scoopxyz:
I'm sold on your suggestion, I'll revise the PR with that

dbr:
Would it make sense to disable `-Werror` based on [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) ?

PR #560, opened 2018-09-01
Fix oiio dependent apps compilation

remia:
This could be used to fix the compilation errors when building OCIO 1.1 (while waiting for 2.0) application using OIIO >= 1.8.5. This occured to me lately during a fresh Mac install with brew (`brew install opencolorio --with-python` trigger a compilation that fails in the end).

Would be glad to have others thoughts.

### DIFF
--- a/src/apps/ocioconvert/CMakeLists.txt
+++ b/src/apps/ocioconvert/CMakeLists.txt
@@ -11,6 +11,10 @@ if (OIIO_FOUND)
     
     add_executable(ocioconvert ${share_src_files} main.cpp)
     
+    if(OIIO_VERSION AND NOT ${OIIO_VERSION} VERSION_LESS 1.8.5)
+        set_property(TARGET ocioconvert PROPERTY CXX_STANDARD 11)
+    endif()
+    
     set_target_properties(ocioconvert PROPERTIES COMPILE_FLAGS -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE})
 
     target_link_libraries(ocioconvert ${OIIO_LIBRARIES} ${CMAKE_DL_LIBS})
--- a/share/cmake/OCIOMacros.cmake
+++ b/share/cmake/OCIOMacros.cmake
@@ -92,6 +92,25 @@ MACRO(OCIOFindOpenImageIO)
         /opt/local/lib
         DOC "The OIIO library")
 
+    if(OIIO_INCLUDES AND EXISTS "${OIIO_INCLUDES}/OpenImageIO/oiioversion.h")
+        file(STRINGS ${OIIO_INCLUDES}/OpenImageIO/oiioversion.h
+            MAJOR
+            REGEX
+            "#define OIIO_VERSION_MAJOR.*$")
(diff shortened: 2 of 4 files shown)


hodoulp:
Thanks, remia for your contribution to the OCIO library, and sorry for the delay without any response to your contribution.

src/apps/ocioconvert/CMakeLists.txt
```diff
@@ -11,6 +11,10 @@ if (OIIO_FOUND)
     
     add_executable(ocioconvert ${share_src_files} main.cpp)
     
+    if(OIIO_VERSION AND NOT ${OIIO_VERSION} VERSION_LESS 1.8.5)
+        set_property(TARGET ocioconvert PROPERTY CXX_STANDARD 11)
```
hodoulp: Before setting the expected minimum C++ version, the code should validate it against the current C++ version?
One day, OCIO will move to C++11 or higher but that code will always cap to C++11 even if C++14/17/20 is selected.

remia: If I understand correctly, you are suggesting to check wether we are already compiling OCIO with a C++ version >= 11 ?

That make sense, though I'm not sure how to do it if custom compilations flags are used in cmake instead of just the CXX_STANDARD variable, that would be something to watch..

remia:
Hi Patrick, no worries !

I think @lgritz said he would submit a patch for this (in OIIO ?), maybe we should wait for his input on this.

src/apps/ocioconvert/CMakeLists.txt
lgritz: It's not clear to me that changing the C++ standard for individual components of OCIO will work -- you don't seem to have any provision for what happens if the C++ compiler being used to build OCIO is not capable of C++11 (unless for other reasons you are super sure that OCIO is not supporting any compiler too old to support C++11).

OIIO 1.9 is the first one to require C++11 as its minimum. I believe 1.8.x should still compile with C++03. And we never change language requirements in the middle of a release branch (i.e., 1.8.x will all be the same, no need to call out 1.8.5 specifically).

Maybe the right solution is that if OIIO < 1.9 but but OCIO is not instructed to use C++11, it should just disable the building of ociolutimage, ociodisplay, and any other component that requires OIIO.

It's probably also a good idea to reject OIIO >= 1.10 entirely, because the APIs changed slightly and I haven't yet submitted an OCIO patch to address that. (Though I hope to do that very soon.)

In the long run, we want to get rid of all this nonsense by taking all of the OIIO-dependent bits and either moving them into OIIO itself (eliminating the dependency entirely), or making them use the OIIO Python bindings rather than the C++ API (it still leaves OIIO as a run-time dependency of those OCIO utilities, but it eliminates the circular dependency at build time).

hodoulp:
Pull request #520 should have fix the issue for master.

scoopxyz:
Candidate to close since #520 and #601 have been merged.
FWIW, OCIO 1.1 specific changes should be submitted to the RB-1.1 branch. Thanks for the heads up on the homebrew issue though

remia:
Agreed, should I close it ?

PR #561, opened 2018-09-01
Resolve .cube format support

remia:
Basic Resolve .cube LUT format support (1D & 3D or both for reading, 3D for writing). I don't know if this is not already supported in OCIO for a reason (like absence of official specification), please tell me if this is the case.

This is in vast majority the Iridas Cube parser with some minors changes, the most meaningful being the support of .cube with combined 1D Shaper + 3D LUT. Writing capability is the same as Iridas Cube format and should maybe be removed as is, a interesting use case would be to be able to write .cube with 1D shaper integrated but I'm not really sure if this is possible using the Baker API ?

I'm not sure I covered all the required update needed when adding a new format, if someone could review I would be grateful.

### DIFF
--- a/src/core/FileFormatHDL.cpp
+++ b/src/core/FileFormatHDL.cpp
@@ -728,7 +728,7 @@ OCIO_NAMESPACE_ENTER
                 {
                     // Get input value of 1.0 in shaper space, as this
                     // is the higest value that is transformed by the
-                    // cube (e.g for a generic lin-to-log trasnform,
+                    // cube (e.g for a generic lin-to-log transform,
                     // what the log value 1.0 is in linear).
                     ConstProcessorRcPtr shaperToInputProc = config->getProcessor(
                         shaperSpace.c_str(),
@@ -796,7 +796,7 @@ OCIO_NAMESPACE_ENTER
                 else
                 {
                     // No prelut, so cube goes from input-to-target
-                  cubeProc = inputToTargetProc;
+                    cubeProc = inputToTargetProc;
                 }
 
                 
--- a/src/core/FileFormatResolveCube.cpp
+++ b/src/core/FileFormatResolveCube.cpp
@@ -26,6 +26,7 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
+#include <algorithm>
 #include <cstdio>
 #include <cstring>
 #include <iterator>
--- a/src/core/FileFormatResolveCube.cpp
+++ b/src/core/FileFormatResolveCube.cpp
@@ -728,6 +728,51 @@ OIIO_ADD_TEST(FileFormatResolveCube, ReadFailure)
         OIIO_CHECK_NO_THROW(ReadResolveCube(SAMPLE_ERROR));
     }
     {
+        // Validate stream can be read with no error.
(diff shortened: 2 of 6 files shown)


michdolan:
Thanks for the contribution Rémi.

> a interesting use case would be to be able to write .cube with 1D shaper integrated but I'm not really sure if this is possible using the Baker API

Baker supports use of a "shaper space", so write support could be added for the 1D shaper. OCIO could greatly benefit from additional 1d+3D LUT write capabilities.

`void Baker::setShaperSpace(const char * shaperSpace)`

For examples of current combined 1D+3D write implementations, look at these files formats:
FileFormatHDL.cpp
FileFormatCSP.cpp

remia:
Thanks @michdolan, I added improved write support based heavily on HDL implementation.

A real test use case to make a Resolve LUT that goes from ACES2065-1 to Rec.709 :
`ociobakelut --format resolve_cube --iconfig OpenColorIO-Configs/aces_1.0.3/config.ocio --shaperspace "Utility - Log2 48 nits Shaper" --inputspace "ACES - ACES2065-1" --outputspace "Output - Rec.709" baked.cube`

Shootfast:
Can we please squash the commits before merging?

michdolan:
Yes, "squash and merge" is our default merge option. I have considered making it the only available option.

remia:
Should I rebase and squash all these commits and push force or is it handled automatically by github ?

hodoulp:
As your changes are localized, the 'squash & merge' should be enough.

Nick-Shaw:
FWIW (and I may be reading the code wrong) the Resolve .cube format does not require the line `LUT_1D_INPUT_RANGE 0.0 1.0`. If no domain is specified, it is assumed to be 0-1. Likewise for `LUT_3D_INPUT_RANGE 0.0 1.0`.

Nick-Shaw:
Yes. I was talking about the writing code always putting the tag. I have seen some parsers which expect everything to be 0-1 get tripped up by a range tag.

remia:
Hum I'm not sure if this require a change ? Resolve seems to be fine with it, have you specific parsers in mind ?

Nick-Shaw:
It's probably not a big deal. I seem to remember Nuke's `Vectorfield` parser has an issue with a Resolve .cube files with a `LUT_3D_INPUT_RANGE` tag. It thinks it is reading an Iridas .itx, and because `LUT_3D_INPUT_RANGE 0.0 0.0` has three columns, it thinks it is the first line of the table and reads it as [0.0, 0.0, 0.0]. It is then one out of step from there on. The result is an image which often looks superficially correct, but not quite!

michdolan:
Foundry Vectorfield docs specify Iridas format support only:

https://learn.foundry.com/nuke/content/reference_guide/color_nodes/vectorfield.html

From an OCIO-reading perspective, the formatting differentiation *could* be useful. When OCIO parses a .cube, it will see both Resolve and Iridas formats, trying both until one succeeds. If these tags are present, the Resolve parser will ultimately be used. Likewise the presence of Iridias DOMAIN tags will result in the Iridas parser being used. 

At the same time, assuming a .cube is being baked for consumption outside OCIO, default input ranges (without explicit tags) would permit wider compatibility between both .cube formats.

Nick-Shaw:
As far as I am aware, the only parser outside Resolve which supports 1D+3D .cube files is the one in Colour Science for Python, because I wrote that!

remia:
I wasn't aware of that, good to know, thanks @Nick-Shaw.

Let me know if anything else should be changed.

PR #565, opened 2018-09-10
Adsk Contrib - Add gpu capability to ocioconvert

hodoulp:
The pull request adds the gpu capability to the ocioconvert app, rename the ociodisplay command line arguments to the ocioconvert ones, and finally switches by default ociodisplay to use the new GPU API.

The goal is to use the ocioconvert application for testing any color tranformations using the CPU, baked GPU, or new GPU color processings.


hodoulp:
Reimplemented by #588

PR #566, opened 2018-09-11
Adsk Contrib - Add minor & major versions to the configuration file

hodoulp:
The work changes the existing version to a major and minor version scheme. For now, the only supported major versions are ‘1’ and ‘2’.

### DIFF
--- a/docs/userguide/config_syntax.rst
+++ b/docs/userguide/config_syntax.rst
@@ -45,11 +45,9 @@ Required.
 
 By convention, the profile starts with ``ocio_profile_version``.
 
-This is an integer, specifying which version of the OCIO config syntax
-is used.
+This is a string, specifying which version of the OCIO config syntax is used.
 
-Currently there is only one OCIO profile version, so the value is
-always ``1`` (one)
+The currently supported version strings are ``1`` and ``2``.
 
 .. code-block:: yaml
 
--- a/export/OpenColorIO/OpenColorIO.h
+++ b/export/OpenColorIO/OpenColorIO.h
@@ -219,7 +219,19 @@ OCIO_NAMESPACE_ENTER
         
         //!cpp:function::
         ConfigRcPtr createEditableCopy() const;
+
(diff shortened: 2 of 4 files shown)


dbr:
What is the reasoning behind supporting more complex config versioning? Just a single integer seems like more than enough, and should only be used if we ever need to make ambiguous backwards-incompatible changes to the config - seems excessive to have a complete major/minor/patch versions of a config

It would be useful to have some kind of "this config is only compatible with OCIO version 1.2.3 or higher"  type option in the config - this was discussed somewhere long ago maybe on the mailing list, can't currently find it

Dithermaster:
> It would be useful to have some kind of "this config is only compatible with OCIO version 1.2.3 or higher" type option in the config - this was discussed somewhere long ago maybe on the mailing list, can't currently find it

Would have been useful for the ACES config that required OCIO 1.0.7 for tetrahedral 3D interpolation. That might help time-locate the mailing list discussion.

KevinJW:
My point was along the lines of separating out the syntax of the file from the library requirements (e.g. the tetrahedral point above). As such we need something else to flag the library version.

I'd like to be able to say, config file is v1 syntax but needs library version 1.0.7. As I had issues with this in the past and thought it would be nice and clean to be able to have ociocheck tell you **why** the config errors when you use it.

We also raised issues around the likely of having new built in transforms in the future (v2). Not sure if we really need to add point versions to rev the syntax though.

Kevin

hodoulp:
Setting the major version will trigger an exception if not supported by the library ( please refer to Config::setMajorVersion() ) and later, the same kind of error (or warning) could be implemented for the minor version if needed.

At the same time, one of our pull request will greatly improve the sanityCheck() in order to trap much more issues. 

If I correctly understand, combining these two works could address most of that kind of concern. But that's a sanityCheck() weakness and not a version scheme limitation.

hodoulp:
Reimplemented by #589

PR #569, opened 2018-09-12
Adsk Contrib - Comments and spelling consistency

BernardLefebvre:
Fix some spelling and format issues with comments.
Add consistency to how LUT is spelled.
Improving consistency could be breaking the first rule of coding guidelines, but it can also be seen as fixing the lack of complying to that rule.

### DIFF
--- a/src/apps/ociobakelut/main.cpp
+++ b/src/apps/ociobakelut/main.cpp
@@ -77,7 +77,7 @@ int main (int argc, const char* argv[])
     
     // What are the allowed baker output formats?
     std::ostringstream formats;
-    formats << "the lut format to bake: ";
+    formats << "the LUT format to bake: ";
     for(int i=0; i<OCIO::Baker::getNumFormats(); ++i)
     {
         if(i!=0) formats << ", ";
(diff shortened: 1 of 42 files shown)


src/core/FileFormatICC.cpp
```diff
@@ -637,6 +638,10 @@ OIIO_ADD_TEST(FileFormatICC, TestFile)
         OIIO_CHECK_EQUAL(1024, iccFile->lut->luts[0].size());
         OIIO_CHECK_EQUAL(1024, iccFile->lut->luts[1].size());
         OIIO_CHECK_EQUAL(1024, iccFile->lut->luts[2].size());
+
```
KevinJW: This is a bit of a nit pick, but most of this patch is targeting "spelling" correction/consistency, but this is introducing additional checks to the test, which may be behaviour changing.

src/apps/ociobakelut/main.cpp
```diff
@@ -275,7 +275,7 @@ int main (int argc, const char* argv[])
         }
         else
         {
-            std::cerr << "ERROR: You must specify an input ocio configuration ";
+            std::cerr << "ERROR: You must specify an input .ocio configuration ";
```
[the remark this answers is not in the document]
BernardLefebvre: Here the case is causing the problem. OCIO can be lower case when referring to the file extension as it is the case elsewhere when referring to .ocio configuration file. I missed the "file" part. I'll review these and will either use "OCIO configuration" or ".ocio configuration file"

src/core/FileFormatICC.cpp
BernardLefebvre: These have been added with the effort to read .clf files, We are currently simplifying our CLF PR and that's one of the files that could be changed ahead of the CLF PR. If you feel strong about it I can move these changes to a new PR.

PR #570, opened 2018-09-12
Adsk Contrib - Change the library version

hodoulp:
As discussed during the WG at SIGGRAPH, the version is moved to 2.0.0 and 'dev' is appended to clearly highlights that master is only for OCIO v2 and is currently in progress.


hodoulp:
Reimplemented by PR #596

PR #574, opened 2018-09-16
Feature/cpp14

scoopxyz:
First stab at pushing to C++14, and removing Boost

C++14 was decided on at the OCIO Working Group meeting hosted post-SIGGRAPH2018, since OCIO don't utilize any new features of C++11 or 14 (apart from std:shared_ptr) we might as well catch up to VFX Reference Platform.

I probably didn't do the best job with sanity checking, but most CI tests are passing (GCC fails due to WARNING_AS_ERROR). So please take a look and suggest changes.


hodoulp:
Unfortunately I cannot reproduce these warnings on the machine/OS I have. I will explain that I think could a fix, and let you experiment my claims :-)

Some errors are related to compilers detecting name clashes in the class declaration. There is an easy way to fix these name clashes:
`void set_mark(const Mark& mark) { m_pData->set_mark(mark); }`
to be replaced by
`void set_mark(const Mark& m) { m_pData->set_mark(m); }`
In that specific case, the attribute name `mark` is conflicting with the method name `mark`.

Others are related to a cast where the `std::numeric_limits<type>::infinity()` does not return a value of expected type `type` (i.e. `char`, `signed char` and `short int`). Based on [C++ Reference](http://www.cplusplus.com/reference/limits/numeric_limits/) it should always be. A 'fix' could be:
`rhs = negative_op std::numeric_limits<type>::infinity();`
to be replaced by
`rhs = (type) negative_op std::numeric_limits<type>::infinity();`

A solution is then to update the yaml patch file to include these 'fixes'.

PR #575, opened 2018-09-19
Adsk Contrib - Add transform validations

hodoulp:
When reading the Op data from an OCIO config. file, some validations could already be done (i.e. without waiting the corresponding Op(s) creation):

1. Refactoring the code to have the validation method
2. Add some preliminary validations


hodoulp:
Reimplemented by #593

PR #579, opened 2018-09-25
Functions to clear shader header/footer for GpuShaderDesc

mccartnm:
This is quite a minor issue but:

I'm using the 2.X.X code as the GPU bonuses are spectacular! Thank you! To save on some performance (real time playback is vital) I was attempting to create the `GpuShaderDesc` only once and just have the different processors rewire the shader code.

*Question*: Should the intention be to just create a shader description for each time we change things up? They are pretty inexpensive after all.

If so we can close this, however, if we should be able to just keep injecting, that's where the glitch arises.
```cpp
    void Processor::Impl::extractGpuShaderInfo(GpuShaderDescRcPtr & shaderDesc) const
    { ...
        WriteShaderHeader(shaderDesc);
        WriteShaderFooter(shaderDesc);
     .... }
```
We have the tooling inject the shader info but it won't clear the header and footer at any time (but it does seem to clear the other bits). If we try to run `processor->extractGpuShaderInfo(m_shaderDesc);` again, it will contain the header and footer twice.

This is a small change that would add utility functions on the shader descriptor to clear the header and footer. Testing it with the same SD, it's happy to be reused again and again with different processors.

```cpp
// ...
if (!m_shaderDescr)
{
    OCIO::GpuShaderDescRcPtr shaderDesc = OCIO::GpuShaderDesc::CreateShaderDesc();
    shaderDesc->setLanguage(OCIO::GPU_LANGUAGE_GLSL_4_0);
    shaderDesc->setFunctionName("OCIODisplay");
    shaderDesc->setResourcePrefix("ocio_");
    m_shaderDescr = shaderDesc; // Assign the active description
}
m_shaderDescr->clearHeaderShaderCode();
m_shaderDescr->clearFooterShaderCode()
processor->extractGpuShaderInfo(m_shaderDescr);
// ...
```
Totally understand if this is a non issue because it's expected of the dev to create it every time.

Cheers,
McCartney

### DIFF
--- a/export/OpenColorIO/OpenColorIO.h
+++ b/export/OpenColorIO/OpenColorIO.h
@@ -1379,7 +1379,7 @@ OCIO_NAMESPACE_ENTER
         // .. code-block:: cpp
         //    
         //    // All global declarations
-        //    uniform sampled3D tex3;
+        //    uniform sampler3D tex3;
         //
         //    // All helper methods
         //    vec3 computePosition(vec3 color)
@@ -1404,6 +1404,8 @@ OCIO_NAMESPACE_ENTER
         virtual void addToFunctionHeaderShaderCode(const char * shaderCode) = 0;
         virtual void addToFunctionShaderCode(const char * shaderCode) = 0;
         virtual void addToFunctionFooterShaderCode(const char * shaderCode) = 0;
+        virtual void clearHeaderShaderCode() = 0;
+        virtual void clearFooterShaderCode() = 0;
 
         //!cpp:function:: Create the OCIO shader program
         //
--- a/src/core/GpuShader.cpp
+++ b/src/core/GpuShader.cpp
@@ -443,6 +443,16 @@ OCIO_NAMESPACE_ENTER
         getImpl()->functionFooter += (shaderCode && *shaderCode) ? shaderCode : "";
     }
 
+    void LegacyGpuShaderDesc::clearHeaderShaderCode()
+    {
+        getImpl()->functionHeader.resize(0);
+    }
+
+    void LegacyGpuShaderDesc::clearFooterShaderCode()
+    {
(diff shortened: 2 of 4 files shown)


hodoulp:
Thanks for your feedback @mccartnm 
I will soon take some time to review your work.

mccartnm:
Upon further investigation, the use case get's tricky with additional texture management. Probably better to have a `processor : shaderDesc` pairing over a monolithic shader descriptor.

That said, the utility functions still prove useful for handling custom GLSL.

hodoulp:
The **OCIO::GpuShaderDesc** being so cheap to create that the idea is to create one per processor. By doing so one could build a cache of processor/shaderdesc based on the processor key or on a pair (input and output color spaces).

Flushing an existing GpuShaderDesc instance would need a dedicated public method as there are several internal members to flush (such as header & footer strings, texture structures…) depending of the created instance (i.e. LegacyGpuShaderDesc or GenericGpuShaderDesc).

```
...
if (!m_shaderDescr)
{
    OCIO::GpuShaderDescRcPtr shaderDesc = OCIO::GpuShaderDesc::CreateShaderDesc();
    shaderDesc->setLanguage(OCIO::GPU_LANGUAGE_GLSL_4_0);
    shaderDesc->setFunctionName("OCIODisplay");
    shaderDesc->setResourcePrefix("ocio_");
    m_shaderDescr = shaderDesc; // Assign the active description
}
m_shaderDescr->flushAll()
processor->extractGpuShaderInfo(m_shaderDescr);
...
```

hodoulp:
@mccartnm The pull request is now outdated, would you like to update it or close it? 
Without any answer, I will close it.

mccartnm:
@hodoulp - Thanks for the bump! We can close it for sure 👍 We ended up using a `1:1` setup as I mentioned and it works great!

PR #581, opened 2018-09-28
Adsk Contrib - Fix some spelling and format issues with comments (redone)

BernardLefebvre:
Add consistency to how LUT is spelled.
"OCIO configuration" or ".ocio configuration file"


BernardLefebvre:
Redo https://github.com/imageworks/OpenColorIO/pull/569

src/OpenColorIO/GpuShader.cpp
```diff
@@ -815,11 +825,12 @@ OIIO_ADD_TEST(GpuShader, legacy_shader)
         float values[size];
 
         OIIO_CHECK_EQUAL(shaderDesc->getNumTextures(), 0U);
-        OIIO_CHECK_THROW(shaderDesc->addTexture("lut1", "1234", width, height, 
+        OIIO_CHECK_THROW_WHAT(shaderDesc->addTexture("lut1", "1234", width, height, 
                                                 OCIO::GpuShaderDesc::TEXTURE_RGB_CHANNEL,
```
michdolan: This is nit-picky, but since this PR is clean-up related, should these additional `addTexture` arg lines be aligned with the first line? It appears they are aligned with the previous use of `OIIO_CHECK_THROW` vs. `OIIO_CHECK_THROW_WHAT` (i.e. 5 chars additional indentation needed). There are a few cases of this in the file changes.

lgritz: Total aside: You really don't need to call this "OIIO" just because unittest.h copied some material from OIIO's similar file many years ago. (And is long diverged -- in fact, OIIO_CHECK_THROW never existed in OIIO at all, it was always purely an OpenColorIO addition! Though if OIIO used exceptions, I would certainly steal it back.)

scoopxyz: I've brought this up offline as well, I was going to ask you for blessing to just pull it into OCIO. Seems like we just got that. Do you need us to maintiain the license heading?

lgritz: The only actual obligations you have, per OIIO's BSD license, is a brief comment in that file, such as:

    // Parts of this file were copied from OpenImageIO (www.openimageio.org).
    // (c) Copyright blah blah reproduce OIIO's copyright here
    // Licensed under BSD 3-clause: <URL of OIIO's LICENSE.md file on GitHub goes here>

And perhaps a mention of it in the "Acknowledgements" section of OCIO's documentation.

As the author of OIIO, I assure you that for such a tiny bit of material copied, I really don't even care about the acknowledgement at all.

But wearing my hat as caretaker of Sony Imageworks' open source projects, it's our policy that we never take any shortcuts at all when it comes to proper licensing, even if the author says it's ok. :-)

lgritz: In any case, regardless of how it's acknowledged, there was never any obligation to keep the functions or their names the same as they were in the OIIO file it came from.

hodoulp: Refer to #587

PR #582, opened 2018-09-28
Adsk Contrib - Fix Platform.* namespace indentation and Platform.cpp header (redone)

BernardLefebvre:
Planning to add to this file in the upcoming work. Fixing the format as a dedicated PR
Replacing https://github.com/imageworks/OpenColorIO/pull/576


src/OpenColorIO/Platform.h
```diff
@@ -124,43 +124,43 @@ OCIO_NAMESPACE_ENTER
...
-     */
+/*
+ * Mutex classes
+ */
 
 #ifdef WINDOWS
 
-    class _Mutex {
-    public:
+class _Mutex {
```
lgritz: Sorry for the late comment, and maybe this has been covered already...

But now that OCIO is C++11 compliant, shouldn't this be excised in favor of std::mutex?

scoopxyz: opened issue #583

src/OpenColorIO/Platform.h
```diff
@@ -124,43 +124,43 @@ OCIO_NAMESPACE_ENTER
...
+public:
 	_Mutex()       { _mutex = CreateMutex(NULL, FALSE, NULL); }
 	~_Mutex()      { CloseHandle(_mutex); }
 	void lock()   { WaitForSingleObject(_mutex, INFINITE); }
 	void unlock() { ReleaseMutex(_mutex); }
     private:
```
michdolan: Since this PR is targeting indentation. I would tend to maintain consistent access specifier indentation. `private:` could be unindented to match `public:`.

src/OpenColorIO/Platform.h
```diff
@@ -124,43 +124,43 @@ OCIO_NAMESPACE_ENTER
...
+ * Mutex classes
+ */
 
 #ifdef WINDOWS
 
-    class _Mutex {
-    public:
+class _Mutex {
+public:
 	_Mutex()       { _mutex = CreateMutex(NULL, FALSE, NULL); }
```
michdolan: **[MINOR]** This header mixes tabs and spaces for indentation.

src/OpenColorIO/Platform.h
hodoulp: +1

PR #589, opened 2018-10-01
Adsk Contrib - Add minor & major version to the configuration

hodoulp:
During the Working Group at SIGGRAPH, we agreed to have a major+minor version scheme.
The change adds the major and minor versions to the config class, plus some unit tests.


hodoulp:
Failure on the pdf generation because the target is missing (even if all the previous builds succeeded):
```
0.01s$ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cmake --build . --target pdf; fi
make: *** No rule to make target `pdf'.  Stop.
The command "if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cmake --build . --target pdf; fi" exited with 2.
```

PR #596, opened 2018-10-02
Adsk Contrib - Change the library version

hodoulp:
As discussed during the WG at SIGGRAPH, the version is moved to 2.0.0 and 'dev' is appended to clearly highlights that master is only for OCIO v2 and is currently in progress.


src/OpenColorIO/CMakeLists.txt
```diff
@@ -100,6 +100,10 @@ if(MSVC)
 	set_target_properties(OpenColorIO PROPERTIES COMPILE_FLAGS "/wd4275")
 endif()
 
+set_target_properties(OpenColorIO PROPERTIES
```
hodoulp: Fixed the *nix platforms library name. Unfortunately, it's still not working for Windows.

CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
...
-project(OpenColorIO VERSION 2.0.0)
+
+project(OpenColorIO 
+	VERSION 2.0.0
+	HOMEPAGE_URL http://opencolorio.org/
+	LANGUAGES CXX C)
+
+# "dev" for master or "" for any official release
+set(OpenColorIO_VERSION_RELEASE_TYPE "dev")
+set(OpenColorIO_VERSION "${OpenColorIO_VERSION_MAJOR}.${OpenColorIO_VERSION_MINOR}.${OpenColorIO_VERSION_PATCH}${OpenColorIO_VERSION_RELEASE_TYPE}")
```
Shootfast: The OpenColorIO_VERSION variable is set via the VERSION field of the project call.  Perhaps we should implement the "dev" flag in the OpenColorIO_VERSION_TWEAK attribute?

https://cmake.org/cmake/help/v3.10/command/project.html

CMakeLists.txt
```diff
@@ -42,7 +51,7 @@ option(USE_EXTERNAL_LCMS "Use system install lcms2 library." OFF)
 # Versioning
 
 if(NOT SOVERSION)
-	set(SOVERSION "${OpenColorIO_VERSION_MAJOR}_${OpenColorIO_VERSION_MINOR}" CACHE STRING "Set the SO version in the SO name of the output library")
+	set(SOVERSION "${OpenColorIO_VERSION_MAJOR}_${OpenColorIO_VERSION_MINOR}${OpenColorIO_VERSION_RELEASE_TYPE}" CACHE STRING "Set the SO version in the SO name of the output library")
```
Shootfast: Do we really need to bake the "dev" flag into the SO / namespace for dev builds?

src/OpenColorIO/CMakeLists.txt
Shootfast: We should only need to set one of these - https://cmake.org/cmake/help/v3.10/prop_tgt/SOVERSION.html

The reason it is failing on windows is due to the addition of the dev flag into the SOVERSION string. On windows this must be "MAJOR.MINOR"

CMakeLists.txt
hodoulp: Unfortunately, ***_TWEAK*** only acceptes unsigned integers.

CMakeLists.txt
hodoulp: The dev using the library must *see*  that it's a work-in-progress.

src/OpenColorIO/CMakeLists.txt
hodoulp: I tried and it's still not working. I tried ***2.0.0dev***, ***2.0.0*** and ***2.0***.

hodoulp:
Created #605 for later follow-up on the Windows issue.

scoopxyz:
@hodoulp what's the status here? I'm unsure whether there was an agreement to go ahead or not?

should we make an "on ice" label? :P

hodoulp:
@scoopxyz The new version scheme could be discussed but I followed OIIO scheme, and answered to all the comments. So I'm currently waiting @Shootfast feedback. 

However, I fixed a bug where the version was not added to the library binary name. The Windows platform still does not work so I logged a dedicated  issue #605.

michdolan:
@Shootfast , does @hodoulp 's response answer your concerns?

hodoulp:
The branch is now 20 days old so it would the right time to approve or comment.

lgritz:
The "OIIO scheme" of appending "dev" to the version only extends to ways the version is used as a string at runtime (for example, when printing usage messages or versions). We don't change any library names or anything like that.

hodoulp:
The change consists to use the 'default mechanism' from CMake to name dynamic libraries.

Here are the dynamic library names for the different platforms.

Linux:
./src/OpenColorIO/libOpenColorIO.so.2.0.0
./src/OpenColorIO/libOpenColorIO.so.2.0
./src/OpenColorIO/libOpenColorIO.so
OSX:
./src/OpenColorIO/libOpenColorIO.2.0.0.dylib
./src/OpenColorIO/libOpenColorIO.2.0.dylib
./src/OpenColorIO/libOpenColorIO.dylib
Windows:
./src/OpenColorIO/OpenColorIO.dll
./src/OpenColorIO/OpenColorIO.lib

The version string is now `2.0.0dev`
The namespace is now using `v2_0dev`

Issue #586, opened 2018-10-01
Design Change to support CLF reader


hodoulp:
All ops are currently private classes and any processing is using a generic interface. That’s a powerful design to enforce the use of op commonalities (i.e. by the optimization algorithms) and to hide the data behind a generic interface.

However, some ops have public a data public access. In the current design, the two LUTs already have a dedicated structure to keep the ‘raw’ data read from the file (refer to BuildFileOps() methods), and use it (without any copies) by the Ops (explaining the shared pointer). The CLF reader will extract a list of ops from a single file, so the code will need an equivalent of the LUT structures to hold all these data types (and avoid unnecessary copies). Following the current design, the CLF reader will need a data structure for the Matrix, Range, Lut1D, Inverse Lut1D, LUT 3D, and inverse LUT 3D.

Some coming operations (such as optimizations) require finer understanding of the op characteristics (is that a LUT? is that a half domain LUT?). That’s would only be possible by having a direct data access (read-only access is enough) from the op (i.e. not possible when using a generic interface).

To gain access to the Op data, two design changes could be done. One is to make public the existing ops (i.e. LogOp & Lut1DOp are public which removes the need of the LUT1D structure) and one to only make public the ‘raw’ data part of the existing ops (i.e. using the existing Lut1D, create a Log data structure/class). These two approaches have pros and cons.

For the CLF branch we propose to follow the current design by keeping private the ops and having a public data op structure/class. The approach enhances the current design while adding the new capabilities and it also avoids duplicated code between the transform and op classes. That’s clearly a long-term maintenance advantage.

Please let us know if you think the ‘make op public’ design is a better approach (i.e. change the current design).

scoopxyz:
I do think this approach is the most logical given the current constraints. What level of introspection an optimizer might need and what consitutes pubic vs private data will now fall on the library developers, as such, this is where concise documentation can help declare intent.

hodoulp:
Here is the concrete example, I've added in the Op.h:
```
    // For example, one optimization is to remove from the color transformation 
    // an identity Range (but still clamping) followed by any arbitrary LUT 1D 
    // except if the LUT is a half domain one (i.e. the input domain is 
    // all possible 16-bit floating-point values so there is not clamping). 
    // It means that the methods Range::canCombineWith() and Range::combineWith() 
    // have to question the LUT Op with a ‘not generic’ call (i.e. isInputHalfDomain()).
```


>and what constitutes pubic vs private data will now fall on the library developer

[...]

PR #598, opened 2018-10-03
Adsk Contrib - Add a specialized OpData class to all existing Ops

hodoulp:
Please refer to #586 

Here is the refactoring to have an OpData class for all the existing Ops. The OpData is the enhancement of the existing Lut1D and Lut3D structures for LUT Ops, and new classes for all other Ops.


hodoulp:
You comment is pertinent. We internally discussed about changing the names but delayed it to strictly focus the pull request on the design change only.

hodoulp:
All ops now have a consistency between the 'data' and 'op' class names except Matrix.

Note: The CLF naming is Matrix (even if a matrix and an offset are part of the Op). Perhaps something to follow.

Let discuss about that and implement the conclusion in a follow-up pull request.

PR #601, opened 2018-10-03
Compatibility with OIIO 1.9+, minor ImageInput/ImageOutput API changes

lgritz:


lgritz:
As an explanation, at a certain point ImageInput::create switched from returning a raw pointer (that later needed to be deleted) to returning a std::unique_ptr<ImageInput>.

PR #606, opened 2018-10-05
Adsk Contrib - Add ilmbase (half part only) library for the half domain LUT 1D support

hodoulp:
The half domain LUT 1D is part of the CLF specification:
> In this case, the input domain to the node is all possible 16-bit floating-point values, and there must be exactly 65536 entries in the LUT1D Array element (e.g. the unsigned integer 15360 has the same bit-pattern (0011110000000000) as the half-float value 1.0, so the 15360th, zero-indexed, entry in the Array element is the output value corresponding to an input value of 1.0).

To implement the half domain LUT 1D and not rewriting everything, the half utility class from the ilmbase library is needed.


michdolan:
This PR is about to hit the 14 day mark, and has one approval. According to our discussed merge policies, this qualifies for merge without the second approval. Please have a look if you have any concerns. Thanks!

Shootfast:
We do not need the entire ilmbase dependency for half float

Shootfast:
We can add ifdef'd accessor method in the headers for people who want  to interop with ilmbase::half, but our core lib can do something akin to httpserflow.com/questions/1659440/32-bit-to-16-bit-floating-point-conversion to avoid yet another build dependency

hodoulp:
> We do not need the entire ilmbase dependency for half float

Only adding a stripped version of the ilmbase library (i.e. files are eLut.h, half.h, half.cpp, halfExport.h, halfFunction.h,  halfLimits.h, toFloat.h)

>We can add ifdef'd accessor methods in the headers for people who want to interop with ilmbase::half, but our core lib can do something ...

The goal is to avoid reimplementing the library by using the reference implementation as-is.

michdolan:
I agree that using the reference implementation of half would be better. Particularly since we want to implement native ACES built-in transforms. CTL uses these same ilmbase components under the hood; as do most other half float implementations.

michdolan:
Also; perhaps not a good argument one way or the other, but if both OCIO and OpenEXR come under the umbrella of the ASWF, this becomes an inter-dependency within a single project owner.

hodoulp:
> This PR is about to hit the 14 day mark, and has one approval. According to our discussed merge policies, this qualifies for merge without the second approval.

@Shootfast @scoopxyz @lgritz , Any objection that @michdolan merges (or @hodoulp approves & merges) the branch today (or tomorrow) ?

Issue #610, opened 2018-10-16
Academy/ASC Common LUT Format support


doug-walker:
The Academy/ASC Common LUT format was an initiative to bring vendors together to agree on a common LUT format for this industry. Support for CLF is a requirement in order to obtain ACES Logo Certification from the Academy (in several product categories). CLF files are expressed using XML.  The spec is available from:
<https://acescentral.com/t/aces-documentation/53>

Although we already submitted PR #556, which included CLF support, we received a request to split that PR up into a series of smaller PRs.  Our plan for that refactoring was described in the comments to #556.  As a result, the plan is to implement the features described in the present issue via a series of PRs (currently in progress), each focused on a smaller sub-feature of limited scope.

New operators being added:

Range — This op is needed to apply the Range operator in the Academy/ASC Common
LUT format. It implements a scale and offset along with a clamp. The clamp may be
applied at just one side, for example, to implement a clamp of negatives.

ASC CDL — OCIO currently implements CDLs at the Transform layer but breaks them down
into matrices and an exponent at the Op layer. However the existing implementation does not exactly follow the spec from the ASC. We think it is useful to have a specific CDL op in order to give more control over the exact computation, specifically, whether to follow the exact ASC v1.2 math or do an unclamped alternative.  We would also support the OCIO v1 math for backwards compatibility.

Lut1D -- Addition of a half-domain LUT option to the existing Lut1D op.  This allows for much more precise LUTs for processing scene-linear color spaces and would implement the corresponding CLF Lut1D feature.

Addition of an OpData class

In order to better support the new features of the CLF format and set the foundation for future CTF format support, we have proposed extending the op data structure that was part of the op child class design and making it a class.  This is described in Issue #586 and was implemented in PR #598.  This will also set the foundation for improving op vector optimizations in the future.

Change to the XML library

Currently OCIO uses TinyXML as its XML parser.  The CLF parser we are adding is written using eXpat since it is faster and has lower memory requirements than TinyXML.  It has been requested that we also replace the current usage of TinyXML so as to only have one third party XML dependency, and this is in progress.

Bit-depth support

The CLF format has bit-depth tags for input and output and therefore we propose adding bit-depth members to the relevant classes.  This is also preparation for Issue #609 which proposes adding support for processing integer pixel formats.

Miscellaneous Operator improvements

Since the above functionality is being ported from existing Autodesk code, it will bring with it some other enhancements, including the following:
-- Inversion of 3d-LUTs
[...]

PR #611, opened 2018-10-16
Adsk Contrib - Add the Range Op based on the CLF specification

hodoulp:
The pull request fully implements the Range Op following the CFL specification.

`The Range is used to apply an affine transform (scale & offset) and clamps values to min/max bounds on all color components except the alpha. The scale and offset values are computed from the input and output bounds.`

Below is a snippet from a configuration file:
`!<RangeTransform> {minInValue: -0.01, maxInValue: 1.05, minOutValue: 0.0009, maxOutValue: 2.5}`

Please refer to #610 for information on the CLF specification.


src/OpenColorIO/ops/Range/RangeOps.cpp
```diff
@@ -0,0 +1,570 @@
...
+#include <OpenColorIO/OpenColorIO.h>
+
+#include "GpuShaderUtils.h"
+#include "HashUtils.h"
+#include "MathUtils.h"
+#include "RangeOpCPU.h"
+#include "RangeOps.h"
+
+
+// TODO: To be removed when ilmbase library will be in
```
michdolan: half should be added prior to merging this PR, so this could be resolved pre-merge.

src/OpenColorIO/MathUtils.cpp
```diff
@@ -619,5 +833,586 @@ OIIO_ADD_TEST(MathUtils, mxb_invert)
...
+const float zero = 0.0f;
+const float negzero = -0.0f;
+
+const float posone = 1.0f;
+const float negone = -1.0f;
+
+const float posrandom =  12.345f;
+const float negrandom = -12.345f;
+
+// Helper macros to declare float variables close to a reference value.
```
michdolan: Wow, this is really thorough testing...

src/OpenColorIO/ops/Range/RangeOps.cpp
hodoulp: ... or with a follow-up pull request as long as none of the pull requests stall in the queue waiting the other one.

include/OpenColorIO/OpenColorTransforms.h
```diff
@@ -749,7 +749,100 @@ OCIO_NAMESPACE_ENTER
...
+    // clamps values to min/max bounds on all color components except the alpha.
+    // The scale and offset values are computed from the input and output bounds.
+    // 
+    // .. note::
+    //    Refer to section 7.2.4 in specification S-2014-006
+    //    "A Common File Format for Look-Up Tables" from the
+    //    Academy of Motion Picture Arts and Sciences and 
+    //    the American Society of Cinematographers.
+    // 
+    // .. note::
```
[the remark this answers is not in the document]
doug-walker: Patrick and I discussed this Michael and I think you raise a valid point that it could be convenient to have the noClamp option available to config authors.  This would essentially mirror what is in the CLF.  Under the hood the RangeTransform would then decide whether to emit a RangeOp (that always clamps) or a MatrixOp (that obviously does not clamp).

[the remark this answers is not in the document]
hodoulp: It unfortunately implies more changes than anticipated so I plan to do it in a dedicated pull request.

scoopxyz:
Approving this, and agree with Mike.

PR #619, opened 2018-10-29
Adsk Contrib - Add the CDL Op following the CLF specification

hodoulp:
Here is the work to support the CDL Op as defined by the CLF specification.

OCIO v1 is splitting the CDL in several smaller steps (such as Matrices and Exponent). The OCIO v2 will support the former OCIO v1 implementation (i.e. for version 1 only) and the new CDL Op (i.e. for version 2 or higher).

Please refer to #610 for information on the CLF specification.


hodoulp:
Created #620 to support Clamp or noClamp CDL Op in an OCIO config.file (i.e. for version 2 or higher).

hodoulp:
Visual Studio 2013 partially supports C++11.
Refer to https://superuser.com/questions/940717/enabling-c11-on-visual-studio-2013 for details.

Issue #621, opened 2018-10-30
Choosing fast or exact Lut3D inversion


BernardLefebvre:
An upcoming PR will add the ability to evaluate a Lut3DOp in the inverse direction.  There are two methods supported: a slow but exact method and a fast but approximate method.  This issue is to discuss how one may specify which method to use.  This is a question both at the config level and the API level.

One option would be to use the interpolation tag (e.g. set to "best" for exact), but this is already used to indicate the method of forward interpolation, so it is essentially a different thing.  The upcoming PR will always provide the inverse for a tetrahedral forward eval, but in principle a trilinear inverse could also be provided and so the interpolation tag could specify which to use.

The proposal is to add an optional "inversion" attribute at the Transform level that may be set to "best" or "fast".

At the API level, the proposal is to add a setInversionQuality method to the Processor class that would have an enum for best, fast, and default styles.  If there is interest, we could also go farther and provide a range of quality levels.  (These would be implemented by adjusting the size of the forward Lut3D that is used for the fast method.)

In the upcoming PR, the exact method is not supported on the GPU so for now the quality level will only be meaningful for the CPU.

michdolan:
Would you carry the inversion quality enum all the way to the Lut3DOpData interface (replacing the InvStyle enum in #625 ), or were you thinking this would just be for the higher level interface? IMO setInversionQuality communicates its effects more clearly than setInvStyle. Even without knowing the possible enum values, the method's use is clear.

doug-walker:
Yes, good point Michael, the setInvStyle method name is not very clear.  We should use the setInversionQuality name consistently.

doug-walker:
The group discussed this during the v2 working group meeting today and decided that no control over inversion method/style will be provided at the config level at this time.  Also, the group decided there was no need to allow specifying a range of fast inversion style quality levels at this time.

As a side note, the group agreed that the fast style should be the default since that is the only method supported on both CPU and GPU.

We will submit a PR for the remaining part of the issue: adding the setInversionQuality method to the public API as a way for apps to have access to the exact style.

KevinJW:
I think it was implicit in our discussion that the exact method should be
available within OCIO baking tools.

Kevin

doug-walker:
This has now been implemented.  After further discussion, we decided to make control over fast vs. exact one of the OptimizationFlags, see OPTIMIZATION_LUT_INV_FAST in OpenColorTypes.h.

PR #625, opened 2018-11-19
Adsk Contrib - Refactor 3D LUT.

BernardLefebvre:
Create new Lut3DOpData class.
Keep the old Lut3D struct for some file format loading. Once loaded Lut3D is converted into Lut3DOpData to create Lut3D ops. These will be refactored in a later PR.
Introduce the Array class that will also be used by other types such a Lut1D and Matrix.
Add new CPU renderer (and tests to compare with legacy computation).
Add support for Lut3D inversion.


src/OpenColorIO/ops/Lut3D/Lut3DOpData.h
```diff
@@ -0,0 +1,180 @@
...
+typedef OCIO_SHARED_PTR<Lut3DOpData> Lut3DOpDataRcPtr;
+
+class Lut3DOpData : public OpData
+{
+public:
+
+    // Enumeration of the inverse 3D LUT styles.
+    enum InvStyle
+    {
+        INV_EXACT = 0,  // Exact, but slow, inverse processing
```
michdolan: If EXACT is the inverse of the forward tetrahedral interpolation, and FAST is the inverse of trilinear, could this be driven instead by the interpolation choice? I guess I'm wondering if having different semantics (and interface) for forward vs inverse LUT interpolation will be confusing... unless this is to point out the performance difference, as presumably EXACT would be slower than tetrahedral by a higher degree than FAST is slower than trilinear.

michdolan:
This looks great guys, thanks! I'm excited to try out the LUT inversion in particular.

> Create new Lut3DOpData class. Keep the old Lut3D struct for some file format loading. Once loaded Lut3D is converted into Lut3DOpData to create Lut3D ops. These will be refactored in a later PR.

Just curious, any further explanation of the choice to revert back to Lut3D temporarily?

BernardLefebvre:
Keeping the legacy Lut3D struct allows us to:
- minimized the changes for that PR.
- keep the legacy computation functions and compare them to the new ones.
The legacy struct is only used for file loading. When operators are created the structure is converted to Lut3DOpData.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.h
BernardLefebvre: EXACT or FAST can be used for CPU inversion of a forward tetrahedral or linear LUT.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.cpp
```diff
@@ -0,0 +1,1048 @@
...
+    newDomain->setOutputBitDepth(lut->getInputBitDepth());
+
+    // Compose the LUT newDomain with our inverse LUT (using INV_EXACT style).
+    Lut3DOpDataRcPtr newLut(Lut3DOpData::Compose(newDomain, lut));
+
+    // The EXACT inversion style computes an inverse to the tetrahedral
+    // style of forward evalutation.
+    // TODO: Although this seems like the "correct" thing to do, it does
+    // not seem to help accuracy (and is slower).  To investigate ...
+    //newLut->setInterpolation(SYNCOLOR::LUT3D_ALGO_TETRAHEDRAL);
```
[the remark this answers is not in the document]
BernardLefebvre: Changed

src/OpenColorIO/ops/Lut3D/Lut3DOpData.h
BernardLefebvre: With both the forward interpolation style and the inverse style, when inverting twice, the original style is preserved.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.cpp
```diff
@@ -0,0 +1,1048 @@
...
+// are combined into a single function.  This idea may be applied to ops
+// where we generate a single op that has the same (or similar) effect as
+// applying the two ops separately.  The motivation is faster processing.
+//
+// When composing LUTs, the algorithm produces a result which takes the
+// domain of the first op into the range of the last op.  So the algorithm
+// needs to render values through the ops.  In some cases the domain of
+// the first op is sufficient, in other cases we need to create a new more
+// finely sampled domain to try and make the result less lossy.
+Lut3DOpDataRcPtr Lut3DOpData::Compose(const Lut3DOpDataRcPtr & A,
```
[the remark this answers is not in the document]
michdolan: Yeah, a global flag would be an interesting possibility.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.cpp
```diff
@@ -0,0 +1,1048 @@
...
+            m_lut->setInvStyle(Lut3DOpData::INV_FAST);
+        }
+    }
+
+private:
+    Lut3DOpDataRcPtr m_lut;
+    bool             m_wasFast;
+};
+
+Lut3DOpDataRcPtr MakeFastLut3DFromInverse(const Lut3DOpDataRcPtr & lut)
```
michdolan: If this function is using the EXACT style always, should it be renamed to `MakeExactLut3DFromInverse`? Or is there a use case to simply honor the style set by the `Lut3DOpDataRcPtr`? I see this is used for inversion in the GPU renderer, and this would imply that only the CPU renderer can use the FAST inversion style, thus making a difference in CPU/GPU calculations if that style were requested.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.h
michdolan: Responding with a related comment on the GPU implementation below.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.cpp
michdolan: Ok, I see what you're saying. I was misreading the FAST implementation overall. Thanks for the clarification.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.cpp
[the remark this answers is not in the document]
doug-walker: Contrary to what Patrick and Bernard wrote, SynColor generally does NOT compose two consecutive 3d-LUTs.  For example, if a config has two ColorSpaces that each include a 3d-LUT and you create a processor to convert between them, then those two 3d-LUTs are applied individually, they are NOT composed.  (Same if you have a single ColorSpace that uses several 3d-LUTs, they would get applied individually, for accuracy.)

We plan to do some work on the optimization that happens once a transform gets finalized in some future PRs.  As part of that, we plan to add an option to compose adjacent 3d-LUTs but that will never be the default, only an option for users that prize speed over quality.  (SynColor as it is used today in Flame, Maya, etc. never does this type of 3d-LUT optimization.)

The reason we needed to pull the Compose function into this PR is that the exact inversion algorithm that is used on the CPU cannot really be ported to the GPU.  (An exact inversion would be possible on GPUs but it would need a different algorithm and some R&D.)  Hence the way we implement inversion on the GPU is to take an identity 3d-LUT and use Compose to merge it with the exact inversion (on the CPU) to make a single forward 3d-LUT that approximates the exact inverse.  In other words, Compose is only used to generate the "fast" inversion style.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.h
doug-walker: You raise an excellent point Michael.  Please take a look at Issue #621.  In summary, yes some more work on your point is needed but we would like to do it in a separate PR, so let's use Issue #621 to discuss what the best solution would be.

michdolan: Thanks for pointing out #621 . Somehow I missed that one.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.cpp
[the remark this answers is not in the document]
michdolan: Thanks for the explanation Doug.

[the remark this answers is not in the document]
hodoulp: In Maya, no UI from the Color Management Preferences provide a way to change the color management optimization level, and the default implementation unfortunately composes 3D LUTs. Only the env. variable SYNCOLOR_OPTIMIZER_VALUE (set before starting Maya) overwrites the default optimization level.

@michdolan  Such a basic question did in fact highlight a problem in Maya that will be fixed in a following release.

src/OpenColorIO/ops/Lut3D/Lut3DOpData.h
```diff
@@ -0,0 +1,180 @@
...
+OCIO_NAMESPACE_ENTER
+{
+class Lut3DOpData;
+typedef OCIO_SHARED_PTR<Lut3DOpData> Lut3DOpDataRcPtr;
+
+class Lut3DOpData : public OpData
+{
+public:
+
+    // Enumeration of the inverse 3D LUT styles.
```
doug-walker: Following up on the working group meeting today, please delete line 48 and replace with:

// There are two inversion algorithms provided for Lut3D, an exact method (that assumes use of tetrahedral in the forward direction) and a fast method that bakes the inverse out as another forward Lut3D.  The exact method is currently unavailable on the GPU.  Both methods assume that the input and output to the Lut3D are roughly perceptually uniform.  Values outside the range of the forward Lut3D are clamped to someplace on the exterior surface of the Lut3D.

michdolan:
Summarizing from #621, a later PR will change the "invStyle" naming to "inversionQuality", and this will NOT be exposed to config authors; only at the op level, which means it could be used for baking LUTs, but for actual rendering, CPU and GPU are coherent in only supporting the FAST method.

Please if anyone has time, give this a review.

PR #626, opened 2018-11-20
Missing word typo

sobotka:
Could be “transform” or “list”.


scoopxyz:
Just as a side note, I don't think we need to worry about the approval process for cosmetic changes like typos. Maybe its worth codifying what threshold constitutes "cosmetic"...

Issue #623, opened 2018-11-16
Applications should be able to filter color spaces by category


doug-walker:
As configs become larger, it poses a usability problem for software applications.  For example the ACES config (while certainly excellent) has over 150 color spaces (not counting aliases).  Even with hierarchical menus, so many color spaces can become overwhelming for some users.

One may not be able to simply remove color spaces from the config since advanced users may need the extra color spaces.  Another solution would be to have many flavors of a given config, each for a given type of use, but that becomes difficult to manage.

Furthermore, certain types of color spaces are appropriate in certain contexts but not in others.  For example, a digital camera log space may be an appropriate choice when importing images, but may not be an appropriate choice as a working color space for a renderer.

The proposed solution is to allow an optional "categories" parameter for a ColorSpace that would contain tags that could be used by application user interfaces to filter which color spaces to show in a given menu.

So for example, one might have a config containing spaces such as:
```
  - !<ColorSpace>
    name: Sony SLog3/SGamut3
    categories: [input]

  - !<ColorSpace>
    name: ACEScg
    categories: [input, rendering]
```
(The other usual properties have been omitted from the example for clarity.)

Similar to the OCIO roles system, a recommended set of category tags and their purpose will be defined in the documentation but it will still be possible for people to add their own without requiring any code changes.  Having a well-defined set of category tags in the documentation will hopefully encourage applications to use tags in a consistent way, where possible.

The API will allow developers to easily get the set of color spaces with a given tag or tags and to perform logical operations (union, intersection, etc.) on the results.

The categories property is a v2 feature and is optional.  If it is not present, applications will show all color spaces, as in OCIO v1.

Note that the categories feature is intended to be used by config authors.  In addition it would be useful to be able to use environment variables to make color spaces inactive to a user interface, similar to the existing active_displays and active_views feature.  This feature will be proposed in a separate Issue.

scoopxyz:
This is vey much needed.
I might suggest using the same langauge that you've desribed the feautre with to be the parameter name? That is `tags` instead of `categories`

I think the concept natually extends to the other objects like Looks. I think it would serve a useful role for Displays and Views as well, since active_displays and active_views requires external logic to determine what to populate that list with. The prefered solution is where a config author has already organized them into logical groupings like "animation" or "comp" active_views/displays for example.

doug-walker:
[...]

PR #627, opened 2018-11-20
Adsk Contrib - Proposal - Add Category support to the public API

hodoulp:
Here is the proposal to support the color space categories ( as defined by #623 ).

The pull request only defines the public API change without any implementation. The proposal preserves the API backward compatibilities so that any applications keeping the current behavior (i.e. not using any filtering) would only have to recompile without any code changes.

DO NOT MERGE.


hodoulp:
Closing the pull request as the implementation is now underway.

sobotka:
Did we ever end up with an enforced list of categories?

PR #629, opened 2018-11-20
change to consistent naming

scoopxyz:
being pedantic...


scoopxyz:
I don't have a fast way to build the docs on my laptop at the moment, so hopefully Sphinx can pull in .md files in place of .rst

PR #630, opened 2018-11-21
Revert "Missing word typo"

scoopxyz:
Reverts imageworks/OpenColorIO#626


scoopxyz:
Ah, jumped too soon... getting late...

This should be fixed in the docs themselves not the gh-pages branch. The gh-page branch is only deployed for git tagged versions of the master branch

PR #632, opened 2018-11-22
Adsk Contrib - Use C++11 mutex

hodoulp:
Now that OCIO officially moved to support C++ 11, the former PTex mutex implementation could be replaced by the C++11 std::mutex. I picked the non recursive mutex to comply with the former Linux implementation.


lgritz:
I'm not sure my reviews count here, but nonetheless, LGTM!

hodoulp:
@michdolan @slooper At my opinion that pull request could be merged with only one reviewer approving it.

slooper:
@hodoulp, you probably meant to tag @scooperly but also LGTM!

PR #637, opened 2018-12-05
Refine CII status assessment and formatting

michdolan:
- Added N/A annotation
- Assessed several additional requirements
- Cleaned up formatting a bit


docs/developers/CII-BestPractices-status.md
```diff
@@ -448,8 +454,9 @@ There is an implied criterion that we should mention here:
 <li><a name="warnings_strict"></a>It is SUGGESTED that projects be maximally strict with warnings in the software produced by the project, where practical. (N/A allowed.) <sup>[<a href="#warnings_strict">warnings_strict</a>]</sup><dl><dt><i>Details</i>:<dt> <dd>Some warnings cannot be effectively enabled on some projects. What is needed is evidence that the project is striving to enable warning flags where it can, so that errors are detected early.</dd></dl></li>
 </ul>
 
-❓**OCIO needs somebody to assess**
-usually to address use of 3rd party library headers that trigger warnings.
+❓ **OCIO needs somebody to assess** Usually to address use of 3rd party library
+   headers that trigger warnings.
+
```
hodoulp: No strict warning detection enabled by default in the OCIO CI builds (i.e. one was in place but disabled following the file restructuration). So, I would suggest a FIXME for that topic.

docs/developers/CII-BestPractices-status.md
```diff
@@ -467,50 +474,49 @@ usually to address use of 3rd party library headers that trigger warnings.
 
 </ul>
 
-
 <b><i>Use basic good cryptographic practices</i></b>
```
hodoulp: I would suggest "**N/A**" for all the security topics.

hodoulp: OCIO reads some external files introducing some security concern to any apps using it. Please forget my comment around security topics as it partially applies.

docs/developers/CII-BestPractices-status.md
michdolan: Thanks @hodoulp . I updated this topic with a TODO since this item is suggested rather than required.

PR #638, opened 2018-12-05
Fix wrong display returned by getDefaultDisplay()

Nazg-Gul:
The issue was caused by wrong logic around parsing display/views
override environment variable: if it is missing, it was converted
to an empty string. That empty string was considered an override
of one element display, with an empty name.

Now splitting empty string into vector will give empty result,
which seems to be what is expected.

### DIFF
--- a/src/OpenColorIO/Config.cpp
+++ b/src/OpenColorIO/Config.cpp
@@ -274,11 +274,17 @@ OCIO_NAMESPACE_ENTER
         {
             std::string activeDisplays;
             Platform::Getenv(OCIO_ACTIVE_DISPLAYS_ENVVAR, activeDisplays);
-            SplitStringEnvStyle(activeDisplaysEnvOverride_, activeDisplays.c_str());
+            activeDisplays = pystring::strip(activeDisplays);
+            if (!activeDisplays.empty()) {
+                SplitStringEnvStyle(activeDisplaysEnvOverride_, activeDisplays.c_str());
+            }
             
             std::string activeViews;
             Platform::Getenv(OCIO_ACTIVE_VIEWS_ENVVAR, activeViews);
-            SplitStringEnvStyle(activeViewsEnvOverride_, activeViews.c_str());
+            activeViews = pystring::strip(activeViews);
+            if (!activeViews.empty()) {
+                SplitStringEnvStyle(activeViewsEnvOverride_, activeViews.c_str());
+            }
             
             defaultLumaCoefs_.resize(3);
             defaultLumaCoefs_[0] = DEFAULT_LUMA_COEFF_R;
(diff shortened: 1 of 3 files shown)


hodoulp:
@Nazg-Gul  Thanks for debugging the issue and proposing a solution.
Could add a unit test to 'demonstrate' the fix?

Nazg-Gul:
@hodoulp. did my best in writing the tests, hope they make sense :)

src/OpenColorIO/Config.cpp
```diff
@@ -2585,4 +2589,129 @@ OIIO_ADD_TEST(Config, range_serialization)
...
+        "    isdata: false\n"
+        "    allocation: uniform\n";
+
+    // First, check without environment variable override.
+    {
+        std::istringstream is(SIMPLE_PROFILE);
+        OCIO::ConstConfigRcPtr config;
+        OIIO_CHECK_NO_THROW(config = OCIO::Config::CreateFromStream(is));
+        OIIO_CHECK_EQUAL(std::string(config->getDefaultDisplay()), "sRGB");
+    }
```
hodoulp: In order to target the bug fix, you could add:
OIIO_CHECK_EQUAL(config->getNumDisplays(), 0);

src/OpenColorIO/Config.cpp
```diff
@@ -2585,4 +2589,129 @@ OIIO_ADD_TEST(Config, range_serialization)
...
+        const std::string active_displays(
+            std::string(OCIO::OCIO_ACTIVE_DISPLAYS_ENVVAR) +
+            "=AdditionalDisplay,ArtificialDisplay");
+        putenv(const_cast<char *>(active_displays.c_str()));
+
+        std::istringstream is(SIMPLE_PROFILE);
+        OCIO::ConstConfigRcPtr config;
+        OIIO_CHECK_NO_THROW(config = OCIO::Config::CreateFromStream(is));
+        OIIO_CHECK_EQUAL(std::string(config->getDefaultDisplay()), "AdditionalDisplay");
+    }
```
hodoulp: In order to target the bug fix, you could add (i.e. pseudo code):

  OIIO_CHECK_EQUAL(config->getNumDisplays(), 2); 
  OIIO_CHECK_EQUAL(std::string(config->getDisplays(0)), std::string("AdditionalDisplay"));
  OIIO_CHECK_EQUAL(std::string(config->getDisplays(1)), std::string("ArtificialDisplay"));
}
// Now, test with an empty environment variable
{
  putenv(OCIO::OCIO_ACTIVE_DISPLAYS_ENVVAR="");
  OIIO_CHECK_EQUAL(config->getNumDisplays(), 0);
}

sobotka:
Updated to fix segfault when no active_* is listed in the configuration.

See https://github.com/sobotka/blender/commit/a60536956e69af3101b2bf29edb7a76fa4e74956

Nazg-Gul:
@sobotka that is a fix for a temporary workaround on Blender side for until proper fix is accepted and released in the upstream. It is unneeded in the context of this fix.

@hodoulp, not sure why it is expected to have zero displays in the fist patch hunk. More like it should be all three screens? Is it expected to test for just displays/views order, or also for an active display/view?

hodoulp:
@Nazg-Gul  
1. The first comment I fixed the typo.
2. The second comment is about creating a unit test failing without the fix, and now succeeding.

Nazg-Gul:
@hodoulp, added a test for number of order of displays now. But needed to fix order of arguments in `IntersectStringVecsCaseIgnore()`. The new one makes more sense -- use predictable and controllable vector to grant ordering.

As for empty string used for environment variable -- i am not sure it's easy to do withing an existing Platform.cpp. The issue is: missing and empty environment variables are returned as an empty string. Can see two possible ways
- Make `Getenv` to return boolean value, truth if variable was found.
- Add separate `Hasenv`, but then windows code-path will become annoying. I wouldn't know how to avoid memory allocation for a variable value just to see whether its set or not will work.

Also, views ordering doesn't seem to be ensured anywhere, so for those i've simply left a comment in the test,

hodoulp:
@Nazg-Gul 
> ... But needed to fix order of arguments in IntersectStringVecsCaseIgnore(). The new one makes more sense -- use predictable and controllable vector to grant ordering.

Good catch :) 
Look for methods 'Config::getDefaultDisplay()' & 'Config::getDefaultView()' which are demonstrating the fix i.e. the idea is to preserve the order (as defined in the config. file) of the active displays or views.

michdolan:
@Nazg-Gul , when you get a chance, can you submit an Imageworks CLA for contributions to OCIO?

http://opensource.imageworks.com/cla/

Eventually we will all have to do another CLA for ASWF, but until that transition becomes official, the Imageworks CLA needs to be signed before we can merge anything. Thanks!

Nazg-Gul:
@michdolan, i have CLA signed back in November 2014, but that was for OSL (think it doesn't matter and is an agreement with SPI, not with a specific project?). One thing though: the mailing address did change since then. So, shall i sign and send a new one, or that will confuse some system?

@hodoulp, your patch seems more comprehensive on the testing side :) I'll try to give that a test in our environment ASAP. But then, what is the way to move forward? As in, what do you expect me to do/wrap up and what you'd prefer to pick up and finish yourself? :)

hodoulp:
@Nazg-Gul 
> But then, what is the way to move forward? As in, what do you expect me to do/wrap up and what you'd prefer to pick up and finish yourself? :)

As a dev. I something feel that sending code far better explain a point than any long sentences. Please do not see the patch as a way to complete our work. It's a way to communicate the idea around having unit tests 'documenting' use cases (and problematic ones).

Adding the unit tests from Config.cpp and the fix for Getenv() will complete the review at my point of view.

michdolan:
@Nazg-Gul , if you already provided a CLA for OSL contributions, you should be good. Thanks!

hodoulp:
Hi @Nazg-Gul, 
You find several bugs affecting `master` and `RB-1.1`.
Do you think you will have some time to finalize the pull request?

Nazg-Gul:
@hodoulp,. yep. Will review the status and address feedback tomorrow.

Nazg-Gul:
@hodoulp, i've made a separate commit with the GetEnv. Figured, no real need to combine it into same commit. used your tests for display and verified the things we were testing against are covered bu them. Also did more comprehensive tests of views.

Noticed that order of views always follows order of definition in the config file. Not sure what is the designed behavior here actually.

src/OpenColorIO/Config.cpp
```diff
@@ -2585,4 +2591,370 @@ OIIO_ADD_TEST(Config, range_serialization)
...
+    {
+        std::string myProfile = 
+            SIMPLE_PROFILE_HEADER
+            +
+            "active_displays: []\n"
+            "active_views: []\n"
+            + SIMPLE_PROFILE_FOOTER;
+
+        const std::string active_displays(
+            std::string(OCIO::OCIO_ACTIVE_VIEWS_ENVVAR) + "="); // No value, but misleading space
```
hodoulp: The space explaining the need of that test case, is missing. 
The string should be "= " (instead of "=").

Nazg-Gul:
@hodoulp, ah you're right. Corrected and pushed a fix.

PR #639, opened 2018-12-05
Adsk Contrib - Add the Gamma Op as defined by the Adsk CTF specification

hodoulp:
Here is the work to support the Gamma Op as defined by the Autodesk CTF specification.

Note 1: For backward compatibility the Exponent Op is still present and using the OCIO v1 implementation. Later pull request will reimplement the Op using the ACES Gamma Op.

Note 2: Following Slack discussion around the op name, no consensus was found so we proceed with the CLF name for now (i.e. Exponent Op could still be used as it will stay in OCIO v2).

Note 3: The Python bindings will come later.


scoopxyz:
Hey Patrick et. al.

I like the additional extension to support Gamma and Pseudo-Gamma functions, though I'm a bit iffy about having both an Exponent and Gamma Transform defined. Is the intent to deprecate Exponent after the initial release of 2.0?

hodoulp:
The ExponentTransform will stay in OCIO v2 to guaranty a full backward compatibility for existing OCIO v1 config. files (i.e. the exponent keyword) and OCIO v1 implementation (i.e. ExponentTransform is part of the public API). But we plan to reimplement it using the new Gamma Op.

Deprecating an Op was never done before so we (i.e. the OCIO community) should have a discussion on that topic first.

scoopxyz:
Ok, well it doesnt seem worthwhile to carry Exponent Transform indefinitely into the future if it provides no value other than backward compatibility. Agreed, we'd need to think about how to gracefully handle that.

michdolan:
As long as it's marked as deprecated in code and docs, keeping around some minimal interface for backwards compatibility for another major version (i.e. till OCIO v3) is probably a reasonable move.

src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp
```diff
@@ -0,0 +1,441 @@
...
+#include "SSE.h"
+
+
+OCIO_NAMESPACE_ENTER
+{
+
+// Note: The parameters are validated when the op is created so that the
+// math below does not require checks for divide by 0, etc.
+
+class CDLOpCPU;
```
michdolan: Are these CDLOp forward declarations necessary?

src/OpenColorIO/ops/Gamma/GammaOpData.cpp
```diff
@@ -0,0 +1,1248 @@
...
+                                         getRedParams(), getGreenParams(),
+                                         getBlueParams(), getAlphaParams() );
+}
+
+bool GammaOpData::isInverse(const GammaOpData & B) const
+{
+    const Style styleA = getStyle();
+    const Style styleB = B.getStyle();
+
+    // Note: It's possible that someone could create something where they
```
michdolan: Perhaps this edge case should trigger a warning, so its not quietly unhandled.

include/OpenColorIO/OpenColorTransforms.h
```diff
@@ -501,6 +501,70 @@ OCIO_NAMESPACE_ENTER
...
+        // a basic power function or a piecewise function (L*, sRGB).
+        virtual GammaStyle getStyle() const;
+        //!cpp:function::
+        virtual void setStyle(GammaStyle style);
+
+        //!cpp:function:: Set the exponent value for the power function for R, G, B, A.
+        // .. note::
+        //    By convention, the gamma values should be >= 1. (Where possible, set 
+        //    the transform direction to inverse rather than using gamma values less than 1.)
+        void setGammaValues(const double * vec4);
```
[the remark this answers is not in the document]
hodoulp: Good point.

src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp
hodoulp: Good point.

src/OpenColorIO/ops/Gamma/GammaOpData.cpp
hodoulp: When reading from config. files or CTF file, the validation traps the constraints (refer to GammaOpData::validateParameters() ).

src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+        _mm_storeu_ps(rgba, data);
+
+        rgba += 4;
+    }
+#else
+    float * rgba = rgbaBuffer;
+
+    const float red[5] 
+        = { float(m_red.scale),  float(m_red.offset),
```
scoopxyz: Is there a reason for carrying the `double` render parameters everywhere if they're just going to cast to `float`?

src/OpenColorIO/ops/Gamma/GammaOpUtils.cpp
```diff
@@ -0,0 +1,248 @@
...
+{
+    const double offset = std::max(p[1], EPS);
+    return 1. + offset;
+}
+
+
+
+};
+
+void ComputeParamsFwd(const GammaOpData::Params & gParams,
```
scoopxyz: Does this actually work?

Using the sRGB values provided elsewhere (`gamma=2.4, offset=0.055`) I don't get the correct breakpoint between the linear and power component of the curve. I get `0.03928571428571429` when calculated outside (in python numpy)...

That said the unittests should probably check against these known "truths" instead of always using arbitrary gamma and offset combinations.

scoopxyz:
Another fundamental difference I see is that the "Gamma" function's "gamma" is just a power function that is (`x^gamma`) rather than what I'm more familiar of with gamma meaning (`x^1/gamma)?

Also, the mixed terminology of `Exponent`, `Gamma`, `Power` have now all made it into the code base. This only serves to confuse IMHO.

src/OpenColorIO/ops/Gamma/GammaOpUtils.cpp
doug-walker: Sean, could you elaborate on why you think the break point is wrong?  For the sRGB case, the monCurveBreakFwd function above would return  0.055/(2.4-1) = 0.03928571428571429, exactly what you wrote.

doug-walker:
Sean, we felt it was better to use the "gamma" value directly as the exponent.  This is more in line with the way OCIOv1 Exponent works as well as CDL.  Also, if someone says 'the gamma of the monitor is 2.4,' then they mean the monitor is applying x^2.4, and not x^(1/2.4).  

Using the inverse is something that grading systems sometimes do since the gamma is the midtones control and colorists like it if the picture gets lighter when they raise it.  But from a color management point of view, I don't think it makes sense to use the inverse notation.

As to the terminology, I think the use of Power is fine and unambiguous.  It's also not at the public API level.  Exponent we are open to completely removing, except in the Yaml parser for backwards compatibility.  Gamma we are open to rename but when I raised the issue on Slack no clear consensus formed around an alternative.

src/OpenColorIO/ops/Gamma/GammaOpUtils.cpp
scoopxyz: It should be 0.04045 no?

scoopxyz:
Yeah, fair enough that you've asked before.

Yes, let deprecate the exponent ops, a get formal vote on what to call these ops moving forward. Next meeting?

src/OpenColorIO/ops/Gamma/GammaOpUtils.cpp
scoopxyz: Ah, im finding a descrepency based on what source you look at, fun.

So assuming the W3 site is correct then this should be fine.

src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp
hodoulp: We are slowly transitioning the Ops (i.e. OpData) to double so that the GPU implementation benefits from the double precision, which imposes the RenderParams structure to handle double values. But the CPU implementations (i.e. SEE and not SEE) still computes from float values (i.e. the pixel) needing a cast of the Gamma parameters outside the for-loop processing. As the CPU performance is critical, the transition to a double precision computation within the for-loop should be investigated first.

src/OpenColorIO/ops/Gamma/GammaOpUtils.cpp
Dithermaster: Is this discussion around the Rec.709/Rec.2020 gamma curve? When the industry moved to higher bit-depths they found that the constants were not high enough precision and there was a discontinuity at the break point. So they changed the constants to have another decimal place, which fixed that problem. The improved constants are fine when used with older content and lower bit-depths, so no need to keep both around; just use the improved ones.

doug-walker: Dithermaster, we designed the code to avoid the problem you mention.  In various standards there are a lot of constants shown but in reality there are only two free parameters and the rest are calculated by someone and specified to various degrees of precision.  In our implementation the user specifies the two free parameters (the offset and exponent) and the other constants (break point, etc.) are calculated to machine precision so that the two curve segments will match precisely.

scoopxyz:
My approval is only pending the resolution of what to call the Op/Transform everything else LGTM

JGoldstone:
Not that I am on the path to approval, and I can’t comment on the code in the way others have, but:

Please let’s not call this things `GammaTransform` and `GammaOp`. My position is this: we should always encourage users and especially contributors to reference foundational documents, which to me means documents approved by international standards bodies. None of them are using “gamma” to include a piecewise function as this PR suggests we do. I think that’s a hint that _we shouldn’t either_; further I think we should take our best shot at aligning themselves with their usage. Not that they are all perfectly consistent but here’s a sampling (skewed towards HDR standards since that’s what I deal with most):

- The ICC spec calls this sort of thing a parametricCurve (ICC.1:2004-10, §10.15) and they have five different variants. The first is a pure exponent form. All the others are piecewise and take between 3 and 7 parameters: CIE 122-1966 is gamma, a and b; IEC 61966-3 adds a ‘c’; IEC 61966-2.1 (sRGB) adds a d, and an unnamed fourth piecewise definition adds an e and f.

- ARIB STD-B67 (_Essential Parameter Values for the Extended Image Dynamic Range Television (EIDRTV) System for Programme Production_), Table 2 (p.4), has a two-section-piecewise _Reference non-linear transfer function (opto-electronic transfer function, OETF)_.

- The ASC regarded ‘Gamma’ as such an overloaded word that they skipped it, saying _Because the terms Lift, Gain and Gamma have long established definitions, and because those definitions may vary in detail from system to system and manufacturer to manufacturer, to avoid controversy, competition and confusion, the ASC uses a set of three transfer functions with unique names and orthogonal (non-overlapping) definitions_ and in the next paragraphs define _Offset_, _Slope_ and _Power_.

- CEA-861.3, _HDR Static Metadata Extensions_, has two tables (Table 3 and Table 8) of _Electro-Optical Transfer Function_, and each has its first two entries as _Traditional Gamma - SDR Luminance Range_ and _Traditional Gamma - HDR Luminance Range_ with the third entry being _SMPTE ST 2084_. To me this says the general curve is a transfer function and gamma functions are special cases thereof.

- The DCI spec (1.2, 3/7/2008, Table 11) lists a _Transfer Function_ parameter with a nominal value being _Gamma 2.6_.

- The EBU _Report on Display Gamma / EOTF Measurement_, authored principally by Richard Salmon of the BBC, seems to sometimes use Gamma as a particular type of EOTF, and sometimes will say _the EOTF curve (Gamma) was…_. Not helpful. EBU Tech 3335, _Methods of measuring the imaging performance of television cameras for the purposes of characterization and setting_, principally written by Alan Roberts of the BBC, titles their section 2.1 _Opto-Electronic transfer curve (Gamma)_, thus basically equating the two.

- ISO 14524, _Photography — Electronic still-picture cameras — Methods fo measuring opto-electronic conversion functions (OECFs)_ is notable for saying _conversion function_ instead of _transfer function_, but in other ISO documents they seem to use plain old _transfer function_ just for optics, e.g. _modulation transfer function_. They do say _function_ one way or the other and never use _gamma_.

- ITU-R BT.709 §1, p. 2, uses _Opto-electronic transfer characteristics_ to describe a piecewise function, and uses the phrase _gamma pre-corrected signals_ in §4, p.3, to describe what the display will be fed as a signal; this to me implies the display is expected to behave in a single-exponent-parameter fashion.

- ITU-R BT.1361, _Worldwide unified colorimetry and related characteristics of future television and imaging systems_, the crazy uncle at Thanksgiving member of the ITU standards family, in Annex 1 (_Extended colour gamut system using negative RGB signals_), says _opto-electronic transfer characteristics (gamma curve)_. But to call BT.1361 an outlier would be serious understatement.

- Critically ITU-R BT.1886, _Reference electro-optical transfer function for flat panel displays used in HDTV studio production_, never uses the term _gamma_ at all. It’s all about _Reference electro-optical transfer function_ with a normative exponent of 2.4, and is a simple pure exponent model for an idealized display which emits no light when not told to do so.

- Oh and there’s a great diatribe by Alvy Ray Smith from his early days at Microsoft, titled simply _Gamma Correction_, [here](http://alvyray.com/Memos/CG/Microsoft/9_gamma.pdf).

All this is to say that standards bodies (and groups like the ASC informally promoting something as ‘standard’), with few sane exceptions and outside BBC house style (cf. _Hybrid Log Gamma_), avoid using _gamma_ for anything other than the exponent of the simplest power function. 

_Transfer function_ is the most common generic term that would be inclusive of piecewise functions. Unfortunately we’d be stuck with `TransferFunctionTransform` which sounds really awkward. I’m going to suggest following the ICC on this and going with `ParametricCurveTransform`.

If I’m suggesting this at this late date, when all we are discussing is the name, it behooves me to make a complete proposal. So:

`GammaTransform` becomes `ParametricCurveTransform`
`GammaStyle` becomes `ParametricCurveStyle`
`ParametricCurveStyle` has two values (for now), `PARAMETRIC_CURVE_EXPONENT` and `PARAMETRIC_CURVE_LINEAR_AND_EXPONENT`
`GammaStyleToString` becomes `ParametricCurveStyleToString`
`GammaStyleFromString` becomes `ParametricCurveStyleFromString`
`ops/Gamma` becomes `ops/ParametricCurve`
`ops/Gamma/GammaOpCPU.cpp` becomes `ops/ParametricCurve/ParametricCurveOpCPU.cpp`
`ops/Gamma/GammaOpData.cpp` becomes `ops/ParametricCurve/ParametricCurveOpData.cpp`
`ops/Gamma/GammaOpUtils.cpp` becomes `ops/ParametricCurve/ParametricCurveOpUtils.cpp`
`ops/Gamma/GammaOps.cpp` becomes `ops/ParametricCurve/ParametricCurveOps.cpp`
YAML style parameters are updated to `exponent` and `inear_and_exponent` (or is that `linearandexponent`?), and to `exponent` instead of `gamma`.
`GammaType` becomes `ParametricCurveType`
`BuildGammaOps` becomes `BuildParametricCurveOps`
`CreateGammaOp` becomes `CreateParametricCurveOp`
`GammaTransform::setGammaValues()` becomes `ParametricCurveTransform::setExponentValues()`
`GammaTransform::getGammaValues()` becomes `ParametricCurve::getExponentValues()`

The only weakness I can see in the above is one that exists in the current PR as well: the assumption in the `ParametricCurveStyle` enum that if something isn’t a simple power function, it must be describing a monitor. That seems overly specific, and even if the renaming proposed above is rejected, I think the assumption that if it’s not a simple exponent applied to a base value, this much be about monitors, should be examined.

michdolan:
I like the ParametricCurve name. Something similar was proposed on Slack at one point, but that discussion fizzled out. One challenge with this naming from an architectural perspective is that it broadens the scope of this Op considerably. My understanding (correct me if I'm wrong here Doug) is that there were plans on having a number of parametric curve ops. A "gamma" op and a "log" op for example, that go beyond the current ExponentOp and LogOp.

If we made this a ParametricCurveOp, I feel like it invites these separate ops to be unified into a more general definition. The challenge with that is keeping the interface simple enough to be usable without something like the AllocationTransform's "vars" property, which can be overloaded with N arguments based on the transform variant. That's not meant as an argument against @JGoldstone 's very well evidenced point; just meant to imply that this discussion probably has wider implications beyond this Op/Transform name.

sobotka:
Won’t be at the meeting, and am but a flea here, however in 100% agreement that “gamma” feels like an anachronistic step backwards.

Dithermaster:
Excellent reference list, but it's missing Charles Poynton, PhD, who wrote the book (or at least a book, "A Technical Introduction to Digital Video") on digital television. He says the Rec. 709 transfer function curve is used for gamma correction to/from linear, and it includes a linear section (making it piecewise): https://poynton.ca/notes/colour_and_gamma/GammaFAQ.html

So like I said before, if this Op is only ever used for that, I'm OK with naming it GammaOp. But if it's ever going to grow to include other transfer functions, then it needs a new name.

src/OpenColorIO/ops/Gamma/GammaOpUtils.cpp
Dithermaster: > Dithermaster, we designed the code to avoid the problem you mention.
Very cool!

KelSolaar:
Completely agree with @JGoldstone and I would also rather prefer to see a dog called a dog.

When I read *gamma*, I think about a power function, not a piece-wise function. Our OCIO config at work defines _colorspaces_ that are true gamma functions, e.g. a pure 2.2 and I'm sure anybody would be confused by the introduction of a `GammaOp` that is modeled by a piece-wise function under the hood.

My \[x^{2.2}\] cents :)

sobotka:
>The luminance generated by a physical device is generally not a linear function of the applied signal. A conventional CRT has a power-law response to voltage: luminance produced at the face of the display is approximately proportional to the applied voltage raised to the 2.5 power. The numerical value of the exponent of this power function is colloquially known as gamma. This nonlinearity must be compensated in order to achieve correct reproduction of luminance.

Poynton also references REC.709’s characteristics as it is in it and the specification’s nomenclature as *transfer function*, with no reference to “gamma” in the defining specifications themselves.

doug-walker:
Thomas, I would argue that "GammaTransform" is very much like "calling a dog a dog".  To continue your analogy, I think a pure power function and the sRGB function are like two different breeds of dog.  We don't have the development resources to build separate ops for these two things, so the name will inevitably not be overly specific (i.e. it has to be a "dog" rather than a "beagle").

In comparison, the sRGB curve and a camera log curve are both piecewise parametric curves, but to me one is a dog and the other is a cat.  I would rather have a "dog" op and a "cat" op, which each implement a few related mathematical forms.  To me, "ParametricCurve" is too vague.  It is more like a "FourLeggedMammal" op.

Anyway, we will discuss it during the working group meeting today.

src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp
doug-walker: Patrick and I discussed this and he will look at adding a commit for this.

src/OpenColorIO/ops/Gamma/GammaOpData.cpp
doug-walker: Michael, does Patrick's reply resolve your concern?

michdolan:
@doug-walker , in your design of this transform, do you have plans of implementing other display transfer functions here as well (i.e. HLG, or a full BT.1886 impl)?

src/OpenColorIO/ops/Gamma/GammaOpData.cpp
michdolan: Yes, thanks.

KelSolaar:
Should you call it *GammaPieceWiseOp* would be clear on what it does, *GammaOp* simply is not. Somebody reading a config without reading the docs might be fooled. There is ambiguity and the fact we are debating about it is a sign :)

hodoulp:
Hi Guys, 
It could help the discussion if you could take some time to think about potential good names and propose them (or send them if you could be in the WG meeting).

doug-walker:
Patrick, Bernard, and I discussed this and we could support Joseph's proposal of DisplayOETFTransform.  (In hopes of avoiding spending our entire meeting arguing over the perfect name.)

src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp
hodoulp: Done

michdolan:
Time to merge this one. Friendly 24 hour warning in case anyone has additional comments.

PR #643, opened 2019-01-04
Propose additional OCIO TSC members

michdolan:
Proposing to add two additional members to the initial OCIO TSC:

Carl Rand - Foundry
Kevin Wheatley - Framestore


lgritz:
ok by me

hodoulp:
Very good candidates as TCS members.

Issue #591, opened 2018-10-01
Python unit tests fail to run on Windows


hodoulp:

scoopxyz:
Are you referring to a specific build combination or are the tests silently failing? Here's a copy-paste from the latest Appveyor build on Master:
```
    Start 2: test_python
2: Test command: C:\Python27-x64\python.exe "C:/source/ocio/tests/python/OpenColorIOTestSuite.py" "C:/source/ocio/_build" "Release"
2: Test timeout computed to be: 10000000
2: test_interface (MainTest.MainTest) ... ok
2: test_interface (ConstantsTest.ConstantsTest) ... ok
2: test_interface (ConfigTest.ConfigTest) ... ok
2: test_is_editable (ConfigTest.ConfigTest) ... ok
2: test_interface (ContextTest.ContextTest) ... ok
2: test_interface (LookTest.LookTest) ... ok
2: test_interface (ColorSpaceTest.ColorSpaceTest) ... ok
2: test_interface (TransformsTest.TransformsTest) ... ok
2: test_interface (GpuShaderDescTest.GpuShaderDescTest) ... ok
2: test_interface (Baker.BakerTest) ... ok
2: 
2: ----------------------------------------------------------------------
2: Ran 10 tests in 0.000s
2: 
2: OK
2/2 Test #2: test_python ......................   Passed    0.25 sec
100% tests passed, 0 tests failed out of 2

```

hodoulp:
1) Compiling using ***Visual Studio*** now needs lines below to successfully run Python unit tests:
```
    -DPYTHON_LIBRARY=C:\Python27\libs\python27.lib ^
    -DPYTHON_INCLUDE_DIR=C:\Python27\include ^
    -DPYTHON_EXECUTABLE=C:\Python27\python.exe ^
```
2) Compiling with ***jom*** or ***ninja***, Python unit tests now always fail (with or without previous lines).
```
3: Traceback (most recent call last):
3:   File "D:/OpenSource/ocio_adsk_publ_1/tests/python/OpenColorIOTestSuite.py", line 18, in <module>
3:     import PyOpenColorIO as OCIO
3: ImportError: No module named PyOpenColorIO
3/3 Test #3: test_python ......................***Failed    0.12 sec
```

PR #646, opened 2019-01-09
Adsk Contrib - Fix Windows command line build

hodoulp:
The Windows Python Binding unit tests failed because of the difference between running from Microsoft Visual Studio and from a console (i.e. Vc++ using ninja for example). The expected output directory is not the same as Microsoft Visual Studio happens the build type by default.


hodoulp:
Fixing #591

PR #647, opened 2019-01-15
Add draft ASWF charter, CLA, DCO

lgritz:
Commit the draft charter, CLAs, and DCO to the repo so that it's easy to reference them.

For now, in `docs/ASWF-WIP`, will move them to a more official sounding place when ASWF votes to admit the org.


lgritz:
These are the versions we've had circulating and approved, and are the standard ones from ASWF templates or copied from OpenVDB (with minimal changes for our names, licenses, etc.). I just realized they were never committed to our repo. There is nothing new to see here.

PR #648, opened 2019-01-17
Adsk Contrib - Add Category support to Color Spaces

hodoulp:
Please refer to #623 to have the complete explanation, and #627 for the first API proposal.

Here is the full implementation of the 'categories' support for Color Spaces. One could first have a look to the unit tests in the file ColorSpaceSet.cpp to have examples on how to use this new mechanism.

Note: To limit the size of the review, the Python bindings are planned for another pull request.

### DIFF
--- a/include/OpenColorIO/OpenColorIO.h
+++ b/include/OpenColorIO/OpenColorIO.h
@@ -304,30 +304,52 @@ OCIO_NAMESPACE_ENTER
         // 
         // ColorSpaces
         // ^^^^^^^^^^^
-        
+
+        //!cpp:function:: Get all color spaces having a specific category
+        //                in the order they appear in the config file.
+        //
+        // .. note::
+        //    If the category is null or empty, the method returns 
+        //    all the color spaces like :cpp:func:`Config::getNumColorSpaces` 
+        //    and :cpp:func:`Config::getColorSpaceNameByIndex` do.
+        //
+        // .. note::
+        //    It's worth noticing that the method returns a copy of the 
+        //    selected color spaces decoupling the result from the config. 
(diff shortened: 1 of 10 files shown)


hodoulp:
Here is the first C++11 vs Compiler issue that I fixed using a gcc/clang 5.x version. The side effect is a compilation failure for older compiler versions.
Is there any point supporting older gcc/clang version (i.e. lower than version 5)?

lgritz:
We're still primarily using gcc 4.8.

hodoulp:
Fixed.

*Note: The VxF platform recommendation is now gcc 6.3.1*

lgritz:
Yes, vfx platform is 6.3, but CentOS 7 that many of the studios are using still has 4.8 as the default compiler.

scoopxyz:
LGTM.

I will spawn another ticket with my remaining question, namely.
* What categories are required to exist by ociocheck?

include/OpenColorIO/OpenColorIO.h
```diff
@@ -693,9 +747,105 @@ OCIO_NAMESPACE_ENTER
...
+    //!rst:: .. _colorspaceset_section:
+    // 
+    // ColorSpaceSet
+    // *************
+    // The *ColorSpaceSet* is a set of color spaces (i.e. no color space duplication) 
+    // which could be the result of :cpp:func:`Config::getColorSpaces`
+    // or built from scratch.
+    // 
+    // .. note::
+    //    The color spaces are decoupled from the config. ones i.e. any
```
michdolan:
```suggestion
    //    The color spaces are decoupled from the config ones, i.e., any 
```

michdolan:
I would be ok with turning it off.

PR #651, opened 2019-01-23
Adsk Contrib - Lut3D inversion tests slow in debug on windows

BernardLefebvre:
MSVC implementation of vector does many checks in debug and this makes the LUT 3D inversion tests too slow to be enabled by default in debug for windows.
Note: this was originally part of PR #640.


michdolan:
Thoughts on this @BernardLefebvre or @hodoulp ?

BernardLefebvre:
There are TODO: in the code and tests are running in release. Disabling them in debug because slow tests might be painful for developers working on other features. Every CI is running is release and is running the tests, if they were to fail, devs could locally re-enable them in debug to investigate.

PR #656, opened 2019-01-29
Adsk Contrib - CLF parsing

BernardLefebvre:
This PR implements Issue #610.  It adds a new FileFormat to read Academy/ASC Common LUT Format (CLF) files.  This fully implements the S-2014-006 spec from the Academy.

Note: you will see references to CTF in the code.  The CLF support is based on the SynColor CTF reader, since CLF is a subset of CTF.  The Autodesk Color Transform Format (CTF) is a superset of CLF that adds support for all OCIO ops (e.g. including Exponent and Log) as well as some ops that are useful for accurate ACES implementations.  Because CTF requires adding some additional ops to OCIO, this PR does not enable CTF support yet.  That will come in a future PR.

This PR adds a lot of unit tests and the tests/data/unittestsfiles.tar.gz has many new example CLF files for parsing tests.  There are files that excercise all parts of the spec, including good files and files with syntax errors to test the exception handling.  We put a fair amount of work into trying to provide useful error messages when a file cannot be parsed, indicating the line number and the exact nature of the issue.


doug-walker:
Good point Michael.  There are indeed some messages that should be changed to "CTF/CLF ...".  Bernard will look into that.

doug-walker:
Yes, excellent question Michael.  Versioning of file formats like this is inevitable and is challenging to handle correctly.  We tried to be rigorous and put in place the mechanism to be able to handle multiple versions of the spec.  So in some places you will see multiple handlers for a given op if its definition was changed over time.  For example, when CLF went from v1 to v2 there were some changes made by the committee.  The code is able to handle this as well as the various versions of CTF, as we added new features.  There are also unit tests with files of various versions.

The Academy has just convened a working group to discuss the next version of CLF, so I think we should expect the spec to continue to evolve.  However, if any changes are made to the S-2014-006 spec, they will need to increment the compCLFversion attribute (I will insist on that!).  So as changes are made, it will obviously mean adding more code to the parser, but the foundation we have in place should be able to handle it.

michdolan:
Thanks for the explanation. The code seems really rigorous, which is great, but definitely non-trivial to digest. I can see where criticism about CLF/CTF being "OCIO within OCIO" comes from, but assuming applications without OCIO support CLF, being able to more precisely represent an OCIO transform stack in a single file is a big gain IMO.

doug-walker:
Yes, and with the rise of web-based services, etc., the ability to serialize an arbitrary OCIO transformation in a well-defined way is very useful.

hodoulp:
Just to mention that for now, any new Op has an associated transform so that a CLF (and later CTF) file could be converted to a list of transforms in an OCIO config file if needed.

scoopxyz:
I'll work on getting my eyes on this early this week.

hodoulp:
@michdolan @scoopxyz  That pull request hits the 'two-weeks' delay so that's now a good candidate for a merge.

michdolan:
@scoopxyz was planning to have a look this week. Two more days enough Sean?

michdolan:
I'm planning on merging this Tuesday of next week unless there is additional feedback.

PR #658, opened 2019-02-04
Update CLAs per Sony legal department edits

lgritz:
The earlier one was copied straight from OpenVDB, which Sony Legal
decided was too closely tied to OpenVDB's use of the Mozilla Public
License and that they preferred the standard LF templates.

This draft brings it more in line with the standard Linux Foundation
CLAs found here:
 https://github.com/AcademySoftwareFoundation/tac/tree/master/process

The new CCLA is exactly the LF one found above (with the proper name and
license substitutions). The new ICLA is very close to the CCLA, whereas
the LF template is more minimal; Sony Legal liked how the LF CCLA
spelled out some definitions more precisely and modeled after that.

### DIFF
--- a/docs/ASWF-WIP/CLA-corporate.md
+++ b/docs/ASWF-WIP/CLA-corporate.md
@@ -1,42 +1,49 @@
 Corporate Contributor License Agreement ("Agreement")
 
 Thank you for your interest in the OpenColorIO Project a Series of LF
-Projects, LLC (hereinafter "Project") which has selected the BSD 3-clause
-(hereinafter "BSD") license for its inbound contributions. The terms You,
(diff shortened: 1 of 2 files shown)


scoopxyz:
Presumably we all need to sign the updated CLA? When/ how?

I've create the ticket #659 to action adding documentation around the CLA process.

lgritz:
Let's get final approval from LF first, but then yes, start working on getting your orgs to sign the new one. Presumably the deadline is when the repository actually changes from underneath github.com/imageworks to github.com/AcademySoftwareFoundation -- once it's no longer being hosted by SPI, submissions should require the new CLA.

PR #666, opened 2019-02-12
Adsk Contrib - Rename LUT inversion quality member to clarify the use of the member.

BernardLefebvre:
Rename LUT member and associated enum that control the quality of LUT inversion.
This is part of issue #621 but it does not add a way to control the inversion quality through the API.


include/OpenColorIO/OpenColorTypes.h
```diff
@@ -263,6 +261,25 @@ OCIO_NAMESPACE_ENTER
...
+    // quite accurate even for scene-linear values, but for Lut3D the baked
+    // version is more of an approximation.  The DEFAULT is the fast method
+    // since it is the only one available on both CPU and GPU.
+    //
+    enum LutInversionQuality
+    {
+        LUT_INVERSION_BEST = 0,
+        LUT_INVERSION_FAST,
+        LUT_INVERSION_DEFAULT
+    };
```
michdolan: To be consistent with Interpolation, I feel like this should be:

LUT_INVERSION_EXACT = 0,
LUT_INVERSION_FAST,

LUT_INVERSION_DEFAULT = 254,
LUT_INVERSION_BEST = 255

BEST would be an alias to the "best" suitable inversion quality (EXACT), and DEFAULT an alias to the default inversion quality (FAST). Neither should describe an inversion quality themselves.

michdolan:
Thanks @BernardLefebvre . LGTM.

PR #670, opened 2019-02-20
Adsk Contrib - Add a possibility to disable a GPU test at compile time.

BernardLefebvre:
A GPU test can be disabled and still visible (as disabled) in the output.


michdolan:
This PR could be merged with single approver. I'll merge it in 24 hours if no issues are raised.

PR #671, opened 2019-02-20
Adsk Contrib - Add LUT tests that are using CTF parsing.

BernardLefebvre:
Adding several LUT tests.
Fix LUT 1D GPU so that floats are sanitized like they are on the CPU code path.


BernardLefebvre:
Closing for now. Missed internal code review.

PR #673, opened 2019-02-22
Update OCIO mail lists to ASWF domain

michdolan:
Notice of the mail list change has been sent to ocio-dev and ocio-users. This PR updates OCIO docs to reflect the change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,11 +14,11 @@ save yourself a lot of time by asking first.
 
 How do you talk to us? There are several ways to get in touch:
 
-* [ocio-dev](https://groups.google.com/forum/#!forum/ocio-dev):
+* [ocio-dev](https://lists.aswf.io/g/ocio-dev):
 This is a development focused mail list with a deep history of technical
 conversations and decisions that have shaped the project.
 
-* [ocio-users](https://groups.google.com/forum/#!forum/ocio-users):
+* [ocio-user](https://lists.aswf.io/g/ocio-user):
 This is an end-user oriented mail list, focused on how to use OCIO’s features
 within a host application. Common topics include crafting configs, DCC behavior,
 and general color questions.
@@ -34,7 +34,7 @@ GitHub **issues** are a great place to start a conversation! Issues aren’t
 restricted to bugs; we happily welcome feature requests and other suggestions
 submitted as issues. The only conversations we would direct away from issues are
 questions in the form of “How do I do X”. Please direct these to the ocio-dev or
-ocio-users mail lists, and consider contributing what you've learned to our
+ocio-user mail lists, and consider contributing what you've learned to our
 docs if appropriate!
 
 ## Getting Started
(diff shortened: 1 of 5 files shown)


michdolan:
@scoopxyz what's the process for deploying these changes to gh-pages?

docs/configurations/index.rst
```diff
@@ -8,7 +8,7 @@ and how to create new ones.
 
 OCIO Configurations can be downloaded here: `.zip <http://github.com/imageworks/OpenColorIO-Configs/zipball/master>`__ `.tar.gz <http://github.com/imageworks/OpenColorIO-Configs/tarball/master>`__ (OCIO v1.0+)
 
-If you are interested in crafting custom color configurations, and need assistance, please contact: `ocio-dev <http://groups.google.com/group/ocio-dev>`__\.
+If you are interested in crafting custom color configurations, and need assistance, please contact: `ocio-dev <https://lists.aswf.io/g/ocio-dev>`__\.
```
doug-walker: This could probably be debated either way, but to me, people creating configs would be better considered as "users" and so the better list would be ocio-user.  The ocio-dev list seems like it is more for people developing the library itself.  Not a strong objection on my part, just wanted to raise the question.

doug-walker: Oh, and I just noticed in the next file it points people creating configs to ocio-user, so this convinces me even more that this one should point to ocio-user.

CONTRIBUTING.md
```diff
@@ -14,11 +14,11 @@ save yourself a lot of time by asking first.
 
 How do you talk to us? There are several ways to get in touch:
 
-* [ocio-dev](https://groups.google.com/forum/#!forum/ocio-dev):
+* [ocio-dev](https://lists.aswf.io/g/ocio-dev):
 This is a development focused mail list with a deep history of technical
```
doug-walker: This may confuse people given that the old ocio-dev archives have not been migrated over yet.  Perhaps at the end of the sentence add something like: 
[NOTE: The ocio-dev google group archives are in process of being migrated to the new aswf list.]

michdolan: Alternatively, we can wait to merge/deploy these changes until after the Google groups are moved.

docs/configurations/index.rst
michdolan: Good call. I will change this.

scoopxyz:
I added this in the Travis build script a while back:
```
deploy:
  provider: pages
  skip_cleanup: true
  github_token: $GITHUB_TOKEN
  on:
    tags: true
  local_dir: ./docs/build-html
  fqdn: opencolorio.org
```

Which deploys the docs to gh-pages whenever a version is tagged on Master. Since we don't want to do that obviously you can just locally build the sphinx page, then push the commit to the `gh-pages` branch and upload.

scoopxyz:
Oh, also make sure you keep the `google2560df9c4b755196.html` and `CNAME` file in the branch otherwise the site crashes...

Learned from experience ;)

CONTRIBUTING.md
doug-walker: Either way is fine with me.

PR #674, opened 2019-02-22
Adsk Contrib - Rework LogOpData to get ready for Log CTF parsing.

BernardLefebvre:
The OCIO Log Op had the capability to do Cineon-type lin-to-log transforms but the parameters were never exposed at the Transform level so that they could be set via a Config file.  This PR adds that capability.

Following up on the discussion at the last v2 working group about the ExponentTransform, we decided that rather than add parameters to the LogTransform, we would create a new Transform called LogAffineTransform (since the math is an affine transform, a log, and then another affine transform).   Both the Log and LogAffine Transforms create the same Op under the hood.

The updated Op is also a prerequisite for the CTF file format support that is in progress.

The PR also adds SSE and non SSE code path CPU and GPU implementation as well as many unit tests.


BernardLefebvre:
Will remove the enum from the public API.

hodoulp:
Next Monday will be the time to merge this one. Friendly 24 hour warning in case anyone has comments.

PR #677, opened 2019-02-24
Add USE_SSE Checks To Fix Compile on Linux

zicklag:
Fixes #617.

I'm not 100% sure I'm understanding this correctly, but when SSE is enabled, in `Lut1DOp.cpp` the compiler complains because `Lut1D_Nearest()` is defined but not used. The function *is* used in the unit tests, so we include it only when SSE is disabled or we are compiling unit tests.

For `MathUtils.cpp` it is a similar story, but for some reason, even with SSE disabled, the compiler will complain that the `GetMxbResult()` function is defined but not used for the DLL (.so) build. The function *is* used for the static build of the library, so we include it in the static build and the unit tests.

The tests will run successfully, but I've only tested on Linux.

Tell me if I'm missing anything! :smiley:

PR #678, opened 2019-02-24
Cherry picked fixes for proposed 1.1.x release

michdolan:
This PR merges the two fixes from the master branch proposed for a new 1.1.x patch.

The changes to the apps (related to OpenImageIO) add a requirement for C++11 that was not previously present, so that impact needs weighing before we decide to merge this. I have successfully built and tested this against the latest OIIO release on Linux, but can also confirm it will not build with an earlier CXX standard. One solution may be to support both paths with some CMake magic.

Commits cherry picked from:
https://github.com/imageworks/OpenColorIO/commit/718cc7f9f7ebb189643b8691df7ef4d9a735adac
https://github.com/imageworks/OpenColorIO/commit/1ff22741ff577bd1be504452eb3fa4ce1a652370
(plus a later correction made to two ImageInput/ImageOutput mismatches)

### DIFF
--- a/src/apps/ocioconvert/main.cpp
+++ b/src/apps/ocioconvert/main.cpp
@@ -33,6 +33,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <OpenColorIO/OpenColorIO.h>
 namespace OCIO = OCIO_NAMESPACE;
 
+// NB: OIIO_VERSION >= 10903 requires C++11 or later
 #include <OpenImageIO/imageio.h>
 #include <OpenImageIO/typedesc.h>
 #if (OIIO_VERSION < 10100)
--- a/src/apps/ocioconvert/main.cpp
+++ b/src/apps/ocioconvert/main.cpp
@@ -321,7 +321,7 @@ int main(int argc, const char **argv)
         f->write_image(OIIO::TypeDesc::FLOAT, &img[0]);
         f->close();
 #if OIIO_VERSION < 10903
-        OIIO::ImageInput::destroy(f);
+        OIIO::ImageOutput::destroy(f);
 #endif
     }
     catch(...)
--- a/src/apps/ocioconvert/main.cpp
+++ b/src/apps/ocioconvert/main.cpp
@@ -119,7 +119,11 @@ int main(int argc, const char **argv)
     std::cerr << "Loading " << inputimage << std::endl;
     try
     {
+#if OIIO_VERSION < 10903
+        ImageInput* f = OIIO::ImageInput::create(inputimage);
+#else
         auto f = OIIO::ImageInput::create(inputimage);
+#endif
         if(!f)
         {
             std::cerr << "Could not create image input." << std::endl;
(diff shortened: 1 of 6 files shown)


michdolan:
Hitting same CI failures as #677 

RB-1.1 CI env might just need some tweaking. AppVeyor is failing as the VS2015 image is missing Boost. The Travis Mac Clang build is failing to "find a suitable distribution" of markupsafe.

lgritz:
Oh, oops, I didn't realize I was changing C++ standard requirements. (C++ 03 compat in 2019, ok...)

I don't know how to reconcile that. OIIO 2.x certainly is only C++11 or higher, but a patch release should definitely not force a different compiler.

michdolan:
I wonder if we could extend the use of the OIIO version macro to revert to the C++ 03 friendly constructors when appropriate. That or an explicit opt-in cmake property to choose.

michdolan:
I'll cherry pick #682 into this PR once that is merged into master

hodoulp:
> I'll cherry pick #682 into this PR once that is merged into master

As no C++ version is explicitly mentioned in the makefiles, the gcc default version is C++98 (for gcc version < 6.1) while Vc++ default is latest (i.e. depends on the Visual Studio version).  Bottom line, the ‘auto’ is problematic in that context.

gcc link: http://gcc.gnu.org/projects/cxx-status.html (bottom of the page)

michdolan:
'auto' in this case is a 'unique_ptr', so in either case, C++11 is the minimum.

With the latest commit ( https://github.com/imageworks/OpenColorIO/commit/65a08dbedb771adc62b35b71e2a8b637776b8504 ), C++11 is only required when building with OIIO_VERSION >= 10903. The default behavior would be to not build at all unless OIIO_VERSION < 10903 were being used, which is consistent with OCIO 1.1. 

Keeping the OIIO related commits would allow OCIO to be built against the latest OIIO optionally. The builder would have to explicitly specify a newer C++ standard on the commandline with -DCMAKE_CXX_FLAGS, but they would have the option to do so. We could update the build instructions and make this requirement clear. 

The alternative is to remove these commits from the potential patch release and not support a newer OIIO until OCIO v2, which is acceptable if this is deemed beyond the scope of a patch.

michdolan:
...or we target a 1.2 release and officially support building with a newer C++ standard, without changing library functionality.

hodoulp:
One potential solution to benefit from OIIO newer versions for OCIO v1, would be to explicitly enforce the C++11 constraint.

While keeping the current changes, one could add the following code somewhere in the ocioconvert and ociodisplay main files:

```
#if OIIO_VERSION >= 10903 && __cplusplus < 201103
#error "While not officially supported, compile the library with C++11 (or higher) to benefit from newer OIIO versions"
#endif
```

michdolan:
I tested the added #error directive with builds against older and newer OIIO. To note, when I did try to build OCIO with OIIO 2.0 and C++03, Both the compiler and OIIO gave clear errors about the incompatibility before OCIO did.

michdolan:
This PR will be merged in 24 hours if there is not additional feedback.

Issue #675, opened 2019-02-22
Add support for commonly used color functions


doug-walker:
There are a number of color functions that get used over and over and that currently need to be implemented as 1d or 3d LUTs.  Representing them as LUTs is undesirable since it is less accurate and often has problems with over-range values.  Also, on the GPU simply doing the math is often faster and reduces the usage of texture memory.

In theory, these functions could be implemented via a general purpose color language such as CTL.  However, in practice such languages are difficult to implement in a way that is fast, cross-platform, and robust.  Also, that approach does not necessarily provide for an inverse transform direction, which is vital for a number of important use-cases.

Therefore, we have found it very useful to provide a set of hard-coded color functions as common building block transforms.  It is straight-forward to do this in a way that is high-performance, cross-platform, robust, supports CPU and GPU, and that includes inverses.  (And although the formulas themselves are hard-coded, they take parameters that are adjustable and therefore cover a range of specific cases that all use the same underlying math.)

Because there is a fair amount of overhead involved in adding new ops, we prefer to combine similar functions into a single op.  However, if the community prefers to have more granularity at the user-facing Config file level, this is possible by exposing the functions as separate Transforms.  For example the ExponentTransform and ExponentWithLinearTransform are separate Transforms at the config level but under the hood they both instantiate a common Op type.

As described in the OCIO v2 proposal, we also want to provide optional built-in color space conversions, but a separate issue will be logged for that.  The present issue is to capture the work at the Transform and Op level to implement specific formulas or processing operators and is a prerequisite for efficient implementation of the built-in color space conversions.  (Several of the individual transforms/ops will typically need to be stacked together to do color space conversions.)

Here are the proposed new transforms (in addition to the existing ExponentTransform, LogTransform, etc.):

ExponentWithLinearTransform
-- Does functions such as sRGB and L*.

ExponentWithOffsetTransform
-- Implements some commonly used ICC parametric functions.

LogAffineTransform
-- Implements the classic Cineon lin-to-log type curves.

LogCameraTransform
-- Implements most camera OETFs including LogC, SLog3, and also ACEScct.

LogHDRTransform
-- Implements PQ and the HLG-OETF.

FixedFunctionTransform
[...]

PR #680, opened 2019-02-25
Adsk Contrib - Add the FixedFunctionOp support

hodoulp:
Please refer to #675 to have the complete rational behind the work. 
The pull request is only focusing on the **FixedFunctionTransform** part.


KevinJW:
As mentioned in yesterday's call, my concern with this is it is a collection of 'stuff', into which we just keep adding more stuff?

If it only had ACES transforms then maybe it has better cohesion, design wise, as it wouldn't keep growing.

The surround calculation taking a parameter also slightly stands out as being not a **fixed** function. Plus it isn't really ACES specific. Perhaps this would be better implemented in a future Rec 2100, along with other ITU HDR 'stuff', or do people think having two transforms with 'enum' style parameters is actually worse??

This reminds me of the discovery problem in dynamic languages like Python where kwargs, tells you nothing about what it is supposed to do.

I think this concern also extends to the config side, I think it might be useful to write out a config that mocks out the required transforms to do one or two of the current ACES RRT+ODT transforms and see what the long term will look like.

Kevin

doug-walker:
Thanks for the feedback Kevin.  Here's my take on the points you raised:

I agree that it is a high priority to keep the Transform API easy to understand, while at the same time adding the functions we need.  One approach would be to split FixedFunction into a bunch of atomic transforms that only do one thing.  The reason we did not propose that approach in this case is because of the sheer number of Transforms that would create, many of which would only be used in very specific circumstances.  

For transforms that are used a lot, I agree it is useful to have them as distinct entities.  So for example, Exponent + ExponentWithLinear, and Log + LogAffine + LogCamera, are all exposed as separate transforms rather than combining them into just a multi-purpose Exponent and Log transform, since these will get used in almost every config.  For the more rarely used functions, my opinion is that it is better to group those into one, or at most a small number of, multi-function transforms, to avoiding cluttering the API with a huge number of transforms, of which only a handful are used regularly.

I thought about calling it ACESTransform, but ACES is not the only standard that requires a few similar special functions.  Rec.2100 is an example where an algorithm is needed that is virtually the same as an algorithm needed for ACES, so it made more sense to group the code into a common module to avoid code duplication.  (Rec.2100 HLG contrast adjustment vs. ACES Output Transform surround adjustment.)

Per your question about what an ACES RRT+ODT would look like, for ACES v1.0/1.1 it would basically be this:  FixedFunction (glow v1), FixedFunction (red modifier v1), Range, Matrix, Range, Matrix, Lut1D, Matrix, Matrix, Matrix, FixedFunction (surround v1), Matrix, Matrix, Matrix, Range, Exponent.  (Some of the matrices would obviously get combined during optimization, but it's clearer to represent them separately.)

Regarding your observation that one of the functions has the ability to take a parameter, the "FixedFunction" name was not meant to imply that no parameters were allowed.  It was meant to convey the idea that these are highly specialized, rather than general purpose functions.  As always, naming is hard and if the group is able to coalesce around a better name, we are open to that.

Doug

hodoulp:
Following the TSC meeting, here is a friendly reminder that the branch will be merged this week unless someone has comments.

hodoulp:
`Travis` & `AppVeyor` are blocked without any ability to resume the builds. So, I will push a 'dummy commit' hoping to unlock the builds.

hodoulp:

include/OpenColorIO/OpenColorTransforms.h
```diff
@@ -563,8 +563,63 @@ OCIO_NAMESPACE_ENTER
...
+        virtual void validate() const;
+
+        //!cpp:function::
+        virtual FixedFunctionStyle getStyle() const;
+        //!cpp:function:: Select which algorithm to use.
+        virtual void setStyle(FixedFunctionStyle style);
+
+        //!cpp:function::
+        size_t getNumParams() const;
+        //!cpp:function:: Set the parameters (for functions that require them).
```
michdolan: Are the required params for each FixedFunction style documented somewhere (outside of looking at the lower level implementation classes)? AllocationTransform, which shares a similar interface, was never super clear in regards to how many and what the params are for each allocation option. With this Transform, there are many more styles, so I would propose a brief overview of the params for each here in OpenColorTransforms.h, or along with the enum in OpenColorTypes.h, with the intent being to be visible to the API docs.

I also think the explanation about which ACES versions the specific functions are defined in (like in FixedFunctionOpData.h) would be useful in the API docs as well, to be obvious about "03" referring to ACES 0.3/0.7 and "10" to ACES 1.0.

src/OpenColorIO/ops/FixedFunction/FixedFunctionOpData.cpp
```diff
@@ -0,0 +1,386 @@
...
+            throw Exception(ss.str().c_str());
+        }
+    }
+    else
+    {
+        if(m_params.size()!=0)
+        {
+            std::stringstream ss;
+            ss  << "The style '" << ConvertStyleToString(m_style, true) 
+                << "' must have zero parameter but "
```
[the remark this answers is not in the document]
hodoulp: Fixed

include/OpenColorIO/OpenColorTransforms.h
hodoulp: Good catch. We will improve the OCIO public API documentation.

hodoulp: Done.

hodoulp:
Warning about the pull request merge by tomorrow morning.

PR #681, opened 2019-02-25
Adsk Contrib - Improve error messages for the config file parsing.

hodoulp:
While working on another feature, we discovered that the config parser did not add the line number to the warnings & errors. We internally discussed about the findings and agreed that the line number is beneficial on long-term (i.e. for example, the ACES 1_0_3 config has ~ 4k lines).

Here is the work to haver the line number and to improve some error messages.


src/OpenColorIO/OCIOYaml.cpp
```diff
@@ -1739,8 +1779,8 @@ OCIO_NAMESPACE_ENTER
                     if(second.Type() != YAML::NodeType::Map)
                     {
                         std::ostringstream os;
-                        os << "'displays' field needs to be a (name: key) map.";
-                        throw Exception(os.str().c_str());
+                        os << "The value type of 'displays' key needs to be a map.";
+                        throwValueError(node.Tag(), first, os.str());
                     }
```
michdolan: This is nit picky, but some of the throwValueError msg language is inconsistent:

```
305: << ", the value parsing of the key '" << keyName
1718: os << "The value type of 'environment' key needs to be a map.";
1764: os << "The value type of the 'roles' key needs to be a map.";
1782: os << "The value type of 'displays' key needs to be a map.";
```
Perhaps instead:
```
305: << ", the value parsing of the '" << keyName << "' key "
1718: os << "The value type of the 'environment' key needs to be a map.";
1764: os << "The value type of the 'roles' key needs to be a map.";
1782: os << "The value type of the 'displays' key needs to be a map.";
```
Or:
```
305: << ", the value parsing of the key '" << keyName
1718: os << "The value type of the key 'environment' needs to be a map.";
1764: os << "The value type of the key 'roles' needs to be a map.";
1782: os << "The value type of the key 'displays' needs to be a map.";
```
Or any other consistent wording you prefer.

hodoulp: Good point. I will fix that.

PR #684, opened 2019-02-27
Adsk Contrib - Remove unused functions

BernardLefebvre:
Some function in MathUtils.cpp are already defined and implemented (with different case) in header file.


hodoulp:
This PR qualifies for single approver rules and could be merged next week unless there is additional feedback.

michdolan:

PR #687, opened 2019-02-27
Adsk Contrib - Update Windows documentation

hodoulp:
The Windows batch file example was for OCIO v1.
I've updated the content to OCIO v2, added dependencies for GPU unit tests and added some explanations.


michdolan:
This PR qualifies for single approver rules. I'll merge it in 24 hours unless there is additional feedback.

PR #688, opened 2019-02-27
Adsk Contrib - Extract test files.

BernardLefebvre:
Having all test files in a single archive makes it challenging to do changes in in different PR.
This is removing the archive and ectracting to current tests files so that we can track changes with upcoming PR.


hodoulp:

hodoulp:

Issue #690, opened 2019-02-28
How to manage Third-Party libraries?


hodoulp:
The question is the same for all external dependencies, but solution could be different depending of the third-party library usage within OCIO.

OCIO has three main types of third-party library dependencies:
1. Libraries for the OCIO library such as yaml, eXpat
2. Libraries for the OCIO apps such as lcms2, glew, OIIO, OpenEXR
3. Libraries for documentation such as JinJa, Sphinx

Note: Java support is unclear as the jar file is seven years old.

Whatever would be the long-term solution to manage third-party libraries (from points 1 & 2), the 'mechanism' must support:
- Select a specific version
i.e. to ensure a quality level, use third-party library versions validated by the CI builds i.e. avoid 'latest version' behavior
- Apply an arbitrary patch file
i.e. Add specific compilation/link flags, Add custom namespace, Change code to remove warnings/build breaks...
- Compile on the local machine using local OCIO CMake settings
i.e. propagate debug/release mode, use local compiler (i.e. both are extremely sensible for the Windows platform)
- Allow 'use external libraries'
i.e. Linux common make behavior (Refer to the next point for Windows platform as no such /usr/local equivalent exists)
- Allow install within the OCIO build directory (i.e. should be the default behavior)
i.e. global install on Windows platform is problematic as everything could end up in system32 with binary name conflicts. Plus, the build directory must ideally be self-contained (and 3rd-party lib. be static libraries) which is mandatory when working on different projects with library overlaps.

Compiling Open Source project is generally challenging on Windows, so I propose that OCIO library successfully compiles by default imposing eXpat & yaml presence.

Please let us know your ideas/comments,
Patrick

**Below is the original email thread:**

```
On Tue, Feb 19, 2019 at 3:16 PM Larry Gritz <lg@imageworks.com> wrote:
I think that we had some consensus that if those binary dependencies could be hosted someplace reliable, then it should be fairly easy for us to download them at build time rather than have them in the main OCIO repo. The biggest concern was that some of the external dependencies might be either unsupported, or not supported for the particular commit or version that we desire, but that would be alleviated if we know those exact versions were available even if "we" somehow were responsible for hosting them (outside the OCIO repository proper).



On Tue, Feb 19, 2019 at 12:07 PM Steve Winslow <swinslow@linuxfoundation.org> wrote:
Hello all,

I understand that John has already discussed with you some of the feedback from the intake license scan for OpenColorIO. Here are a few additional details:

[...]

PR #691, opened 2019-03-04
Add notes for 03-04-2019 TSC meeting

michdolan:


docs/tsc/meetings/2019-03-04.md
```diff
@@ -0,0 +1,92 @@
...
+    - **TODO**: Michael to create new Python API design and submit PR for feedback/discussion/approval.
+
+* Apple has deprecated OpenGL
+    - Mark: One option could be to build a bridget between Vulcan and Metal. Proposes use of MoltenVK.
+    - Sean: SPIRV-Cross also worth investigation.
+    - Doug: We're not sure *when* OpenGL support will actually be removed.
+        - It's a matter of when, not if...
+        - Agrees MoltenVK is a good consideration.
+    - Carl: Different parties within Apple have different opinions of timeline for dropping support.
+    - Mark: Nothing in OCIO's design is preventing Metal support from being added now.
```
hodoulp: *Note:* That should be the approach to investigate first as the library is built to support several different languages and language versions.

docs/tsc/meetings/2019-03-04.md
```diff
@@ -0,0 +1,92 @@
...
+    - Larry: Circle has better, faster CPUs than Travis
+    - **TODO**: Once ASWF/LF commit to CI choice, Mark/Sean to investigate transition to new CI infastructure
+        - If OpenVDB adopts new system first, we can reference their setup.
+
+* Moving OCIO to ASWF Git organization
+    - Legal approvals for OCIO domain tranfer are complete!
+    - The GitHub account move should be fairly straightforward, but may bring down CI builds for a day or two diring transition.
+    - **TODO**: LF needs Travis and AppVeyor accounts created before we can move repo, to host CI builds.
+    - **TODO**: Larry/Michael to work with John/Thanh to facilitate the account move.
+    - There may be a requirement to fix some of the embedded external dependencies *before* the repo move to satisfy detected license issues (from the license scan).
```
hodoulp: *Note:* I've created #690 to centralize the discussions on that topic.

docs/tsc/meetings/2019-03-04.md
```diff
@@ -0,0 +1,92 @@
...
+        - Supports GPUs when hosted on our servers, but pricing may be a concern.
+        - Proposed AppVeyor for Windows support for now.
+    - OpenVDB reports that Circle significantly faster than Travis
+    - Larry: Circle has better, faster CPUs than Travis
+    - **TODO**: Once ASWF/LF commit to CI choice, Mark/Sean to investigate transition to new CI infastructure
+        - If OpenVDB adopts new system first, we can reference their setup.
+
+* Moving OCIO to ASWF Git organization
+    - Legal approvals for OCIO domain tranfer are complete!
+    - The GitHub account move should be fairly straightforward, but may bring down CI builds for a day or two diring transition.
```
hodoulp: Typo 'diring' -> 'during' ?

hodoulp: *Comment:* The two repos must coexist for a while as many pull requests already exist and we also have several internal code reviews.

docs/tsc/meetings/2019-03-04.md
```diff
@@ -0,0 +1,92 @@
...
+
+* Apple has deprecated OpenGL
+    - Mark: One option could be to build a bridget between Vulcan and Metal. Proposes use of MoltenVK.
+    - Sean: SPIRV-Cross also worth investigation.
+    - Doug: We're not sure *when* OpenGL support will actually be removed.
+        - It's a matter of when, not if...
+        - Agrees MoltenVK is a good consideration.
+    - Carl: Different parties within Apple have different opinions of timeline for dropping support.
+    - Mark: Nothing in OCIO's design is preventing Metal support from being added now.
+        - Doug: Adds that per-language texture support is already present as well.
```
doug-walker: That's not quite what I said.  OCIO already has the ability to emit different shader code based on language (e.g. GLSL vs HLSL).  The point that I was trying to make was that it is not only a question of the shader language, it is also a question of how to manage textures under a low-level paradigm such as Vulkan/Metal and this will also need investigation.

Also, my statement farther up of "when, not if" was not about when Apple will remove support for OpenGL.  What I meant, was that OCIO supporting low-level graphics APIs including Vulkan and Metal is a question of when, not if.  In other words, there is no doubt that we will need that for our products at some point.

doug-walker: So, I'll propose the following summary.  Here you could write:
- Doug: Investigation is needed not only on the shader language but also on how textures will be managed under a low-level graphics API.

And above you could write:
- Doug: We're not sure yet *when* OpenGL support will actually be removed by Apple.
   - From my point of view, OCIO will definitely need to support low-level graphics APIs such as Vulkan and Metal, so the question is more when than if.
   - Agreed MoltenVK is one option to investigate.

michdolan: Thanks for the clarification @doug-walker

PR #700, opened 2019-03-16
Adsk Contrib - Various adjustments to the Windows DLL

hodoulp:
The pull request includes:
1. Any Windows DLL must have some extra information embedded in the library binary. Waiting the completion of the ASWF transition, some are still missing. But the mechanism is now in place.
2. To avoid the infamous DLL hell on Windows change the binary file name to mimic the Linux good practice.  Refer to #605.
3. Add the WIN32_LEAN_AND_MEAN define to strip 'not often used' Windows libraries which is defined as a good practice. However, I double-checked the binary size without noticing any size change.


hodoulp:
The pull request is the continuation of #596.

src/OpenColorIO/CMakeLists.txt
```diff
@@ -143,13 +156,29 @@ if(OCIO_USE_SSE)
...
+        PRIVATE
+            XML_STATIC
+    )
+
+    if(BUILD_SHARED_LIBS)
+
+        # Impose a versioned name on Windows to avoid binary name clashes 
+        # between different versions of the same library.
+
+        # Note: Do not append the PATCH to allow Apps use the
```
michdolan:
```suggestion
        # Note: Do not append the PATCH to allow Apps to use the
```

hodoulp: Fixed

hodoulp:

hodoulp:

PR #702, opened 2019-03-19
Adsk Contrib - Improve the compilation options

hodoulp:
The work only adjusts the generated makefiles to:
1. Avoid using compiler specific C++ extensions (good practice)
2. Request the shared library to have position-independent code i.e. gcc -fPIC option (good practice)


hodoulp:

PR #703, opened 2019-03-20
Adsk Contrib: Add a file to control eol of test files.

BernardLefebvre:
Make sure that git does not control the end of line character used for test files. 
We favour lf on all platforms for most files.
One matrix_windows.clf is used to test crlf on all platforms.


.gitattributes
```diff
@@ -0,0 +1,15 @@
+# Control the end-of-line (i.e. eof) type for all the test files.
```
michdolan:
```suggestion
# Control the end-of-line (i.e. eol) type for all the test files.
```

Should this be eol (instead of eof)?

hodoulp:

hodoulp:

PR #706, opened 2019-03-20
Adsk contrib - Fix a bit depth issue for the Range Op

hodoulp:
Here is the work done:
1. The Range Op input and output bit depths were lost when cloning an instance. I fixed the problem and added a unit test to validate the fix.
2. I changed some Ops to mimic the 'good practice' for clone methods to unify code across Ops & decrease bugs.


hodoulp:

hodoulp:

PR #707, opened 2019-03-21
Adsk Contrib - Fix GPU tests build

BernardLefebvre:
A bad merge conflict resolution did break the GPU tests build (that is not built by the build system).


hodoulp:
***Being a build break fix:*** 24-hour merge warning. Please have a look if you haven't yet.

PR #712, opened 2019-03-27
Adsk Contrib - Add Reference OpData

BernardLefebvre:
This is a step towards completing the CTF file format reader.  Specifically, this PR implements the CTF Reference op feature that allows a CTF file to incorporate a list of ops from another CTF file.  

The PR implements CTF parsing of the Reference element.  A Reference OpData class is added that holds the info extracted from the CTF file. The implementation relies heavily on the existing FileTransform, so we are not adding a ReferenceTransform or even a ReferenceOp.

As part of this work, we also needed to upgrade the search path mechanism.  The existing mechanism did not work properly on Windows since it represents all paths as one long string using ':' as a separator.  We added the option to represent the paths in the config file as a list of separate strings and added calls in the public API for interacting with the path in that way.


hodoulp:

hodoulp:
24-hour reminder of the coming merge. Please have a look if you haven't yet.

Issue #709, opened 2019-03-26
Build error on Ubuntu 18.04 (Bionic Beaver).


KelSolaar:
Hi,

I'm trying to build within a Docker container to get `ociolutimage` working on macOs without busting my Homebrew sandcastle. I'm getting those errors when building with Ubuntu 18.04 (Bionic Beaver):

```
[ 52%] Building CXX object src/OpenColorIO/CMakeFiles/OpenColorIO.dir/ops/Gamma/GammaOpCPU.cpp.o
/home/OpenColorIO/src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp: In member function 'virtual void OpenColorIO::v2_0dev::GammaBasicOpCPU::apply(float*, long int) const':
/home/OpenColorIO/src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp:217:19: error: 'powf' was not declared in this scope
         rgba[0] = powf(pixel[0] * m_inScale, m_redGamma) * m_outScale;
                   ^~~~
/home/OpenColorIO/src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp: In member function 'virtual void OpenColorIO::v2_0dev::GammaMoncurveOpCPUFwd::apply(float*, long int) const':
/home/OpenColorIO/src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp:317:33: error: 'powf' was not declared in this scope
         const float data[4] = { powf(pixel[0] * red[0] + red[1], red[2]) * m_outScale,
                                 ^~~~
/home/OpenColorIO/src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp: In member function 'virtual void OpenColorIO::v2_0dev::GammaMoncurveOpCPURev::apply(float*, long int) const':
/home/OpenColorIO/src/OpenColorIO/ops/Gamma/GammaOpCPU.cpp:422:33: error: 'powf' was not declared in this scope
         const float data[4] = { powf(pixel[0] * m_inScale, red[0]) * red[1] - red[2],
                                 ^~~~
src/OpenColorIO/CMakeFiles/OpenColorIO.dir/build.make:1454: recipe for target 'src/OpenColorIO/CMakeFiles/OpenColorIO.dir/ops/Gamma/GammaOpCPU.cpp.o' failed
CMakeFiles/Makefile2:846: recipe for target 'src/OpenColorIO/CMakeFiles/OpenColorIO.dir/all' failed
[...]

PR #715, opened 2019-03-28
Adsk Contrib - Improve Apps error reporting

hodoulp:
Following investigations around the issue #709, I discovered that some apps succeeded while producing a 'faulty' output i.e. ociolutimage. So, I've updated some Apps to correctly report errors.

*Note:* That does not 'fix' the problem reported by #709


src/apps/ociodisplay/main.cpp
```diff
@@ -129,13 +129,20 @@ static void InitImageTexture(const char * filename)
...
-#if (OIIO_VERSION >= 10800)
+            const bool ok = f->read_image(
+#if OIIO_VERSION >= 10800
                 OIIO::TypeFloat, 
 #else
                 OIIO::TypeDesc::TypeFloat, 
 #endif
                 &img[0]);
+
+                if(!ok)
```
michdolan: Lines 140-144 indentation should match line 132

src/apps/ocioconvert/main.cpp
```diff
@@ -436,7 +436,20 @@ int main(int argc, const char **argv)
         img.resize(imgwidth*imgheight*components);
         memset(&img[0], 0, imgwidth*imgheight*components*sizeof(float));
         
-        f->read_image(OIIO::TypeDesc::TypeFloat, &img[0]);
+        const bool ok = f->read_image(
```
michdolan: Can we use OIIO::TypeDesc::FLOAT here directly to avoid the version check? The f->write_image calls are using it, for example.

lgritz: Yeah, that effectively means the same thing.

lgritz: This is only necessary, BTW, if you are trying to support OIIO <= 1.7, which even I no longer do.

hodoulp: Thanks. I will clean the code.

src/apps/ociodisplay/main.cpp
hodoulp: Fixed

src/apps/ocioconvert/main.cpp
hodoulp: Fixed

PR #717, opened 2019-04-01
Adsk Contrib - Create an OpRcPtrVec managing the Op bit depth

hodoulp:
All Ops now have input and output bit depth so the OpRcPtrVec must manage consistency between the op bit depths. The output bit depth should match the input bit depth of the following op.


scoopxyz:
Just confirming that in practice only the first and last Op would potentially be an integer data type and that a transform wouldn't be able to "truncate" a bit-depth in the middle of an Op chain?

hodoulp:
The content of a CTF file is up to the user. However, the processing only manages any arbitrary bit-depth for the first (i.e. the input bit-depth), and the last (i.e. the out bit-depth) Ops and the intermediate processing is always in RGBA F32 (i.e. all the other ops).

*Note:* The pull request #699 better highlights the point.

doug-walker:
Sean, good question.  Please keep in mind that the bitDepth attributes in a CLF file do NOT specify processing precision.  Rather, they specify how to interpret the parameter values.  For example, if you have a Lut3D with an outBitDepth of 10i, then you would expect the Array values to nominally be scaled between 0 and 1023.  It does NOT mean that you're supposed to process/apply the LUT at 10 bit precision.

The main purpose of this PR is to put in place a consistent way of enforcing these conventions when ops are added or removed from an OpVec.  For example, if you insert an op into an OpVec, you'd like to ensure there is a call the adjacent op's setBitDepth to give it a chance to make any necessary adjustments to its parameter scaling.

Ideally the "bitDepth" attributes in the CLF spec would have been named "parameterScale" or something like that.  But as we know, naming is hard.

As Patrick wrote, PR 699 deals with the processing precision and there is a separate PixelFormat variable used to control that.  These are related but different topics.

src/OpenColorIO/ops/Truelight/TruelightOp.cpp
```diff
@@ -61,21 +61,27 @@ OCIO_NAMESPACE_ENTER
                         TransformDirection direction);
             virtual ~TruelightOp();
             
-            virtual OpRcPtr clone() const;
+            OpRcPtr clone() const override;;
```
michdolan: double ";" after some of the `override` additions in this file.

hodoulp: Fixed.

hodoulp:
Unfortunately, the fix involves much more work than expected and conflicts with recent pull requests.  So, I close the pull request and will create a new one later.

PR #718, opened 2019-04-02
TSC Meeting notes

hodoulp:


hodoulp:
This PR qualifies for single approver rule and could be merged this week if no objections.

michdolan:
My 2 cents on these PRs is that we merge them after 48 hours if nobody has voiced objections. Notes can always be amended later if needed. Open to other opinions of course.

hodoulp:
Good. I will merge it tomorrow unless there are any objections.

scoopxyz:
Yeah, these are pretty low impact, not sure if we can do conditionwl rules. But 48 hours is more than enough time.

lgritz:
IMHO, all the usual rules should be exempted for PRs that are purely administrative or internal documentation that can't break builds, introduce bugs, or confuse users. I don't think it even needs a single "approval", if the person writing the notes already has commit privileges.

I think it's more than sufficient to post a PR with the simple message "Here are the minutes from this week's meeting, please suggest edits in the next 48 hours, at which time I will merge." And then do it.

There is already a flood of truly important messages about neglected PRs and deadlines thereof. There's no need to dilute that stream with repeated notices about meeting minutes. There is no negative consequence at all of merging quickly and without fanfare, and occasionally needing to make corrections in a subsequent commit.

My view on it, anyway.

PR #720, opened 2019-04-03
Rebuild gh-pages from OCIO 1.1.1

michdolan:
The 1.1.1 release didn't deploy the updated docs to gh-pages because the current Travis build only builds the PDF docs. This PR updates gh-pages from my local build of the html docs.


michdolan:
@scoopxyz , do you mind taking a look at this since you have done previous gh-pages deployments?

Issue #609, opened 2018-10-16
Support for integer pixel formats


doug-walker:
OCIO currently only accepts floating-point image buffers. However in many scenarios one is starting with an integer buffer. Furthermore, the first operation often performed is a 1d-LUT. So rather than accepting an integer buffer and doing a fast look-up, OCIO currently converts to float and does interpolation rather than a simple look-up.

We propose to add direct support for 8, 10, 12, and 16-bit integer and 16-bit half-float formats with both RGB and RGBA layouts. (We could also add support for packed 10-bit formats and BGR formats, is that of interest?)

When a 1d-LUT is the first operator in the chain, it will be interpolated to the needed bit-depth and then a look-up will be done on the pixels. This gives a big performance boost with no loss of precision.

Support for these formats will also be done for output, making it possible for OCIO to provide a correctly rounded and scaled integer result when needed. This can also provide a performance gain by avoiding a separate copy/conversion of the image buffer.

This would not affect internal calculations, which would continue to be done at 32-bit float precision.

doug-walker:
This issue has now been implemented.

PR #722, opened 2019-04-05
Adsk Contrib - Add native bit depth support to the 1D LUT Op

hodoulp:
From the issue #609:
>OCIO currently only accepts floating-point image buffers. However, in many scenarios one is starting with an integer buffer. Furthermore, the first operation often performed is a 1d-LUT. So rather than accepting an integer buffer and doing a fast look-up, OCIO currently converts to float and does interpolation rather than a simple look-up.

The complete implementation of the feature requires several steps. But the pull request #699 added the structural changes:
>The pull request has **all the changes to the OCIO v2 public API to natively support the pixel formats while maintaining full backward compatibility with version 1**.  As the final implementation would have created a huge pull request, only minimal set of changes (with lot of TODOs and comments) are in to limit the size and ease the review.

The current pull request adds native bit depth support to the 1D LUT Op to benefit from the look-up table for integer bit depths i.e. one major performance optimization for the CPU processing. Several unit tests now validate the new capability.

*Note*:
One or two pull requests are still expected to complete the feature.


hodoulp:

PR #723, opened 2019-04-08
Adsk Contrib - Enable 'warnings as error' for the Linux CI builds

hodoulp:
Enable 'warning as error' for the Linux CI builds to trap all potential issues using the compiler capabilities. Only the Linux CI builds are warning free for now.

*Note:* By default, the 'feature' is OFF so anyone cloning and compiling the library could not be blocked because of an untested combination of platform & compiler versions.


CMakeLists.txt
```diff
@@ -54,8 +54,8 @@ if(MSVC)
 	# Because our Exceptions derive from std::runtime_error we can safely disable this warning
 	set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} /wd4275")
 else()
-	#TODO: Fix warnings before enabling the flag.
-	#set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} -Wall")
+    # TODO: OCIO being under active development, unused functions are fine for now.
```
michdolan: [minor] Noting that this line uses spaces for indentation, where others are using tabs.

hodoulp: Even if the file is using tabulations, the coding style requires to not use them. So, I will fix that for short-term i.e. put a tabulation instead of spaces.

hodoulp:

Issue #577, opened 2018-09-20
Config MatrixTransform generation low precision


sobotka:
When using the MatrixTransform to generate a configuration line, the precision appears fixed at 6 decimal places. The configuration's YAML seems to be able to handle much more than this, including scientific notation if hand coded.

Is there a precision option I'm missing?

KevinJW:
Relates to #389 

Kevin

hodoulp:
I did some investigations on that topic to better understand the Matrix problem with the OCIO config file.

Here is some code to understand the float & double precisions using `std::numeric_limits<T>`:
```
#include <iostream>
#include <limits>

int main()
{
   std::cout << "float  => "
        << " max_digits10 = " << std::numeric_limits<float>::max_digits10 << ", "
        << " digits10 = " << std::numeric_limits<float>::digits10 << ", "
        << " digits = " << (2 + std::numeric_limits<float>::digits * 3010/10000)
        << std::endl;

   std::cout << "double => "
        << " max_digits10 = " << std::numeric_limits<double>::max_digits10 << ", "
        << " digits10 = " << std::numeric_limits<double>::digits10 << ", "
        << " digits = " << (2 + std::numeric_limits<double>::digits * 3010/10000)
        << std::endl;

    return 1;
}
```
The output is:
```
float  =>  max_digits10 = 9,  digits10 = 6,  digits = 9
double =>  max_digits10 = 17,  digits10 = 15,  digits = 17
```

When changing the code (i.e. OCIO & yaml following #389 changes) to use the `std::numeric_limits<float>::max_digits10`, the serialization/deserialization of floats have some rounding issues (i.e. various serialization/deserialization unit tests fail because of  string comparisons for float values: '0.11' versus '0.110000007'`).

So, I think that the good long-term solution is to have a Matrix of doubles (instead of floats) so double precision values will be saved/loaded by the current yaml version. However, the MatrixTransform public API would have to support doubles (and floats for backward compatibility).

**Note:**
The new Range Op only supports double values without any serialize/deserialize rounding issue (i.e. a Range unit test is validating some double values with 14 bit precision).

hodoulp:
@sobotka Is the problem now fix in master following the merge of the pull requests #724 & #883?

sobotka:
Will give it a roll this weekend. Thanks so much for pushing some changes @hodoulp !

doug-walker:
This was fixed in PR #724.

PR #724, opened 2019-04-08
Adsk Contrib - Improve the Matrix value precision in config file

hodoulp:
Improve the precision of the Matrix values in a config file to overcome the 6-digits limitation from YAML current version.  As the underlying MatrixOp was already moved to double to increase internal computation (i.e. combine, ...) precision, the MatrixTransform now provides double precision access.

*Note:* That's a better long-term solution compare to 'patch' the current YAML library as it was previously discussed.  With the current implementation, the precision is now preserved whatever is the used YAML version.

This will also address Issue #577.


doug-walker:
@KevinJW, this is our preferred solution to solving this issue.  Since you wrote PR #389, could you please review?

src/OpenColorIO/MathUtils.h
```diff
@@ -30,101 +30,89 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 #define INCLUDED_OCIO_MATHUTILS_H
 
-#include <cmath>
-#include <vector>
+#if defined(_WIN32) || defined(_WIN64) || defined(_WINDOWS) || defined(_MSC_VER)
+#ifndef WINDOWS
+#define WINDOWS
+#endif
+#endif
+
```
KevinJW: Is this to avoid min/max pollution or something?  would defining  NOMINMAX be less ugly if so?

hodoulp: That's dead code I forgot to remove in my various attempts to fix the Clang 9 warnings.

hodoulp: Fixed.

hodoulp:
@KevinJW I did the change, and plan to merge it next tuesday. Please let me know if you have any comment.

hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch could be merged this week. Unless there are some objections I will merge it tomorrow afternoon (New-York timezone).

PR #727, opened 2019-04-12
Fix minor typos

remia:
*  Missing fields in `Baker` assignement operator
*  Missing non const method raising a warning when using for range loop : `warning: 'begin' and 'end' returning different types ('std::__1::__wrap_iter<std::__1::basic_string<char> *>' and 'std::__1::__wrap_iter<const std::__1::basic_string<char> *>') is a C++17 extension [-Wc++17-extensions]`

### DIFF
--- a/src/OpenColorIO/Baker.cpp
+++ b/src/OpenColorIO/Baker.cpp
@@ -79,6 +79,8 @@ OCIO_NAMESPACE_ENTER
             {
                 config_ = rhs.config_;
                 formatName_ = rhs.formatName_;
+                type_ = rhs.type_;
+                metadata_ = rhs.metadata_;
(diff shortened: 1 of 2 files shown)


hodoulp:
Thanks @remia to propose fixes.
Could you explain on which platform/compiler versions you found these issues?

remia:
@hodoulp the warning is present on macOS Mojave 10.14.4 using the XCode compiler :
```
Apple LLVM version 10.0.1 (clang-1001.0.46.3)
Target: x86_64-apple-darwin18.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

PR #728, opened 2019-04-13
Add openglbuilder target install

michdolan:
OpenGLBuilder is really useful for integrating the new GPU renderer. Prior to the the repo refactor it was installed to share/, but now it is no longer being installed. This PR restores the install behavior, and fixes a few typos.

Issue #726, opened 2019-04-11
CMake issue with existing OCIO installs


doug-walker:
Because of the way CMake is currently set up, it will sometimes add /usr/local/include to the include path in order to pick up IlmBase or GLEW.  If there happens to also be an old install of OpenColorIO also in /usr/local/include, then the make will fail since the old/installed versions of the OpenColorIO headers are picked up rather than the development versions.

Note that simply uninstalling the existing OCIO from /usr/local may not be an ideal option since it may be a dependency for something else, e.g. OpenImageIO.  For example, someone may want to do development work on OCIO v2 but still have OCIO v1 installed for use with OIIO.

This issue is related to issue #690.  

One work-around (other than to remove the installed version of OCIO) was to comment out all of share/cmake/FindIlmBase.cmake 
and to replace the line:
    find_package(GLEW REQUIRED)
with:
if(NOT APPLE)
    find_package(GLEW REQUIRED)
endif()
in the files: 
src/apps/ocioconvert/CMakeLists.txt
src/apps/ociodisplay/CMakeLists.txt
src/utils/openglbuilder/CMakeLists.txt
tests/gpu/CMakeLists.txt

doug-walker:
Fixed by PR #730.

PR #730, opened 2019-04-17
Adsk Contrib - Fix install issues around the OpenGLBuilder utility

hodoulp:
Prior to the repo. refactoring the OpenGLBuilder utility was installed in the `/share` directory. But it's now no longer installed anywhere. Following investigations around this issue, we discovered few inconsistencies and limitations. The focus of these adjustments is mainly to ease the maintenance and the install of the oglBuilder library:
1.	Consistency in the library names i.e. all names now start with `OpenColorIO`,
2.	Consistency in the install paths i.e. all include files use the same base path,
3.	More granularity control over needed external libraries (i.e. experienced conflicts with an installed`OCIOv1` #726),
4.	Add the OpenColorIO namespace to enforce version control between OpenColorIO libraries.
 

*Note:* Merged changes from PR #728 to have several typo fixes.


hodoulp:

PR #731, opened 2019-04-18
Adsk Contrib - Fix CTF Reference "inverted" parsing issue and edit a test file to validate.

BernardLefebvre:
Fix CTF Reference "inverted" parsing issue and edit a test file to validate.
inverted="true" results into an inverted reference, anththing else, included inverted="false" results into a forward reference.


hodoulp:

PR #732, opened 2019-04-18
Fix typo in FormatRegistry::getFormatNameByIndex

remia:


hodoulp:
Good catch. Would you mind creating (or updating) a unit test to demonstrate the fix?

hodoulp:
@michdolan When finalized, I propose that the fix be integrated in RB-1.1.

remia:
Came up with this unit test, not really elegant but should catch the error (at least for me).

michdolan:
@hodoulp agreed. This is a good candidate for a 1.1.2

michdolan:
24 hour merge notice.

hodoulp:
Sorry to be late. I will have a look today.

src/OpenColorIO/transforms/FileTransform.cpp
```diff
@@ -985,10 +985,34 @@ OIIO_ADD_TEST(FileTransform, AllFormats)
     OIIO_CHECK_ASSERT(FormatExtensionFoundByName("spi3d", "spi3d"));
     OIIO_CHECK_ASSERT(FormatExtensionFoundByName("spimtx", "spimtx"));
     OIIO_CHECK_ASSERT(FormatExtensionFoundByName("vf", "nukevf"));
+}
+
+OIIO_ADD_TEST(FileTransform, FormatByIndex)
+{
+    OCIO::FormatRegistry & formatRegistry = OCIO::FormatRegistry::GetInstance();
 
+    for (unsigned int i = 0; i < formatRegistry.getNumRawFormats(); ++i)  {
```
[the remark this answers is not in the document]
hodoulp: BTW, some Linux platforms also failed (i.e. Travis CI builds).

src/OpenColorIO/transforms/FileTransform.cpp
```diff
@@ -985,10 +985,34 @@ OIIO_ADD_TEST(FileTransform, AllFormats)
...
+
+        if (i < formatRegistry.getNumFormats(OCIO::FORMAT_CAPABILITY_READ)) {
+            OIIO_CHECK_ASSERT(formatRegistry.getFormatNameByIndex(OCIO::FORMAT_CAPABILITY_READ, i) != "");
+            OIIO_CHECK_ASSERT(formatRegistry.getFormatExtensionByIndex(OCIO::FORMAT_CAPABILITY_READ, i) != "");
+        }
+        else {
+            OIIO_CHECK_ASSERT(formatRegistry.getFormatNameByIndex(OCIO::FORMAT_CAPABILITY_READ, i) == "");
+            OIIO_CHECK_ASSERT(formatRegistry.getFormatExtensionByIndex(OCIO::FORMAT_CAPABILITY_READ, i) == "");
+        }
+    }
```
hodoulp: In the unit test, the variable `i` represents the maximum number of FileFormats. And the same variable is also used as an index for the 'read' file formats and the 'write' file formats. However, a file format could only offer a 'read' capability so the indexes ('read' vs. 'write' file formats) could be different...

hodoulp: I did a copy & paste of the unit test, and it fails on my Windows machine in debug mode.  OIIO_CHECK_ASSERT() are all comparing pointers instead of strings.

src/OpenColorIO/transforms/FileTransform.cpp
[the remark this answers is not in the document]
remia: Thanks @hodoulp, will fix this.

What I'm not sure is why the last commit I made (fd0223d5c7335003c39a6bcf8e4b012ccf4b18a3) was successful and not yours. Is it because of the rebase from master you made (that introduce new warnings as errors flags) ?

src/OpenColorIO/transforms/FileTransform.cpp
remia: Agreed that the test was not clear nor necessarily valid, I updated it to check read and write capabilities independently.

src/OpenColorIO/transforms/FileTransform.cpp
```diff
@@ -985,10 +985,33 @@ OIIO_ADD_TEST(FileTransform, AllFormats)
     OIIO_CHECK_ASSERT(FormatExtensionFoundByName("spi3d", "spi3d"));
     OIIO_CHECK_ASSERT(FormatExtensionFoundByName("spimtx", "spimtx"));
     OIIO_CHECK_ASSERT(FormatExtensionFoundByName("vf", "nukevf"));
+}
+
+void GetFormatByIndex(OCIO::FormatRegistry &reg, int cap)
```
hodoulp: [Minor] The method is doing validations so a better name could be 'ValidateFormats'

remia: Updated the name.

PR #733, opened 2019-04-19
Adsk Contrib - Fix some Mac GPU unit tests

doug-walker:
Fixed the GPU unit tests that have been failing on the Mac.

Also adjusted the order of steps in the GPU tests to try and follow a more uniform sequence, to improve the readability.

Added two comments to clarify the Lut3D array ordering, requested on Slack by Remi.


hodoulp:

PR #737, opened 2019-04-26
Adsk Contrib - Add extra information to ociodisplay & ocioconvert

hodoulp:
When running `ociodisplay` or `ocioconvert` knowing the exact OIIO & OCIO versions could be useful to report any issue. The pull request adds these versions plus the OpenGL version supported by the graphic card.  The latter added to a dump of the GLSL shader program would definitively help reporting GPU errors.


hodoulp:

src/apps/ocioconvert/main.cpp
```diff
@@ -385,13 +396,39 @@ int main(int argc, const char **argv)
...
+        std::cout << std::endl;
+        std::cout << "OIIO Version: " << OIIO_VERSION_STRING << std::endl;
+        std::cout << "OCIO Version: " << OCIO::GetVersion() << std::endl;
+        const char * env = getenv("OCIO");
+        if(env && *env)
+        {
+            try
+            {
+                std::cout << std::endl;
+                std::cout << "OCIO Configuration = '" << env << "'" << std::endl;
```
michdolan: [minor] For consistency, perhaps format these lines with ":" instead of "="?

`std::cout << "OCIO Configuration: '" << env << "'" << std::endl;`

src/apps/ociodisplay/main.cpp
```diff
@@ -758,13 +736,52 @@ int main(int argc, char **argv)
...
+    }
+
+    if(g_verbose)
+    {
+        const char * env = getenv("OCIO");
+
+        if(env && *env)
+        {
+            std::cout << std::endl;
+            std::cout << "OCIO Configuration = '" << env << "'" << std::endl;
```
michdolan: [minor] Same comment as for ocioconvert for consistency with ":"

hodoulp: Fixed

src/apps/ocioconvert/main.cpp
hodoulp: Fixed

PR #739, opened 2019-04-26
Adsk Contrib - Fix some TruelightOp issues

hodoulp:
While implementing the bit-depth support, I found issues around the TruelightOp implementation. That op was not correctly updated following the latest refactoring #699. As there are no access to the library and no unit tests, it could happen again.


hodoulp:

doug-walker:
Please review by the end of the week as we have another PR that this is blocking.  Thanks!

PR #740, opened 2019-04-29
Setup SonarCloud static analysis via Travis CI

michdolan:
This PR adds C++ static analysis to the project, courtesy of SonarCloud and Travis CI. Additional commits will be needed to complete this work, based on test results; I may need to close and reopen it a few times to coax Travis into triggering rebuilds.

The goal is to hook this to a weekly Travis cron job, which will run our typical build matrix, but also run analysis (on just the Linux Clang build for now) and push the results here:
https://sonarcloud.io/dashboard?id=imageworks_OpenColorIO

This will have to be reconfigured when the project repo moves to the ASWF organization, but the changes should be minimal.

This PR also proposes some general improvements to the Travis build configuration:
- Adds job names, which allow use of $TRAVIS_JOB_NAME for build conditions
- Reworks the OSX GCC build, which appears to have actually been a second Clang build due to "gcc" being an alias to "clang", per the Travis docs. I removed the third OSX job, since I bumped the other two to use the same Xcode version. Perhaps we should add a job with a more recent macOS version to replace it?
- Added the new OCIO copyright/license statement

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,11 +69,6 @@ before_install:
       wget https://cmake.org/files/v3.12/cmake-3.12.2-Linux-x86_64.sh;
       sh cmake-3.12.2-Linux-x86_64.sh -- --skip-license --prefix=${TRAVIS_BUILD_DIR}/_cmake;
       export PATH="${TRAVIS_BUILD_DIR}/_cmake/bin:$PATH";
-      # if [[ "$TRAVIS_JOB_NAME" == linux_*_sonar ]]; then
-      #   wget https://sonarcloud.io/static/cpp/build-wrapper-linux-x86.zip;
-      #   unzip build-wrapper-linux-x86.zip;
-      #   export PATH="$PATH:$PWD/build-wrapper-linux-x86";
-      # fi
     fi
   # OSX ones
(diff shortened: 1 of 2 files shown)


michdolan:
After additional digging, turns out sonar-scanner can't be run for external (from a fork) PRs, due to security reasons. That means we'll need to merge this before we can see if it's working, but it also means I should be able to revert my last commit where I was manually installing build-wrapper, since the sonarcloud addon was not loading. I'll just comment out that part, since we may end up needing it if we use a CI system other than Travis (i.e. with no convenient addon).

michdolan:
As discussed in TSC meeting, since we have GCC CI through our Linux CI jobs, we can focus the macOS jobs on more platform specific compilers, so I have reconfigured Travis to include two macOs clang builds:
- OSX 10.12 (Sierra)
-OSC 10.14 (Mojave)

michdolan:
Closing this, since the ASWF CI efforts are leaning toward Azure Pipelines. I'll open a new PR with SonarCloud and Azure in coming weeks.

hodoulp:
Some parts (at least the Python bindings) are compiled in C++17 where C++11 is the selected version of the library. **To be investigated**.

PR #748, opened 2019-05-09
Adsk Contrib - Create an OpRcPtrVec managing the op bit-depths

hodoulp:
The direct usage of a `std::vector` for the list of ops limited the ability to cleanly add bit-depth adjustments. So, the pull request adds a specialized list of ops to handle bit-depths while preserving the type name and the C++11 `std::vector` interface (e.g. to limit code changes). The class now adjusts and validates the bit-depths between ops (e.g. two unit tests demonstrate the new behavior). It means that the output and input bit-depths of two following ops must be the same.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch could be merged this week. Unless there are some objections I will merge it tomorrow afternoon (New-York timezone).

PR #753, opened 2019-05-27
Adsk Contrib - Rework ParseNumber.

BernardLefebvre:
Rework ParseNumber to avoid calling strlen and modifying the buffer.
Add tests.
Fix a file used by unit test.


hodoulp:

hodoulp:
Here is a friendly reminder that the branch will be merged this afternoon (New-York time).

PR #755, opened 2019-05-28
Limit current CI builds to master and RB- branches

michdolan:


hodoulp:
What about the `gh-pages` branch?

michdolan:
The gh-pages branch is static, since it represents the live website. There's really no way to detect a problem until it is merged. The gh-pages deployment comes from the whitelisted branches here, when a new tag is created, so those source PRs, on master and RB-* are where the doc build testing would be verified.

PR #760, opened 2019-06-04
Adsk Contrib - Fix some code smells

hodoulp:
Following the first complete `Sonar Cloud` static analysis on the OCIO source code, I checked the content of the report to see the kind of output it generates, and its usability. And I discovered several code smells which could be easily changed. I did not intent to fix all of them because it would have taken too much time.   The main goal is only to decrease the number to later trap important ones.


hodoulp:

hodoulp:
Python bindings seem to compile with C++17 instead of the expected C++11. The last commit adds the warning disabling while the long-term solution is to fix the C++ version discrepancy.

`[ 64%] Building CXX object src/bindings/python/CMakeFiles/PyOpenColorIO.dir/PyAllocationTransform.cpp.o
In file included from /home/travis/build/imageworks/OpenColorIO/src/bindings/python/PyAllocationTransform.cpp:29:`
`In file included from /usr/include/python2.7/Python.h:85:`
`/usr/include/python2.7/unicodeobject.h:534:5: error: 'register' storage class
      specifier is deprecated and incompatible with C++17
      [-Werror,-Wdeprecated-register]`
    register PyObject *obj,     /* Object */
 `   ^~~~~~~~~`

michdolan:
I got the impression that the C++17 warning is just Clang being forward thinking in its compiler warnings. I don't think GCC has the warning.

src/OpenColorIO/Exception.cpp
```diff
@@ -31,24 +31,34 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 OCIO_NAMESPACE_ENTER
 {
   
-    Exception::Exception(const char * msg) throw()
-    : std::runtime_error(msg)
-    {}
+Exception::Exception(const char * msg)
+    :   std::runtime_error(msg)
```
michdolan: Just a style question. Further down in this PR, an initializer list has one space between ":" and first initializer, but here there are three. Should those be consistent?

hodoulp: Yes, I will check and fix.

hodoulp: A 'tab' but preserve the existing 'style' in the file you change. That's will be another 'rule' to define when we will put in place `clang-tidy`.

michdolan: Agreed

src/OpenColorIO/Mutex.h
```diff
@@ -59,12 +59,18 @@ OCIO_NAMESPACE_ENTER
 
     /** Automatically acquire and release lock within enclosing scope. */
     template <class T>
-    class AutoLock {
+    class AutoLock
```
lgritz: In C++11, this should be std::lock_guard

PR #761, opened 2019-06-05
Adsk Contrib - Fix some Op equality bugs

hodoulp:
Following the last completed `Sonar Cloud` static analysis on the OCIO source code, I checked the content and noticed a bug around the `Op::operator==()`. Digging into the 'bug', I found that some cases were not working but the declaration is the right one.

The fix will not remove the detected 'bug' but it will still fix some existing bugs.

Note: I refactored a little bit the Ops to have the same layout for the equality operator. So, the next new Op could safely copy & paste the code of an existing Op.


hodoulp:

PR #738, opened 2019-04-26
Update CCLA from changes needed to work in the CLA system

jmertic:
Signed-off-by: John Mertic <jmertic@linuxfoundation.org>

PR #762, opened 2019-06-06
Update the CCLA with the extra words from the e-signature page.

lgritz:
One paragraph is additionally listed when you go to sign the actual
new CLA, which clarifies the role of the CLA manager. Adding it to the
copy of the CLA in our repo just so it exactly matches what people will
see and sign when they use the e-signature system.

### DIFF
--- a/docs/aswf/CLA-corporate.md
+++ b/docs/aswf/CLA-corporate.md
@@ -30,6 +30,18 @@ to the Project shall be:
 Submitted under a Developer's Certificate of Origin v. 1.1 (DCO); and
 Licensed under the BSD-3-Clause License.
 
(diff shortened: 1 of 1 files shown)


lgritz:
To explain, this is a clarifying edit to the *future* CCLA that will be needed when we move the repo out of the Imageworks account and into the ASWF account on GitHub. Until then, it's still the same old CLA we have always used that is in effect. And I'm not changing anything unilaterally, I'm just making sure that the copy we have in the repo exactly matches the one that's on the LF web site e-signature page; we didn't update it after that was added as they refined the e-signature process.

lgritz:
Oops, I did not notice that this was a duplicate of #738 
Closing
Somebody please review and merge #738

Issue #756, opened 2019-06-03
Platform::Getenv uses potentially stale data on Windows


llatta:
The Windows implementation of the Platform::Getenv function uses the _dupenv_s to read environment variables from the C runtime. The C runtime environment variable functions read data that is cached at process start from the real Windows process environment (ie GetEnvironmentVariable). If after process start any place calls Windows SetEnvironmentVariable, or setenv in a different C runtime (debug vs release runtime, or older versions of MS CRT) it modifies the environment variables in a way that is not visible to the cached state in the used C runtime. So then Platform::Getenv returns stale data that is not correct any more.

We should change the implementation to read the real process environment directly, via GetEnvironmentVariable. Here's a suggested implementation, and it's simpler too: https://github.com/materialx/MaterialX/blob/master/source/MaterialXFormat/Environ.cpp#L18

hodoulp:
Thanks for the feedback.

hodoulp:
Hi @llatta 
I did some tests on my side and `OCIO::Platform::Getenv` does not see env. variables set by `SetEnvironmentVariable`.  So I will change the implementation.

Question:
> or setenv in a different C runtime (debug vs release runtime, or older versions of MS CRT)

Any reference?

hodoulp:
Merged in `master` i.e. OCIOv2

llatta:
Thanks for the fix!

Sorry just realized I never replied to the question for more info on the different C runtime problem. That was how I originally ran into the issue. I had compiled OCIO with the release CRT (of VS 2015), and the code using it was linked against the debug CRT. The code linked against debug CRT code was calling ::setenv("OCIO", ...) and that change wasn't picked up by OCIO when it got initialized. Your fix should address this, as either CRT's setenv does call through to the Windows SetEnvironmentVariable

hodoulp:
Merged in `RB-1.1`i.e. OCIOv1

PR #763, opened 2019-06-11
Adsk contrib - Improve the Windows GetEnv() implementation

hodoulp:
Here is the improvement following the issue #756 .


hodoulp:
Friendly reminder that this PR qualifies for single approver rule and could be merged next week unless there is feedback.

Issue #767, opened 2019-06-19
ClearAllCaches doesn't seem to be working in the same way since 1.0.9


fnalex:
Reloading a .ccc file doesn't seem to work the same in 1.0.9 and 1.1.x

The following code will create a file at /tmp/example.ccc and then create a CDL transform using the id cc03343. Then it will overwrite the file at /tmp/example.ccc with a different slope; clear all the caches and create a new transform from the same filepath with the same id.

```
#include <string>
#include <fstream>
#include <iostream>
#include <OpenColorIO/OpenColorIO.h>

namespace OCIO = OCIO_NAMESPACE;

static const std::string kContentsA = {
"<ColorCorrectionCollection>\n"
"    <ColorCorrection id=\"cc03343\">\n"
"        <SOPNode>\n"
"            <Slope>0.1 0.2 0.3 </Slope>\n"
"            <Offset>0.8 0.1 0.3 </Offset>\n"
"            <Power>0.5 0.5 0.5 </Power>\n"
"        </SOPNode>\n"
"        <SATNode>\n"
"            <Saturation>1</Saturation>\n"
"        </SATNode>\n"
"    </ColorCorrection>\n"
"    <ColorCorrection id=\"cc03344\">\n"
"        <SOPNode>\n"
"            <Slope>1.2 1.3 1.4 </Slope>\n"
"            <Offset>0.3 0 0 </Offset>\n"
"            <Power>0.75 0.75 0.75 </Power>\n"
"        </SOPNode>\n"
"        <SATNode>\n"
"            <Saturation>1</Saturation>\n"
"        </SATNode>\n"
"    </ColorCorrection>\n"
"</ColorCorrectionCollection>\n"
};

static const std::string kContentsB = {
"<ColorCorrectionCollection>\n"
"    <ColorCorrection id=\"cc03343\">\n"
"        <SOPNode>\n"
"            <Slope>1.1 2.2 3.3 </Slope>\n"
"            <Offset>0.8 0.1 0.3 </Offset>\n"
"            <Power>0.5 0.5 0.5 </Power>\n"
"        </SOPNode>\n"
"        <SATNode>\n"
"            <Saturation>1</Saturation>\n"
"        </SATNode>\n"
"    </ColorCorrection>\n"
"    <ColorCorrection id=\"cc03344\">\n"
"        <SOPNode>\n"
"            <Slope>1.2 1.3 1.4 </Slope>\n"
[...]

PR #768, opened 2019-06-21
Adsk Contrib - ClearAllCaches() missed a CDLTransform internal cache.

hodoulp:
The public API method to clear all the caches from OCIO missed a CDL Transform internal cache. The reported issue #767 highlights the problem.

The pull request consists to add the missing cache in ClearCDLTransformFileCache() (i.e. used by ClearAllCaches() ) and to add unit tests (one using the example described in the issue).

### DIFF
--- a/docs/tsc/meetings/2019-04-15.md
+++ b/docs/tsc/meetings/2019-04-15.md
@@ -1,0 +1,59 @@
+Apr 11, 2019
+
+Host: Michael Dolan
+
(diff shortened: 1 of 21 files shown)


scoopxyz:
Hey Patrick just a heads up that some of your commits aren't being associated with your GitHub account. Should be fine when they get squashed, but just something to keep an eye out for

PR #772, opened 2019-07-17
Adopt OIIO unittest.h as OCIO UnitTest.h

michdolan:
As part of the larger effort of cleaning out ext/, this PR moves ext/oiio/src/include/unittest.h to src/OpenColorIO/UnitTest.h and renames all macros and functions from OIIO_ to OCIO_. A few other stray OIIO references are also resolved.

Amidst the mass rename, there is a related fix preventing a -Wmaybe-uninitialized compiler warning that appeared. That change is to the `ExposureContrastOpData.cpp` `style` test.

This PR will have merge conflicts with #771 which I will resolve between their merges.

### DIFF
--- a/src/OpenColorIO/Baker.cpp
+++ b/src/OpenColorIO/Baker.cpp
@@ -275,10 +275,10 @@ OCIO_NAMESPACE_EXIT
 #ifdef OCIO_UNIT_TEST
 
 namespace OCIO = OCIO_NAMESPACE;
-#include "unittest.h"
+#include "UnitTest.h"
 
 /*
-OIIO_ADD_TEST(Baker_Unit_Tests, test_listlutwriters)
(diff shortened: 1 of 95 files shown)


src/OpenColorIO/Baker.cpp
```diff
@@ -275,10 +275,10 @@ OCIO_NAMESPACE_EXIT
 #ifdef OCIO_UNIT_TEST
 
 namespace OCIO = OCIO_NAMESPACE;
-#include "unittest.h"
+#include "UnitTest.h"
 
 /*
-OIIO_ADD_TEST(Baker_Unit_Tests, test_listlutwriters)
+OCIO_ADD_TEST(Baker_Unit_Tests, test_listlutwriters)
```
lgritz: Not only is this ok with me, but I continue to be mystified how it got this name in the first place, since OIIO never had an ADD_TEST!

PR #775, opened 2019-07-22
Security policy proposal

michdolan:
This is a proposed draft for a new SECURITY.md outlining the OCIO security policy. Once we have agreed on the included topics and wording, this will complete all CII badge requirements. I used the OpenVDB policy as a starting point (preserving some sentences verbatim when they were more generic).

Please suggest edits, additional or less topics, or any other feedback you have.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -27,28 +27,27 @@ None
 
 None
 
+## File Format Expectations
 
-
-
-
-
-
-
-***File Format Expectations***
-
-Attempting to read a .vdb file will:
-* Return success and produce a valid VDB data structure in memory
+Attempting to read a .ocio file will:
+* Return success and produce a valid Config data structure in memory
 * Fail with an error
-* Execute forever
-* Run out of memory
 
-The last two options may be surprising.  VDBs, however, are designed
-as open-ended containers of production data that may be terabytes in size.
+The bulk of data read comes from transform files referenced by a Config. 
+These files may be arbitrarily large, and may live on any accessible volume. 
+Referenced file paths support environment variable expansion, making OCIO's 
+Processor invokation susceptible to environmental redirection. Files are only 
+read when needed for finalizing a Processor, so no transform files should be 
+accessed as a result of loading a Config.
 
 It is a bug if some file causes the library to crash.  It is a serious
 security issue if some file causes arbitrary code execution.
 
-***Runtime Library Expectations***
+OpenColorIO will attempt to associate a file's data and layout with a 
+registered file format, regardless of a file's extension. Invalid input data
+results in an error.
+
+## Runtime Library Expectations
 
 We consider the library to run with the same privilege as the linked
 code.  As such, we do not guarantee any safety against malformed
@@ -64,7 +63,7 @@ We do not consider this as severe as file format issues because
 in most deployments the parameter space is not exposed to potential
 attackers.
 
-***Proper Data Redaction***
+## Proper Data Redaction
 
 A common concern when working with sensitive data is to ensure
 that distributed files are clean and do not possess any hidden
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -0,0 +1,76 @@
+# Security and OpenColorIO
+
+The OpenColorIO Technical Steering Committee (TSC) takes security very
+seriously. We strive to design secure software, and utilise continuous
```
retokromer: `utilise ` (UK) > `utilize` (US)

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+library to crash. It is a security issue if calling a function with 
+well-formed arguments causes arbitrary code execution.
+
+We do not consider this as severe as file format issues because in most 
+deployments the parameter space is not exposed to potential attackers.
+
+## Proper Data Redaction
+
+A concern when reading sensitive data in a computer program is to ensure 
+that data is only used and accessable in intended ways. OpenColorIO's public
```
retokromer: `accessable` > `accessible`

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+interface is not granted access to internal data structures read from 
+transform files. Proprietary LUTs and related data are only exposed inasmuch 
+as their source is accessable on the filesystem.
```
retokromer: ditto

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+file and search paths, to avoid the possibility of a maliciously modified 
+environment redirecting file reads to an insecure location.
+
+It is a bug if some file causes the library to crash. It is a serious
+security issue if some file causes arbitrary code execution.
+
+OpenColorIO will attempt to associate a file's data and layout with a 
+registered file format, regardless of a file's extension. Only expected data 
+structures will be read. It is a bug if reading invalid or malformed data from 
+a file does not result in an immediate error.
```
doug-walker: OCIO tries all the format readers until one succeeds.  No error is raised if a format fails and then another succeeds.  Also, we purposely allow some malformed data, e.g. common misspellings such as SatNode/SATNode.  So I think the last sentence should be modified.  

Suggest: "It is a bug if reading seriously invalid or malformed data from a file does not result in an error from the file format reader ultimately used with a file."

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+It is a bug if calling a function with well-formed arguments causes the 
+library to crash. It is a security issue if calling a function with 
+well-formed arguments causes arbitrary code execution.
+
+We do not consider this as severe as file format issues because in most 
+deployments the parameter space is not exposed to potential attackers.
+
+## Proper Data Redaction
+
+A concern when reading sensitive data in a computer program is to ensure
```
doug-walker: While the first sentence is not wrong, I don't think OCIO is used for any "sensitive" data.  I think this whole paragraph should be rephrased.  I mean OCIO does not take any serious steps to protect the data since that data is all coming from unencrypted readable files on the file system anyway.

SECURITY.md
```diff
@@ -0,0 +1,76 @@
...
+well-formed arguments causes arbitrary code execution.
+
+We do not consider this as severe as file format issues because in most 
+deployments the parameter space is not exposed to potential attackers.
+
+## Proper Data Redaction
+
+A concern when reading sensitive data in a computer program is to ensure 
+that data is only used and accessable in intended ways. OpenColorIO's public
+interface is not granted access to internal data structures read from
```
doug-walker: Given that the API allows you to make a Processor from a FileTransform, to find out what ops are used in a Processor, and it allows you to send pixels through a Processor, and it allows you to make a GLSL program of the Processor, I don't think it makes sense to say that we don't allow access to the data read from transform files.

But in any case, I don't understand why this a security risk for the API to expose data read from transform files.  In some sense, that's the whole purpose of the library.  ;-)

michdolan: I was thinking of a case where a LUT is under heavy NDA, and considering the idea that the raw LUT data could be somewhat protected from being ripped via the OCIO Python bindings. It's a far fetched use case for sure. But you raise a great point that in OCIO v2 terms, the LUT would be fully exposed via the GPU renderer since it would not longer be baked into a combined-op 3D texture. 

I'm okay with removing the Data Redaction section if we think it's unnecessary or untrue, both of which could be the case.

PR #778, opened 2019-07-22
Adsk contrib - Fix most of the warnings detected by SonarCloud

hodoulp:
Following the integration of `SonarCloud` in the build system, the static analysis now reports much more warnings from the source code. To limit the number of warnings and better highlight the critical ones, the pull request fixes most of them.  While fixing these trivial warnings I fixed a subtle one in (ops/Exponent/ExponentOps. cpp line 76).


hodoulp:

PR #781, opened 2019-07-24
update to Cmake 3.11

retokromer:
As suggested in https://github.com/imageworks/OpenColorIO/pull/776#issuecomment-514502914: 
> Noting that PR #771 bumps the minimum CMake version to 3.11, which is the current minimum of OpenEXR 2.3.0 (that OCIO uses Half from).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.11)
 
 project(OpenColorIO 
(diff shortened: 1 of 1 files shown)


michdolan:
This change will be made in #771 already actually:

https://github.com/imageworks/OpenColorIO/pull/771/files#diff-af3b638bc2a3e6c650974192a53c7291

I did not however update the macOS docs your other PR changed to 3.10

retokromer:
Thank you! Then this is useless. Sorry for the noise.

Issue #676, opened 2019-02-22
Add support for Autodesk CTF file format


doug-walker:
The Color Transform Format (CTF) is the native LUT format for Autodesk products including Maya, Flame, and Lustre.  

This format is a superset of the Academy/ASC Common LUT Format (CLF).  The extensions to CLF that comprise the CTF format were added primarily for two purposes:

1)  To provide more accurate color processing by avoiding the need to represent certain functions as LUTs.  This is particularly useful for providing compact and accurate ACES Output Transform implementations.

2)  To allow any OCIO operators to be serialized losslessly to a text-based (XML) format.  (CTF is the only format we're aware of that allows this.)

Adding CTF support would be useful for users whose workflows interact with Autodesk applications.  It is potentially also useful for scenarios that would require losslessly serializing an OCIO processor into an XML format.  For example, when debugging or when implementing cloud services.

hodoulp:
@doug-walker being part of the ACES CLF WG, he proposed some ops/capabilities he designed for the Autodesk CTF format, be considered for the ACES CLF format.

doug-walker:
Closing as this has now essentially been implemented.  There are a few edge cases of Autodesk CTFs that may not load, but those would be better tracked as separate issues.

PR #782, opened 2019-07-26
Adsk Contrib - CLF/CTF format write support

BernardLefebvre:
This PR adds the ability to write the Academy/ASC CLF format and Autodesk CTF format.  It is a continuation of issues #610 and #676 and builds on the earlier reader work.

A new write method is also introduced.  Whereas Baking refers to a process of reducing a processor's ops into an alternate form that may be saved into a given format, Writing does not attempt any alteration of the original operators.  For CTF, all OCIO ops (except Truelight) can be losslessly written to the format.  For CLF, the write method will throw if the processor contains ops that the CLF format cannot represent.  A new ociowrite command line tool has also been added.

A new FormatMetadata class is added to hold rich metadata from formats such as CLF/CTF.  It is now possible to access the various descriptive metadata attributes and elements from the XML encoding of these formats.  Furthermore, that metadata may be set in preparation for writing.

Enhancements are also made to the GroupTransform and a processor may now be converted to a GroupTransform.  A followup PR is planned to add additional capabilities and tests related to writing LUTs in CLF/CTF.

As usual, a lot of unit tests have been added for this feature.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch could soon be merged unless there are some objections.

src/OpenColorIO/fileformats/FormatMetadata.h
```diff
@@ -0,0 +1,148 @@
...
+#include <vector>
+
+#include <OpenColorIO/OpenColorIO.h>
+
+#include "PrivateTypes.h"
+
+OCIO_NAMESPACE_ENTER
+{
+
+static constexpr const char * METADATA_INFO = "Info";
```
michdolan: Should the metadata keys have consistent casing? Or are Info and Description special cases (CLF/CTF specific)?

i.e.
Info, Name, Id, Description
OR
info, name, id, description

doug-walker: Michael, fair question, but we are following the CLF spec, those are the spellings used in S-2014-006.  There is a logic to it though since the lower-case ones (id, name) are XML attributes whereas the upper-case ones (Info, Description) are XML elements.

michdolan: Ok, that makes sense, and leads to my second question. Is it confusing to lump both under the same `METADATA_*` naming scheme given that they are not intended to be used in the same way. Maybe it doesn't matter since they are not publicly facing, but outside of reading the explanation in FormatMetadata.h, it could be unclear that they carry a more specific intended use than just as metadata keys.

doug-walker: I think it is more clear where they are used since ID and NAME show up in getAttribute type calls and INFO and DESCRIPTION show up in getElement type calls.  But I agree this header could use more explanation.  

Bernard, what I would suggest is the following:
// INFO and DESCRIPTION  are CLF XML elements described in S-2014-006.
... the info and desc declarations
// NAME and ID are CLF XML attributes described in S-2014-006.
... the name and id declarations
// ROOT is simply a placeholder ... (4 line comment from below)
... the root declaration

michdolan: That would help clarify. Thanks!

doug-walker:
Patrick, Bernard, and I have been working on some updates to this PR.  Please stay tuned for another commit with some fixes and other changes.

src/OpenColorIO/fileformats/FormatMetadata.h
BernardLefebvre: done. Will be part of an upcoming commit.

hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch could be merged somewhere week unless there are some objections.

michdolan:
I like the new approach to bit-depth handling. It feels more natural/consistent to isolate it within the file IO.

hodoulp:
As this work is now a blocker, here is the friendly reminder that we plan to merge the pull request this Thursday afternoon (NYC timezone).

PR #785, opened 2019-07-31
Create COMMITTERS.md

jmertic:
Signed-off-by: John Mertic <jmertic@linuxfoundation.org>


hodoulp:
**[Question]** After reading the list of names, I think that we should update the permissions before merging anything. What do you think?

michdolan:
Is your thought to reduce admins to those who have committed in the last 2 or so years?

hodoulp:
That's a topic for a TSC meeting. But the idea would be to have some kind of exit process after X years without any activities (i.e. any kind of activities from commits to TSC participation).
The goal is not to 'downgrade' someone but to keep/increase his interest in OCIO or to promote some else.

michdolan:
Sounds good @hodoulp . Let's discuss Monday.

PR #792, opened 2019-08-08
Adsk Contrib - Fix doc generation on Windows

hodoulp:
The pull request fixes the documentation generation under Windows. Mainly two problems were present:
1) The default installation path of `pip` (i.e. the package manager for Python) is different between Windows and Linux. I changed `findSphinx.cmake` method to use the right path on Windows.
2) The second problem was around the `Sphinx` variable resolution. The commands to build the documentation (i.e. docs/CMakeLists.txt at lines 109 & 136) use the `sphinx` variable and indicate a dependency on `Sphinx`. That was not working i.e. the variable resolution was not done when calling the command. I fix that problem by explicitly adding the dependency before the commands.

*Note:* ubuntu platform (under docker) fails to update 'texlive' (i.e. a latex compiler needed by the documentation generation) so I was unable to test the fix on Linux. I will monitor the CI builds.

*Note:* While investigating the doc. generation, I looked at the generated errors and fixed some common errors to improve the `rst` documentation and learn the `rst` syntax.


michdolan:
Regarding point 2:

the `SPHINX_EXECUTABLE `variable used in docs/CMakeLists.txt is globally cached in ext/CMakeLists.txt line 51. I put it there along with all of the other ext/ find/install functionality. It should always be defined prior to this CMake module running, but I realize where it's implemented is not clear at all.

I suggest we move the entire Sphinx find block from ext/CMakeLists.txt into docs/CMakeLists.txt. The previous Sphinx find functionality lived here, and I moved it in an attempt to consolidate everything to ext/. I was on the fence about doing it, and probably shouldn't have. This is the block that could be moved to replace your added `find_package(Sphinx)` call, minus the lines that are unnecessary given this move:

```
### External Python dependencies ###

include(FindPythonPackage)

# Sphinx
# https://pypi.python.org/pypi/Sphinx
set(SPHINX_MIN_VERSION 1.8.5)
find_package(Sphinx ${SPHINX_MIN_VERSION} REQUIRED)
```

hodoulp:
Good idea. I will investigate.

hodoulp:
*Note:* When generating the documentation, the `Sphinx` package is always downloaded and updated. That's slowing down the doc generation. Good for another pull request.

michdolan:
The Sphinx find module will always download and install it if OCIO_INSTALL_EXT_PACKAGES == ALL. If == MISSING, it should only install when sphinx-build is not found in its expected location. If == NONE (the default option) it will not install. In our CI, I have it always set to ALL, but in your environment, you could use one of the other options.

hodoulp:
There is a problem as the default value of `OCIO_INSTALL_EXT_PACKAGES` is `MISSING` i.e. CMakeLists.txt at line 100.

michdolan:
Good catch.

hodoulp:

PR #797, opened 2019-08-13
CI branch cleanup master

michdolan:
This PR removes the Travis and AppVeyor config files from the master branch, since they have been replaced by azure-pipelines. Azure targets RB-* branches, but will only be included in newer releases, so will never run for existing RB-1.* branches. A second pull request will focus the RB branch Travis and AppVeyor config files to only build RB-1.* branches, so the old CI system can still be used for patch releases.


michdolan:
Note: I've re-enabled Travis and AppVeyor webhooks, so they will be triggered for CI until this PR is merged.

PR #799, opened 2019-08-13
Swap SPDX license identifiers

michdolan:
This PR updates the remaining source files with a SPDX reference to the BSD-3-Clause license, and our new contributor inclusive copyright statement.

Intentionally omitted sources:

- TruelightTransform: Another PR is coming which removes these from OCIO.
- argparse, strutil, UnitTest: these were pulled from OIIO, not contributed, so the original license and copyright statement is preserved, which is also BSD-3-Clause. These licenses are also called out in LICENSE.md.

One file in question:

- FindNuke.cmake: This is also BSD-3-Clause, but the original contribution was not from the original author, so we could either conform this as a contribution, or treat it like a pulled source, preserving the existing notice. Arguments for either option?


doug-walker:
I'm not an expert on this and would like to get John's approval on this PR.  But I have a couple of questions:
1) When I read this: <https://spdx.org/ids-how>, it says the SPDX ID should be used to supplement, not replace, existing notices in files.  I do think removing the old notice is nice, in that it cleans things up, but I just want to make sure it's allowed.
2) Larry, regarding your suggestion to add the extra line, when I visit that URL it just seems to give me the BSD-3 license.  So what is that adding beyond the SPDX?  (Not opposed to adding it, just want to understand it better.)

doug-walker:
Also, looking at this: <https://spdx.org/sites/cpstandard/files/pages/files/using_spdx_license_list_short_identifiers.pdf>, it recommends that the LICENSE file be updated to say the individual files use the SPDX identifier.

michdolan:
Good catch about that recommendation @doug-walker . I am also interested in @jmertic take on Larry's suggestion, and these changes overall. I'll update this based on future discussion.

Posting the language from the link here for reference:

> It is recommended to also make a reference that SPDX
> license identifiers are being used where the license(s) for
> the project are documented ( “LICENSE”, “COPYING”, etc.
> ).

jmertic:
So first off, the representation of the SPDX short form identifier and Copyright statement are both inline with the recommended guidance from the ASWF TAC - so you should be good to go there.

@lgritz suggestion to adding a link to the LICENSE file in the repo could be problematic if there is a difference between that file and the SPDX license definition ( or if one develops over time, such as if the project changes licenses ). I'd generally recommend sticking with the SPDX header only to alleviate any future confusion.

As for a reference in the LICENSE file to the SPDX header being used - seems like a good idea but would want @swinslow to chime in with the right wording to use.

Let me know if I addressed all the questions :-)

michdolan:
The only other question I have is in regards to this file:

https://github.com/imageworks/OpenColorIO/blob/master/share/cmake/modules/FindNuke.cmake

It's BSD-3-Clause, but was contributed by someone other than the author (at least directly). Is this safe to update to the general copyright statement, or should we preserve it's current state?

jmertic:
@michdolan I'd leave it in it's current state, just because it came from another source than the author.

swinslow:
Hi all, several very good questions in this thread. Although I can't provide you with legal advice (and you may want to talk to your own legal counsel), here's what I'd say re: community approaches to license and copyright notices:

* **Third party copyright statements**: Please note that you should **not** _remove or replace_ a third party's copyright statement, without their express permission to do so. It is fine to _supplement_ notices where accurate, but the rule of thumb is generally to leave third party's statements exactly as they are.

* **Including URL to LICENSE file**: I think it's fine to include the URL if you want. A minor note, I suppose it's possible that e.g. the URL would go stale if the license filename changes in the future. The SPDX license identifier by itself is meant to be stable, by referring to the [specific entry](https://spdx.org/licenses/BSD-3-Clause.html) on the [SPDX license list](https://spdx.org/licenses/). So either approach is fine I would say, whether to include the link to the project's license or not.

* **FindNuke.cmake**: If this file came from an upstream source who wasn't the person contributing it to OpenColorIO, I would recommend leaving its notices as-is.

* **Removing license texts**: When adding SPDX short-form IDs, different projects have taken different approaches for what to do with the old notices. I'd say most projects have tended to leave the old notices in place.

* **Explaining the SPDX short-form identifiers in the LICENSE file**: The PDF linked above is a bit out of date, as it was developed when these short-form IDs were just starting to be used. I think it's fine to include an explanation of the identifiers in the LICENSE file, but I wouldn't say it's necessary.

Apologies for over-wordiness and let me know if I missed your question  :)

michdolan:
@swinslow one other question. In our LICENSE.md file we include selected license blocks for third-party or upstream contributions (not from the author) that exist somewhere in the source tree. Would you say that's a good rule to follow? i.e. is it necessary to include all of these license/copyright statements in there (FindNuke.cmake for example)? Or is our blanket BSD-3-Clause license block in LICENSE enough?

michdolan:
As a general note, almost every statement this PR touches is from Imageworks or Autodesk (one from CineSite). Perhaps we should talk to our legal teams respectively to see if they approve of the change, and if not, leave them as-is, or supplement just the SPDX line.

swinslow:
@michdolan It's a tough call. I would say it is helpful to downstream users because it saves them the trouble of having to scan or review the full sources looking for licenses. The flip side is that often it becomes out-of-sync when new third-party code is added in the coming months or years, and contributors don't know or remember to update the main LICENSE file accordingly.

This is yet another area where practices are all over the map. What I might suggest is that:
1) It likely isn't _necessary_ to reproduce all those license blocks in the main LICENSE file -- as long as they're in the sources, people can find them;
2) since you've already done so, it doesn't cause any harm to keep them there; and
3) many downstream users will have compliance processes that involve them scanning or reviewing the full sources anyway, so they always have that option in either case.

Long-winded way of saying, I think the blocks in the LICENSE file are fine to keep there.

And responding to your follow-up comment just now -- yes, I would recommend that you'll want to talk with your own legal counsel about whether they are comfortable with the changes to your own copyright notices. I am happy to chat with your counsel if it is helpful, and to explain more about how this practice has been implemented in several of our other projects.

jmertic:
The only counterpoint is - GitHub has a nice facility for identifying the license name based on the content in the LICENSE file ( more details at https://help.github.com/en/articles/licensing-a-repository#detecting-a-license ). Not having the LICENSE file not matching the normal template will not make that work and might make it harder if someone is looking for a project filtered by license.

You could always add license information for additional components in a separate THIRD_PARTY file.

lgritz:
@jmertic That's what I did for OIIO -- LICENSE.md for just our BSD, LICENSE_THIRD_PARTY.md for the concatenated and explained licenses of any other bits of code we have embedded.

michdolan:
I like that idea of splitting the licenses into two files. That seems way more clear.

michdolan:
I matched @lgritz language from OIIO for the OCIO LICENSE-THIRD-PARTY.md. I couldn't come up with a clearer way to word it...

michdolan:
TODO: Add note to README about origination of project from Imageworks, since copyright statement is being generalized. Consider adding an acknowledgements section for companies and individuals who have contributed.

michdolan:
Please see README.md (bottom in particular) for a new acknowledgement section that aims to attribute Imageworks and Autodesk for their respective contributions to OCIO 1.0 and 2.0. This should be considered a draft which can be revised per legal needs, etc.

jmertic:
Looks good @michdolan !

README.md
```diff
@@ -11,55 +11,91 @@ OpenColorIO
...
 OpenColorIO is released as version 1.0 and has been in development since 2003.
-OCIO represents the culmination of years of production experience earned on such
-films as SpiderMan 2 (2004), Surf's Up (2007), Cloudy with a Chance of Meatballs
-(2009), Alice in Wonderland (2010), and many more. OpenColorIO is natively
-supported in commercial applications like Katana, Mari, Nuke, Silhouette FX, and others.
+OCIO represents the culmination of years of production experience earned on
+such films as SpiderMan 2 (2004), Surf's Up (2007), Cloudy with a Chance of
+Meatballs (2009), Alice in Wonderland (2010), and many more. OpenColorIO is
+natively supported in commercial applications like Katana, Mari, Nuke,
+Silhouette FX, and others.
```
doug-walker: It might be worth mentioning Maya here too since it is one of the other big apps in the VFX world.

README.md
```diff
@@ -11,55 +11,91 @@ OpenColorIO
...
+  * aces_1.0.2
+  * aces_1.0.1
+  * aces_0.7.1
+  * aces_0.1.1
+* Other
+  * nuke-default
 
----
-Images from "Cloudy With A Chance of Meatballs" Copyright 2011 Sony Pictures Inc. All Rights Reserved.
+Acknowledgements
```
doug-walker: Thanks Michael, looks great to me.  I've also forwarded to our legal support in case they have any input on the copyright aspect.

README.md
michdolan: That crossed my mind while revising this. Completely agree. I'll add it.

doug-walker:
Michael, please hold off merging this, I just got an email from our legal.  I'll follow up with you separately.

doug-walker:
Hi Michael, following up on my recent comment, I met with our legal team to discuss this.  They gave me their approval, so we have no objections to merging this PR.

michdolan:
Ok, great! Thanks Doug! I'll update this branch with new code needing the license update, and we can merge after that.

michdolan:
Ok, this has been brought up to date. I'll plan on merging it tomorrow evening unless there's additional feedback.

PR #803, opened 2019-08-19
Adsk Contrib - Addressing TODO from SSE.h

BernardLefebvre:
- Use constants instead of defines.
- Change sseSelect arguments order.


hodoulp:
Friendly reminder that this PR qualifies for single approver rule so it could now be merged soon.

PR #807, opened 2019-08-22
Adsk Contrib - Fix build breaks on macOS

hodoulp:
Here is an attempt to check the GPU capability of the macOS virtual machines used by the Azure CI build system.


hodoulp:
Added some outputs to have the GPU unit test context, here are the results:
`2: GL Vendor:    Apple Inc.`
`2: GL Renderer:  Apple Software Renderer`
`2: GL Version:   2.1 APPLE-16.7.4`
`2: GLSL Version: 1.20`

hodoulp:
We think that it is not worth investing more time to enable GPU unit tests on an emulator (and having to support it). So, I disabled the GPU unit tests, but I still preserve the branch to merge small glitches I fixed.

PR #809, opened 2019-08-26
Adsk Contrib - Case for GPU source files

BernardLefebvre:
This fixes inconsistency in the GPU tests source files, some were using "_Test" instead of "_test"


hodoulp:

PR #810, opened 2019-08-28
Adsk Contrib - Move tests out of source files.

BernardLefebvre:
Demonstrate with 2 files how we can move test code out of the main source code.
- File are smaller.
- #ifdef are removed.
- Code from a cpp file of the libraire implementation can still be accessed from the corresponding test file.
- Code from the librairie does not hold the test code, thus code analysis is not biased by test code.


hodoulp:
No. That's a 'normal' pull request but we first want the community feedback before doing too much work. The idea is to merge the pull request and to slowly transition the other unit tests.

hodoulp:

PR #811, opened 2019-08-28
Adsk Contrib - Impose explicit namespaces

hodoulp:
Using implicit namespaces could lead to 'tricky' bugs so it should not be used. 

The code could take the wrong method if there are two methods (with same name but different behaviors from different namespaces) with implicit namespaces (i.e. `using namespace xx;` & `using namespace yy;`).

Note that the pull request also removes trailing whitespaces & tabulations of the changed files only.


hodoulp:

PR #812, opened 2019-08-30
Adsk Contrib - Refactor RangeTransform to improve the RangeOp creation

hodoulp:
The pull request fixes a design weakness where the data exchange between a Transform instance (i.e. class from the public API) and its corresponding Op instance (i.e. private class) was manually done. The problem is that any new/update on the Op data could be lost by the implementation. Please to refer to the former `BuildRangeOps` implementation as example. 

The new design implements the data exchange using the corresponding OpData. It means one line using the OpData instead of several lines to copy data one-by-one. The pseudo code is now:
```
void BuildRangeOps(.., const RangeTransform & transform, ..)
{
   ..
   const RangeOpData & data
        = dynamic_cast<const RangeTransformImpl*>(&transform)->data();
   CreateRangeOp(ops, data.clone(), combinedDir);
   ..
}
```

**Note:** To highlight the new design `RangeTransform` is used as example. If accepted, following pull request(s) will refactor all the other transforms.

Note that the pull request also removes trailing whitespaces & tabulations of the changed files only. Select `Hide withespace changes` using the gear to mute these differences.


michdolan:
I should see if this plays more nicely with pybind11, since the destructor is public in your proposed design.

hodoulp:
The contructor and destructor of the 'new' RangeTransform class are ( and must be ) not accessible. That's a mandatory design requirement for any class part of a public C++ API. It means that it will not help the Python binding.

michdolan:
If I explicitly wrap the RangeTransformImpl though (since it's now a derived class), handling the RangeTransform down casting, that interface is public.

michdolan:
Nevermind, I see that class is not exposed in the OCIO public interface.

hodoulp:
@michdolan Did I answer to your questions concerning the pull request?

PR #814, opened 2019-08-30
WIP: pybind11 bindings initial work

michdolan:
This draft PR demonstrates the approach I have been exploring to create new Python bindings which will natively support Python 2 and 3 with pybind11. 

Overall the amount of code needed to create the bindings is significantly less, so will result in smaller and fewer source files.

Like our previous bindings, this approach uses a wrapper object to hold a shared_ptr to either an editable OCIO object, or a constant OCIO object. This allows us to more-gracefully control editability of the underlying OCIO objects from Python. I initially wanted to avoid using a wrapper, since pybind11 supports shared_ptr as a native holder type, but hit walls with needing our classes to have public destructors. Feel free to propose another approach if I overlooked a better alternative.

Python tests are currently disabled, but the portions I have wrapped so far are fully functional and provide a more consistent experience with the C++ API, albeit remaining Pythonic. In particular note the differences in PyStatic.cpp, where OCIO's enums are moved outside of OCIO.Constants, and now behave like enums. Please have a look at PyTransform.cpp as well. I would recommend looking at it without the diff, as it is only similar in name to the previous file. All of the OCIO transforms could be contained in this file.

Once we settle on the best design, I'll chug through the rest of the classes.


src/bindings/python/PyTransform.cpp
```diff
@@ -1,336 +1,131 @@
...
+                self.getTransformRcPtr()->push_back(transform.getConstRcPtr());
+            })
+        .def("clear", [](PyGroupTransform & self) 
             {
-                pyobj = PyObject_New(PyOCIO_Transform,
-                    (PyTypeObject * ) &PyOCIO_MatrixTransformType);
-            }
-            else if(ConstRangeTransformRcPtr rangeTransform = \
-                DynamicPtrCast<const RangeTransform>(transform))
+                self.getTransformRcPtr()->empty();
```
michdolan:
```suggestion
                self.getTransformRcPtr()->clear();
```
Noting my error...

doug-walker:
Thanks for your work on this Michael!   Patrick and I were talking about this today and he will reach out to you next week to discuss.  It would be very nice to find a way to be able to have the bindings auto-generated.

hodoulp:
[Minor] The VC++ compilation produces two warnings:
```
F:\OpenSource\ocio_michdolan\src\bindings\python\PyTransform.cpp(30): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
F:\OpenSource\ocio_michdolan\src\bindings\python\PyTransform.cpp(68): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
```

PR #815, opened 2019-09-02
Add bare AppVeyor config

michdolan:
This is an attempt to re-enable AppVeyor for only RB-1.x CI. Not having a config on master causes PR CI failures instead of preventing AppVeyor web hooks from having an effect.


michdolan:
I'm going to merge this to prevent CI failures for other PRs, now that AppVeyor is re-enabled.

jmertic:
Curious on the use case for AppVeyor - I thought Azure Pipelines supported Mac, Windows, and Linux?

michdolan:
AppVeyor is still being used for CI on RB-1.x branches. We determined to
keep it rather than rework CI for our few remaining patch releases. Master
and RB-2.x branches are all Azure. The config is needed in master because
enabling the AppVeyor web hooks will trigger CI builds for branches without
the config using a default setup. This config prevents that.

PR #816, opened 2019-09-03
Adsk Contrib - Improve the Windows/Vc++ pdb support

hodoulp:
The dynamic library name now includes the major & minor to avoid the known Windows [DLL hell](https://en.wikipedia.org/wiki/DLL_Hell). However, the pdb file name was not changed falling in the same trap than the DLL name clash. The pull request changes the pdb filename to follow the dynamic library name pattern.


hodoulp:

PR #824, opened 2019-09-09
Adsk Contrib - Fix build breaks of the Linux static library compilation

hodoulp:
Following the issue #821, @michdolan, @jiapei100 and I did some investigations to understand the problem. And we found the cause which is around `-fPIC` & `pytstring`. Even if I was not able to reproduce the original problem in my environment (docker & ubuntu without any external library installed), I still reproduced another one (also detected by @jiapei100) and fixed it.

The requirement is that:

1. Any external libraries compiled for OCIO **must** have `-fPIC` (i.e. Position-Independent Code). 
2. As the Python bindings generate a dynamic library on Linux, `-fPIC` is also mandatory for the OCIO library even if compiled as static library.

The pull request fixes the static library compilation under Linux, but it could also address the original build break (i.e. dynamic library build break on Linux when using external libraries). @michdolan & @jiapei10 could validate my fix in your environment?


hodoulp:

PR #826, opened 2019-09-10
Adsk Contrib - CDL rework for description.

BernardLefebvre:
CDL in OCIO V1 was handling only 1 descritpion. The first description in the SOPNode.
Extending CDL description support using the FormatMetadata.
Legacy get/setDescription on CDLTransform is kept for backward compatibility but using setDescription will clear all other formatMetadata.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch could be merged somewhere next week unless there are some objections.

PR #769, opened 2019-06-26
Adsk Contrib - Add bit-depth support to apply methods

hodoulp:
The pull request implements an important part of #609 .

> OCIO currently only accepts floating-point image buffers. However, in many scenarios one is starting with an integer buffer. Furthermore, the first operation often performed is a 1d-LUT. So rather than accepting an integer buffer and doing a fast look-up, OCIO currently converts to float and does interpolation rather than a simple look-up.

Some ground works were already done to ease the bit-depth support integration and reduce the final pull request size. Please refer to #699 and #722 .

> When a 1d-LUT is the first operator in the chain, it will be interpolated to the needed bit-depth and then a look-up will be done on the pixels. This gives a big performance boost with no loss of precision.

The implemented approach is now using the first (or last) op in the color transform (i.e. processor instance) to perform the input and output bit-depth conversions when needed.

**Public API:**
To remove the existing limitation of the OCIOv1 API, the new design provides several improvements:
- Ability to specify an input and output ImageDesc supporting bit-depths:
`void CPUProcessor::apply(ImageDesc & imgDesc) const;`
`void CPUProcessor::apply(const ImageDesc & srcImgDesc, ImageDesc & dstImgDesc) const;`
- Ability to control the optimization (what optimization(s) to use or not), and finalization (fast or exact inverse LUT support):
`ConstCPUProcessorRcPtr Processor::getDefaultCPUProcessor() const;`
`ConstCPUProcessorRcPtr Processor::getOptimizedCPUProcessor(OptimizationFlags oFlags, FinalizationFlags fFlags) 
 const;`
- Ability to process diffferent input and output bit-depths:
`ConstCPUProcessorRcPtr Processor::getOptimizedCPUProcessor(BitDepth inBitDepth, BitDepth outBitDepth, OptimizationFlags oFlags,  FinalizationFlags fFlags) const;`

**Reviewers:**
I would suggest you start from the public API before digging in the implementation. One could start with the OpenColorIO.h file, the Config class (header + implementation), the Processor class (header + implementation), and the new CPUProcessor class (header + implementation and associated unit tests).

**Pseudo-code** (from ocioconvert):
`OIIO::ImageSpec spec;`
`OCIO::ImgBuffer img;`
`...`
`img.allocate(spec);`
`...`
`OCIO::ConstConfigRcPtr config = OCIO::GetCurrentConfig();`
`OCIO::ConstProcessorRcPtr processor = config->getProcessor(...);`
[...]

PR #827, opened 2019-09-11
Adsk Contrib - Improve the performance of integer bit-depth processing

hodoulp:
The pull request complements #769 by implementing new optimizations around the integer bit-depth processing. The targeted use case is to process an integer bit-depth image and to output it in 32-bit float.
Some ground works were already done on the bit-depth support. Please refer to #699, #722 and #769.

Please refer to #769 to have the previous improvements and here are the new results.

**Measures (RB-1.1 vs branch)** using spi-vfx:

| Conversion | bit-depths | Improvement |
| ----------- | ------------ | ----------- |
| lnf to lgf | UI16 to UI16 | **~14x** |
| lnf to lgf | UI16 to F32 | **~31x** |
| lgf to lnf | UI16 to UI16 | **~9x** |
| lgf to lnf | UI16 to F32 | **~21x** |

Note that the ImageDesc class from the public API can now be used to implement any user custom image layout and be correctly processed by the library. That scenario was forbidden in the previous implementation.

Note that the ImgBuffer class from oiiohelper and ocioperf were improved to process any kind of image bit-depths (i.e. in fact `uint8`, `uint16`, `half` and `float`).


lgritz:
Nice perf results!

doug-walker:
In case it may be helpful, I'm attaching a diagram that illustrates the way memory buffers are used in the new CPUProcessor design.

[CPUProcessor-buffer-usage.pdf](https://github.com/imageworks/OpenColorIO/files/3649960/CPUProcessor-buffer-usage.pdf)

PR #829, opened 2019-09-13
Adsk Contrib - CTF write fixes

BernardLefebvre:
Fix CTF version when saving CDL.
Fix 3D LUT tetrahedal intepolation saving in CTF.
Save an identity matrix when saving an empty transform in CTF.
Enhance ociowrite verbose mode.


michdolan:
If there is no additional feedback on this, we'll merge it tomorrow.

Issue #804, opened 2019-08-20
Minimum gcc version/code license query


KevinJW:
Whilst looking at one of the pull requests I noticed the following code : https://github.com/imageworks/OpenColorIO/blob/e8a0513a661ef570097a35d66f66d05c0378b789/src/OpenColorIO/SSE.h#L48

It begs two (real) questions:

1) What is the minimum gcc version we should support in v2

2) Are there any license implications from the "Copy & Paste"

3) is the code from 4.1.2 or 4.1.1 :-)

Kevin

KelSolaar:
The VFX Reference Platform recommends 6.3.1 for 2018, 19 and 20 with some notes: https://vfxplatform.com/#footnote-gcc6

hodoulp:
Good catch @KevinJW , these lines could be removed.

PR #831, opened 2019-09-14
Remove outdated macros specific to GCC 4 and GCC 5

matthewdcong:
This commit addresses #804. The VFX Reference Platform recommends GCC 6.3.1 for CY2018 and later so these macros are no longer necessary.


hodoulp:
Thanks @matthewdcong for your help.

doug-walker:
We discussed this during the TAC meeting today and agreed that GCC 4 & 5 support may be removed from the master branch.  Thanks Matthew for the PR!

doug-walker:
Also, as was noted on vfx-platform-discuss today, devtoolset-6 (and hence gcc 6) is no longer available with CentOS 7.7.  So that's one more reason why it should be safe to remove these even earlier gcc versions.

Issue #766, opened 2019-06-17
Replace AutoLock with std::lock_guard


michdolan:
Per @lgritz observation in PR #670 : in C++11, AutoLock (src/OpenColorIO/Mutex.h) should be replaced with std::lock_guard.

thirdknife:
Is this issue still open? PR was merged in master.

hodoulp:
@thirdknife This one is already fixed by #832

PR #832, opened 2019-09-14
Replace AutoLock with std::lock_guard

matthewdcong:
Addresses issue #766.


src/OpenColorIO/Mutex.h
```diff
@@ -57,30 +57,14 @@ OCIO_NAMESPACE_ENTER
...
-
 #ifndef NDEBUG
     // add debug wrappers to mutex
     typedef DebugLock<std::mutex> Mutex;
 #else
     typedef std::mutex Mutex;
 #endif
 
-    typedef AutoLock<std::mutex> AutoMutex;
+    typedef std::lock_guard<std::mutex> AutoMutex;
```
hodoulp: A minor change better explaining the design in `debug` mode is to replace `std::mutex` by `Mutex` at line 67.

hodoulp:
While reviewing your changes to remove legacy code, I discovered that the `debug` version of the mutex suffers from several weaknesses.

If you have some time it would be great to take the opportunity to improve the debug version while doing the cleanup.

Below is the original code:
```
#ifndef NDEBUG
    // In debug mode, try to trap recursive cases and lock/unlock debalancing cases
    template <class T>
    class DebugLock : public T {
        public:
            DebugLock() : _locked(0) {}
            ~DebugLock() { assert(!_locked); }
            void lock()   { assert(!_locked); _locked = 1; T::lock(); }
            void unlock() { assert(_locked); _locked = 0; T::unlock(); }
            bool locked() { return _locked != 0; }
        private:
            int _locked;
    };
#endif
```

The weaknesses are:
1. Inheriting from 'Standard Template Library' classes could be problematic as these classes were never designed to be parent classes (i.e. some may not even have a virtual destructor, etc). A better design design is to have a member variable of type `std::mutex`.

2. The `try_lock` method is missing opening the door to bugs.

3. The counter `_locked` is a `signed integer` acting as a boolean, so its type should be `bool`.

A better design could be:
```
#ifndef NDEBUG
    // In debug mode, try to trap recursive cases and lock/unlock debalancing cases
    class DebugLock
    {
    public:
            DebugLock() = default;
            DebugLock(const DebugLock &) = delete;
            DebugLock& operator=(const DebugLock &) = delete;
            ~DebugLock() { assert(!m_locked); }

            void lock()   { assert(!m_locked); m_mutex.lock(); m_locked = true; }
            void unlock() { assert(m_locked); m_mutex.unlock(); m_locked = false; }
            void try_lock() { assert(!m_locked); m_locked = m_mutex.try_lock(); }

            bool locked() const { return _locked; }

        private:
            std::mutex m_mutex;
            bool       m_locked = false;
    };
#endif
```
Here is the reference https://en.cppreference.com/w/cpp/thread/mutex

If you do not have time, we can still create a defect for later improvement.

Anyway thanks for helping us.

matthewdcong:
Sure, I'd be happy to fix that up!

matthewdcong:
Thanks for the suggestions Patrick! I've incorporated your design in my latest commit.

Issue #747, opened 2019-05-04
[v1.1.1] Compilation with -Wall fails with -Werror=cast-function-type


sobotka:
Compilation in the tag v1.1.1 branch with GCC 8.2.0+ fails with -Werror=cast-function-type.

hodoulp:
Based on investigations done by @matthewdcong for #838, these warnings should be gone.
@sobotka Could validate the point and close the issue?

sobotka:
Looks like `RB-1.1` needs the C11 flags set on a clean compile. It errors our with extensive output, first being:
```
/src/apps/ocioconvert/main.cpp:36:
/usr/include/c++/8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
```

hodoulp:
`RB-1.1` is not C++11, and the error (line 36 of `[]/src/apps/ocioconvert/main.cpp`) is related to an OIIO header file. So, the warnings could only be related to the OIIO version you are using. The OIIO library already transitioned to C++11 or higher long-time ago which could explain the warnings.

Anyway that does not seem to be an OCIO issue if `line 36` is the only trigger of all the warnings.

sobotka:
Not sure. I tried a clean compile after a clean clone of the repository, and flipping to the `RB-1.1` branch. Compiled OCIO for years, and can't recall this issue.

The YAML now simply triggers a warning, but the rest of the compilation bombs out with `gcc (Debian 8.3.0-6)`.

sobotka:
OIIO is master. Is there a means to negotiate the conflict?

OCIO seems to have nerfed the original error, but hard to be certain it compiles cleanly.

hodoulp:
> OIIO is master. Is there a means to negotiate the conflict?

Yes. Even if OCIOv1 is not C++11, you could compile it (apps only or apps & core) with the C++11 flag (i.e. add `set(CMAKE_CXX_STANDARD 11)` to the appropriate `CMakeLists.txt`). I would personally limit the change to the apps in order to avoid other potential warnings from the core library.

sobotka:
So I've added that C++11 setting to three apps and it compiles RB-1.1, which dovetails into meaning that this issue is also closed due to the warnings as errors.

michdolan:
Can this be closed?

sobotka:
[...]

PR #838, opened 2019-09-20
Address two warnings in GCC 8/9

matthewdcong:
GCC 9.20 warns about a `const char* -> char*` cast which casts away constness. This was required for the `putenv` function but can be resolved by using the more modern `setenv` function instead.

The GCC 8+ function pointer cast warnings mentioned in #747 appear to be a red-herring. Please refer to https://bugs.python.org/issue33012. Thus, I've updated the CXXFLAGS in the CMakeLists.txt to suppress the warning along with a comment explaining the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,10 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     # Enable a bunch of compiler warnings...
     # http://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wshadow -Wconversion -Wcast-qual -Wformat=2")
+    # GCC 8 and above has a new heuristic to detect invalid pointer
+    # casts. However, this results in false positives on the Python
+    # C API which we compile against to generate Python bindings.
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-cast-function-type")
     # set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pedantic")
 endif(CMAKE_COMPILER_IS_GNUCXX)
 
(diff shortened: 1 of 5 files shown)


hodoulp:
@matthewdcong Another useful pull request to help us maintaining the OCIOv1 release branch  👍

CMakeLists.txt
```diff
@@ -148,6 +148,10 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     # Enable a bunch of compiler warnings...
     # http://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wshadow -Wconversion -Wcast-qual -Wformat=2")
+    # GCC 8 and above has a new heuristic to detect invalid pointer
+    # casts. However, this results in false positives on the Python
+    # C API which we compile against to generate Python bindings.
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-cast-function-type")
```
hodoulp: The warning disabling only applies to the Python bindings, so it should be in the `src/pyblue/CMakeLists.txt` file. The idea is to control which code really needs it to avoid hiding some potential issue on other parts of the library.

src/core/Config.cpp
```diff
@@ -1915,11 +1915,13 @@ OIIO_ADD_TEST(Config, EnvCheck)
     "  - !<View> {name: Raw, colorspace: raw}\n"
     "\n";
     
-    
-    std::string test("SHOW=bar");
-    putenv((char *)test.c_str());
-    std::string test2("TASK=lighting");
-    putenv((char *)test2.c_str());
+#ifdef WINDOWS
```
hodoulp: There is already a method `Platform::getenv()` in `src/core/Platform.h`, so the code could be improved by adding a `Platform::setenv()` to localize the platform differences even if it's only needed by unit tests for now.

CMakeLists.txt
matthewdcong: Good point. I've moved it to `src/pyglue/CMakeLists.txt`.

src/core/Config.cpp
matthewdcong: Added `Platform::setenv()` and changed the unit test to use it.

PR #840, opened 2019-09-21
Port GCC 9 compiler fixes from RB-1.1 to master

matthewdcong:
This changelist implements the equivalent of `Platform::setenv` mentioned in #838 for the master branch. In addition, it also addresses a potential buffer overrun which was triggering a warning in GCC9.


src/OpenColorIO/fileformats/FileFormatDiscreet1DL.cpp
```diff
@@ -807,7 +807,8 @@ void TestToolsStripBlank(
     const std::string & stringResult)
 {
     char stringToStrip[200];
-    stringCopy(stringToStrip, stringToStripChar, 200);
+    stringCopy(stringToStrip, stringToStripChar, 199);
+    stringToStrip[199] = '\0';
```
KevinJW: I'm wondering if the NULL termination should be forced into the stringCopy macro, or if we should evaluate if a safe version of the copy function could be used directly (do all the platforms support one?)

src/OpenColorIO/fileformats/FileFormatDiscreet1DL.cpp
```diff
@@ -818,7 +819,8 @@ void TestToolsStripEndNewLine(
...
```
KevinJW: Same as above

src/OpenColorIO/fileformats/FileFormatDiscreet1DL.cpp
hodoulp: @matthewdcong The GCC9 found a valid 'Linux' issue where the method (a define declared from line 799 to 803) only truncates the string i.e. without adding the null terminated character. 

And @KevinJW is right. The method design is 'weak' as demonstrated by the existing code, and a platform agnostic method must be used.

The solution is to use a **'snprintf'** (i.e. a platform agnostic method) which adds a null termination. Most of the time, the code could be:
`snprintf(dest_buff, sizeof(dest_buff), "%s", src_string);`
( or `sprintf(dest_buff, "%.*s", sizeof(dest_buff) - 1, src_string);`  with a precision ). 

As the destination buffer size is known at compile time, the code could be:
```
constexpr const size_t dstStringSize = 200;
char dstString[dstStringSize];
snprintf(dstString, dstStringSize, "%s", srcString);
```

matthewdcong: Agreed, it's a valid Linux issue that would manifest if the length of `stringToStrip` was equal to or greater than 200.

`snprintf` definitely seems like the right approach here -- I'll incorporate that into my changes. Thanks for the feedback!

hodoulp:

matthewdcong:
I've followed the instructions and added the signoff. I think perhaps this PR was submitted before the signoff was required -- not sure when that got added.

michdolan:
Yes, looks good. The sign-off requirement went into effect Monday. Thanks Matthew!

PR #847, opened 2019-10-02
Adsk Contrib - Ability to customize the default logging mechanism.

hodoulp:

As the library could be integrated in various applications the default logging mechanism must be customizable. The common use case would be to use a DCC logging mechanism if any. By default, the library uses the `std::cerr`.

The pull request adds a public mechanism to register a custom logging method.
Refer to [include/OpenColorIO/OpenColorTypes.h](include/OpenColorIO/OpenColorTypes.h) at line 231 to have the callback details, and to [include/OpenColorIO/OpenColorIO.h](include/OpenColorIO/OpenColorIO.h) at line 138-141 to have the public method declarations.


hodoulp:

PR #849, opened 2019-10-02
Adsk Contrib - Add Raw configuration access

hodoulp:

If the configuration cannot load successfully, apps need a fallback to still have something to use. 

When loading a configuration using the `OCIO` env. variable, the existing `raw` configuration is used if `OCIO` points to nothing (i.e. empty or a 'not accessible' file). But an app could load a 'faulty' configuration file, so it also needs a fallback configuration.

The pull request adds a public method to create a `raw` configuration in memory. I've also added a unit to validate the method.


hodoulp:

Issue #382, opened 2014-12-10
ociobakelut doesn't error if writing to non-existent folder


dbr:
With OCIO 1.0.9, ociobakelut claims to have written the file even if the output path is not writeable (non-existent parent folder or unsuitable permissions), and incorrectly exits with a zero exit-code

For example:

```
$ ociobakelut --inputspace lin --outputspace project --format houdini --v /tmp/nonexistantdirectory.lut
...
[OpenColorIO INFO]: Wrote '/tmp/nonexistantdirectory.lut'

$ ociobakelut --inputspace lin --outputspace project --format houdini --v /root/test.lut
...
[OpenColorIO INFO]: Wrote '/root/test.lut'
```

YashAgrawal0:
Is this issue still open?

doug-walker:
Yes, I believe this is still a useful issue to work on Yash.

YashAgrawal0:
@dbr  @doug-walker  

When I run this command : 

>  ociobakelut --inputspace lin --outputspace project --format houdini --v /tmp/nonexistantdirectory.lut

I got the following errors :  

```
[OpenColorIO Info]: Color management disabled. (Specify the $OCIO environment variable to enable.)
OCIO Error: Error baking houdini:Could not find source color space 'lin'.
```

Can you help me resolve these?

dbr:
Oh, “lin” and “project” are color spaces in our internal OCIO config, but they can be changed to any other space and the problem be the same.

You’d need to download one of the example configs (or write one - the actual content of the colour spaces don’t have any bearing on the error so long as they are syntactically valid), then point the command to that config file either via the OCIO end variable or the `--iconfig` argument). The docs on opencolorio.org should cover this (can’t link to them as Github pages is currently down for some reason)

YashAgrawal0:
In my machine I ran ociobakelut command for both the cases when output file path is specified in a non-existent folder and in a directory with root permissions, in both the cases I got the error  `non-writable path` which was the expected behaviour.

Refer to the [image](https://user-images.githubusercontent.com/24262584/78782486-be801180-79bf-11ea-9a7b-53a733231ca4.png) for more info.

dbr:
Ahh! I only just noticed this is already fixed in 01246674361fcee31be8aba2eee74722f1f018c7 via pull request #854

PR #854, opened 2019-10-04
Check fail bit to catch non-writable file paths

matthewdcong:
Addresses #382: If a non-writable path is specified during the construction of an `std::ofstream`, the `std::fail` bit is set to true. By querying this before attempting to write to the path, we can report the correct error message and return 1.

Tested using the aces.1.0.3 configuration with writable and non-writable paths.


hodoulp:

PR #855, opened 2019-10-04
Fix segfault when baking without a config

### LINKED ISSUE #388: PyOpenColorIO Baker segfault
```
>>> import PyOpenColorIO
>>> PyOpenColorIO.version
'1.0.9'
>>> b = PyOpenColorIO.Baker() 
>>> b.setFormat("cinespace")
>>> b.bake()
Segmentation fault
```

The specific format isn't important, happens with "lustre", "hoduini" etc - but if unset it gives a useful `PyOpenColorIO.Exception: The format named '' could not be found.` error

matthewdcong:
Fixes #388. The file format implementations call:
`ConstConfigRcPtr config = baker.getConfig()` followed by `config->getProcessor`. However, if a configuration has not been set for the Baker, config will be a null `shared_ptr` and the `getProcessor` call with segfault. Added a unit test to verify that an exception is properly thrown and verified that the Python script in the issue no longer segfaults.

### DIFF
--- a/src/OpenColorIO/Baker.cpp
+++ b/src/OpenColorIO/Baker.cpp
@@ -114,7 +114,22 @@ OCIO_NAMESPACE_ENTER
     
     void Baker::setFormat(const char * formatName)
     {
-        getImpl()->formatName_ = formatName;
+        FileFormat* fmt = FormatRegistry::GetInstance().getFileFormatByName(formatName);
+        FormatInfoVec formatInfoVec;
+        fmt->getFormatInfo(formatInfoVec);
+        for(unsigned int i=0; i<formatInfoVec.size(); ++i)
+        {
+            if(formatInfoVec[i].capabilities & FORMAT_CAPABILITY_BAKE)
+            {
+                getImpl()->formatName_ = formatName;
+                return;
+            }
+        }
+
+        std::ostringstream os;
+        os << "File format " << formatName;
+        os << " does not support baking.";
+        throw Exception(os.str().c_str());
     }
     
     const char * Baker::getFormat() const
@@ -216,9 +231,7 @@ OCIO_NAMESPACE_ENTER
 
         if(!getConfig())
         {
-            std::ostringstream err;
-            err << "No OCIO config has been set";
-            throw Exception(err.str().c_str());
+            throw Exception("No OCIO config has been set");
         }
    
         try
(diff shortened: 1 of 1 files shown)


src/OpenColorIO/Baker.cpp
```diff
@@ -213,7 +213,14 @@ OCIO_NAMESPACE_ENTER
             err << "' could not be found. ";
             throw Exception(err.str().c_str());
         }
-        
+
+        if(!getConfig())
+        {
+            std::ostringstream err;
+            err << "No OCIO config has been set";
+            throw Exception(err.str().c_str());
```
hodoulp: These three lines could be merged in a single line like:
`throw Exception("No OCIO config has been set.");`

src/OpenColorIO/Baker.cpp
```diff
@@ -388,6 +395,16 @@ OCIO_ADD_TEST(Baker_Unit_Tests, bake)
 
 }
 
+OCIO_ADD_TEST(Baker_Unit_Tests, bake_empty_config)
```
hodoulp: [Minor] To better follow other unit test names, the 'major' name is the class name i.e. 'Baker', and the 'minor' name is the test itself i.e. 'empty_config' or 'missing_config'.

src/OpenColorIO/Baker.cpp
```diff
@@ -388,6 +395,16 @@ OCIO_ADD_TEST(Baker_Unit_Tests, bake)
...
+{
+    // Verify that running bake with an empty configuration
+    // throws an exception and does not segfault.
+    OCIO::BakerRcPtr bake = OCIO::Baker::Create();
+    bake->setFormat("cinespace");
```
hodoulp: [Request] The `setFormat()` method does not validate that the format accepts the 'bake' mode.  That could be something interesting to add if you have time.

hodoulp:

src/OpenColorIO/Baker.cpp
matthewdcong: Merged.

src/OpenColorIO/Baker.cpp
matthewdcong: Got it, thanks! Changed the major name to `Baker` for both the `bake` and `empty_config` tests.

src/OpenColorIO/Baker.cpp
matthewdcong: Added the validation.

matthewdcong:
Thanks for reviewing, Patrick! I've made the changes suggested and would be happy to backport this to OCIOv1 in the near future.

PR #856, opened 2019-10-04
Fix spelling mistake in config class

matthewdcong:
Signed-off-by: Matthew Cong <1372750+matthewdcong@users.noreply.github.com>

intenal -> internal. Noticed it while greping through the codebase.


hodoulp:

PR #857, opened 2019-10-04
AZP configuration updates

michdolan:
Currently we have a forced CI cron build every Sunday. With this PR, Monday-Saturday now have a full CI build too IF code has been changed. This will update static analysis more regularly, until we get the quality gate to a place where it can be enabled as a PR check.

I also moved the "latest" builds to use the ci-common Docker image and removed explicit lib paths from their cmake overrides. They pointed to the default install location for each lib, so weren't necessary.

Implemented macOS python install improvement suggested by Christina Tempelaar-Lietz (OpenEXR). This removes a lot of warnings from the install log, potentially speeding up mac builds.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -109,7 +109,7 @@ jobs:
   pool:
     vmImage: 'ubuntu-16.04'
   container: $[ variables['containerImage'] ]
-  #condition: in(variables['Build.Reason'], 'Manual', 'Schedule')
+  condition: in(variables['Build.Reason'], 'Manual', 'Schedule')
 
   steps:
   - template: share/ci/templates/checkout.yml
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,6 +34,9 @@ schedules:
       include:
         - master
 
+variables:
+- group: sonar
+
 jobs:
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


hodoulp:
Here is the problem:
```
09:52:33.231 ERROR: Error during SonarQube Scanner execution
java.lang.IllegalStateException: You're not authorized to run analysis. Please contact the project administrator.
```

michdolan:
SonarCloud is back up and running:
https://sonarcloud.io/dashboard?id=AcademySoftwareFoundation_OpenColorIO

I've re-enabled the condition limiting it to the cron builds, but the above is the output from the run in this PR.

zachlewis:
Re: the macos python script, for posterity --

It seems that the SDK headers installer provided with xcode-10 is only intended as a stopgap for builds that rely on macros with hardcoded references to /usr/include; the installer isn't provided for xcode-11 / macos-10.15. 

Instead, it might make more sense to pass the SDK path to clang via the -isysroot compiler flag: e.g.,
export CFLAGS= "-I$(brew --prefix openssl)/include -Wno-nullability-completeness-isysroot $(xcrun --show-sdk-path)"


Source: https://apple.stackexchange.com/questions/337940/why-is-usr-include-missing-i-have-xcode-and-command-line-tools-installed-moja

michdolan:
Thanks Zach! Timely insight as our macOS CI hit a wall related to this exact thing today, because of an update of our hosted CI macOS agent. I'll give this a whirl.

Issue #318, opened 2013-06-08
OpencolorIO failed to build with yaml-cpp 0.5.1


Pinkbyte:
build output:

/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.cpp: In function 'void OpenColorIO::v1::LogUnknownKeyWarning(const string&, const YAML::Node&)':
/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.cpp:45:16: error: no match for 'operator>>' in 'tag >> key'
/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.cpp:45:16: note: candidates are:
/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.h:80:49: note: void OpenColorIO::v1::operator>>(const YAML::Node&, OpenColorIO::v1::ColorSpaceRcPtr&)
/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.h:80:49: note:   no known conversion for argument 2 from 'std::string {aka std::basic_string<char>}' to 'OpenColorIO::v1::ColorSpaceRcPtr& {aka std::tr1::shared_ptrOpenColorIO::v1::ColorSpace&}'
/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.h:82:49: note: void OpenColorIO::v1::operator>>(const YAML::Node&, OpenColorIO::v1::GroupTransformRcPtr&)
/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.h:82:49: note:   no known conversion for argument 2 from 'std::string {aka std::basic_string<char>}' to 'OpenColorIO::v1::GroupTransformRcPtr& {aka std::tr1::shared_ptrOpenColorIO::v1::GroupTransform&}'
/var/tmp/portage/media-libs/opencolorio-1.0.8/work/opencolorio-1.0.8/src/core/OCIOYaml.h:84:49: note: void OpenColorIO::v1::operator>>(const YAML::Node&, OpenColorIO::v1::TransformRcPtr&)
[...]

PR #859, opened 2019-10-05
Replace deprecated method GetMark

melMass:
Related to #318 

I could not build on MacOs using YAML-cpp 0.6.3 
The `GetMark` method is not in YAML-cpp since 0.5 it has been replaced by `Mark`

### DIFF
--- a/src/OpenColorIO/OCIOYaml.cpp
+++ b/src/OpenColorIO/OCIOYaml.cpp
@@ -116,7 +116,7 @@ OCIO_NAMESPACE_ENTER
             if (!node.Read<bool>(x))
             {
                 std::ostringstream os;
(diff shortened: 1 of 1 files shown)


melMass:
As the CI failed I looked at the log and the CI Container is set to use YAMLCPP 0.3 so it failed. When trying to implement a define for backward compatibility I realized there was already one...
For some reason, the `OLDYAML` definition was not triggered on mac. I'll pull request if I find the culprit.

hodoulp:
Thanks @melMass to put some time in the OpenColorIO project. @michdolan also faced some challenges around the 'yaml' library few weeks ago when he reorganized the third-party library dependencies. 

Anyway, if you have any question, please feel free to contact us.

PR #861, opened 2019-10-07
Create EasyCLA-Validation.md

jpalmerLinuxFoundation:


hodoulp:
@jpalmerLinuxFoundation Thanks your contribution to OpenColorIO project.

The change is small but I do not understand it. Could you add some details on the rational the change?    @jmertic Is that related to the problem you mentioned during the TSC meeting?

jpalmerLinuxFoundation:
> @jpalmerLinuxFoundation Thank you for your contribution to OpenColorIO project.
> 
> The change is small but I do not understand it. Could you add some details on the rational the change? @jmertic Is that related to the problem you mentioned during the TSC meeting?

Yes, this is to troubleshoot EasyCLA. We determined that the app was not installed, so now we're re-installing and testing it.

jpalmerLinuxFoundation:
This PR can be deleted.

PR #862, opened 2019-10-07
Update Docker build files to use branch release 2.0 of OpenImageIO

thirdknife:


hodoulp:
Thanks @thirdknife to improve the docker file with a better OIIO version. We may have notice that `boost` is now not used in the `master` branch. If you have some time, could take that opportunity to also remove the `boost` step in the docker files?

PR #865, opened 2019-10-08
After Effects and Photoshop plug-ins updated for v2

fnordware:


fnordware:
GitHub says I have to do this --signoff thing, but when I followed its instuctions it sent me into a long sequence of conflicts involving every commit I've done. Then I tried running the rebase command with -X theirs and it added a whole bunch of other commits to my list.

Any suggestions?

hodoulp:
@fnordware Could you provide details about the branch itself? 
When did you create it? from which repo? etc.

But one thing I can highlight is that all commits must now be `signoff`. I personally squash all the commits before creating the pull request.

fnordware:
I basically made a branch called "v2" that merged in everything from the OpenColorIO master branch. I worked in a branch from that, "v2_ae", rebasing a few times to bring myself up to date.

How do I squash commits together? I basically just want everything to wind up in its current state with the --signoff thing on, preferably without having to go through oodles of merge conflicts to get there.

fnordware:
OK, I started with a clean branch, copied the latest files into place, and make a new commit with everything. And --singoffed! Should be easier to swallow now.

vendor/aftereffects/OpenColorIO_AE_Context.cpp
```diff
@@ -623,28 +628,43 @@ void OpenColorIO_AE_Context::setupDisplay(const char *input, const char *device,
...
+void OpenColorIO_AE_Context::setupLUT(OCIO_Invert invert, OCIO_Interp interpolation)
 {
     OCIO::FileTransformRcPtr transform = OCIO::FileTransform::Create();
     
     if(interpolation != OCIO_INTERP_NEAREST && interpolation != OCIO_INTERP_LINEAR &&
-        interpolation != OCIO_INTERP_TETRAHEDRAL && interpolation != OCIO_INTERP_BEST)
+        interpolation != OCIO_INTERP_TETRAHEDRAL && interpolation != OCIO_INTERP_CUBIC &&
+        interpolation != OCIO_INTERP_BEST)
     {
         interpolation = OCIO_INTERP_LINEAR;
```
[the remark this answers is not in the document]
fnordware: That code is meant to catch the case where an unfamiliar value gets sent in, correcting it to my default, which is Linear.

But I guess you're touching on a design choice. I don't provide "Default" as an option, I default to Linear. Personally I preferred that, not knowing what "Default" would mean. On the other hand, I do provide "Best" as an option.

I guess for now the default for all LUTs _is_ Linear, so maybe it's fine?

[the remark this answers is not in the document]
michdolan: Ok. If it relates more to your menu, and the user sees the linear default being chosen there, that makes total sense.

I was considering a case where a OCIO default could someday change (i.e. tetrahedral for 3D LUT), which could then lead to a config with a invalid interpolation value behaving differently in these plugins vs in other DCCs. For now, linear is correct though, yes.

This note in the OCIO headers illustrates the current state of things:

```
    // INTERP_DEFAULT will choose the default interpolation type for the requested
    // context:
    //
    // 1D LUT INTERP_DEFAULT: LINEAR
    // 3D LUT INTERP_DEFAULT: LINEAR
    //
    // INTERP_BEST will choose the best interpolation type for the requested
    // context:
    //
    // 1D LUT INTERP_BEST: LINEAR
    // 3D LUT INTERP_BEST: TETRAHEDRAL
    //
    // Note: INTERP_BEST and INTERP_DEFAULT are subject to change in minor
    // releases, so if you care about locking off on a specific interpolation
    // type, we'd recommend directly specifying it.
```

vendor/photoshop/OpenColorIO_PS_Context.cpp
```diff
@@ -140,7 +152,11 @@ OpenColorIO_PS_Context::getLUTProcessor(OCIO::Interpolation interpolation, OCIO:
     
     OCIO::ConstProcessorRcPtr processor = _config->getProcessor(transform);
     
-    return processor;
+    OCIO::ConstCPUProcessorRcPtr cpu_processor = (direction == OCIO::TRANSFORM_DIR_INVERSE ?
+                                                    processor->getOptimizedCPUProcessor(OCIO::OPTIMIZATION_DEFAULT, OCIO::FINALIZATION_EXACT) :
+                                                    processor->getDefaultCPUProcessor());
+
```
hodoulp: The code above seems strange to use the Inverse or Forward direction to select the finalization flag?

vendor/photoshop/OpenColorIO_PS_Context.cpp
```diff
@@ -258,3 +274,19 @@ OpenColorIO_PS_Context::getDefaultTransform(const std::string &device) const
...
+    value = (env != NULL ? env : "");
+#else
+    char env[1024] = { '\0' };
+
+    const DWORD result = GetEnvironmentVariable(name, env, 1023);
+
+    value = (result > 0 ? env : "");
+#endif
+}
```
hodoulp: The env. access implementation is in several files. Do you think OCIO should have Getenv() & SetEnv() is the public API?

vendor/photoshop/OpenColorIO_PS_Context.cpp
fnordware: I figured for Photoshop we may as well to the exact inverse on a LUT. I guess I could use the EXACT flag on the forward transform as well, but figured it wouldn't change anything.

vendor/photoshop/OpenColorIO_PS_Context.cpp
fnordware: I made my own function so I could use GetEnvironmentVariable() on Windows instead of std::getenv(), which was not updating. Would use an OpenColorIO version if there was one. I especially thing the $OCIO variable should be formalized somewhere. You can see I made this to do that:
```
static void getenvOCIO(std::string &value) { getenv("OCIO", value); }
```

vendor/photoshop/OpenColorIO_PS_Context.cpp
doug-walker: Yes Brendan, the EXACT flag has no effect on forward evals (they are always exact).  If it's working fast enough for you, that's fine to use EXACT.  Give it a try on some large images to make sure.

fnordware: OK, I changed it in b6cc8072ee55915c4cba36d206e7609378cd3975 so I'm always using EXACT in Photoshop for both forward and reverse. I figure for still images Photoshop users won't mind waiting a little longer for their LUTs to run. In After Effects I leave EXACT as an optional thing.

hodoulp:
LGTM

PR #868, opened 2019-10-15
Adsk Contrib - Remove bit-depth from ops.

BernardLefebvre:
To better support the CLF/CTF file formats which contain input and output bit-depths on each op, a previous PR added the bit-depths to the definition of all ops.  (This is also how SynColor works.)  This implied, for example, that LUT entries in the ops were scaled based on the bit-depths.  However, we decided that it would be a better/simpler design if we have the CLF/CTF reader and writer normalize LUT values on-the-fly by the bit-depth (see fileformats/ctf/CTFReaderHelper.cpp).  One benefit of this is that OCIO developers that don't care about CLF/CTF bit-depth, now don't have to deal with it or consider it.

We did preserve the ability to get/set the file bit-depth for those developers that want to know or control the bit-depth when reading/writing CLF/CTF files and various other LUT formats.

A previous PR added bit-depth support to the ImageDesc and CPUProcessor.  Since most ops already worked in a normalized manner, this PR does not change evaluation too much.  The one op that works directly on all bit-depths is LUT1D, so this PR has more changes there.  As a result, we also added a bunch of additional LUT1D unit tests in this PR.


hodoulp:

PR #869, opened 2019-10-16
Adsk Contrib - Add 10 & 12 bit-depth support

hodoulp:

Now that the bit-depth support #769 and the bit-depth integer optimizations #827 are in, the current work adds the 10-bit and 12-bit bit-depths and it also improves some unit tests to better test integer bit-depths.


hodoulp:

PR #870, opened 2019-10-18
Adsk Contrib - Transforms interface

BernardLefebvre:
Rework GroupTransform interface and remove transform functions using float when there is a double version.

(The decision to remove the float versions was made by the group during the working group meeting at SIGGRAPH.)


src/bindings/python/PyExponentTransform.cpp
```diff
@@ -124,9 +125,10 @@ OCIO_NAMESPACE_ENTER
         {
             OCIO_PYTRY_ENTER()
             ConstExponentTransformRcPtr transform = GetConstExponentTransform(self);
-            std::vector<float> data(4);
-            transform->getValue(&data[0]);
-            return CreatePyListFromFloatVector(data);
+            double data[4];
+            transform->getValue(data);
+			std::vector<double> dataVec{ data, data + sizeof(data) / sizeof(double)  };
```
michdolan: [minor] This line has tab indent but neighbors use spaces.

include/OpenColorIO/OpenColorTransforms.h
```diff
@@ -864,15 +848,9 @@ OCIO_NAMESPACE_ENTER
         TransformRcPtr & getTransform(int index);
 
         //!cpp:function::
-        int size() const;
-        //!cpp:function:: Adds a copy of transform to the group.
-        void push_back(const ConstTransformRcPtr & transform);
-        //!cpp:function:: Adds transform to the group.
-        void push_back(TransformRcPtr & transform);
-        //!cpp:function:: Clears transforms (not the FormatMetadata).
-        void clear();
```
michdolan: I like the new GroupTransform interface, but wondering if we should keep `clear()` or another way to remove transforms. I haven't had a need for it in practice myself yet, but there could be cases.

michdolan: Could also be a pop or insert method or something. Not to go overkill in making a vector interface... since it seems like your goal is to get away from it.

BernardLefebvre: The goal was to simply the interface. Most common use is to create a group transform and to add transforms to it, so there is no real need for erase, insert etc.
If a empty group transform is needed, user can use GroupTransform::Create.

src/bindings/python/PyExponentTransform.cpp
BernardLefebvre: tabs have been replaced by spaces

include/OpenColorIO/OpenColorTransforms.h
michdolan: Ok. Seems fair. Users can iterate one GroupTransform and selectively add transforms to a new empty GroupTransform if variation is needed.

src/OpenColorIO/CPUProcessor.cpp
```diff
@@ -1680,7 +1680,7 @@ OCIO::ConstCPUProcessorRcPtr BuildCPUProcessor(OCIO::TransformDirection dir)
     OCIO::ConfigRcPtr config = OCIO::Config::Create();
 
     OCIO::MatrixTransformRcPtr transform = OCIO::MatrixTransform::Create();
-    const float offset4[4] = { 1.4002f, 0.4005f, 0.8007f, 0.5007f };
+    const double offset4[4] = { 1.4002, 0.4005, 0.8007, 0.5007 };
```
KevinJW: should this be constexpr? I think there are a number of those in this PR.

src/OpenColorIO/Config.cpp
```diff
@@ -43,9 +43,9 @@ OCIO_NAMESPACE_ENTER
         };
         
         // These are the 709 primaries specified by the ASC.
-        const float DEFAULT_LUMA_COEFF_R = 0.2126f;
-        const float DEFAULT_LUMA_COEFF_G = 0.7152f;
-        const float DEFAULT_LUMA_COEFF_B = 0.0722f;
+        const double DEFAULT_LUMA_COEFF_R = 0.2126;
+        const double DEFAULT_LUMA_COEFF_G = 0.7152;
+        const double DEFAULT_LUMA_COEFF_B = 0.0722;
```
KevinJW: constexpr?

src/OpenColorIO/CPUProcessor.cpp
BernardLefebvre: changed

src/OpenColorIO/Config.cpp
BernardLefebvre: changed

hodoulp:

PR #874, opened 2019-10-24
Adsk Contrib - Remove useless warnings from make files.

hodoulp:

The pull request removes lot of warnings from various make files, which do not add anything to the build. All theses warnings are from third-party libraries.

The idea is to only display warnings to be fixed at some point.

Note that `yaml 0.3.0` makefile generation & compilation still display some warnings that only a code patch could fix.


hodoulp:
Friendly reminder that this PR qualifies for single approver rule so it could be merged this week.

PR #324, opened 2013-08-22
added support for activeViewsEnvOverride_ in getActiveViews()

compso:
added fix for using OCIO_ACTIVE_VIEWS env variable to filter views, related to issue https://github.com/imageworks/OpenColorIO/issues/67

Issue #791, opened 2019-08-08
[v1.1.1] Default display is alphabetical first if active_displays is not set


fnRaihanKibria:
According to https://opencolorio.org/userguide/config_syntax.html : "Unless the active_displays and active_views sections are defined, the first display and first view will be the default."

However using a config without an active_displays and active_views section results in the alphabetical first display being the default. Iterating through the displays also lists them in alphabetical order rather than the order of definition in the config file, although I'm not sure if that's intentional or not.

The views on the other hand are ordered correctly as they are in the config and default to the first in that order.

Config file snippet:
```
displays:
  zz:
    - !<View> {name: z1, colorspace: raw}
    - !<View> {name: y1, colorspace: Gamma2.4}
    - !<View> {name: x1, colorspace: rec709}
    - !<View> {name: a1, colorspace: sRGB}
  yy:
    - !<View> {name: z2, colorspace: raw}
    - !<View> {name: y2, colorspace: Gamma2.4}
    - !<View> {name: x2, colorspace: rec709}
    - !<View> {name: a2, colorspace: sRGB}
  xx:
    - !<View> {name: z3, colorspace: raw}
    - !<View> {name: y3, colorspace: Gamma2.4}
    - !<View> {name: x3, colorspace: rec709}
    - !<View> {name: a3, colorspace: sRGB}
  aa:
    - !<View> {name: z4, colorspace: raw}
    - !<View> {name: y4, colorspace: Gamma2.4}
    - !<View> {name: x4, colorspace: rec709}
    - !<View> {name: a4, colorspace: sRGB}
```

Test program:
```
#include <iostream>
#include <OpenColorIO/OpenColorIO.h>

namespace OCIO = OCIO_NAMESPACE;

int main()
{
    std::cout << "Using OCIO Version : " << OCIO_VERSION << std::endl;
    try
    {
        OCIO::ConstConfigRcPtr config = OCIO::GetCurrentConfig();
        std::cout << "default display: " << config->getDefaultDisplay() << std::endl;

        const int num_disps = config->getNumDisplays();
        std::cout << "number of displays: " << num_disps << std::endl;

        for(int i=0; i < num_disps; ++i) {
            const char* display = config->getDisplay(i);
            std::cout << "display " << i << " default view: " << config->getDefaultView(display) << std::endl;
            
            const int num_views = config->getNumViews(display);
            for(int k=0; k < num_views; ++k) {
                const char* view = config->getView(display, k);
[...]

Issue #67, opened 2011-02-02
OCIO_ACTIVE_VIEWS does not filter view list


jeremyselan:
The ACTIVE_DISPLAYS envvar is used to both filter, and prioritize, the list of allowed displays. The ACTIVE_VIEWS envvars is only used for prioritization. Make it do both.

doug-walker:
This was fixed in PR #875.  In OCIO 2.x, the Config getViews method only returns active views.

PR #875, opened 2019-10-24
Adsk Contrib - Fix the active views sorting and filtering behavior

hodoulp:

Refer to #67.

The active display list filters and orders the list of accessible displays present in a config file. The active display list could be defined using `active_displays` present in a config file and/or using the env. variable `OCIO_ACTIVE_DISPLAYS` (i.e. both mechanism could coexist).

The problem is that the active view list failed to filter and order the list of active views for a specific display even if `active_views` or `OCIO_ACTIVE_VIEWS` mechanisms also exist.

The pull request fixes the active view list issues, and it also adjusts unit tests to validate the fix.

**Note** that some inconsistent behaviors still exist for both active displays and views. In order to speedup the merge, these inconsistencies were not addressed here. It requires further discussions with the community to define the expected behavior before updating the code.

**Note** that an additional unit test highlights the known display name order issue (refer to #791). Some investigations raised a potential `yaml` limitation so it requires a discussion around the current `yaml` version before any code changes.


hodoulp:

hodoulp:
The pull request fixes #324

PR #876, opened 2019-10-25
Fix macOS CI python install script

michdolan:
Implemented fix proposed by @zachlewis , which will hopefully resolve the current macOS 10.14 CI failure, resulting from a macOS hosted agent update.

Signed-off-by: michdolan <michdolan@gmail.com>

### DIFF
--- a/share/ci/scripts/macos/install_python.sh
+++ b/share/ci/scripts/macos/install_python.sh
@@ -6,13 +6,18 @@ set -ex
 
 PYTHON_VERSION="$1"
 
-brew install pyenv openssl
+brew install pyenv openssl xqurtz
 
-# Include openssl, disable nullability warning, to greatly reduce log noise during Python install
+# Greatly reduce log warnings during Python install
+export CFLAGS="-I$(brew --prefix openssl)/include -Wno-nullability-completeness"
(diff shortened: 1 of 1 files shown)


michdolan:
The CI issue is fixed. There was a non-criritcal error in the python built when trying to build tkinter (arguably not needed here), but seeing if I can get a complete python build for completeness...

share/ci/scripts/macos/install_python.sh
```diff
@@ -6,24 +6,16 @@ set -ex
...
-if [[ "$MACOS_MAJOR" -gt 9 && "$MACOS_MINOR" -gt 13 ]]; then
-    sudo installer \
-        -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_${MACOS_MAJOR}.${MACOS_MINOR}.pkg \
-        -target /
-fi
-
 brew install pyenv openssl
 
 # Greatly reduce log warnings during Python install
 export CFLAGS="-I$(brew --prefix openssl)/include -Wno-nullability-completeness"
```
hodoulp: $CLFAGS is missing at the end of the line. Even if the Azure macOS succeeded, could that be a problem?

PR #879, opened 2019-10-30
Adsk Contrib - Add baking support for CLF/CTF.

BernardLefebvre:
Add baking support for CLF/CTF.
The CLF/CTF formats will make use of the baker metadata "id" and "name" attributes and children elements "Description", "InputDescriptor", "OutputDescriptor", and "Info"
Formats such as CSP, IridasCube, and ResolveCube use baker metadata for comments.


hodoulp:

PR #880, opened 2019-10-31
Fix AZP macOS dependency install failure

michdolan:
The build failures this fixes appear to be caused by another AZP macOS agent update, related to both CMake and Python. This PR removes our custom install of CMake (since AZP has a compatible version installed already now), and refines the pyenv build dependencies. Both of these were contributing to the dependency install task failing for macOS. 

Signed-off-by: michdolan <michdolan@gmail.com>

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -184,21 +184,16 @@ jobs:
   - template: share/ci/templates/checkout.yml
   - bash: |
       share/ci/scripts/macos/install.sh
+      share/ci/scripts/macos/install_python.sh 2.7.16
     displayName: Install dependencies
-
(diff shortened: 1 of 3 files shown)


hodoulp:
@michdolan That's the second time the macOS virtual machine configuration changed causing us trouble with the builds. None of them were blocking us but one-day it could be a blocker. Are you aware of any way to have some 'notifications' of coming 'virtual machine and/or configuration' changes?

michdolan:
They do have an RSS feed:
https://status.dev.azure.com/_rss

I'll reach out to LFRE to see if they have any visibility into coming updates.

PR #882, opened 2019-11-04
2019-11-04 TSC meeting notes

michdolan:
Signed-off-by: michdolan <michdolan@gmail.com>

PR #883, opened 2019-11-07
Adsk Contrib - Fix the display order issue

### LINKED ISSUE #517: Doesn't build with yaml-cpp 0.6.0
Attempted build on Fedora Rawhide with GCC 8 preview:

`
[ 73%] Linking CXX executable ocio_core_tests
cd /builddir/build/BUILD/OpenColorIO-1.1.0/build/src/core_tests && /usr/bin/cmake -E cmake_link_script CMakeFiles/ocio_core_tests.dir/link.txt --verbose=1
/usr/bin/c++  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -mcet -fcf-protection -Wall -Wextra -Wshadow -Wconversion -Wcast-qual -Wformat=2 -msse2 -O2 -g -DNDEBUG  -Wl,-z,relro  -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -rdynamic CMakeFiles/ocio_core_tests.dir/__/core/AllocationOp.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/AllocationTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Baker.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/CDLTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Caching.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ColorSpace.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ColorSpaceTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Config.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Context.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Display.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/DisplayTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Exception.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ExponentOps.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ExponentTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormat3DL.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatCC.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatCCC.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatCDL.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatCSP.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatHDL.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatIridasCube.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatIridasItx.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatIridasLook.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatPandora.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatSpi1D.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatSpi3D.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatSpiMtx.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatTruelight.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileFormatVF.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/FileTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/GpuShaderDesc.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/GpuShaderUtils.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/GroupTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/HashUtils.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ImageDesc.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ImagePacking.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/LogOps.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/LogTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Logging.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Look.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/LookParse.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/LookTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Lut1DOp.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Lut3DOp.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/MathUtils.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/MatrixOps.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/MatrixTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/NoOps.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/OCIOYaml.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Op.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/OpOptimizers.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ParseUtils.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/PathUtils.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Platform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Processor.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/ScanlineHelper.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/Transform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/TruelightOp.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/TruelightTransform.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/UnitTest.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/md5/md5.cpp.o CMakeFiles/ocio_core_tests.dir/__/core/pystring/pystring.cpp.o  -o ocio_core_tests -ltinyxml -lyaml-cpp 
make[2]: Leaving directory '/builddir/build/BUILD/OpenColorIO-1.1.0/build'
CMakeFiles/ocio_core_tests.dir/__/core/OCIOYaml.cpp.o: In function `YAML::Exception::Exception(YAML::Mark const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/usr/include/yaml-cpp/exceptions.h:122: undefined reference to `vtable for YAML::Exception'
CMakeFiles/ocio_core_tests.dir/__/core/OCIOYaml.cpp.o: In function `YAML::RepresentationException::RepresentationException(YAML::Mark const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/usr/include/yaml-cpp/exceptions.h:155: undefined reference to `vtable for YAML::RepresentationException'
CMakeFiles/ocio_core_tests.dir/__/core/OCIOYaml.cpp.o: In function `YAML::Exception::Exception(YAML::Mark const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/usr/include/yaml-cpp/exceptions.h:122: undefined reference to `vtable for YAML::Exception'
CMakeFiles/ocio_core_tests.dir/__/core/OCIOYaml.cpp.o: In function `YAML::RepresentationException::RepresentationException(YAML::Mark const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/usr/include/yaml-cpp/exceptions.h:155: undefined reference to `vtable for YAML::RepresentationException'
CMakeFiles/ocio_core_tests.dir/__/core/OCIOYaml.cpp.o: In function `YAML::Exception::Exception(YAML::Mark const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/usr/include/yaml-cpp/exceptions.h:122: undefined reference to `vtable for YAML::Exception'
CMakeFiles/ocio_core_tests.dir/__/core/OCIOYaml.cpp.o: In function `YAML::RepresentationException::RepresentationException(YAML::Mark const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/usr/include/yaml-cpp/exceptions.h:155: undefined reference to `vtable for YAML::RepresentationException'
/usr/bin/ld: ocio_core_tests: hidden symbol `_ZTVN4YAML9ExceptionE' isn't defined
/usr/bin/ld: final link failed: Bad value
`

hodoulp:

The pull request starts from investigations to only upgrade the Yaml third-party library from 0.3 to 0.6.3 (i.e. latest one) to finally fixes various long-standing issues:

- Imposes double serialization precision to 15 digits [src/OpenColorIO/OCIOYaml.cpp](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/display_order?expand=1#diff-74c88273b31122385791a6a558e04b87R2494).
- Fixes the display order issue. The Yaml 0.3 implementation uses an `std::map` for the Yaml:Map implementation (breaking the config file order as the `std::map` imposes an alphabetical order by design). Starting from version 0.6.0, the implementation is now using a `std::vector< std::pair<> >` preserving the original order ([yaml-cpp/node/detail/node_data.h](https://github.com/jbeder/yaml-cpp/blob/master/include/yaml-cpp/node/detail/node_data.h#L117)). With a small change ([src/OpenColorIO/Display.h](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/display_order?expand=1#diff-5a3dc2b9904de7f343ff19ea82bfcf6dR42)), OCIO now preserves the display order from the Yaml parsing.

For detailed explanation of the Yaml changes, please refer to:
https://github.com/jbeder/yaml-cpp/pull/386
https://github.com/jbeder/yaml-cpp/pull/379

- Adds display checks in [Config::sanityCheck()](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/display_order?expand=1#diff-25497053ae4f9672912e8c24d14c8a66R616) to validate that active display exists.
- Cleanups the code to only support Yaml recent version (i.e. drop support of versions lower than 0.6.0). Decision made at the last TSC meeting (i.e. 2019-11-04 #882)

The work should fix #517, #389, #577 and #791.


hodoulp:
Yaml library fails on parsing configuration files when compiled with Clang 7 or higher in release mode only. I did some successful tests with clang 6.

Problematic Azure builds are: 
- Linux CentOS 7 Clang 7 in Release only
- macOS 10.14 AppleClang 10.0 in Release only

hodoulp:
For **Yaml 0.6.x** compiled with **Clang 7.x** (and higher), the `Yaml::Sequence` has a bug when accessing an element using an index!

Note: My experiments indicate that it works if an iterator is used (instead of an index).

hodoulp:
There are unfortunately more issues between Yaml 0.6.x and Clang 7 (and higher) optimizations. Even the latest changes to improve the error handling failed to fix all 'understandable' issues and to trap the remaining problem(s). As the config. file parsing does not need specific optimizations (i.e. read performance not being a concern), the last change disables the Clang code optimizations for the [OCIOYaml.cpp](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/883/commits/cd4fe0bf8d001cebb415e51bcfca3b0d93993559#diff-74c88273b31122385791a6a558e04b87R54) file only.

Please let me know if you have any concerns.

KevinJW:
So to make doubles work fully the  magic number of digits should typically be 17 as given by std::numeric_limits<T>::max_digits10

if we want to serialise in memory values to file and back again.

Kevin

hodoulp:
The OCIOv1 code only serializes float values using the ‘default’ Yaml 0.3.0 precision of 6 digits. So, it was clamping some Op values. The OCIOv2 code with Yaml 0.6.3 now only serializes double values with a ‘imposed’ 15-digit precision clearly improving the overall serialization accuracy.

The code does not use the 17-digit default double precision (nor the 16-digit) because of the imprecision of any floating-point values impacting most of the serialization/deserialization. For example, serializing the LUMA values wrote `[0.21260000000000001, 0.71519999999999995, 0.0722]` (and `[0.2126, 0.7151999999999999, 0.0722]` with a 16-digit precision)  instead of the expected `[0.2126, 0.7152, 0.0722]`. 

Note that the available bit precision for double values are:
```
#include <iostream>
#include <limits>

int main()
{
  std::cout << "double -> " << std::numeric_limits<double>::digits10 << "!\n";
  std::cout << "double -> " << (std::numeric_limits<double>::digits * 3010/10000) << "!\n";
  std::cout << "double -> " << std::numeric_limits<double>::max_digits10 << "!\n";
}
```
which outputs:
```
double -> 15!
double -> 15!
double -> 17!
```
*Note* that using the example from [std::numeric_limits<T>::max_digits10](https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10) with the LUMA red value with a double type, provides the same precision issue.


Being practical we think that the 15-digit serialization provides enough precision for any Op values. Do you see use cases requiring higher precision?

KevinJW:
Patrick,

I guess it depends what the 'true' or source of the values are. If you start with a lower precision text representation like your example, it makes sense to try keep that representation.

If you were instead computing a value internally to OCIO and trying to bake out the result, then I'd be more in favour to preserve the full bit pattern in the serialisation process, so that any image processing produces the same results both before and after.

I think this means there is no single answer that works for everybody. Perhaps with the updated CLF output baking my alternate case is better represented by that method, thus sticking with digits10 for the YAML case is the right way.

Probably worth adding to the documentation pile

Kevin

KevinJW:
This may be pushing the issue to the limit, so this should be seen as very much dealing with an edge case...

What about some kind of flag similar to the other quality/accuracy options that could be part of the config object, so you can say 'my config requires 'X' digits, or maybe passed in to the OCIOYaml::write function (better?), that defaults to the digits10, but could be set by the caller to override?

I'm assuming that for the Processor::write() we will have to use max_digits10 as that is what it's documentation would suggest is required, "Writing (as opposed to Baking) is a lossless process.".

Kevin

hodoulp:
> Probably worth adding to the documentation pile ...

Having an explanation of the default 'double' precision serialization is anyway something important to mention.

> What about some kind of flag similar to the other quality/accuracy options that could be part of the config object, ...

That could be something interesting to propose.

michdolan:
Configurable precision could be a good option. Would you reference that precision in the config file though so that others reading the config could preserve additional precision? I wonder though how often such a feature would be used. You might just end up with the default camp and the max precision camp...

hodoulp:
The two last commits remove the last 'problematic' code, so the optimization disabling (i.e. disabled all optimizations for Clang & OCIOYaml.cpp) is now removed.

The following Yaml code:

Case 1: It stops the loop before the end!
```
const YAML::Node node;
for (unsigned i = 0; i < node.size(); ++i)
{ ... }
```
Case 2: It generates an invalid node (i.e. *it)!
```
const YAML::Node node;
for (YAML::const_iterator it = node.begin();  it != node.end();  ++it)
{ ... }
```
Case 3: it works like that:
```
const YAML::Node node;
for (const auto & v : node)
{ ... }
```

The code `x = node.as<std::string>();` throws an exception (i.e. the node is invalid) and `node.Tag()` (used to build a error message) crashes. The problem is related to [Intel intrinsics](https://software.intel.com/sites/landingpage/IntrinsicsGuide/) AVX (used by Clang in optimization mode) with some 'invalid' YAML::Node instances generated by the faulty loops.
```
#0  __memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:287
#1  0x00007f88f738be48 in std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007f88f737b153 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x000000000066599b in operator<<<char, std::char_traits<char>, std::allocator<char> > ()
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/basic_string.h:6328
#4  load () at /home/devel/git/ocio_adsk_priv/src/OpenColorIO/OCIOYaml.cpp:152
#5  0x000000000065d417 in load () at /home/devel/git/ocio_adsk_priv/src/OpenColorIO/OCIOYaml.cpp:2166
#6  0x000000000065ad01 in open () at /home/devel/git/ocio_adsk_priv/src/OpenColorIO/OCIOYaml.cpp:2479
#7  0x00000000004744a0 in CreateFromStream () at /home/devel/git/ocio_adsk_priv/src/OpenColorIO/Config.cpp:376
#8  ociotest_Config_internal_raw_profile () at /home/devel/git/ocio_adsk_priv/src/OpenColorIO/Config.cpp:1759
#9  0x0000000000af1207 in operator() () at /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/std_function.h:687
#10 main () at /home/devel/git/ocio_adsk_priv/tests/cpu/UnitTest.cpp:71
```

**Note:**
If you are curious, use https://cppinsights.io/ to understand the differences between the three loops and https://godbolt.org/ to see differences between the assembly implementations. 

**Note:** 
These implementation differences are something I've already investigated (in my [blog](https://hodoulp.blogspot.com/2019/09/the-c98-and-c11-for-each-range.html)) following performance hits of the new [range-based for loop](https://en.cppreference.com/w/cpp/language/range-for). It seems that even different 'for-each range' implementations should also be investigated as well as `gcc` vs. `clang` vs. `vc++` implementations of these loops.

PR #884, opened 2019-11-11
Adsk Contrib - Enhance optimization

BernardLefebvre:
This PR is a major update to OpOptimizer.cpp and all of the Op methods it relies on.  This is the code that attempts various adjustments to a vector of ops to try and reduce processing time while delivering the same, or nearly the same, results.

We reviewed how optimization related methods such as isInverse, canCombineWith, combineWith, etc. were implemented and made enhancements to allow for more optimizations and to simplify the code where possible.  For example, the CombineWith method of Range op is now fully implemented and allows two arbitrary Range ops to be combined into one as well as combining a Range and a Lut1D where possible.

A large number of unit tests were added to validate many types of optimization scenarios.

In a previous PR we had added Optimization and Finalization "flags" (bit patterns implemented as an enum) that control the process of making a CPU or GPUProcessor from a Processor.  After a lot of discussion we decided to fold the Finalization flags into the Optimization flags, so now there is just a single enum used for both since the two concepts are so closely related/overlapping.

Each type of optimization is now controllable from this enum and may be turned on/off individually from the public API.  This should allow plenty of control for debugging should any optimization-related issues arise.  For example, the OPTIMIZATION_NO_DYNAMIC_PROPERTIES bit in the enum allows dynamic properties to be treated as not dynamic and ops using them could be optimized.

This work has built on the previous PRs that introduced the CPU/GPUProcessors, added support for integer pixel processing, improved the ImageDesc, and optimized the CPU processing code.  Although we do have some additional ops to add, with this Op optimizer work now complete we feel that the overall way the CPU and GPU renderers work is now substantially in the final form for OCIO v2 and we welcome your feedback.


hodoulp:
As discussed during the last WG meeting, here is a friendly reminder that the branch will be merged next week unless there are some objections.

src/OpenColorIO/OpOptimizers.cpp
```diff
@@ -112,32 +223,30 @@ OCIO_NAMESPACE_ENTER
...
+            if (IsCombineEnabled(type1, oFlags) && op1->canCombineWith(op2))
             {
                 tmpops.clear();
-                first->combineWith(tmpops, second);
+                op1->combineWith(tmpops, op2);
 
                 // tmpops may have any number of ops in it. (0, 1, 2, ...)
-                // (size 0 would occur potentially iff the combination
-                // results in a no-op)
+                // (size 0 would occur potentially iff the combination results in a no-op).
```
michdolan: typo: iff -> if

lgritz: "iff" is a frequent abbreviation (in math and programmer speak) for "if and only if". Is that intentionally what is meant here?  (If it is the intended meaning, then the word "potentially" should be removed, and maybe it's not wise to use that abbreviation.)

PR #888, opened 2019-11-20
Azure Build - Update the 'latest' build to use the latest Yaml.

hodoulp:

Even if I cannot reproduce the problem with a centOS docker, the problem is around the CentOS 'latest' build (i.e. test OCIO with all the latest code from its third-party dependencies).  So, I changed the Yaml settings to use the latest Yaml code.

Unfortunately, I have no way to test the change and validate if it fixes the problem. But the change is still needed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -96,7 +96,6 @@ jobs:
 # ----------------------------------
 # Latest dependency tags (cron only)
 # ----------------------------------
-# TODO: Set yaml-cpp to "latest" when 0.6.x is supported
 
 - job: Latest
   strategy:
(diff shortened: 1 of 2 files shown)


hodoulp:
For the `centOS` image & 'latest build', the problem is that cmake failed because it cannot find a valid Yaml version (found 0.3.0 but expecting 0.6.3). However, it correctly behaves for CentOS image with 'normal' build. Both are executing the same cmake code!

```
CMake Error at /usr/local/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find YamlCpp: Found unsuitable version "0.3.0", but required is
  at least "0.6.3" (found /usr/local/include)
  /usr/local/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:376 (_FPHSA_FAILURE_MESSAGE)
  share/cmake/modules/FindYamlCpp.cmake:113 (find_package_handle_standard_args)
  ext/CMakeLists.txt:15 (find_package)


-- Configuring incomplete, errors occurred!
```

So I created a `centOS` docker image and followed step-by-step the 'latest' build instructions. `cmake` succeeded as well as the compilation.

doug-walker:
Merging 887 and 888 to help Patrick address a build issue.

PR #890, opened 2019-11-25
Add mission statement

michdolan:
There's an ASWF goal for projects to write a mission statement, which is less about explaining what the project *is*, and more about establishing our goals as a project in maintaining and advancing the project. It's a commitment to OCIO users to help guide our work and decision making.

This attempt at writing the OCIO mission statement is intended as a first draft. It's something that should be collaborated on so that we all agree with it and are committed to it, so please suggest edits and voice your opinions.

OpenEXR has gone through this process and is a great example: https://github.com/AcademySoftwareFoundation/openexr/blob/master/README.md


README.md
```diff
@@ -38,6 +38,32 @@ OpenColorIO is free and open source software ([LICENSE](LICENSE)), and
 one of several projects actvively sponsored by the ASWF
 ([Academy Software Foundation](https://www.aswf.io/)).
 
+OpenColorIO Project Mission
+---------------------------
+
+The OpenColorIO project is committed to providing an industry standard solution 
+for highly precise, performant, and consistent color mangement across digital 
+content creation applications and pipelines.
```
michdolan: I feel like this should be following by a clarifying statement on what the "industry" is, and perhaps clarifying the critical role OCIO currently plays in the industry (and thus why these goals are important), but I was struggling with the right wording...

michdolan: maybe I should just replace "industry standard" with "best in class"

README.md
```diff
@@ -38,6 +38,32 @@ OpenColorIO is free and open source software ([LICENSE](LICENSE)), and
...
+
+OpenColorIO aims to:
+
+* be stable, secure, and thouroughly tested on all platforms
```
doug-walker: "all platforms" --> "Linux, Mac, and WIndows"

(Just to be a little more specific since IOS etc. are not currently a priority.)

README.md
```diff
@@ -38,6 +38,32 @@ OpenColorIO is free and open source software ([LICENSE](LICENSE)), and
...
+* be performant on modern CPUs and GPUs
+* be simple, scalable, and well documented
+* be compatible with critical color and imaging standards
+* provide lossless color processing wherever possible
+* maintain config backwards compatability across major versions
+* have every new feature carefully reviewed by industry leaders
```
doug-walker: Should we clarify what we mean by "industry"?  Should we assert a film/tv/games focus?

michdolan: Yes, agreed.

README.md
michdolan: Great point.

README.md
KelSolaar: `mangement` —> management

README.md
michdolan: Keeping as-is for now.

README.md
michdolan: Changed

README.md
michdolan: Changed and fixed.

michdolan:
Below says checks waiting for status, but they all passed (see checks tab). Since this is just an .md file, I'm not going to bother about it. Hopefully GH actions will resolve this sort of issue.

PR #892, opened 2019-11-26
Adsk Contrib - Namespace and code formating

BernardLefebvre:
This PR implements the namespace proposal that was recently discussed by Larry and Patrick on the ASWF email list.  Since that involved touching nearly every file in the repo, we wanted to take the opportunity to also do some cleanup of some formatting/naming issues to be more consistent with the existing OCIO guidelines.  This PR does _not_ make any changes to functionality.

- OCIO Namespace: The old namespace macro hid the "namespace" keyword and this was unfriendly to external tools. There is no change for the API or customization via the build environment. The new approach looks like this:
```
namespace OCIO_NAMESPACE
{
}
```
- Following up on the proposal in PR #810, all CPU tests have been moved to the cpu directory.
- Folder names are made more consistently lower case to be more in line with the repo reorganization Mark did earlier in the year.
- <Name>Ops*.* files which used a mix of singular and plural "Ops" have been renamed to the singular form <Name>Op*.* (except NoOps.* because it actually holds several ops).
- Namespace indentations have been removed to be consistent with the coding guidelines in the documentation.
- Class members ending with “\_” have been renamed to start with “m\_”  to be consistent with the coding guidelines in the documentation.
- A few supporting files have been moved to a more logical location.

Because this PR changes so many files it will cause tricky merge issues with the other development we have underway.  Since this is basically just changes to organization and formatting and not functionality, we would like to merge this within two days rather than waiting our usual two weeks.  We would appreciate your prompt review.


hodoulp:
Here is a friendly reminder that we plan to merge the pull request tomorrow afternoon as agreed at the last TSC meeting.

PR #896, opened 2019-11-29
Adsk Contrib - Add more FixedFunction algorithms

hodoulp:

The pull request adds more predefined algorithms to the FixedFunction op.  The new styles added to the `FixedFunctionTransform` are listed below:
```
FIXED_FUNCTION_RGB_TO_HSV,          //! Classic RGB to HSV function
FIXED_FUNCTION_XYZ_TO_xyY,          //! CIE XYZ to 1931 xy chromaticity coordinates
FIXED_FUNCTION_XYZ_TO_uvY,          //! CIE XYZ to 1976 u'v' chromaticity coordinates
FIXED_FUNCTION_XYZ_TO_LUV           //! CIE XYZ to 1976 CIELUV colour space (D65 white)
```

For the implementation details, you could have a look to the [GPU implementation](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/new_fixed_functions?expand=1#diff-9571918cbfd4c7f8556692ef4793e445R219) or the [CPU implementation](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/new_fixed_functions?expand=1#diff-bf3a7cd1d60d7e308a83b291e80e9266R702).

To validate the implementations several [CPU unit tests](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/new_fixed_functions?expand=1#diff-42315a6dd13142b7196fe7a82ec0f5d2R297) and [GPU unit tests](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/new_fixed_functions?expand=1#diff-a2b26d08dd941771040b73ab3beb0963R247) for the all styles (in both direction) were added.

Please note that the build now detects and fails on missing enumeration values in a switch. Please always flavor the 'enumerations/switch' writing so the compiler could help all of us.


src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -628,6 +699,344 @@ void Renderer_REC2100_Surround::apply(const void * inImg, void * outImg, long nu
...
+
+        float d = X + 15.f * Y + 3.f * Z;
+        d = (d == 0.f) ? 0.f : 1.f / d;
+        const float u = 4.f * X * d;
+        const float v = 9.f * Y * d;
+
+        const float Lstar = ( Y <= 0.008856451679f ) ? 9.0329629629629608f * Y : 
+                                                     1.16f * powf( Y, 0.333333333f ) - 0.16f;
+        const float ustar = 13.f * Lstar * (u - 0.19783001f);   // D65 white
+        const float vstar = 13.f * Lstar * (v - 0.46831999f);   // D65 white
```
KevinJW: D65  looks like it should be a parameter at some point. Also I think this is kind of a very basic colour appearance model, which I could see being extended to include the other CAMs Lab, CIECOM02, etc etc. At which point the parameter would be common to these?

doug-walker: Thanks for the comment Kevin.  We want to keep it simple for now but there should be no problem extending it in the future to take a parameter as long as the default remains D65.

michdolan:
The Windows CI failure looks like a webservice fluke. The GCC builds appear CTL related though. 

Other than the CI failure, this LGTM.

hodoulp:

PR #899, opened 2019-12-02
Adsk Contrib - Add inactive color spaces

hodoulp:

The pull request adds the ‘inactive color space’ feature which allows configuration file creator to keep some color spaces while hiding them from the user. The color spaces can now be active (i.e. visible) or inactive (i.e. not visible but still accessible). A scenario is to keep deprecated color spaces for backward compatibility purpose.

The creator can set one or several color spaces as inactive which hides them from a user request on color spaces. Using the existing API, a user can only access active color spaces but still create a color transformation using inactive color space names. However, the new API offers the runtime customization of the list of inactive color spaces.

The feature implementation allows a complete customization of a configuration file:
+ a new field `inactive_colorspaces` lists the inactive color spaces. When the field is not present or empty, all the color spaces are active i.e. visible from the user.
+ a runtime customization using the env. variable `OCIO_INACTIVE_COLORSPACES` which supersedes the configuration content.
+ a user request (i.e. using [Config::setInactiveColorSpaces()](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/inactive_color_spaces_2?expand=1#diff-6d6a1794fccb2d819c76633c51318336R439)) supersedes the env. variable and the configuration file content.

The legacy API only lists active color spaces. 
```
for (int idx=0; idx<getNumColorSpaces(); ++idx)
{
    const char * csName = getColorSpaceNameByIndex(idx);
    // do something…
}
```
The following code lists all the color spaces from a specific criterion. The API calls request an enumeration value to specify what color spaces to get i.e. active, inactive or all.
```
for (int i = 0; i < c->getNumColorSpaces(COLORSPACE_ALL); ++i)
{
    const char* name = c->getColorSpaceNameByIndex(COLORSPACE_ALL, i);
    // do something…
}
```

As usual several [unit tests](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/inactive_color_spaces_2?expand=1#diff-fdc39e716c100c9095fd34735d9a2f05R2783) validate the feature.


michdolan:
This will be a very welcome feature :)

hodoulp:

Issue #893, opened 2019-11-27
Improving how file paths are mapped to colorspaces


doug-walker:
OCIO currently supports a workflow where users may embed a ColorSpace name string in their file paths and this allows applications to automatically associate imported files with the correct ColorSpace.

However some users do not have the flexibility to alter the paths to the media files in this way.  There is a need for some additional capabilities that would allow simple functionality such as:
-- map all paths that contain the string "texture" to a specific ColorSpace
-- map all files with the extension "exr" to a specific ColorSpace

In addition, it would be very helpful for a config to always specify what ColorSpace to use by default when an application is importing a file and the normal rules are not matched.  Some applications cannot work in strictparsing mode, i.e., they cannot always put up a dialog box to ask users what ColorSpace to use, and so there must be a fallback.  It is better for the config to specify this explicitly rather than an app making a random choice (e.g. the first ColorSpace in the config).

This issue proposes a new feature called "File Rules" to provide these capabilities.

First a quick recap of what is currently in place.  OCIO currently offers the parseColorSpaceFromString method on the Config class that takes a file path string and searches it to find a ColorSpace name from the loaded config.  If a match is found it returns the ColorSpace name.  If no match is found and the strictparsing token in the config is false, it returns the ColorSpace name specified by the default role.  Otherwise it returns an empty string.

We propose adding a new section to the config called "file_rules".  Here is an example:

```
file_rules:
    !<Rule> {name: LogC, extension="*", pattern="*LogC*", colorspace=ARRI LogC}
    !<Rule> {name: OpenEXR, extension="[eE][xX][rR]", pattern="*", colorspace=ACEScg}
    !<Rule> {name: TIFF, regex=".*\.TIF?F$", colorspace=sRGB}
    !<Rule> {name: ColorSpaceNamePathSearch}
    !<Rule> {name: Default, colorspace=Raw}

```
The File Rules are a set of mappings that are evaluated from the top down.  The first rule to match is what determines which ColorSpace is returned.

There are four types of rules available.  Each rule type has a name key that may be used by applications to refer to that rule.  Name values must be unique.  The other keys depend on the rule type.

1)  Basic Rules
This is the basic rule type that uses Unix glob style pattern matching and is thus very easy to use.  It contains the keys:
name: Name of the rule
pattern: Glob pattern to be used for the main part of the name/path.
[...]

PR #904, opened 2019-12-20
Adsk Contrib - Adding file rules to config.

BernardLefebvre:
This PR implements issue #893, there is more detail there as well, but in summary:

The file rules are a set of filepath to color space mappings that are evaluated from first to last. The first rule to match is what determines which color space is returned. There are four types of rules available. Each rule type has a name key that may be used by applications to refer to that rule. Name values must be unique. The other keys depend on the rule type:

- Basic Rule: This is the basic rule type that uses Unix glob style pattern matching and is thus very easy to use. It contains the keys:
  * name: Name of the rule
  * colorspace: Color space name to be returned.
  * pattern: Glob pattern to be used for the main part of the name/path.
  * extension: Glob pattern to be used for the file extension. Note that if glob tokens are not used, the extension will be used in a non-case-sensitive way by default.

- Regex Rule: This is similar to the basic rule but allows additional capabilities for power-users. It contains the keys:
  * name: Name of the rule
  * colorspace: Color space name to be returned.
  * regex: Regular expression to be evaluated.

- OCIO v1 style Rule: This rule allows the use of the OCIO v1 style, where the string is searched for color space names from the config. This rule may occur 0 or 1 times in the list. The position in the list prioritizes it with respect to the other rules. It has the key:
  * name: Must be "ColorSpaceNamePathSearch".

- Default Rule: The file_rules must always end with this rule. If no prior rules match, this rule specifies the color space applications will use. It has the keys:
  * name: must be "Default".
  * colorspace : Color space name to be returned.


include/OpenColorIO/OpenColorIO.h
```diff
@@ -634,6 +652,148 @@ class OCIOEXPORT Config
...
+    //!cpp:function::
+    FileRulesRcPtr createEditableCopy() const;
+
+    //!cpp:function:: Does include default rule. Result will be at least 1.
+    size_t getNumEntries() const noexcept;
+
+    //!cpp:function:: Get name of the rule.
+    const char * getName(size_t ruleIndex) const;
+
+    //!cpp:function:: Setting pattern will erase regex.
```
michdolan: In my gut I feel like a `setRuleType` method (which would serialize to the config in a type attr) would be cleaner than relying on mutually exclusive attributes (clearing regex when glob is used and vice-versa). It would also prevent users from needing to manually enter either of the reserved rule type names, which could be set automatically when either of those types are used. You could then only have a `pattern` and `extension` attr (no `regex`, since it would be implied by the type and use the `pattern` slot). Just my two cents...

include/OpenColorIO/OpenColorIO.h
```diff
@@ -634,6 +652,148 @@ class OCIOEXPORT Config
...
+//   * colorspace: Color space name to be returned.
+//   * regex: Regular expression to be evaluated.
+// 
+// - OCIO v1 style Rule: This rule allows the use of the OCIO v1 style, where the string
+//   is searched for color space names from the config. This rule may occur 0 or 1 times
+//   in the list. The position in the list prioritizes it with respect to the other rules.
+//   It has the key:
+//   * name: Must be "ColorSpaceNamePathSearch".
+// 
+// - Default Rule: The file_rules must always end with this rule. If no prior rules match,
```
michdolan: We should mention `strict_parsing` here, noting that this rule will return an empty string when it is enabled, assuming this is still true.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -634,6 +652,148 @@ class OCIOEXPORT Config
...
+    const char * getRegex(size_t ruleIndex) const;
+    //!cpp:function::
+    void setRegex(size_t ruleIndex, const char * regex);
+
+    //!cpp:function:: Set the rule's color space (may also be a role).
+    const char * getColorSpace(size_t ruleIndex) const;
+    //!cpp:function::
+    void setColorSpace(size_t ruleIndex, const char * colorSpace);
+
+    //!cpp:function:: Get number of key/value pairs.
```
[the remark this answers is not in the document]
BernardLefebvre: ok

include/OpenColorIO/OpenColorIO.h
BernardLefebvre: strict parsing is actually not used here. If no color space is found in the path the rule will not be triggered.

include/OpenColorIO/OpenColorIO.h
BernardLefebvre: Pattern/extension and regex do not use the same syntax/language, it would be strange to share the string, so changing the type would require the string to be cleared.
Maybe we could expose the reserved names in the API.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -634,6 +652,148 @@ class OCIOEXPORT Config
...
+    //!cpp:function:: Adds a key/value or replace value if key exists. Setting a NULL or an
+    // empty value will erase the key.
+    void setCustomKey(size_t ruleIndex, const char * key, const char * value);
+
+    //!cpp:function:: Insert a rule at a given ruleIndex. Rule currently at ruleIndex
+    // will be pushed to index: ruleIndex + 1.
+    // Name must be unique.
+    // - "Default" is a reserved name for the default rule. The default rule is automatically
+    //   added and can't be removed.
+    // - "ColorSpaceNamePathSearch" is reserve for a rule that will use
```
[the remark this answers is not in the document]
BernardLefebvre: ok

include/OpenColorIO/OpenColorIO.h
michdolan: Are we no longer honoring the previous functionality of parsing for a path returning a empty string when strict parsing is on? I thought we were trying to preserve that for those that want it.

include/OpenColorIO/OpenColorIO.h
michdolan: Though if the type is mutually exclusive, the pattern would be clearly one or the other. I'm not stuck on that idea though. I think the thing that is a bit confusing with this class is that it can be many pre-defined things, each of which is fully determined by a subset of the attributes. Sometimes name drives behavior, other times regex, and other times pattern and extension. If we add more rules later it could balloon more. Willing to be wrong here of course, but maybe should be discussed more.

michdolan: Also, exposing reserved names is a great idea.

include/OpenColorIO/OpenColorIO.h
BernardLefebvre: If you use Config::parseColorSpaceFromString nothing has changed, strictParsing will continue to work as before.
If you use FileRules ( FileRules::getColorSpaceFromFilepath ) then strictParsing has no effect.

michdolan: Great to know, thanks!

include/OpenColorIO/OpenColorIO.h
BernardLefebvre: Instead a function was added so that user does not have to specify the name of the rule.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -634,6 +652,153 @@ class OCIOEXPORT Config
...
+    // - "ColorSpaceNamePathSearch" is also a reserved name (see insertPathSearchRule below).
+    // Will throw if ruleIndex is not less than :cpp:func:`FileRules::getNumEntries`.
+    void insertRule(size_t ruleIndex, const char * name, const char * colorSpace,
+                    const char * pattern, const char * extension);
+    //!cpp:function::
+    void insertRule(size_t ruleIndex, const char * name, const char * colorSpace,
+                    const char * regex);
+    //!cpp:function:: Inserts a rule that uses cpp:func: 'Config:parseColorSpaceFromString'
+    // to search the path for any of the color spaces named in the config (as per OCIO v1).
+    void insertPathSearchRule(size_t ruleIndex);
```
michdolan: Since the "Default" rule is automatically added, can we also add a convenience function like:

`setDefaultRuleColorSpace(const char * colorSpace)`

That would be clearer than having to call this to set the default rule colorspace, since index and name are pre-defined (presumably index means nothing in this case, since it will always fall last):

`insertRule(3, "Default", "my_colorspace");`

include/OpenColorIO/OpenColorIO.h
```diff
@@ -634,6 +652,153 @@ class OCIOEXPORT Config
...
+//   in the list. The position in the list prioritizes it with respect to the other rules.
+//   StrictParsing is not used. If no color space is found in the path, the rule will not
+//   match and the next rule will be considered.
+//   It has the key:
+//   * name: Must be "ColorSpaceNamePathSearch".
+// 
+// - Default Rule: The file_rules must always end with this rule. If no prior rules match,
+//   this rule specifies the color space applications will use. It has the keys:
+//   * name: must be "Default".
+//   * colorspace : Color space name to be returned.
```
michdolan: Based on comment below, we could mention the `setDefaultRuleColorSpace` function here too.

include/OpenColorIO/OpenColorIO.h
BernardLefebvre: Good idea!

include/OpenColorIO/OpenColorIO.h
michdolan: Thanks. Works well!

include/OpenColorIO/OpenColorIO.h
michdolan: Looks good. Thanks Bernard!

hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can now be merged unless there are some objections.

doug-walker:
Thanks for the many helpful comments Michael, it definitely helped us improve the API.

PR #905, opened 2019-12-20
Adsk Contrib - Refactor Transforms that are using OpData

BernardLefebvre:
Refactor transforms that are using OpData internally in the same say RangeTransform was refactored with PR #812.
When OpData is encoding some direction either using a style or a direction member, use that instead of adding another direction to the transform.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can now be merged unless there are some objections.

PR #909, opened 2020-01-09
Adsk Contrib - Add some string utilities

hodoulp:
The OpenColorIO library is currently using the useful `pystring` library for lot of string manipulations. However, its use is problematic for the OpenColorIO utility static libraries such as `oiiohelpers` because it introduces an indirect dependency.

As OpenColorIO needs are limited, a string utility library (header only) now offers some basic string manipulation capabilities and it can be use anywhere in the project.

_Note 1:_   `pystring` also offers path manipulation so the OpenColorIO library is still using it.

_Note 2:_   A following pull request will also use the new string manipulation library.


lgritz:
I would recommend just folding into ocio itself. I think ext only makes sense for things that are imported more or less unmodified -- a pure dependency, but one we have some need to embed. If we're taking 5 or 6 functions (which frankly are very nearly generic string work anyway), that doesn't seem necessary.

hodoulp:
> Yes, if all OCIO was using from pystring was this small number of simple inline functions, I totally agree that we should remove the dependency.
OCIO library still needs some path helpers so I did not changed the library itself.

hodoulp:
> Is it being considered ext because it was derived from pystring?

No. After discussions we though it's a convenient location even if it twists a little bit the original meaning of the layer (i.e. the directory name). It's an external dependency to any OCIO layers but not one 'coming' from an external third-party library. So, I added some explanations [here]( https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/909/files#diff-9e4edc0a8604dcf9fd1a3806ca97a57dR35).

> Would it make sense to just add it directly to the OCIO source instead?

I do not think.

The point is that these utilities are used by the library itself, the 'utils', the unit tests and finally the apps. The last three are built on top of the library itself. And the utilities are intensively using STL (to benefit from latest C++11 features). It imposes to localize the utilities somewhere outside the OCIO library (to avoid the public API barrier i.e. constraints).

_Note_: Having all these utilities in a header file is only the temporary benefit, later utilities could request an implementation file to avoid introducing too many implementation details and/or explicit dependencies.

_Note_: Using STL in a public API should be forbidden (as much as possible) because it could raise serious issues when building a dynamic library (when the implementation is not in the header file) i.e. memory management, inheritance, etc. For example, it’s not recommended to allocate memory inside a dynamic library (std::vector, std::map, etc), and deallocate the memory outside the dynamic library (i.e. by the destructor, resize(), etc.). Another example, it's usually not recommended to inherit from any STL classes (i.e. containers, etc.) because all these classes were not designed to be inheritable. The most famous known issue is around the lake of virtual destructors for some STL classes (i.e. could cause slicing issues). BTW, the current OCIO::Exception highlights the topic. 

The topic is quite old one and the latest C++ versions could have or not change some of the discussion aspect. Any hints?

**In conclusion**, by moving these utilities outside the library, we enforce these rules for short-term and long-term without any excessive short-term cost.

An adjustment could be to change the directory name to better highlight its new purpose and have an explanation for each included sub-directory. Any suggestions?

My suggestion is to keep the directory name and add more explanation of the layer purpose if needed.

michdolan:
Thanks for the explanation @hodoulp . I see your point. Another option could be src/utils. Those are intended to be shared across the library and apps (though right now only the apps use them), but are clearly internal code. Pystring was there at one point IIRC. Since removing our dependencies from ext, it seems like there is not much there, but during build ext still houses all the dependent external packages. That would be my only apprehension. We kept sampleicc there because it was a modified external dependency.

hodoulp:
> Another option could be `src/utils`. Those are intended to be shared across the library and apps (though right now only the apps use them), ...

That's also an alternative we discussed internally. But current utils contain only ones built on-top of the library and adding the string helpers could introduce some confusion about the layer purpose. 

>but during build ext still houses all the dependent external packages.

Good point.

The directories represent some design layers and should have a single purpose which is clear & easy to understand. Below should be the end goal for the layers:

|----------------|
|----- APPS ----|   <-- ociodisplay, ocioconvert, etc.
|----------------|
|----- UTILS ---|   <-- oglbuilder, oiiohelpers. The code can use 'STR' & 'LIB' but not 'APPS'.
|----------------|
|----- LIB ------|   <-- 'OpenColorIO' i.e. the library itself. The code can use 'STR" but not 'UTILS' or 'APPS'.
|----------------|
|----- STR -----|   <-- The new string utils i.e. any layers can use these utilities.
|----------------|

Having 'STR' and 'UTILS' in the same directory introduces some confusion around the layer purpose.

If 'utils' is renamed 'libutils' (or 'ocioutils') then it's less confusing to reuse 'utils' layer name for all the low-level utilities such as the string utilities.

michdolan:
Good point about the layers (good for a design doc somewhere)

PR #918, opened 2020-01-31
Adsk Contrib - Add support for CLF v3

BernardLefebvre:
Add support for Academy/ASC Common LUT Format v3.  This spec is part of the new ACES 1.2 release.

The v3 spec adds two new Process Nodes: Exponent and Log.  These allow many transforms (e.g. for display devices and many digital cameras) to be represented without needing 1d-LUTs.

The new Process Nodes are similar to what OCIO had but there are also some enhancements.  So we extended the OCIO transforms/ops to be able to represent the new CLF features.

Also v3 attempts to simplify the spec in some ways.  For example, certain edge cases involving the Range node are now disallowed.

Here are the main changes:
– The new Exponent Process Node has multiple options for handling negative values. (These were based on the feedback of the OCIO v2 working group at SIGGRAPH 2019.) These get exposed using the OCIO Exponent and ExponentWithLinear Transforms.  OCIO already supported the CLF spec's basicFwd, basicRev, moncurveFwd, moncurveRev and we added support for the Mirror and Passthru styles. 

– The new Log Process Node has a lot of styles and these are handled in the OCIO LogOp. We already supported the CLF spec's log10, antiLog10, log2, antiLog2, linToLog, and logToLin, and added support for cameraLinToLog and cameraLogToLin. The two new styles are exposed as a new LogCameraTransform in OCIO.

– Range no longer allows the case where only the min or max are provided but they are unequal.  Also, the case where no min/max values are provided is not allowed.

The new ops will also be allowed in the CTF format too (v2 and higher) so that it continues to be able to serialize any OCIO Processor.

We added unit tests to verify proper handling of the compCLFversion attribute for v2 and v3 files, along with some other unit tests.  (We have another PR about unit tests upcoming.)

The new CLF Process Nodes are similar to what was in earlier versions of the CTF format but the syntax is a bit different.  We allow reading either the old or new syntax where it makes sense:
– Matrix Array dim now has 2 integers rather than 3
– Gamma and GammaParams elements become Exponent and ExponentParams
– The "moncurve" is now camelCase, e.g. "moncurveFwd" style is now "monCurveFwd"


hodoulp:

include/OpenColorIO/OpenColorTransforms.h
```diff
@@ -867,6 +887,81 @@ class OCIOEXPORT LogAffineTransform : public Transform
...
+    //!cpp:function::
+    virtual double getBase() const noexcept = 0;
+    //!cpp:function::
+    virtual void setBase(double base) noexcept = 0;
+
+    //!rst:: **Get/Set values for the R, G, B components**
+    //
+
+    //!cpp:function::
+    virtual void getLogSideSlopeValue(double(&values)[3]) const noexcept = 0;
```
michdolan: Just a general comment, not specific to this PR. I'm noticing we have some inconsistency in Transform method naming in the API. Some set/set methods use the `Value` suffix, and some don't. My opinion is that Value is unnecessary in these method names. So:

`getLogSideSlopeValue` could be `getLogSideSlope`

Again, not suggesting this be changed now necessarily, but would be good to do a consistency pass on the API before v2 releases.

PR #919, opened 2020-01-31
Adsk Contrib - Create optimized processor

BernardLefebvre:
Add a function to create an optimized version of a Processor.
This can be used for debug purposes to inspect how a Processor is optimized (for CPU or GPU). Processor could be optimized and written (or converted to transform).


hodoulp:

PR #922, opened 2020-02-03
Adsk Contrib - Processor between two configs

BernardLefebvre:
Adds a function to create a processor that can be used to go from one color space in a config to another color space in a different config.

Both config have to define a role named "aces_interchange"

This can be used to bring assets of a project using a config, to another project using a different config. For details, please refer to issue #927.


hodoulp:
@michdolan @KelSolaar Following the agreement around a new method with user-defined interchange roles (or color spaces) for source and destination, @BernardLefebvre pushed a commit for review.

PR #923, opened 2020-02-03
Adsk Contrib - Improve the GPU public API

hodoulp:

The pull request improves the support of the two main use cases when integrating the GPU color transformation in a DCC.
The first one (i.e. currently implemented in `master` and used by `ociodisplay`) is to collect all the information and to call directly OpenGL. The app `ociodisplay` uses the `GpuShaderDesc` helper class to collect the information and the `oglbuilder` library to build the GPU shader program.
The second use case is when a DCC already manages the GPU (through a library abstracting the hardware for example). The latter case only needs to receive requests as a library already implements `GpuShaderDesc` & `oglbuilder`. So, the DCC would implement a custom class inheriting from the new OCIO class `GpuShaderCreator` to use its own library (i.e. no need of `GpuShaderDesc` & `oglbuilder`).

While modifying code around the GPU generation some cleanups were done:
- Improve the 'cacheId' generation,
- Improve the comments in the shader program,
- Refactor some GPU Op code to be consistent with other Ops.

### DIFF
--- a/include/OpenColorIO/OpenColorIO.h
+++ b/include/OpenColorIO/OpenColorIO.h
@@ -1368,12 +1368,10 @@ public:
     //!cpp:function::
     bool hasChannelCrosstalk() const;
 
-    //!cpp:function:: The returned pointer may be used to set the value of any
-    //                dynamic properties of the requested type.  Throws if the
-    //                requested property is not found.  Note that if the
-    //                processor contains several ops that support the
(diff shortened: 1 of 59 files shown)


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can be merged next Monday unless there are some objections.

Issue #929, opened 2020-02-07
Python binding documentation is broken on Windows


hodoulp:
When investigating the defect #774, I found that the generation of the Python binding documentation is broken on Windows.

The developer C++ & Python binding API documentations are auto-generated using `sphinx` and `rst`. The C++ part is working well but the Python one is broken. Both are correctly working in `RB-1.1`.

That's a CMake configuration issue on the Windows platform.

PR #931, opened 2020-02-12
Adsk Contrib - Fix the generation of the Python documentation

hodoulp:

The pull request fixes the documentation generation of the Python bindings where the path (to the Python bindings) was wrong. Depending of the CMake generator used, the location is different (i.e. the adjustment was missing).

Being in the documentation problem, I also fixed all the existing warnings of the public API documentation.


hodoulp:
Refer to #929

PR #934, opened 2020-02-18
Feature/unittests

meimchu:
- /tests/python/ColorSpaceTests: Rewrote and reorganized.
- /tests/python/OpenColorIOTestSuite: Automate detecting unit test cases from python files within test folder.
- BakerTest: Replaced Baker.py with BakerTest.py for naming consistency.


lgritz:
@meimchu I think for the DCO, you just need to make sure there is a "Signed-off-by: " on your comments. Follow these instructions: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/934/checks?check_run_id=451928481

When you make the commits, just get in the habit of adding `-s` to your commit commands and it will add your userid/email as the sign-off.

lgritz:
@meimchu As for the CLA... I checked and you are listed on our CLA whitelist. But you may need to click the instructions above where it says "please click here to be authorized."

tests/python/ColorSpaceTest.py
```diff
@@ -1,31 +1,221 @@
...
+        for is_data in is_datas:
+            self.cs.setIsData(is_data)
+            self.assertEqual(is_data, self.cs.isData())
+
+        # Wrong type tests
+        wrong_is_datas = [['test'],
+                          'test']
+        for wrong_is_data in wrong_is_datas:
+            with self.assertRaises(TypeError):
+                self.cs.setIsData(wrong_is_data)
```
michdolan: I'm assuming this is failing in CI because most types can evaluate to bool, which this method is expecting.

tests/python/ColorSpaceTest.py
```diff
@@ -1,31 +1,221 @@
...
+
+        # Random string tests
+        for i in range(0, 10):
+            name = generate_name(10)
+            self.cs.setAllocation(name)
+            self.assertEqual('unknown', self.cs.getAllocation())
+
+        # Wrong type tests
+        with self.assertRaises(TypeError):
+            self.cs.setAllocation(None)
```
michdolan: This raises ValueError

michdolan:
Thanks for your work on this @meimchu !

Before you get too far I would recommend running your tests against my ocio_pybind branch instead of master at this point:
https://github.com/michdolan/OpenColorIO/tree/ocio_pybind

The bindings and their behavior will change in various way with the new bindings, including potentially how they handle types. I'll hopefully get the new bindings in a PR in the next few weeks, after which your tests could be submitted. You can make test changes before then of course, but I just wouldn't want you to have to redo a lot of work.

michdolan:
Structurally this is a great direction I think - breaking out the tests more granualrly. Nice work.

tests/python/ColorSpaceTest.py
```diff
@@ -1,31 +1,221 @@
...
+        """
+        Test the isEditable() and createEditableCopy() methods.
+        """
+
+        # Basic test
+        self.assertTrue(self.cs.isEditable())
+
+        cs_copy = self.cs.createEditableCopy()
+        self.assertTrue(isinstance(cs_copy, OCIO.ColorSpace))
+        self.assertTrue(cs_copy, self.cs)
```
michdolan: Should this be `assertEqual`?

tests/python/ColorSpaceTest.py
```diff
@@ -1,31 +1,221 @@
...
+            self.cs.setDescription(name)
+            self.assertEqual(name, self.cs.getDescription())
+
+    def test_bitdepth(self):
+        """
+        Test the setBitDepth() and getBitDepth() methods.
+        """
+
+        # Known constants tests
+        bit_depths = [OCIO.Constants.BIT_DEPTH_UNKNOWN,
```
michdolan: In the new pybind bindings you *should* be able to iterate on the enum instead of listing all the possible values here, as an example of somrhting that will change/improve.

PR #935, opened 2020-02-18
Adsk Contrib - Cleanup platform.[h|cpp] files

hodoulp:

The pull request only cleanups the `Platform.h` & `Platyform.cpp` to remove any legacy code. And it changes the code to flavor the 'std' methods instead of the C counterparts. The two method sets have identical behaviors except few edge cases like error, exceptions, etc.

There is an improvement around the temporary file where some Linux settings (i.e. OS specific) could change the removal policy. So, I've updated the unit tests to impose the cleanup.


hodoulp:

doug-walker:
This is blocking us and it's been two weeks, so I'm going ahead and merging.

PR #937, opened 2020-02-25
Adsk Contrib - Centralize string manipulation helpers

hodoulp:

The work continues the refactoring started by #909 to replace `pystring` string helpers by the new header-only `StringUtils`. These methods greatly benefit from C++11 new features and being header-only, they can be used by any layer (except in the public API).

Note: `pystring` should now only be used for the file-path manipulations.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can be merged EOF unless there are some objections.

PR #938, opened 2020-02-26
Adsk Contrib - Adjust FileRules API

BernardLefebvre:
FileRules was exposing some functions that were using a reference to a Config. This was not aligned with other API functions. These have been removed from the API.
Functions are now part of Config.

### DIFF
--- a/include/OpenColorIO/OpenColorIO.h
+++ b/include/OpenColorIO/OpenColorIO.h
@@ -593,6 +593,20 @@ public:
     //    The argument is cloned.
     void setFileRules(ConstFileRulesRcPtr fileRules);
 
+    // !cpp:function:: Get the color space of the first rule that matched filePath.
+    const char * getColorSpaceFromFilepath(const char * filePath) const;
+
+    //!cpp:function:: Most applications will use the preceding method, but this method may be
+    // used for applications that want to know which was the highest priority rule to match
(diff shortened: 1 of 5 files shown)


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can be merged next week unless there are some objections.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -593,6 +593,20 @@ class OCIOEXPORT Config
     //    The argument is cloned.
     void setFileRules(ConstFileRulesRcPtr fileRules);
 
+    // !cpp:function:: Get the color space of the first rule that matched filePath.
+    const char * getColorSpaceFromFilepath(const char * filePath) const;
+
+    //!cpp:function:: Most applications will use the preceding method, but this method may be
+    // used for applications that want to know which was the highest priority rule to match
+    // filePath.  The getNumCustomKeys(ruleIndex) and custom keys methods may then be used
```
michdolan: Since this method was moved, we might want to clarify where `getNumCustomKeys(ruleIndex)` is defined. Maybe just change to `FileRules::getNumCustomKeys(ruleIndex)`?

BernardLefebvre: good idea

Issue #926, opened 2020-02-04
Adding display-referred color spaces


doug-walker:
This proposal is to add a fundamental new capability to OCIO.  It goes beyond a simple new feature and reflects an enhancement to the underlying architecture.  As such this is a lengthy proposal which begins with some background information to define the problem and then proceeds with a specific proposal.  

The issue is divided into sections, so you may skip down to the config and API proposal sections if you are so inclined.  

Please note that the proposal presented below would be an optional feature and config authors would not be required to use it or to otherwise modify their configs if they don't find it useful.

**BACKGROUND**

ISO standard 22028-1 introduced the notion of "image state" and scene-referred and output-referred (or, more specifically, display-referred) color encodings.  That document forms the foundation for many other color-management-related standards that are essential to the VFX industry including those for ACES (i.e., the SMPTE ST 2065 series of standards) and HDR video (i.e., ITU-R BT.2100, BT.2390, etc.).

One of the core ideas of these standards is that there is a special type of transform needed to convert from a scene-referred encoding to a display-referred encoding.  In ISO 22028-1 and ACES, this is called a "rendering transform" (which unfortunately overlaps with the use of "rendering" elsewhere in VFX).  In the ITU standards it is called the "opto-optical transfer function" or OOTF.

Many VFX artists are missing this key conceptual piece of information and incorrectly assume that there is no fundamental difference between scene-referred and display-referred encodings.  Much of the information on the web about the so called "linear workflow" for CG is wrong due to this misunderstanding.

OCIO was quite forward-thinking in terms of introducing the notion of a "view" that is a separate thing from a "display".  Hopefully this is gradually helping artists realize that the proper conversion from a scene-referred color space to a display is generally not simply a gamma correction (or inverse EOTF, to use the ITU term of art).  

However, currently in OCIO the view/display terminology is basically just a UI-related naming feature and the underlying architecture does not reflect these concepts.  For example, there is no notion of a display ColorSpace.  There is also no notion of a "view transform" that connects a scene-referred and display-referred color space.  The display/view section of the config is simply a way of stating that certain of the color spaces are actually transforms to a display space that incorporate a specific view.  For example, in the spi-vfx config there is a "Film" view for a DCIP3 and an sRGB display.  These point to transforms that convert from the scene-referred reference space all the way to a display (p3dci8 and srgb8).  This is inconvenient for several reasons.

[...]

PR #939, opened 2020-02-27
Adsk Contrib - Add the Display reference space

BernardLefebvre:
Introduce display reference space:
- Add ability to use display color spaces, view transforms in configs.
- Update ColorSpaceTransform, DisplayTransform to handle display reference space.
- Color space can either be scene-referred or display referred.

See issue #926 for more details.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can be merged next week unless there are some objections.

Issue #620, opened 2018-10-29
CDL - Add a flag to clamp (or not) the values


hodoulp:

doug-walker:
This was implemented by PR #943.

PR #943, opened 2020-03-03
Adsk Contrib - Complete the CDL style support

BernardLefebvre:
CDL style was introduced earlier but there was some missing pieces:
* Add missing support for CDL style in config files.
* Change the default to no-clamp so that the default is aligned with v1 (i.e. to be backward compatible as requested during the last Working Group meeting).
* Add tests.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can be merged this week unless there are some comments or objections.

doug-walker:
This implemented issue #620.

PR #944, opened 2020-03-03
Adsk Contrib - Add arguments to unit test binaries

hodoulp:

The OpenColorIO project now has lot of unit tests (`cpu` has about 800 unit tests) making longer to run them all. On the Windows platform, the debug mode makes it quite long to run them all (because of the Microsoft std::vector implementation in debug mode i.e. not related to OpenColorIO) and debugging only one unit test on any platform becomes painful.

The unit tests binaries (i.e. the `cpu` and `utils` ones) now have some command line arguments to help:
1. Stop on the first error `--stop_on_error`
2. Run only a subset of the unit tests `--run_only [Group]/[Name]` Note: strings are case insensitive.

Here are some command line examples:
- `.\tests\cpu\test_cpu_exec.exe --run_only filerules` executes all the unit tests from the group `filerules`.
- `.\tests\cpu\test_cpu_exec.exe --run_only \clone` executes all the unit tests named `clone` from any group.
- `.\tests\cpu\test_cpu_exec.exe --run_only filerules\clone` only executes the unit test `clone` from the group `filerules`.

Note: The implementation consists to reuse the argument helper classes which were originally only used by the apps. The files were 'copy & paste' as-is as the license disclaimer is not the OpenColorIO one.


hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can be merged this week unless there are some comments or objections.

Issue #572, opened 2018-09-14
ociocheck doesn't fully validate views/looks?


KevinJW:
Adding this whilst I remember...

I have not looked into the cause but I sometimes run into errors when using configs that use looks on views, where the file name in a file transform used in the look has a typo. I guess that ociocheck doesn't actually validate the looks/views?

It would be nice if ociocheck could validate that the view can be used (or not), in this case that the FileTransform does refer to a file that actually exists (assuming I'm not using context variables obviously).

Kevin

YashAgrawal0:
@KevinJW , @hodoulp  

I would like to work on this issue. can you help me how do I get started?

KevinJW:
So as an example, I think when we run ociocheck against this file, it should complain that the look refers to a file transform that does not exist.


[ocio_check_look_config.ocio.txt](https://github.com/AcademySoftwareFoundation/OpenColorIO/files/4274975/ocio_check_look_config.ocio.txt)

YashAgrawal0:
@KevinJW Can you eleaborate a bit more as I am new to OCIO, I've installed OCIO as per the instructions given in 
`https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/docs/installation.rst`.

Now what do I need to do to tackle this issue?

KevinJW:
so if you have a working build then you should have a tool called ociocheck, the first step would be to run the file I attached through ociocheck. There are a number of ways to do this but the simplest would be to pass the file in using the --iconfig <file> option.

The goal is to make this actually fail for the right reason, which is that there is no file called "file_that_does_not_exist.spimtx".

You will need to add some code roughly here - https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/src/apps/ociocheck/main.cpp#L211

You could setup a processor to convert apply each look and see if it errors. Obviously ideally we'd like a test that would detect this failure that ran automatically during the execution of the test suite, but that would be further down the line.

Next you might want to look at the definitions of the displays and views and also figure out how to iterate across them and try instance each of those as well.

There is code showing how to create DisplayTransform objects in the ociodisplay tool directory, which might be useful.

Kevin

YashAgrawal0:
@KevinJW  Thanks for the update.


Currently I am getting the output for  ` ociocheck --iconfig ocio_check_look_config.ocio.txt `
[...]

PR #945, opened 2020-03-06
ociocheck now validate views/looks

YashAgrawal0:
Fixes : #572 issue. 

Ociocheck is now able to validate looks and check for the existence of the file provided in the FIleTransform.

Now if the ociocheck is run against a file which does not exist in the same directory then it will throw an error : 

```
ERROR : 
The specified file reference  'file_that_does_not_exist.spimtx' could not be located. The following attempts were made: /home/yash/Desktop/file_that_does_not_exist.spimtx

```

 As shown in this [image](https://user-images.githubusercontent.com/24262584/76064715-3b2e6300-5fb0-11ea-8d81-47e4f354331c.png).

### DIFF
--- a/docs/CompatibleSoftware.rst
+++ b/docs/CompatibleSoftware.rst
@@ -259,8 +259,8 @@ Natron
 'Open Source Compositing Software <http://natron.fr>'
 
 
-Nuke
-****
+Nuke (Foundry)
+**************
 
 `Compositor - The Foundry <http://www.thefoundry.co.uk/products/nuke>`__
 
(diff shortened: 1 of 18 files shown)


doug-walker:
Hi @YashAgrawal0, the improvement to ociocheck looks good, thanks!

However, it would be nice to put the changes to the compatible software list into a separate PR.

lgritz:
This PR contains changes related to validating views, but also contains all the documentation fixes you are working on.

Please use a *different* topic branch for every separate PR you are working on, otherwise it's very confusing to review the code because we don't know which changes belong to which PRs.

PR #947, opened 2020-03-07
ociocheck will now prompt for missing file

YashAgrawal0:
Fixed : #572 . 

Signed-off-by: YashAgrawal0 <yashagrawal048@gmail.com>


src/apps/ociocheck/main.cpp
```diff
@@ -209,6 +209,21 @@ int main(int argc, const char **argv)
...
+
+                OCIO::ConstTransformRcPtr transform = look->getTransform();         
+
+                try
+                {
+                    OCIO::ConstProcessorRcPtr process = config->getProcessor(transform);
+                }
+                catch(OCIO::Exception & exception)
+                {
+                    std::cout << "\nERROR : " << std::endl;
```
michdolan: I tried this out on a config and it works well, but the formatting is inconsistent with other errors. This is what the output looks like:

```
** Looks **
di

ERROR :
The specified file reference  'look_di.cc' could not be located.
```

Similar ColorSpace errors are formatted like (see lines 186-188):

```
** Looks **
di -- error
    The specified file reference  'look_di.cc' could not be located.
```

General errors, on the other hand, are _mostly_ formatted like (see lines: 267, 270):

```
** Looks **
di
ERROR: The specified file reference  'look_di.cc' could not be located.
```

I'm not overly concerned with which error format is used for looks, but would suggest making it consistent with one of the existing styles for readability.

hodoulp: The error messages are clearly inconsistent across the file but also across apps.

The formatting could be:
`std::cerr << "ERROR: " << "Some explanation" << std::endl;`

src/apps/ociocheck/main.cpp
```diff
@@ -209,6 +209,21 @@ int main(int argc, const char **argv)
...
             {
                 std::cout << config->getLookNameByIndex(i) << std::endl;
+
+                OCIO::ConstLookRcPtr look = config->getLook(config->getLookNameByIndex(i)); 
+
+                OCIO::ConstTransformRcPtr transform = look->getTransform();         
+
+                try
+                {
+                    OCIO::ConstProcessorRcPtr process = config->getProcessor(transform);
```
hodoulp: The `Look` could have forward or inverse transform so the test must validate the right one. If `transform` is null then `look->getInverseTransform()` must be used.

hodoulp:
@YashAgrawal0 Your branch is updated to the latest commit. Before changing anything in your local branch you have to update it (i.e. `git update`).

src/apps/ociocheck/main.cpp
YashAgrawal0: @hodoulp 

when I am using 
'''
       OCIO::ConstTransformRcPtr  invTransform = look->getInverseTransform();         
'''

it results in Segmentation fault (core dumped). 

I think its because inverse transform is not defined in the config and we are trying to access it causing segmentation fault to occur.

hodoulp: On a valid instance, `ConstTransformRcPtr Look::getInverseTransform()` could return an empty [std::smart_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr) but it should not crash. However, if the code later deferences it than it crashes.

You should use a debugger to see where the problem is coming from.

YashAgrawal0: @hodoulp 

Can you tell me how to use debugger in ocio?

hodoulp: @YashAgrawal0 

Assuming that you are working on Linux:
1. Compile in 'debug' mode i.e. look at [CMakeLists.txt](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/CMakeLists.txt), and search for `CMAKE_BUILD_TYPE`
2. Use your favorite debugger, to start ociocheck.exe i.e. gdb for example on Linux platform.
>gdb .\src\apps\ociocheck\ociocheck

Lot of documentations are available on the web, but you can start with these gdb commands: [r|run], [n|next], [p|print].

src/apps/ociocheck/main.cpp
```diff
@@ -209,6 +209,40 @@ int main(int argc, const char **argv)
             for(int i=0; i<config->getNumLooks(); ++i)
             {
                 std::cout << config->getLookNameByIndex(i) << std::endl;
+
+                OCIO::ConstLookRcPtr look = config->getLook(config->getLookNameByIndex(i)); 
+
+                OCIO::ConstTransformRcPtr transform = look->getTransform();         
+                OCIO::ConstTransformRcPtr invTransform = look->getInverseTransform();         
+
+                if(transform!=NULL)
```
hodoulp: [Coding Style] The std::shared_ptr supports [boolean cast](en.cppreference.com/w/cpp/memory/shared_ptr/operator_bool) to check if null or not. 
So the code can be simplified like `if (transform)`

src/apps/ociocheck/main.cpp
```diff
@@ -209,6 +209,40 @@ int main(int argc, const char **argv)
...
+                        std::cout << "src file found" << std::endl;
+                    }
+                    catch(OCIO::Exception & exception)
+                    {
+                        std::cerr << "ERROR: " << exception.what() << std::endl;
+                        errorcount += 1;
+                    }
+                }
+
+                if(invTransform!=NULL)
```
hodoulp: Same comment as the one at line 218

src/apps/ociocheck/main.cpp
```diff
@@ -209,6 +209,40 @@ int main(int argc, const char **argv)
             for(int i=0; i<config->getNumLooks(); ++i)
             {
                 std::cout << config->getLookNameByIndex(i) << std::endl;
+
+                OCIO::ConstLookRcPtr look = config->getLook(config->getLookNameByIndex(i)); 
+
+                OCIO::ConstTransformRcPtr transform = look->getTransform();         
+                OCIO::ConstTransformRcPtr invTransform = look->getInverseTransform();
```
hodoulp: [Coding Style] You should keep close related code. This line could be moved at line 231 ( while preserving the empty line from the ending curly bracket i.e. `}` )

PR #949, opened 2020-03-09
Upgrade EOL AZP agents

michdolan:
CI builds are now failing for the two OS versions AZP is retiring. I have swapped both of these for the respective new platform Microsoft hosted agents: macOS 10.15, and Windows 2019.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -167,15 +167,15 @@ jobs:
 - job: macOS
   strategy:
     matrix:
+      10.15 AppleClang 11.0:
+        agentImage: 'macOS-10.15'
+        buildType: Release
       10.14 AppleClang 10.0:
         agentImage: 'macOS-10.14'
         buildType: Release
       10.14 AppleClang 10.0 (Debug):
         agentImage: 'macOS-10.14'
         buildType: Debug
-      10.13 AppleClang 10.0:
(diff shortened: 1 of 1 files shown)


michdolan:
I don't have time to investigate the Windows 2019 agent failure, so am disabling it for now. Either we can revisit when there is time, or hold out for a switch to GitHub Actions, which will require general updates anyway.

michdolan:
I disabled required checks for the two failing builds, so technically we should be unblocked already, but this PR at least adds an additional macOS build, which can be added as a requirement after merging.

hodoulp:
Thanks for the investigation.

I notice that the other Windows images have the compiler version in the name but not the new one
i.e. `vs2017-win2016` versus `windows-2019`. Any hints?

As this is a blocking situation I approved the pull request.

michdolan:
`windows-2019` is the new naming AZP is using (which is more consysent with their naming on other platforms). Each hosted agent has a particular tag, and as of this build they updated the naming scheme. Here's an overview of all available agents, with the installed software that each has:

https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops

Looks like GitHub actions uses identical tags:
https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

The label I gave the job calls out the specific MSVC version, which I looked up.

PR #951, opened 2020-03-12
Adsk Contrib - Add new options to ocioconvert

BernardLefebvre:
Currently ocioconvert only works from a config file and src/dst colourspaces.  This adds two other ways of working:

1)  Specify a LUT file (e.g. a CLF).
or
2)  Specify a source and view + display (rather than src and dst).

The current syntax is:
usage: ocioconvert [options]  inputimage inputcolorspace outputimage outputcolorspace

The new syntax also allows:
usage: ocioconvert [options] --lut LUTNAME  inputimage outputimage 
or 
usage: ocioconvert [options] --view  inputimage inputcolorspace outputimage DISPLAYNAME  VIEWNAME

### DIFF
--- a/src/apps/ocioconvert/main.cpp
+++ b/src/apps/ocioconvert/main.cpp
@@ -357,33 +357,96 @@ int main(int argc, const char **argv)
     bool usegpuLegacy = false;
     bool outputgpuInfo = false;
     bool verbose = false;
+    bool useLut = false;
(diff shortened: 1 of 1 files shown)


doug-walker:
We need this for the ACES CLF working group.  Since it is a minor enhancement to an existing app, we will merge this towards the end of the week, if there are no other comments/suggestions.

PR #954, opened 2020-03-12
Add cmake support for OCIO_LIBNAME_SUFFIX

lgritz:
This patch from Lee Kerley makes it easier for people making custom builds to add an optional suffix to the library name itself.

The purpose is to prevent a studio build for internal tools from conflicting with a different build of OCIO that might be embedded in a DCC app. The existing option to customize the inner namespace does a good job preventing symbol clashes at link time, but does not address the problem where both the DCC app and the plugin we write end up callled libOpenColorIO.so, so whichever is first in the `LD_LIBRARY_PATH` will win and the other won't be found. 

This is a patch against RB-1.1, because that's what we need ASAP, but after this is merged, we will submit another one against master (the build system has changed enough it won't just cherry pick).

Signed-off-by: Lee Kerley <lkerley@imageworks.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,10 @@ if(NOT OCIO_NAMESPACE)
 endif(NOT OCIO_NAMESPACE)
 messageonce("Setting Namespace to: ${OCIO_NAMESPACE}")
 
+set(OCIO_LIBNAME_SUFFIX "" CACHE STRING
+    "Specify a suffix to all libraries that are built")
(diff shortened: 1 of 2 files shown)


lgritz:
The system won't let me review this, because I'm the one who submitted the PR. But actually, Lee wrote it, I just filed the PR to save him some time, so consider it reviewed and approved from me.

This is an important fix for us, so if at all possible, it would be really helpful to get the reviews and merge on this ASAP.

PR #955, opened 2020-03-13
Add cmake support for OCIO_LIBNAME_SUFFIX

lgritz:
This patch from Lee Kerley makes it easier for people making custom
builds to add an optional suffix to the library name itself.

The purpose is to prevent a studio build for internal tools from
conflicting with a different build of OCIO that might be embedded in a
DCC app. The existing option to customize the inner namespace does a
good job preventing symbol clashes at link time, but does not address
the problem where both the DCC app and the plugin we write end up
callled libOpenColorIO.so, so whichever is first in the
`LD_LIBRARY_PATH` will win and the other won't be found.

Signed-off-by: Lee Kerley <lkerley@imageworks.com>


lgritz:
This is basically the same as #954, but applied to master (where the cmake files have shifted around a bit, so it wasn't a direct cherry-pick).

lgritz:
I don't even begin to understand why it says Lee is not authorized on the CLA when we did another PR earlier today. I'm not sure what is different about this one.

michdolan:
In the previous PR, Lee's GH ID was `leekerley` instead of `lkerley`.

src/OpenColorIO/CMakeLists.txt
```diff
@@ -218,11 +218,10 @@ if(WIN32)
 endif()
 
 set_target_properties(OpenColorIO PROPERTIES
-	COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS}")
-
-set_target_properties(OpenColorIO PROPERTIES
-    VERSION ${OpenColorIO_VERSION}
-    SOVERSION ${SOVERSION})
+    OUTPUT_NAME   OpenColorIO${OCIO_LIBNAME_SUFFIX}
```
hodoulp: Something I learn. 

[Not mandatory] As you are reworking the `OUTPUT_NAME`, you could:
1. Change `LIBNAME_SUFFIX` to be ` _${OpenColorIO_VERSION_MAJOR}_${OpenColorIO_VERSION_MINOR}`
i.e. prepended an underscore;
2. Append `${LIBNAME_SUFFIX}` (i.e. which will be `_2_0` for Windows and empty for all other platforms)
i.e. `OUTPUT_NAME   OpenColorIO${OCIO_LIBNAME_SUFFIX}${LINAM_SUFFIX}`;
3. Copy & paste lines 206 to 210 to replace the comment at line 133;
4. Remove lines 204 to 215.

I think that it nicely completes the CMakeLists.txt cleanup around the right output name for the library.

lgritz: Are these things you want me to do in this PR, or is this just a mental note for things to improve further later?

lgritz:
@michdolan Where does the GH ID even occur? In both cases, I submitted the PR, and in both cases, Lee had the same signed-off-by. I'm confused about how I would even tell it what GitHub ID is associated with him.

michdolan:
Good question. The sign-off is identical, but if you look at the GitHub user associated with the commits in the two PR's, they are different account. Here are the two accounts:

https://github.com/leekerley
https://github.com/lkerley

My guess is the first one is associated with the SPI CCLA, but not the second. Maybe just add both to the CCLA whitelist?

src/OpenColorIO/CMakeLists.txt
hodoulp: In the PR if possible.

lgritz:
But I never gave either GH ID, so how did it ever associate this with GH lkerley? And how would it have associated the last one with Lee if I'm the one who submitted the PR?

Unless... lkerley is also Lee from long ago and forgotten, leekerley is new Lee, but they both use the same email and it recognized it in the DCO sign-off and just found one account associated with it, nondeterministically?

lgritz:
OK, mystery solved.

GH will use the email in the DCO signed-off-by field, or maybe the --author for the commit, to associate it with a GH account, even if you didn't use your GH account to post the PR (for example, if I did it on behalf of Lee). But I guess I used his home email for one and work for the other? So anyway, he somehow has two GH accounts, one forgotten, and I just need to add the other to our CLA whitelist.

lgritz:
I will also take a look at @hodoulp's comments later, so don't merge yet even after I do that

src/OpenColorIO/CMakeLists.txt
lgritz: Pushed an update that implements these suggestions. Also fixed version.rc.in.

PR #956, opened 2020-03-13
Updated documentation to incorporate changes from SPI to ASWF 

rashil2000:
Also updated footers of docs pages


michdolan:
Thanks for the submission @rashil2000 ! Your help is very much appreciated. I'm closing this PR for now for the following reasons, and providing direction on how to re-submit later:

- The current OCIO website is built from the `RB-1.1` branch, not `master` (this will change when we near the v2 release)
- We should wait to update the site until #953 is approved and merged, and then we would need to make those same changes to `RB-1.1` via a second PR first.
- It `looks` like the content of this PR is coming from the doc build directory, instead of the installed docs. Make sure you run `make install` following the build and update `gh-pages` from that content, to avoid picking up `.doctrees`, etc. The install-sourced PR would likely not add or remove any files to the branch; only edit them.

But submitting the website changes to `gh-pages` is the correct approach when ready, so good job following through with the deployment.

rashil2000:
@michdolan I had a confusion regarding doctrees too. I saw that they already existed in the gh-pages branch, that's why I pulled them from the doc build directory.
Anyways you were right. Let's wait for #953 to get merged first.

michdolan:
Sounds good. You are right about the .doctrees. I think the reason I thought that is that this build is from master which has additional content from all the v2 work. Thanks again.

rashil2000:
@michdolan Should I start making the required changes here?

michdolan:
Yes, if you have the time to make these updates to RB-1.1 so they can be made live on the website that would be great. Thanks!

rashil2000:
@michdolan Where should I pick the content from this time, the doc build directory or the install directory?

Issue #927, opened 2020-02-04
Providing an interchange mechanism between configs


doug-walker:
**BACKGROUND**

An OCIO config defines its own self-contained universe of color spaces with no explicit connection to other configs or other forms of color management.  It has been said that this is perhaps simultaneously both its best and worst feature.    ;-)

It is great in the sense that the config author defines every color space and how they connect to each other.  There are no unknowns and everything is under control.

However there are cases where it would be extremely useful (or even essential) to be able to convert from a color space in one config to either a color space in another config, or to color spaces as they may exist in other applications/systems.  The lack of this possibility with current configs is a major blocker to using OCIO in a number of workflows today.

Some examples:
1. Many studios create image and CG assets that their artists would like to reuse between shows/projects.  These assets may be tagged as being in specific color spaces that were defined in the config for that project.  Often a new project will have its own config, which may not be fortunate enough to define all the color spaces needed to interpret these older assets.
2. New standards such as USD and MaterialX allow work to be broken down into pieces that may then be assembled or incorporated by reference into larger works.  Currently, there is no way to bring in a component that uses a color space from a separate config that does not also exist in the main config.
3. Assets (images, CG, materials, etc.) may be tagged as being in a color space that may not be associated with any specific config.  For example they may use ACES color spaces, ITU color spaces, or color spaces associated with a given vendor (ARRI, RED, Sony, etc.).  Applications may sometimes need to work with these assets in a way that interoperates with assets associated with the currently loaded config even if that config does not provide a color space for it.
4. It is currently non-trivial for an application to determine whether a color space in one config is actually the same as a color space in another config.

**PROPOSAL**

The proposal is to use the existing role mechanism to refer to one or more standard well-known and precisely defined color spaces.  If two configs both define a common "interchange" role, a conversion from color space A to color space B may proceed by using the first config to go from CS A to the interchange space and then using the second config to convert from the interchange space to CS B. 

The most well-defined standard scene-referred color space today is ACES2065-1, which is defined in SMPTE standard ST2065-1.  The proposal is to allow this to be used as an interchange color space.  The proposed name of the role is: "aces_interchange".

Other interchange color spaces could be defined over time, if needed.

A new getProcessor method will be added to the Config class that would take the source and destination configs and color spaces and return a processor that is able to perform the requested conversion.  (Note that once a Processor has been built, it is essentially just a sequence of ops that is independent of any config.)

[...]

PR #957, opened 2020-03-13
Adsk Contrib - Add exchange role for display-referred color space

BernardLefebvre:
Now that reference type has been added to color spaces, we can enhance the feature to create processor between configs.
When first color space is scene-referred, the role for the exchange color space is `aces_interchange`.
When going from one config to another, if the first color space is display-referred, the default role of the exchange color space is `cie_xyz_d65_interchange`.


doug-walker:
This completes the implementation of issue #927.

PR #958, opened 2020-03-14
Added a check for duplicate keys (fixes #536)

### LINKED ISSUE #536: Config::sanityCheck() bug
Small bug I encountered today, _sanityCheck_ doesn't catch the following case:

```
- !<Look>
  name: primary_dn
  process_space: log
  transform: !<FileTransform> {src: my_file.cdl, interpolation: linear}
  transform: !<FileTransform> {src: my_file.cdl, interpolation: linear}
```

Where the YAML "transform" tag is defined twice, behaviour was that it didn't error it just grabbed the last defined "transform" and used that FileTransform.

rashil2000:

### DIFF
--- a/src/OpenColorIO/OCIOYaml.cpp
+++ b/src/OpenColorIO/OCIOYaml.cpp
@@ -2,8 +2,9 @@
 // Copyright Contributors to the OpenColorIO Project.
 
 #include <cstring>
-#include <OpenColorIO/OpenColorIO.h>
 #include <unordered_set>
+
+#include <OpenColorIO/OpenColorIO.h>
 
 #include "Display.h"
 #include "FileRules.h"
@@ -282,7 +283,9 @@ inline void CheckDuplicates(const YAML::Node & node)
         const YAML::Node & first = iter.first;
         load(first, key);
         if (keyset.find(key) == keyset.end())
+        {
             keyset.insert(key);
+        }
         else
         {
             std::ostringstream os;
--- a/src/OpenColorIO/OCIOYaml.cpp
+++ b/src/OpenColorIO/OCIOYaml.cpp
@@ -4,6 +4,7 @@
 #include <cstring>
 
 #include <OpenColorIO/OpenColorIO.h>
+#include <unordered_set>
 
 #include "Display.h"
 #include "FileRules.h"
@@ -268,6 +269,29 @@ inline void throwValueError(const std::string & nodeName,
         << "' from '" << nodeName << "' failed: " << msg;
 
     throw Exception(os.str().c_str());
+}
+
+// Duplicate Checker
+
+inline void checkDuplicates(const YAML::Node& node)
+{
+    std::string key;
+    std::unordered_set<std::string> keyset;
+
+    for (const auto & iter : node)
+    {
+        const YAML::Node& first = iter.first;
+        load(first, key);
+        if (keyset.find(key) == keyset.end())
+            keyset.insert(key);
+        else
+        {
+            std::ostringstream os;
+            os << "Key-value pair with key '" << key;
+            os << "' specified more than once. ";
+            throwValueError(node.Tag(), first, os.str());
+        }
+    }
 }
 
 // View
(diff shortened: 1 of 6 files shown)


src/OpenColorIO/OCIOYaml.cpp
```diff
@@ -4,6 +4,7 @@
 #include <cstring>
 
 #include <OpenColorIO/OpenColorIO.h>
+#include <unordered_set>
```
hodoulp: We order all the includes, and also regroup them based on 'family'. So, you should move that one up to line 5 (i.e. keep the empty line with the existing include at line 6).

src/OpenColorIO/OCIOYaml.cpp
```diff
@@ -270,6 +271,29 @@ inline void throwValueError(const std::string & nodeName,
     throw Exception(os.str().c_str());
 }
 
+// Duplicate Checker
+
+inline void checkDuplicates(const YAML::Node& node)
```
hodoulp: The global methods must start with an upper case letter (i.e. we did not revisit all the existing files to conform them to the coding style) and add a space between `YAML::Node` and `&`.

src/OpenColorIO/OCIOYaml.cpp
```diff
@@ -270,6 +271,29 @@ inline void throwValueError(const std::string & nodeName,
...
+{
+    std::string key;
+    std::unordered_set<std::string> keyset;
+
+    for (const auto & iter : node)
+    {
+        const YAML::Node& first = iter.first;
```
hodoulp: Same comment than the one at line 276.

tests/cpu/Config_tests.cpp
```diff
@@ -181,6 +181,36 @@ OCIO_ADD_TEST(Config, simple_config)
...
+        "    name: raw_duplicated\n"
+        "    name: raw\n"
+        "\n";
+
+    std::istringstream is;
+    is.str(SIMPLE_PROFILE);
+    OCIO::ConstConfigRcPtr config;
+    OCIO_CHECK_THROW_WHAT(OCIO::Config::CreateFromStream(is),
+                              OCIO::Exception,
+                              "Key-value pair with key 'name' "
```
hodoulp: When the argument list is on several lines, they must be aligned, and the max. line length is 100 characters. You should adjust the code accordingly.

tests/cpu/Config_tests.cpp
```diff
@@ -181,6 +181,36 @@ OCIO_ADD_TEST(Config, simple_config)
     OCIO_CHECK_NO_THROW(config->sanityCheck());
 }
 
+OCIO_ADD_TEST(Config, simple_config_with_duplicates)
+{
+
+    constexpr char SIMPLE_PROFILE[] =
+        "ocio_profile_version: 2\n"
+        "search_path: \n"
```
hodoulp: I missed than one, please add `luts` as the directory where to find the LUT files. 
The line must be:  `"search_path: luts\n"`.

src/OpenColorIO/OCIOYaml.cpp
rashil2000: I had a confusion regarding this too. I tried to visited all the functions on `OCIOYaml.cpp` to check for both the capitalization and the space before the '&' symbol. Some functions had the space, others didn't. So I made a guess.

hodoulp: That's the coding style but not all files are following it. So, only modifying the code you changed is enough.

src/OpenColorIO/OCIOYaml.cpp
hodoulp: To clarify the point:
```
#include <cstring>
#include <unordered_set>

#include <OpenColorIO/OpenColorIO.h>
```

src/OpenColorIO/OCIOYaml.cpp
```diff
@@ -270,6 +270,29 @@ inline void throwValueError(const std::string & nodeName,
...
+{
+    std::string key;
+    std::unordered_set<std::string> keyset;
+
+    for (const auto & iter : node)
+    {
+        const YAML::Node & first = iter.first;
+        load(first, key);
+        if (keyset.find(key) == keyset.end())
+            keyset.insert(key);
```
hodoulp: Even a one-line branch needs to be surrounded by curly brackets.

PR #959, opened 2020-03-15
added some info

jaichhabra:
added some info on

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -8,8 +8,7 @@ Introduction
 ============
 
 OpenColorIO (OCIO) is a complete color management solution geared towards motion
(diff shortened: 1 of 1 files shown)


michdolan:
Thanks for the submission @jaichhabra . These additions are duplicating content from `index.rst` . I think it's better to keep this page unique. 

I will however suggest a doc page that could use some updates that you could work on. This page: https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/docs/FAQ.rst has a table of supported LUT formats which is out of date. You could look at this directory: https://github.com/AcademySoftwareFoundation/OpenColorIO/tree/master/src/OpenColorIO/fileformats for all the currently supported LUT formats, and add what is missing.

PR #968, opened 2020-03-17
Fixes #527: Parsing error on whitespace

### LINKED ISSUE #527: Parsing error on whitespace
`direction:inverse` fails while `direction: inverse` succeeds silently.

@Shootfast reports that ociocheck also fails to note the problem.

ishitarb5:
Checks the key for a colon character to make sure the value is not included within the key and not being misinterpreted as a null value. Therefore, an error message should be returned for a missing colon delimiter. 
Signed-off-by: Ishita Bhimavarapu ishitarb5@gmail.com

### DIFF
--- a/tests/cpu/Config_tests.cpp
+++ b/tests/cpu/Config_tests.cpp
@@ -181,6 +181,39 @@ OCIO_ADD_TEST(Config, simple_config)
     OCIO_CHECK_NO_THROW(config->sanityCheck());
 }
 
+OCIO_ADD_TEST(Config, missing_space_key_value_pair)
+{
+
+    constexpr char SIMPLE_PROFILE[] =
+        "ocio_profile_version: 2\n"
+        "search_path: luts\n"
+        "roles:\n"
+        "  default: raw\n"
+        "file_rules:\n"
+        "  - !<Rule> {name: Default, colorspace: default}\n"
+        "displays:\n"
+        "  Disp1:\n"
+        "    - !<View> {name: View1, colorspace: raw}\n"
+        "active_displays: []\n"
+        "active_views: []\n"
+        "colorspaces:\n"
+        "  - to_reference: !<FileTransform> {direction:inverse}\n"
+        "\n";
+
+    std::istringstream is;
+    is.str(SIMPLE_PROFILE);
+
+    OCIO::ConstConfigRcPtr config;
+    OCIO_CHECK_NO_THROW(config = OCIO::Config::CreateFromStream(is));
+
+    {
+        OCIO::LogGuard log;
+        OCIO_CHECK_NO_THROW(config->sanityCheck());
+        OCIO_CHECK_EQUAL(log.output(), 
+                         "[OpenColorIO Warning]: Unknown key in FileTransform: 'direction:inverse'.\n");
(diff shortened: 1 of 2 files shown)


hodoulp:
@ishitarb5 Thanks for your contribution.

When submitting a change, dev. needs also to provide a unit test to demonstrate the fix or the feature.  Refer to `tests/cpu/Config_tests.cpp` to have example.

Here is a config string that can be used:
````
    constexpr char SIMPLE_PROFILE[] =
        "ocio_profile_version: 2\n"
        "search_path: luts\n"
        "roles:\n"
        "  default: raw\n"
        "file_rules:\n"
        "  - !<Rule> {name: Default, colorspace: default}\n"
        "displays:\n"
        "  Disp1:\n"
        "    - !<View> {name: View1, colorspace: raw}\n"
        "active_displays: []\n"
        "active_views: []\n"
        "colorspaces:\n"
        "  - !<ColorSpace>\n"
        "    name: raw\n"
        "    allocation: uniform\n"
        "\n";
```
You can tweak one line to demonstrate the change.

hodoulp:
BTW, below is the Linux & gcc build break:
````
2020-03-17T03:24:20.8996573Z /__w/1/s/src/OpenColorIO/OCIOYaml.cpp: In function 'void OpenColorIO_v2_0dev::{anonymous}::load(const YAML::Node&, OpenColorIO_v2_0dev::FileTransformRcPtr&)':
2020-03-17T03:24:20.8998654Z /__w/1/s/src/OpenColorIO/OCIOYaml.cpp:1063:48: error: suggest parentheses around '&&' within '||' [-Werror=parentheses]
2020-03-17T03:24:20.8999396Z          if (key.find(":") != std::string::npos && second.IsNull() || !second.IsDefined()) continue;
2020-03-17T03:24:20.9000430Z              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
```

hodoulp:
@sobotka @dbr @michdolan Whatever will be the right fix it relies on the prerequisite that a key cannot contain a colon `:`. Is that true?

sobotka:
I would hope that configs don’t use reserved escape characters. Seems like a quagmire not worth the investment.

tests/cpu/Config_tests.cpp
```diff
@@ -181,6 +181,39 @@ OCIO_ADD_TEST(Config, simple_config)
...
+        "roles:\n"
+        "  default: raw\n"
+        "file_rules:\n"
+        "  - !<Rule> {name: Default, colorspace: default}\n"
+        "displays:\n"
+        "  Disp1:\n"
+        "    - !<View> {name: View1, colorspace: raw}\n"
+        "active_displays: []\n"
+        "active_views: []\n"
+        "colorspaces:\n"
```
michdolan: This is not a valid ColorSpace definition, which is why your test is failing in CI. Please have a look at the other tests in this file to see what is missing. You can run these tests locally to debug the issue. After you build run: `ctest -V`

michdolan:
Closing this for inactivity. Please reopen if you would like to further revise it.

PR #969, opened 2020-03-20
Adsk Contrib - Adjust CDL GPU tests for default transform style change

BernardLefebvre:
Recent CDL changes did modify the default CDL transform style and this implies some changes to the CDL gpu tests.
Also adding some tests.


hodoulp:
As other pull requests depend on this work and it's an easy one, we would like to merge it soon i.e. this week unless objections.

PR #975, opened 2020-03-27
Adsk Contrib - Add version validation

hodoulp:

The pull request adds some version validation when reading or writing a config file. As OCIOv2 introduced many new capabilities changing a config file, it's now important to validate that a config version 1 does not contain any version 2 transforms for example.


src/OpenColorIO/Config.cpp
```diff
@@ -2610,5 +2617,113 @@ ConstConfigRcPtr Config::Impl::Read(std::istream & istream, const char * filenam
     return config;
 }
 
+void Config::Impl::checkVersionConsistency(ConstTransformRcPtr & transform) const
```
michdolan, line 2620: Just a question. Do you think we should add a version compatibility method to the `Transform` interface, so each transform can determine its own compatibility outside of `Config`? I don't have a problem with this implementation; was just thinking that this is the sort of function that can balloon over time.

hodoulp: That's a valid point.
The rational is to avoid having a version related method in the public API (i.e. not useful for any public usage) and to centralize all the checks. There is also the point that it means some code for the first implementation but it should not change a lot in the future.

hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can now be merged this week unless there are some comments.

hodoulp:
This has been up for two weeks, given that there has been no feedback, we will go ahead and merge tomorrow.

PR #977, opened 2020-03-29
Made #953 changes in RB-1.1 release branch

rashil2000:

### DIFF
--- a/docs/conf.py.in
+++ b/docs/conf.py.in
@@ -14,7 +14,7 @@ templates_path = ['templates']
 source_suffix = '.rst'
 master_doc = 'index'
 project = u'OpenColorIO'
-copyright = u'2010, Sony Pictures Imageworks'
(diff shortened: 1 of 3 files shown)


rashil2000:
@michdolan Could you tell me where I should I pick the content from, the doc build directory or the install directory, for making the `gh-pages` branch?

michdolan:
@rashil2000 I said install before, but in looking at the `gh-pages` contents compared to the CMake install command for the docs, it looks like it should be the build dir to match what's currently live.

```
install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/build-html/
        DESTINATION ${CMAKE_INSTALL_PREFIX}/share/doc/OpenColorIO/html
        PATTERN .* EXCLUDE
)
```

If it was install, there would be no files/dirs beginning with `.`. Thanks for making this update!

PR #981, opened 2020-04-02
Re: Updated documentation to incorporate changes from SPI to ASWF

rashil2000:


configurations/index.html
```diff
@@ -120,7 +120,7 @@ <h2>Config Creation<a class="headerlink" href="#config-creation" title="Permalin
       </div>
 
     <div class="footer">
-        &copy; Copyright 2010, Sony Pictures Imageworks.
+        &copy; Copyright Copyright Contributors to the OpenColorIO Project.
```
lgritz, line 123: All of these seem to have the word Copyright doubled.

michdolan: @rashil2000 looks like you'll need to do another RB-1.1 PR to revise this since we missed it in the rst. So anywhere there's:

`u'Copyright Contributors to the OpenColorIO Project'`

Will need to be changed to:

`u'Contributors to the OpenColorIO Project'`

rashil2000: @lgritz @michdolan I've made the requested changes.

PR #986, opened 2020-04-10
Adsk Contrib - Improve op finalization

BernardLefebvre:
Reworking ops finalization:
- ops are only finalized once.
- clarify LUT inversion
- move direction to MatrixOpData and RangeOpData
- change how FixedFunction surround inverse is saved


hodoulp:
This has been up for two weeks, given that there has been no feedback, we will go ahead and merge this week.

PR #987, opened 2020-04-14
Point to Colour-Science repo for reference configs

omenos:
The imageworks/OpenColorIO-Configs repository does not see
any new development or releases. Thomas Mansencal has been
maintaining the configs with newer standards such as
ACES 1.1 and 1.2 for the community.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ Reference Configs
 -----------------
 
 Reference OCIO configuration files and associated LUTs can be found at the
-Imageworks [OpenColorIO-Configs](https://github.com/imageworks/OpenColorIO-Configs)
-repository.
+Colour-Science [OpenColorIO-Configs](https://github.com/colour-science/OpenColorIO-Configs)
+repository, maintained by Thomas Mansencal.
 
 The following reference implementations are provided:
 
@@ -62,6 +62,8 @@ The following reference implementations are provided:
   * spi-anim
   * spi-vfx
 * ACES: Academy Color Encoding System
+  * aces_1.2
+  * aces_1.1
   * aces_1.0.3
   * aces_1.0.2
   * aces_1.0.1


michdolan:
Thanks @omenos . The reason for `OpenColorIO-Configs` being without the new ACES configs is related to a license issue we have been trying to find a solution for with AMPAS. The issue is sort of in a holding pattern, but you can find more information about it here: https://github.com/imageworks/OpenColorIO-Configs/pull/17

I think it's a good idea to link to @KelSolaar 's configs from the OCIO website for visibility (especially since these are shipping with DCCs like Nuke), but I think the link should be added in addition to the `OpenColorIO-Configs` repo, rather than replacing it, with a note that Thomas's fork is where to find these specific ACES configs.

My other comment is that if you want this to appear on the current website, you'll need to submit a PR against the RB-1.1 branch, which is what currently sources the website. `master` will eventually source the site, but closer to the OCIO v2 release later in the summer. Thanks!

KelSolaar:
Ahhhh, the Config Repo Location story! ;)

Related: I will tag the 1.2 as final given nobody complained during the past weeks :)

omenos:
> The reason for OpenColorIO-Configs being without the new ACES configs is related to a license & repo structure discussion we have been having with the ACES team.

Ah, I was not aware of this! Thanks for bringing it to my attention. I always did find it strange that the configs repository hadn't found its way over to the ASWF domain.

> I think it's a good idea to link to @KelSolaar 's configs from the OCIO website for visibility (especially since these are shipping with DCCs like Nuke), but I think the link should be added in addition to the OpenColorIO-Configs repo, rather than replacing it, with a note that Thomas's fork is where to find these specific ACES configs.
> 
> My other comment is that if you want this to appear on the current website, you'll need to submit a PR against the RB-1.1 branch, which is what currently sources the website.

I will make that change, and thanks for the clarification on how the repository is structured. I'm very much looking forward to OCIO v2, a huge shoutout to the everyone working on it for the hard work they've done. It's been amazing for this industry.

@KelSolaar: I've been using it seemingly without issues, but I wonder if anybody with more knowledge/experience has been able to test it in a more thorough fashion across their pipeline. Have you gotten any feedback on it since you initially tagged it?

PR #991, opened 2020-04-17
Add GH Actions CI support

michdolan:
This adds GH Actions CI as a replacement for AZP, per the ASWF decision to switch, and in preparation for adding GPU CI support.

Some notes to help in reviewing the changes:
- GH Actions currently has no support for templates like AZP had, so everything exists in the one yml workflow file. I could have moved some of the commands into .sh scripts, but opted to leverage the dynamic nature of the job matrices, so the workflow is mostly self-contained. The benefit is that it is easier to see what's going on at a glance.
- GH Actions matrices are quite different in design from AZP. In theory their approach is quite nice, as you get the cross product of one or more sequences of options, but I found in practice that it makes it harder to keep the job budget down while running CI. It's easy to end up having 16 jobs when trying to test a number of different configurations. My approach to solving this was to force an AZP style matrix by having the only matrix sequence be a list of build indices, and then using the `include` feature (designed for tacking additional variables onto specific matrix builds) to define each of the index builds independently. The result is fewer builds, but still good representation of OCIO build options. 
- Each platform has Release, Debug, Python 3, Python 2, SSE, no SSE, static, shared, and on Linux GCC and Clang. Some commented workarounds are present which will be removed after a `ci-ocio:2020` image is added to the `aswf-docker` project. I will submit issues/PRs there for needed changes and patch our workflow in later PR(s).
- I have disabled CI when the only changes are .md files to ease PRs for TSC notes and documentation which does not need to be built. Hopefully that will not cause issues with required checks for these PRs. We can easily change that if it does.
- I tested all the builds in my fork successfully, including the latest and SonarCloud builds. Having CI directly in GH is quite nice!


michdolan:
Here is the successful full GH Actions build for reference, since it will not trigger in this PR:
https://github.com/michdolan/OpenColorIO/actions/runs/80819230

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+
+  # ---------------------------------------------------------------------------
+  # Linux
+  # ---------------------------------------------------------------------------
+  # Note: 'runs-on' defines the GH-hosted VM. 'container' is the Docker 
+  #       container where the build occurs.
+  # Note: 'container' images are hosted on DockerHub
+
+  linux:
+    name: Linux CentOS 7 VFX CY2019 ${{ matrix.build-desc }}
```
BernardLefebvre, line 39: If we do not control the versions, we should remove them from the name.

michdolan: We do actually control the Linux version, but it's controlled through the aswf container we spin up rather than through the GH-provided VM. The comment on Linux version is specific to the VM.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+      - name: Create build directories
+        run: |
+          mkdir _install
+          mkdir _build
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
+                -DCMAKE_CXX_STANDARD=14 \
```
hodoulp, line 116: By default, OCIO is C++11 for now. That's a good idea to add a build with C++14, but C++11 must be validated too.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+                -DOCIO_BUILD_DOCS=ON \
```
hodoulp, line 117: The documentation generation could be validated per platform i.e. that's not dependent of the compiler.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+          - build: 3
+            build-desc: (static, no SSE, Python 2.7)
+            build-type: Release
+            python-version: 2.7
+            shared: 'OFF'
+            use-sse: 'OFF'
+            cxx-compiler: g++
+            cc-compiler: gcc
+          - build: 4
+            build-desc: '(Clang 7)'
```
hodoulp, line 70: Are the  '  needed for the description?

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+            build-type: Release
+            python-version: 3.7
+            shared: 'ON'
+            use-sse: 'ON'
+            cxx-compiler: clang++
+            cc-compiler: clang
+          - build: 5
+            build-desc: (Clang 7, Debug)
```
hodoulp, line 78: Where is imposed the version of CLang?

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+    strategy:
+      matrix:
+        build: [1, 2]
+        include:
+          - build: 1
+            build-desc: ''
+            cxx-compiler: g++
+            cc-compiler: gcc
+          - build: 2
+            build-desc: ', Clang 7'
```
hodoulp, line 152: Please remove comma in the description.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+        run: |
+          mkdir _install
+          mkdir _build
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=Release \
+                -DCMAKE_CXX_STANDARD=14 \
+                -DOCIO_BUILD_DOCS=ON \
```
hodoulp, line 191: I do not think that the documentation generation is important here.

hodoulp: .. but C++11 & C++14 are.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=Release \
+                -DCMAKE_CXX_STANDARD=14 \
+                -DCMAKE_CXX_FLAGS="-g -O0 -fprofile-arcs -ftest-coverage" \
+                -DCMAKE_CXX_OUTPUT_EXTENSION_REPLACE=ON \
+                -DCMAKE_EXE_LINKER_FLAGS="-lgcov" \
+                -DOCIO_BUILD_DOCS=ON \
```
hodoulp, line 240: The documentation generation is useless for the static analysis build.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+          mkdir _install
+          mkdir _build
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
+                -DCMAKE_CXX_STANDARD=14 \
+                -DBUILD_SHARED_LIBS=${{ matrix.shared }} \
+                -DOCIO_BUILD_DOCS=ON \
```
hodoulp, line 308: Useful to have at least one build generating the documentation, but not needed for all builds.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+          mkdir _install
+          mkdir _build
+        shell: bash
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
+                -DCMAKE_CXX_STANDARD=14 \
+                -DCMAKE_GENERATOR_PLATFORM=x64 \
```
hodoulp, line 377: C++ 11 must be validated as that's the default version for OCIO.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+      - name: Create build directories
+        run: |
+          mkdir _install
+          mkdir _build
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
+                -DCMAKE_CXX_STANDARD=14 \
```
hodoulp, line 306: C++ 11 must be validated as that's the default version for OCIO.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+      - name: Create build directories
+        run: |
+          mkdir _install
+          mkdir _build
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=Release \
+                -DCMAKE_CXX_STANDARD=14 \
```
hodoulp, line 236: Use C++11 for the static analysis as that's the default C++ version for OCIO.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,397 @@
...
+  push:
+    branches:
+      - master
+      - RB-*
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    branches:
+      - master
+      - RB-*
```
hodoulp, line 17: To correctly understand all the pull requests must use the CI build. It should not be limited to pull requests with a specific branch name.

.github/workflows/ci_workflow.yml
michdolan: Changed. I used C++14 for some of the AZP jobs too, but I can't recall why. There was a reason at one point related to compiler warnings, but doesn't seem to be now.

.github/workflows/ci_workflow.yml
michdolan: Changed so that there is a doc build for each platform and Python version (so two per platform currently). I kept docs enabled for SonarCloud and one latest package build as well.

.github/workflows/ci_workflow.yml
michdolan: I overhauled the description handling so it's much more explicit now

.github/workflows/ci_workflow.yml
michdolan: The Clang version is specific to the docker container from aswf-docker. Those containers  provide GCC 6.3.1 and Clang 7.

.github/workflows/ci_workflow.yml
michdolan: I overhauled the descriptions

.github/workflows/ci_workflow.yml
michdolan: Good point! The goal here was to prevent CI running for updates to older branches, bit that did limit the use quite a bit. I've changed it to explicitly exclude v0 and v1, but allow all others, which will behave with forks much more naturally now.

.github/workflows/ci_workflow.yml
michdolan: Docs need to be tested here because we are also testing latest versions of pybind11 and sphinx. I have changed it to only build for one of the two jobs though.

michdolan: For now I have switched this to C++11. We can add 14 too if needed, but I was trying to keep the build count to a reasonable number. Let me know if you think 14 here is worth another two builds.

michdolan:
GH Actions build status: https://github.com/michdolan/OpenColorIO/runs/610888635?check_suite_focus=true

michdolan:
I am now going to test a full build again by temporarily enabling the scheduled jobs. I will disable them in a follow up commit.

michdolan:
All jobs completed with new changes: https://github.com/michdolan/OpenColorIO/actions/runs/85424974

hodoulp:
[Question] The current build image is the ASWF 2019 one, and you plan to transition the build system to the 2020 one. However, I think that the build system must support the latest VFX Platform recommandation plus 1 or 2 older ones.

[Comment] The new build description is better. It would be nice to have the compiler version if that's possible.

.github/workflows/ci_workflow.yml
```diff
@@ -0,0 +1,426 @@
...
+      - name: Configure
+        run: |
+          cmake ../. \
+                -DCMAKE_INSTALL_PREFIX=../_install \
+                -DCMAKE_BUILD_TYPE=Release \
+                -DCMAKE_CXX_STANDARD=11 \
+                -DOCIO_BUILD_DOCS=${{ matrix.build-docs }} \
+                -DOCIO_BUILD_GPU_TESTS=OFF \
+                -DOCIO_INSTALL_EXT_PACKAGES=NONE \
+                -DOCIO_WARNING_AS_ERROR=ON \
```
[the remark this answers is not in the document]
michdolan: Valid point. Would probably be safer to do so (failing CI less), but it would also mean we would not notice warnings as quickly. That's probably acceptable though.

michdolan:
When I add CY2020 my plan was to keep CY2019 as well. We could add CY2018, but in terms of OCIO dependencies there is not a difference (compilers have not changed between the three years). Even between 2019/2020 the major difference for us is only the Python version.

I could re-add  compiler version, but it will need to be hard-coded since there is no way to inspect it from the docker container. As a counter argument, if you really want to know the specific compiler version, the job is broken up into really granular steps and within the top lines of the "Configure" step you can see the exact compiler version reported by CMake. But I can add it back in.

michdolan:
Also, I just remembered why I was using C++14. It's because VFX ref platform mandates it since CY2018. We can keep with C++11, but  just mentioning it.

.github/workflows/ci_workflow.yml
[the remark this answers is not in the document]
hodoulp: Yes. The goal is to detect build breaks & major issues (i.e. unit test failures).

hodoulp:
> When I add CY2020 my plan was to keep CY2019 as well. We could add CY2018, but in terms of OCIO dependencies there is not a difference [...]

Right, and CY2019 & CY2018 are for OCIOv1.

> Also, I just remembered why I was using C++14. [...]

So, one build per compiler with C++14 is then needed.

> Compiler version in the description

If that's not easy, then forget.

.github/workflows/ci_workflow.yml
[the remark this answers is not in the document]
michdolan: Warnings as error disabled for latest build

michdolan:
GH actions test of new changes:
https://github.com/michdolan/OpenColorIO/runs/614244105?check_suite_focus=true

.github/workflows/ci_workflow.yml
```diff
@@ -354,35 +430,47 @@ jobs:
...
             build-type: Release
             build-shared: 'OFF'
             build-docs: 'ON'
             use-sse: 'OFF'
+            cxx-version: 11
             python-version: 2.7
+          - build: 4
+            build-type: Release
+            build-shared: 'ON'
+            build-docs: 'ON'
```
hodoulp, line 470: Build 1 is already generated the docs so it could be OFF for build 3 & 4.

.github/workflows/ci_workflow.yml
```diff
@@ -280,35 +344,47 @@ jobs:
...
```
hodoulp, line 384: Build 1 already generates the docs so it could be OFF for build 3 & 4.

hodoulp:
Few comments using the latest CI build:

- The cxx is empty in the build title
`Linux CentOS 7 latest ext packages <Clang 7, config=Release, shared=ON, sse=ON, cxx=, python=3.7, docs=OFF>`.
- The docs build cost is about 3 mins per build so one docs build per platform is enough.
- The Python 3 install cost is almost 3 mins so it would be great to have it by default in the ASWF docker image (Python 2 & 3 in the ASWF docker images).
- In the 'latest' builds, there are issues:
  - `eXpat` problems are: builds tests & tools & examples, build type is unknown and builds a shared library.
  - `OpenEXR` problems are: builds tests & examples, and PYILMBASE_ENABLE is enabled (it's causing trouble between boost & python3 but that's known OpenEXR issue [684](https://github.com/AcademySoftwareFoundation/openexr/issues/684))

michdolan:
Most recent GH Actions build:
https://github.com/michdolan/OpenColorIO/actions/runs/86938211

.github/workflows/ci_workflow.yml
michdolan: Done.

.github/workflows/ci_workflow.yml
michdolan: Done.

michdolan:
The most recent commit responds to all comments except resolving the Python 3 install in the `aswf-docker` container. That will be done in a future PR since it will take time to submit that change and roll it into available containers.

PR #993, opened 2020-04-18
Duplicate check in other functions (#536)

rashil2000:
@hodoulp Do I add more tests or the one that was written before would suffice?


tests/cpu/Config_tests.cpp
```diff
@@ -184,7 +184,7 @@ OCIO_ADD_TEST(Config, simple_config)
     OCIO_CHECK_NO_THROW(config->sanityCheck());
 }
 
-OCIO_ADD_TEST(Config, simple_config_with_duplicate)
+OCIO_ADD_TEST(Config, simple_config_with_colorspace_duplicate)
```
hodoulp, line 187: The unit test name can be shorten to 'colorspace_duplicate'

tests/cpu/Config_tests.cpp
```diff
@@ -212,6 +212,39 @@ OCIO_ADD_TEST(Config, simple_config_with_duplicate)
                           "Key-value pair with key 'name' specified more than once. ");
 }
 
+OCIO_ADD_TEST(Config, simple_config_with_cdltransform_duplicate)
```
hodoulp, line 215: The unit test name can be shorten to 'cdltransform_duplicate'.

tests/cpu/Config_tests.cpp
```diff
@@ -212,6 +212,39 @@ OCIO_ADD_TEST(Config, simple_config_with_duplicate)
...
+        "    - !<View> {name: View1, colorspace: raw}\n"
+        "active_displays: []\n"
+        "active_views: []\n"
+        "colorspaces:\n"
+        "  - !<ColorSpace>\n"
+        "    name: raw\n"
+        "    to_reference: !<GroupTransform>\n"
+        "      direction: forward\n"
+        "      children:\n"
+        "        - !<CDLTransform>\n"
```
hodoulp, line 236: It could be simplified to:
` to_reference: !<CDLTransform> {slope: [1, 2, 1], slope: [1, 2, 1]}\n"`

PR #997, opened 2020-04-22
OCIO GOVERNANCE.md and TSC meeting template updates

michdolan:
Update with some new and overdue changes. Adding Mei Chu as SPI's TSC rep, and Sean Cooper as the ACES TAC Rep, as well as updating TSC member companies.


hodoulp:
I was able to restart the faulty build. I'm now waiting the result.

PR #1001, opened 2020-04-25
Added a blank line...

michdolan:
Just testing GH Actions in a PR. Do not merge...


michdolan:
I will submit a PR later which removed AZP, and I think I should move the "latest" and 'SonarCloud" jobs into their own workflow, so they don't clutter up PR checks. GH Actions can support any number of workflows, so good to take advantage of.

Issue #998, opened 2020-04-23
[CMAKE] FindPybind failure on Debian Buster


sobotka:
Platform: Debian Buster amd64

Summary: Build fails with:
```
PyOpenColorIO.h:51:8: error: 'polymorphic_type_hook' is not a class template
 struct polymorphic_type_hook<OCIO::Transform> 
```

Description: The FindPyBind is permitting the build to continue despite the PyBind11 library version being insufficient. As of now, the current version of PyBind11 reported on Debian Buster is:
```
Package: pybind11-dev                    
Version: 2.2.4-2
```
The required version minimum is:
```
# pybind11
# https://github.com/pybind/pybind11
set(pybind11_MIN_VERSION 2.4.3)
find_package(pybind11 ${pybind11_MIN_VERSION} REQUIRED)
```

michdolan:
@sobotka is this working properly in your environment now?

sobotka:
Need to give the Pleasures of Building another whirl...

sobotka:
So sorry Michael for not hammering on this sooner.

I've tested a clean checkout against default cmake finding, and it finds 2.7 on Debian buster, and cleanly compiles.

I manually flipped it over to 3.7, and it too cleanly compiles.

Tremendous work. Thanks so much.

michdolan:
No worries at all. Thanks for reporting the issue!

PR #1002, opened 2020-04-26
pybind11 CMake module refactor

michdolan:
This PR reworks the `Findpybind11.cmake` module to better handle version detection when possible, and warn when it is not. Three methods for locating pybind11 are now used, in this order:

1. If the user specified `-Dpybind11_DIRS=<path>` in the cmake command, try to use that first, and only search that path, ignoring typical system and cmake paths that would be searched otherwise.
2. Try to use a pybind11 CMake config, which should have been installed on the system with pybind11
3. If pybind11 was installed from PyPI using pip, get the path and version from a python process

I've tested these scenarios on my machine and now the pybind11 version number is either discovered, or it is assumed the user knows what they are doing if they override the pybind11 path, and warns them about the version assumption (there doesn't appear to be any version indication in the headers).

This PR should solve #998 


share/cmake/modules/Findpybind11.cmake
```diff
@@ -22,28 +23,88 @@ endif()
...
-            pybind11/pybind11.h
-        HINTS
-            ${_pybind11_SEARCH_DIRS}
-        PATH_SUFFIXES
-            include
-            pybind11/include
-    )
+        if(EXISTS "${pybind11_INCLUDE_DIR}")
+            # Assume user-provided version is OK.
+            # TODO: Can the pybind11 version be inspected from a header?
```
hodoulp, line 42: Yes. The code could parse the file `ext/dist/include/pybind11/detail/common.h` to find the version:
```
#define PYBIND11_VERSION_MAJOR 2
#define PYBIND11_VERSION_MINOR 4
#define PYBIND11_VERSION_PATCH 3
```

share/cmake/modules/Findpybind11.cmake
```diff
@@ -22,28 +23,88 @@ endif()
...
+            )
+        endif()
+    endif()
+
+    if(NOT DEFINED pybind11_VERSION)
+        # Search for a CMake config, which should have been installed with 
+        # pybind11, and will provide version information.
+        find_package(pybind11 ${pybind11_FIND_VERSION} CONFIG)
+
+        if(NOT pybind11_FOUND)
```
hodoulp, line 56: Even if an installed version is found, the version must still be validated.

michdolan: `find_package_handle_standard_args` will validate the version based on the `find_package` call that resolved to run this module.

share/cmake/modules/Findpybind11.cmake
michdolan: Good find. I should have looked in more of the headers! I will get this incorporated.

michdolan: Added.

PR #1005, opened 2020-04-28
Fix CMake OpenImageIO find module warning (Windows, macOS)

michdolan:
This PR resolves this CMake warning that appears on Windows and macOS.

```
CMake Warning (dev) at C:/Program Files/CMake/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (OIIO) does
  not match the name of the calling package (OpenImageIO).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  share/cmake/modules/FindOpenImageIO.cmake:22 (find_package_handle_standard_args)
  src/CMakeLists.txt:8 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```


lgritz:
Not sure what your minimum OIIO is these days, but since 2.1 (the current supported release), OIIO has generated a proper OpenImageIOConfig.cmake (and family) and make it so you don't need a Find module at all, and can just import the targets.

Also, I recommend bumping cmake to 3.12 minimum because that's when they really beefed up automatic use of Packagename_ROOT (either cmake variable or env variable) to do their magic in a wide range of find_foo() built in functions.

michdolan:
Thanks for the suggestions @lgritz . I'm doing some investigation into those changes, and will follow up with another PR.

Issue #980, opened 2020-04-01
Provide a set of common built-in transforms


doug-walker:
Currently all OCIO color spaces need to be provided using external LUT files (except for a few trivial possibilities that may be defined in the config, e.g. using an ExponentTransform). It would be useful to have some built-in transforms that OCIO could generate in memory and on demand. This would provide a number of advantages:

1)  It would be possible to define configs that don’t rely on external LUT files and hence would not break if these files are missing or unreadable from a given machine.
2)  Even where some LUTs are needed, not needing LUTs for the built-in transforms will make config packages smaller.
3)  It would make configs easier to create since authors would only have to create LUTs for transforms that are unique to their project and not have to worry about commonly used transforms.

Of course, there will be no requirement to use any of the built-in transforms. If the config author does not use them in the config then they won’t show up in application menus.  The proposal takes the approach of defining built-in transforms rather than built-in ColorSpaces to avoid name conflicts and to make it clear that it is up to the config author to "opt in" by using them in a ColorSpace definition.

Additional transform elements may be inserted by the config author if the reference space differs from what a given built-in transform expects.

The proposal is to implement all of the standard "building block" transforms used in modern color pipelines, so roughly this would be equivalent to the CTF collection that currently ships with Flame or Maya. This includes matrices for the common primary sets, linearization LUTs for all the main camera log spaces, interchange color spaces, video and digital cinema color spaces, etc.

Please note that this proposal does not add any new Ops, it simply provides a way to create transforms using the existing Ops.  So, just like with FileTransform, ultimately the BuiltinTransform results in the well-known existing Ops that are processed as usual.

Config Details

Here are some examples of how to refer to a built-in transform in a Config file. This would be used in from_reference or to_reference definitions, like FileTransform, ColorSpaceTransform, etc.:
```
   - !<BuiltInTransform> {name: CanonLog3_to_linear}
   - !<BuiltInTransform> {name: ST-2084_EOTF, direction: inverse}
   - !<BuiltInTransform> {name: Red_Log3G10_RedWideGamut_to_ACES2065-1}
```
The name argument would not be case-sensitive.  A function will be provided to iterate over all existing names.

zachlewis:
[...]

PR #1006, opened 2020-04-29
Adsk Contrib - Add Built-in Transform and ociomakeclf app

hodoulp:

The pull request mainly adds the "Built-in Transform" feature as explained by issue #980.

* For a config author the use of a built-in transform is like using a transform. For example `- !<BuiltinTransform> {style: ACEScct_to_ACES2065-1}` adds a transform to convert from `ACEScct` to `ACES2065-1`. Like any other transform the direction is supported.

* The implementation of the built-in transforms are implemented using existing transforms (i.e. built using Ops) so optimizations & finalizations are all done. For example, the  `ACEScct_to_ACES2065-1` is built [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/builtin_transform?expand=1#diff-9674d0f3d8abc42afd63b497346ac9a5R173) in the file src/OpenColorIO/transforms/builtins/ACES.cpp, and the implementation only uses a `LogTransform` and a `MatrixTransform`.

* Some camera built-in transforms are already available (but could be disabled with the cmake option `-DOCIO_ADD_EXTRA_BUILTINS=OFF`). For example, the `SONY_SLOG3-SGAMUT3_to_ACES2065-1` is built [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/master...autodesk-forks:adsk_contrib/builtin_transform?expand=1#diff-a78117aefe7fd2eeebea64d0f9065eaeR77) in the file src/OpenColorIO/transforms/builtins/SonyCameras.cpp.

* The new app `ociomakeclf` converts from any supported LUT file format to the ACES CLF file format and optionally adds conversions from/to ACES2065-1 to make it an LMT. Note that OCIO currently supports the Academy/ASC Common LUT Format v3 (refer to #709 ).
Some usages are:
  * `./src/apps/ociomakeclf/ociomakeclf ./testdata/lut1d_1.spi1d ./res.clf` converts from `spi1d` to the `CLF` file format.
  * `./src/apps/ociomakeclf/ociomakeclf --list` lists the ACES Color Space Conversion transforms (all of which convert to `ACES2065-1`) usable for building an ACES LMT.
  * `./src/apps/ociomakeclf/ociomakeclf ./testdata/discreet-3d-lut.3dl ./res.clf --csc ACEScct` creates a `CLF` file containing three blocks: the conversion from `ACES2065-1` to `ACEScct`, the input LUT file content and finally, the conversion from  `ACEScct` to `ACES-2065-1`. All the blocks could be comprised of one or several transforms.  The result is an ACES LMT file usable with an ACES Metadata File (AMF).

The pull request currently supports a starter set built-in transforms, including all of the ACES CSC transforms. From the public API, the class `BuiltinTransformRegistry` gets access to all the existing styles (i.e. built-in transform names) with a short description. 

This PR puts the foundation mechanism in place for built-in transforms and it should now be straight-forward to add more additional transform instances over time.  The initial set in this PR is intended as an example (which are also useful for building LMTs), not as the complete set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,12 @@ option(OCIO_USE_SSE "Specify whether to enable SSE CPU performance optimizations
 option(OCIO_INLINES_HIDDEN "Specify whether to build with -fvisibility-inlines-hidden" ${UNIX})
 
 ###############################################################################
+# Supplemental builtins
+
+# Add supplemental built-in transformations such as camera related one.
+option(OCIO_ADD_EXTRA_BUILTINS "Specify whether to add supplemental built-in transforms" ON)
+
+###############################################################################
 # Platform & Compiler dependent compilation flags
 
 set(PLATFORM_COMPILE_FLAGS "")
--- a/src/OpenColorIO/CMakeLists.txt
+++ b/src/OpenColorIO/CMakeLists.txt
@@ -143,6 +143,8 @@ if(OCIO_ADD_EXTRA_BUILTINS)
 	)
 
 	list(INSERT SOURCES 0 ${SOURCES_BUILTINS})
+else()
+	message(WARNING "Disabling supplemental built-in transforms removes built-in camera transforms.")
 endif()
 
 if(WIN32 AND BUILD_SHARED_LIBS)
--- a/src/OpenColorIO/CMakeLists.txt
+++ b/src/OpenColorIO/CMakeLists.txt
@@ -108,6 +108,11 @@ set(SOURCES
 	ScanlineHelper.cpp
 	Transform.cpp
 	transforms/AllocationTransform.cpp
+	transforms/builtins/ACES.cpp
+	transforms/builtins/BuiltinTransformRegistry.cpp
+	transforms/builtins/ColorMatrixHelpers.cpp
+	transforms/builtins/OpHelpers.cpp
+	transforms/BuiltinTransform.cpp
 	transforms/CDLTransform.cpp
 	transforms/ColorSpaceTransform.cpp
 	transforms/DisplayTransform.cpp
@@ -127,6 +132,18 @@ set(SOURCES
 	transforms/RangeTransform.cpp
 	ViewTransform.cpp
 )
+
+if(OCIO_ADD_EXTRA_BUILTINS)
+	set(SOURCES_BUILTINS
+		transforms/builtins/ArriCameras.cpp
+		transforms/builtins/CanonCameras.cpp
+		transforms/builtins/PanasonicCameras.cpp
+		transforms/builtins/RedCameras.cpp
+		transforms/builtins/SonyCameras.cpp
+	)
+
+	list(INSERT SOURCES 0 ${SOURCES_BUILTINS})
+endif()
 
 if(WIN32 AND BUILD_SHARED_LIBS)
 
@@ -188,6 +205,13 @@ if(OCIO_USE_SSE)
 	target_compile_definitions(OpenColorIO
 		PRIVATE
 			USE_SSE
+	)
+endif()
+
+if(OCIO_ADD_EXTRA_BUILTINS)
+	target_compile_definitions(OpenColorIO
+		PRIVATE
+			ADD_EXTRA_BUILTINS
 	)
 endif()
 
--- a/src/OpenColorIO/CMakeLists.txt
+++ b/src/OpenColorIO/CMakeLists.txt
@@ -144,7 +144,7 @@ if(OCIO_ADD_EXTRA_BUILTINS)
 
 	list(INSERT SOURCES 0 ${SOURCES_BUILTINS})
 else()
-	message(WARNING "Disabling supplemental built-in transforms removes built-in camera transforms.")
+	message(WARNING "Disabling supplemental built-in transforms removes all built-in camera transforms, limiting OCIO configuration compatibility.")
 endif()
 
 if(WIN32 AND BUILD_SHARED_LIBS)
--- a/src/OpenColorIO/transforms/builtins/ACES.cpp
+++ b/src/OpenColorIO/transforms/builtins/ACES.cpp
@@ -1,0 +1,296 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright Contributors to the OpenColorIO Project.
+
+
+#include <cmath>
+
+#include <OpenColorIO/OpenColorIO.h>
+
+#include "OpenEXR/half.h"
+#include "ops/log/LogOp.h"
+#include "ops/matrix/MatrixOp.h"
+#include "ops/range/RangeOp.h"
+#include "transforms/builtins/ACES.h"
+#include "transforms/builtins/BuiltinTransformRegistry.h"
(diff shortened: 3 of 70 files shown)


CMakeLists.txt
```diff
@@ -111,6 +111,12 @@ endif()
 option(OCIO_USE_SSE "Specify whether to enable SSE CPU performance optimizations" ON)
 option(OCIO_INLINES_HIDDEN "Specify whether to build with -fvisibility-inlines-hidden" ${UNIX})
 
+###############################################################################
+# Supplemental builtins
+
+# Add supplemental built-in transformations such as camera related one.
+option(OCIO_ADD_EXTRA_BUILTINS "Specify whether to add supplemental built-in transforms" ON)
```
michdolan, line 118: What are the benefits of making some transforms optional? This seems like it could be problematic in terms of config compatibility.

src/OpenColorIO/transforms/builtins/ACES.cpp
```diff
@@ -0,0 +1,296 @@
...
+{
+
+// Register all the ACES related built-in tranforms.
+void RegisterAll(BuiltinTransformRegistryImpl & registry) noexcept
+{
+    {
+        auto ACES_AP0_to_CIE_XYZ_D65_BFD_Functor = [](OpRcPtrVec & ops)
+        {
+            MatrixOpData::MatrixArrayPtr matrix
+                = build_conversion_matrix(ACES_AP0::primaries, CIE_XYZ_D65::primaries, ADAPTATION_BRADFORD);
```
michdolan, line 129: Possible candidate for public API. Static `MatrixTransform` constructor from primaries and CAT.

src/OpenColorIO/transforms/builtins/ColorMatrixHelpers.cpp
```diff
@@ -0,0 +1,260 @@
...
+static const Chromaticities red_xy(0.680,  0.320 );
+static const Chromaticities grn_xy(0.265,  0.690 );
+static const Chromaticities blu_xy(0.150,  0.060 );
+static const Chromaticities wht_xy(0.3127, 0.3290);
+
+const Primaries primaries(red_xy, grn_xy, blu_xy, wht_xy);
+}
+
+
+// Here are some notes on how one derives the color space conversion
```
michdolan, line 75: Great explanation!

src/bindings/python/PyBuiltinTransform.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace OCIO_NAMESPACE
+{
+
+void bindPyBuiltinTransform(py::module & m)
+{
+    py::class_<BuiltinTransform, 
+               BuiltinTransformRcPtr /* holder */, 
+               Transform /* base */>(m, "BuiltinTransform")
+        .def(py::init(&BuiltinTransform::Create))
```
michdolan, line 14: Can be done in a later PR, but would be good to have a convenience constructor like the other transforms.

src/bindings/python/PyBuiltinTransformRegistry.cpp
```diff
@@ -0,0 +1,45 @@
...
+    // Wrapper to preserve the BuiltinTransformRegistry singleton.
+    class Wrapper
+    {
+    public:
+        Wrapper() = default;
+        Wrapper(const Wrapper &) = delete;
+        Wrapper & operator=(const Wrapper &) = delete;
+        ~Wrapper() = default;
+
+        size_t getNumBuiltins() const noexcept
```
michdolan, line 21: We should make this interface more Pythonic. This could probably be boiled down a single tuple iterator. I can propose that in a later PR.

CMakeLists.txt
hodoulp: When we presented the feature at the last OCIO working group, some mentioned to avoid having camera related code (i.e. vendor specific code) in the core library. However, it added unnecessary complexity to handle 'normal' built-in transforms versus 'supplemental' built-in transforms (i.e. camera ones) when both are needed most of the time. So, we moved it in the core library with a compilation flag (i.e. could be removed on demand). But the compilation flag in ON by default.

src/OpenColorIO/transforms/builtins/ACES.cpp
hodoulp: Good point. But it should be revisited with the OCIO math library in mind.

src/bindings/python/PyBuiltinTransform.cpp
hodoulp: Good point. In a Later PR.

src/bindings/python/PyBuiltinTransformRegistry.cpp
hodoulp: Good point. In a Later PR.

src/OpenColorIO/transforms/builtins/ColorMatrixHelpers.cpp
hodoulp: @doug-walker did an amazing work on the algorithms & explanations.

src/OpenColorIO/transforms/builtins/ACES.cpp
michdolan: Yes, That makes sense.

src/OpenColorIO/transforms/builtins/ColorMatrixHelpers.cpp
michdolan: Thanks @doug-walker !

CMakeLists.txt
michdolan: The option being default _should_ prevent most issues, I just fear that someone will turn it off and suddenly a config will lack support in that application. What would you think about including a warning if it is off, informing the builder that will introduce potential compatibility issues with some configs?

hodoulp: Good idea.

hodoulp: Done

hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can now be merged this week unless there are new comments.

src/OpenColorIO/CMakeLists.txt
```diff
@@ -128,6 +133,20 @@ set(SOURCES
...
+		transforms/builtins/ArriCameras.cpp
+		transforms/builtins/CanonCameras.cpp
+		transforms/builtins/PanasonicCameras.cpp
+		transforms/builtins/RedCameras.cpp
+		transforms/builtins/SonyCameras.cpp
+	)
+
+	list(INSERT SOURCES 0 ${SOURCES_BUILTINS})
+else()
+	message(WARNING "Disabling supplemental built-in transforms removes built-in camera transforms.")
```
michdolan, line 147: Looks good. I think this should also include a note about OCIO config compatibility to be clear. Something like:

`Disabling supplemental built-in transforms removes built-in camera transforms, limiting OCIO configuration compatibility.`

hodoulp: Agree. Done

PR #1009, opened 2020-05-04
GHA Analysis pybind11 Fix

michdolan:
This PR resolves the error that has been failing the Analysis workflow the last few nights. The failure was due to using a CMake config in the find module where possible, which could define an imported target that I was already defining. Now the find module only defines the target if it doesn't yet exist.

I have also removed the `paths-ignore` option in the CI workflow triggers. Ignore paths currently doesn't work with repos that have required checks setup for branch protection. More on that here:
https://github.community/t5/GitHub-Actions/Feature-request-conditional-required-checks/td-p/36938

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -7,23 +7,6 @@
 name: Analysis
 
 on:
-  push:
-    # Versioned branches and tags are ignored for OCIO <= 1.x.x
-    branches-ignore:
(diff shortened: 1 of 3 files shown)


michdolan:
Successful test `Analysis` run with these changes:
https://github.com/michdolan/OpenColorIO/actions/runs/95517162

hodoulp:
A dedicated PR only containing the 'macos' fix is coming.

PR #1011, opened 2020-05-05
Adsk Contrib - Enhance CLF tests

doug-walker:
The goal is to use the CLF unit tests in OCIO as the collection of CLF files to be used for the ACES CLF implementation working group.

I've moved the intended files into a clf sub-directory to collect them as a group.  I added a further sub-directory for the files that have illegal syntax.

I also refactored FileFormatCTF_tests.cpp to group together similar tests.  In the process I added some new tests and cleaned up some old ones.

I also addressed a few minor issues with the CLF parsing.


doug-walker:
This PR has been available for 10 days.  We would like to merge it early next week if there are no comments.

PR #1015, opened 2020-05-13
Adsk Contrib: Fix View serialization in Config.

BernardLefebvre:
View transform of a view was not saved in config.
Add a test for View serialization.


hodoulp:
@michdolan That's a critical bug in master we would like to merge as soon as possible. Thanks.

PR #1027, opened 2020-06-04
Python unit test module LookTest update

meimchu:
Result of the LookTest module update:
`
test_constructor_with_keyword (LookTest.LookTest) ... ok
test_constructor_without_keyword (LookTest.LookTest) ... ok
test_constructor_without_parameter (LookTest.LookTest) ... ok
test_constructor_wrong_parameter_type (LookTest.LookTest) ... ok
test_description (LookTest.LookTest) ... Segmentation fault: 11
`

The wrong type tests in `setName()`, `setProcessSpace()`, and `setDescription()` will trigger a segmentation fault. That seems like a really harsh rejection and perhaps it would be nice to have it throw exceptions instead of a crash? The rest of the test are fine and should pass without issues.

### DIFF
--- a/src/bindings/python/PyLook.cpp
+++ b/src/bindings/python/PyLook.cpp
@@ -17,7 +17,7 @@ void bindPyLook(py::module & m)
                          const std::string & processSpace,
                          const TransformRcPtr & transform,
                          const TransformRcPtr & inverseTransform,
-                         const std::string & description) 
+                         const std::string & description)
             {
                 LookRcPtr p = Look::Create();
                 if (!name.empty())         { p->setName(name.c_str()); }
@@ -26,24 +26,23 @@ void bindPyLook(py::module & m)
                 if (inverseTransform)      { p->setInverseTransform(inverseTransform); }
                 if (!description.empty())  { p->setDescription(description.c_str()); }
                 return p;
-            }), 
+            }),
              "name"_a = DEFAULT->getName(),
              "processSpace"_a = DEFAULT->getProcessSpace(),
              "transform"_a = DEFAULT->getTransform(),
              "inverseTransform"_a = DEFAULT->getInverseTransform(),
-             "description"_a = DEFAULT->getDescription())  
+             "description"_a = DEFAULT->getDescription())
 
         .def("getName", &Look::getName)
-        .def("setName", &Look::setName, "name"_a)
+        .def("setName", &Look::setName, "name"_a.none(false))
         .def("getProcessSpace", &Look::getProcessSpace)
-        .def("setProcessSpace", &Look::setProcessSpace, "processSpace"_a)
+        .def("setProcessSpace", &Look::setProcessSpace, "processSpace"_a.none(false))
         .def("getTransform", &Look::getTransform)
         .def("setTransform", &Look::setTransform, "transform"_a)
         .def("getInverseTransform", &Look::getInverseTransform)
         .def("setInverseTransform", &Look::setInverseTransform, "transform"_a)
         .def("getDescription", &Look::getDescription)
-        .def("setDescription", &Look::setDescription, "description"_a);
-
+        .def("setDescription", &Look::setDescription, "description"_a.none(false));
     defStr(cls);
 }
 
--- a/src/bindings/python/PyLook.cpp
+++ b/src/bindings/python/PyLook.cpp
@@ -43,6 +43,7 @@ void bindPyLook(py::module & m)
         .def("setInverseTransform", &Look::setInverseTransform, "transform"_a)
         .def("getDescription", &Look::getDescription)
         .def("setDescription", &Look::setDescription, "description"_a.none(false));
+
     defStr(cls);
 }
 
--- a/tests/python/LookTest.py
+++ b/tests/python/LookTest.py
@@ -33,6 +33,7 @@ class LookTest(unittest.TestCase):
         # Wrong type tests.
         with self.assertRaises(TypeError):
             self.look.setName(None)
+            self.look.setName(1)
 
     def test_process(self):
         """
@@ -49,6 +50,7 @@ class LookTest(unittest.TestCase):
         # Wrong type tests.
         with self.assertRaises(TypeError):
             self.look.setProcessSpace(None)
+            self.look.setProcessSpace(1)
 
     def test_description(self):
         """
@@ -65,6 +67,7 @@ class LookTest(unittest.TestCase):
         # Wrong type tests.
         with self.assertRaises(TypeError):
             self.look.setDescription(None)
+            self.look.setDescription(1)
 
     def test_transform(self):
         """
@@ -84,6 +87,7 @@ class LookTest(unittest.TestCase):
         # Wrong type tests.
         with self.assertRaises(TypeError):
             self.look.setTransform(OCIO.ALLOCATION_UNIFORM)
+            self.look.setTransform(1)
 
     def test_inverse_transform(self):
         """
@@ -103,6 +107,7 @@ class LookTest(unittest.TestCase):
         # Wrong type tests.
         with self.assertRaises(TypeError):
             self.look.setInverseTransform(OCIO.ALLOCATION_UNIFORM)
+            self.look.setInverseTransform(1)
 
     def test_constructor_with_keyword(self):
         """
@@ -178,3 +183,4 @@ class LookTest(unittest.TestCase):
 
         with self.assertRaises(TypeError):
             look = OCIO.Look(None)
+            look = OCIO.Look(1)
--- a/tests/python/LookTest.py
+++ b/tests/python/LookTest.py
@@ -1,28 +1,180 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright Contributors to the OpenColorIO Project.
 
-import unittest, os, sys
+import unittest
+import os
+import sys
+
 import PyOpenColorIO as OCIO
+from UnitTestUtils import TEST_NAMES, TEST_DESCS
+
 
 class LookTest(unittest.TestCase):
-    
-    def test_interface(self):
-        lk = OCIO.Look()
-        lk.setName("coollook")
-        self.assertEqual("coollook", lk.getName())
-        lk.setProcessSpace("somespace")
-        self.assertEqual("somespace", lk.getProcessSpace())
-        lk.setDescription("this is a test")
-        self.assertEqual("this is a test", lk.getDescription())
-        et = OCIO.ExponentTransform()
-        et.setValue([0.1, 0.2, 0.3, 0.4])
-        lk.setTransform(et)
-        oet = lk.getTransform()
-        vals = oet.getValue()
-        self.assertAlmostEqual(0.2, vals[1], delta=1e-8)
-        iet = OCIO.ExponentTransform()
-        iet.setValue([-0.1, -0.2, -0.3, -0.4])
-        lk.setInverseTransform(iet)
-        ioet = lk.getInverseTransform()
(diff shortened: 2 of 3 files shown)


michdolan:
If it's crashing that's a bug. I'll take a look at the bindings.

michdolan:
Found the cause of the segfault. For certain argument types, pybind11 will convert `None` to `nullptr`, which OCIO is not happy about. The fix is to explicitly tell pybind11 when an argument can't support this implicit `None` conversion, as documented here:
https://pybind11.readthedocs.io/en/stable/advanced/functions.html#allow-prohibiting-none-arguments

I've tested this in for Look and BuiltinTransform (which had the same issue in my tests), and it fixes it. Following the fix, these calls raise a `TypeError` instead. You can fix this for Look in this PR. Just make the following change to PyLook.cpp:

Change FROM:

```
        .def("getName", &Look::getName)
        .def("setName", &Look::setName, "name"_a)
        .def("getProcessSpace", &Look::getProcessSpace)
        .def("setProcessSpace", &Look::setProcessSpace, "processSpace"_a)
        .def("getTransform", &Look::getTransform)
        .def("setTransform", &Look::setTransform, "transform"_a)
        .def("getInverseTransform", &Look::getInverseTransform)
        .def("setInverseTransform", &Look::setInverseTransform, "transform"_a)
        .def("getDescription", &Look::getDescription)
        .def("setDescription", &Look::setDescription, "description"_a);
```

TO (note the 3 added `.none(false)`):

```
        .def("getName", &Look::getName)
        .def("setName", &Look::setName, "name"_a.none(false))
        .def("getProcessSpace", &Look::getProcessSpace)
        .def("setProcessSpace", &Look::setProcessSpace, "processSpace"_a.none(false))
        .def("getTransform", &Look::getTransform)
        .def("setTransform", &Look::setTransform, "transform"_a)
        .def("getInverseTransform", &Look::getInverseTransform)
        .def("setInverseTransform", &Look::setInverseTransform, "transform"_a)
        .def("getDescription", &Look::getDescription)
        .def("setDescription", &Look::setDescription, "description"_a.none(false));
```

tests/python/LookTest.py
```diff
@@ -84,6 +87,7 @@ def test_transform(self):
         # Wrong type tests.
         with self.assertRaises(TypeError):
             self.look.setTransform(OCIO.ALLOCATION_UNIFORM)
+            self.look.setTransform(1)
```
michdolan, line 90: These added second lines under `assertRaises` won't ever be executed, since the previous line will raise an exception. Applies to all uses in this module. Perhaps instead do:

```
for invalid in (..., ...):
    with self.assertRaises(TypeError):
        self.look.setSomething(invalid)
```

src/bindings/python/PyLook.cpp
```diff
@@ -26,24 +26,23 @@ void bindPyLook(py::module & m)
...
         .def("getProcessSpace", &Look::getProcessSpace)
-        .def("setProcessSpace", &Look::setProcessSpace, "processSpace"_a)
+        .def("setProcessSpace", &Look::setProcessSpace, "processSpace"_a.none(false))
         .def("getTransform", &Look::getTransform)
         .def("setTransform", &Look::setTransform, "transform"_a)
         .def("getInverseTransform", &Look::getInverseTransform)
         .def("setInverseTransform", &Look::setInverseTransform, "transform"_a)
         .def("getDescription", &Look::getDescription)
-        .def("setDescription", &Look::setDescription, "description"_a);
-
```
michdolan, line 46: This blank line should be preserved for readability

meimchu: Resolved in b8a3feb.

tests/python/LookTest.py
meimchu: Resolved in b8a3feb.

tests/python/LookTest.py
```diff
@@ -1,28 +1,186 @@
...
+    def test_inverse_transform(self):
+        """
+        Test the setInverseTransform() and getInverseTransform() methods.
+        """
+
+        # Default initialized inverse transform value is None
+        self.assertIsNone(self.look.getInverseTransform())
+
+        exp_tr = OCIO.ExponentTransform()
+        exp_values = [-0.1, -0.2, -0.3, -0.4]
```
michdolan, line 101: [minor] the inverse exponent values of those used in `test_transform` would be 
`[1.0 / v for v in [0.1, 0.2, 0.3, 0.4]]`

tests/python/LookTest.py
```diff
@@ -1,28 +1,186 @@
...
+                          processSpace='somespace',
+                          transform=exp_tr)
+
+        self.assertEqual(look2.getName(), 'coollook')
+        self.assertEqual(look2.getProcessSpace(), 'somespace')
+        self.assertIsInstance(look2.getTransform(), type(exp_tr2))
+        self.assertIsInstance(look2.getInverseTransform(), type(inv_exp_tr2))
+        self.assertEqual(look2.getDescription(), 'this is a test')
+
+    def test_constructor_without_keyword(self):
```
michdolan, line 147: Recommend using `test_constructor_positional` or `test_constructor_with_positional` instead of `test_constructor_without_keyword`.

tests/python/LookTest.py
```diff
@@ -1,28 +1,186 @@
...
+                         inv_exp_tr,
+                         'this is a test')
+
+        self.assertEqual(look.getName(), 'coollook')
+        self.assertEqual(look.getProcessSpace(), 'somespace')
+        self.assertIsInstance(look.getTransform(), type(exp_tr))
+        self.assertIsInstance(look.getInverseTransform(), type(inv_exp_tr))
+        self.assertEqual(look.getDescription(), 'this is a test')
+
+    def test_constructor_without_parameter(self):
```
michdolan, line 166: I don't think this test is necessary. The granular tests (i.e. `test_name`) have already tested all the values from the default constructor.

tests/python/LookTest.py
```diff
@@ -1,28 +1,186 @@
...
+        """
+
+        # Default initialized transform value is None
+        self.assertIsNone(self.look.getTransform())
+
+        exp_tr = OCIO.ExponentTransform()
+        exp_values = [0.1, 0.2, 0.3, 0.4]
+        exp_tr.setValue(exp_values)
+        self.look.setTransform(exp_tr)
+        oet = self.look.getTransform()
```
michdolan, line 84: [minor] Maybe use `out_exp_tr` instead of `oet` for clarity? Could do this in the inverse test method too.

tests/python/LookTest.py
```diff
@@ -1,28 +1,186 @@
...
+        for name in TEST_NAMES:
+            self.look.setName(name)
+            self.assertEqual(name, self.look.getName())
+
+        # Wrong type tests.
+        for invalid in (None, 1):
+            with self.assertRaises(TypeError):
+                self.look.setName(invalid)
+
+    def test_process(self):
```
michdolan, line 38: Recommend `test_process_space` instead of `test_process`

meimchu: Resolved in 2b67585.

tests/python/LookTest.py
meimchu: Resolved in 2b67585.

tests/python/LookTest.py
meimchu: Resolved in 2b67585.

tests/python/LookTest.py
meimchu: Resolved in 2b67585.

tests/python/LookTest.py
meimchu: Resolved in 2b67585.

meimchu:
Just a quick follow up question. If I were to `print(OCIO.Look())`, this is the result I get:
`<Look name=, processSpace=>`
Should it be more like `<Look name=, processSpace=, transform=None, inverseTransform=None, description=>`

Would it more clear to list them all in the right order for ease of constructor readability?

michdolan:
The `print()` result echos the C++ lib's `ostream operator<<` result,

doug-walker:
It looks like Look.cpp operator<< is unchanged from v1.  As is the case with many of the op<<, some members are not printed if they have empty/default values.  Important members (name and process space in this case) are printed, however, even if empty.  The names seem to be sorted based on importance rather than alphabetically.  Doesn't the order already match the python constructor?  What order do you think would be better?

I'm noticing that we added a description member to Look but we did not update the op<< to print it, so that is a minor bug to be fixed.

meimchu:
Per discussion today in the meeting, I can understand that the string representation of Look() object is just showing the more important bits. Just to elaborate more on what I was referring to earlier:

```
>>> print(OCIO.Look())
<Look name=, processSpace=>
```
By default, it pretty much has nothing. So now I purposely pass in a wrong type parameter so that I know exactly what Look() object parameters there are. I was only wondering if there is a more elegant way to display what parameters are available in this object otherwise (besides using `print dir(lk)`)?

```
>>> lk = OCIO.Look(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. PyOpenColorIO.Look()
    2. PyOpenColorIO.Look(name: str = '', processSpace: str = '', transform: PyOpenColorIO.Transform = None, inverseTransform: PyOpenColorIO.Transform = None, description: str = '')
```

```
>>> lk = OCIO.Look(description="Hello world")
>>> print(lk)
<Look name=, processSpace=>
```
I put in a description parameter and it doesn't get displayed.

But ultimately, I get the desire to only show the more important stuff per today's meeting discussion!

tests/python/OpenColorIOTestSuite.py
```diff
@@ -34,11 +34,11 @@
 import BuiltinTransformRegistryTest
 import BuiltinTransformTest
 import CDLTransformTest
+import LookTest
```
BernardLefebvre, line 37: it could be nice to sort these in alphabetical order

tests/python/OpenColorIOTestSuite.py
```diff
@@ -59,12 +59,12 @@ def suite():
     suite.addTest(loader.loadTestsFromModule(BuiltinTransformRegistryTest))
     suite.addTest(loader.loadTestsFromModule(BuiltinTransformTest))
     suite.addTest(loader.loadTestsFromModule(CDLTransformTest))
+    suite.addTest(loader.loadTestsFromModule(LookTest))
```
BernardLefebvre, line 62: alphabetical order?

doug-walker:
@meimchu, I think you raise a good point that it would be nice for python users to easily see what are the available members to set.  But at the same time, we want to keep the printed values compact and not print seldom used members that are at their default value.

Perhaps the best solution is for the doc string to list the available members and what they do?

tests/python/LookTest.py
```diff
@@ -1,28 +1,173 @@
...
+        self.look.setInverseTransform(exp_tr)
+        inv_oet = self.look.getInverseTransform()
+        self.assertListEqual(inv_oet.getValue(), inv_exp_values)
+
+        # Wrong type tests.
+        for invalid in (OCIO.ALLOCATION_UNIFORM, 1):
+            with self.assertRaises(TypeError):
+                self.look.setInverseTransform(invalid)
+
+    def test_constructor_with_positional(self):
```
michdolan, line 112: This should be `test_constructor_with_keyword` or `test_constructor_keyword`. The next method down should be `test_constructor_with_positional` or `test_constructor_positional`. Positional implies no keyword args are used, and that you are relying on the argument order defined in the function.

PR #1036, opened 2020-06-16
Initial update to Transforms python unit tests:

meimchu:
Breaking them down to individual tests starting with:
* AllocationTransformTest
  * `getNumVars()` method is no longer available. I know we have moved to using looping through `__members__` in previous tests. Just want to confirm that to be the case here as well.
  * `print(ocio.AllocationTransform())` returns error.
* ColorSpaceTransformTest
  * `setSrc()` and `setDst()` wrong type tests will crash python
  * `print(ocio.ColorSpaceTransform())` returns `<ColorSpaceTransform direction=forward, src=, dst=>` which I think is confusing because the order is different from its actual expected parameter order of `src`, `dst` and `dir`. The string representation of the object says `direction` is also different from the actual expected parameter of `dir`. I'm wondering if they could be named the same? I will admit I have yet to test if `ocio.ColorSpaceTransform(direction=ocio.TRANSFORM_DIR_FORWARD)` works. If it does, we can disregard this.


michdolan:
Thanks Mei! Responding to your observations/questions above:

**AllocationTransform:**
`getNumVars()` is intentionally omitted. In the C++ API it is present to allocate the proper array size for fetching the vars. In Python you just call `getVars()` and it returns a list of vars. If you want to know the size, you can call `len(tr.get_vars())`. There are other places where the Python API differs from C++ for similar methods.

The `print(tr)` and `str(tr)` failure for AllocationTransform is harder to track down. I'm looking into it.

**ColorSpaceTransform:**
For `setSrc()` and `setDst()` you can change in C++:
```
        ...
        .def("setSrc", &ColorSpaceTransform::setSrc, "src"_a.none(false))
        ...
        .def("setDst", &ColorSpaceTransform::setDst, "dst"_a.none(false));
```
That should fix the crashes. You will find more cases like this. Feel free to correct them in the bindings as needed.

Regarding `str(tr)`: The string representation of a transform won't necessarily match it's Python constructor parameter order or naming. Parameter naming usually matches the C++ API:
`virtual void setDirection(TransformDirection dir)`, but prints to match the serialized config parameter (`direction` in this case). These strings can be refined over time on the C++ side.

michdolan:
Thinking on the keyword arg naming more, I think you have a good point about matching keyword arg names to the class attribute (i.e. `direction` instead of `dir`) since the constructor is the one case where a developer would be calling out a specific attribute. If we do that, there are several other cases throughout the bindings where that will need to happen too, but can be done in one sweeping change. I'll take a look into it.

meimchu:
- ExponentTransformTest

  - Constructor parameter is `vec4` instead of `value` or something else?

michdolan:
> * ExponentTransformTest
>   
>   * Constructor parameter is `vec4` instead of `value` or something else?

See #1042 which updates all constructor parameter names

meimchu:
- FileTransformTest

  - No CDLStyle parameter in constructor? I took a look at #1042 and didn't notice it added there either. Just curious.

src/bindings/python/PyColorSpaceTransform.cpp
```diff
@@ -29,13 +29,13 @@ void bindPyColorSpaceTransform(py::module & m)
             }), 
              "src"_a        = DEFAULT->getSrc(), 
              "dst"_a        = DEFAULT->getDst(),
-             "dir"_a        = DEFAULT->getDirection(),
+             "direction"_a        = DEFAULT->getDirection(),
```
michdolan, line 32:
```suggestion
             "direction"_a  = DEFAULT->getDirection(),
```

tests/python/ExponentTransformTest.py
```diff
@@ -0,0 +1,153 @@
...
+    TEST_NEGATIVE_STYLE = OCIO.NEGATIVE_MIRROR
+    TEST_DIRECTION = OCIO.TRANSFORM_DIR_INVERSE
+
+    def setUp(self):
+        self.exp_tr = OCIO.ExponentTransform()
+
+    def tearDown(self):
+        self.exp_tr = None
+
+    def test_direction(self):
```
michdolan, line 21: To avoid repeating this method in every transform test, you may want to either:
- Create a Transform `TestCase` base class with this and any other common tests, which child transform classes can then just inherit without explicit implementation.
- Create a utility function to test direction and call it from each transform class.

tests/python/ExponentWithLinearTransformTest.py
```diff
@@ -0,0 +1,196 @@
...
+        Test the setNegativeStyle() and getNegativeStyle() methods.
+        """
+
+        # Default initialized negative style is clamp.
+        self.assertEqual(self.exp_tr.getNegativeStyle(), OCIO.NEGATIVE_LINEAR)
+
+        for negative_style in OCIO.NegativeStyle.__members__.values():
+            # These negative extrapolations are not valid for
+            # MonCurve exponent style.
+            ExceptionNegatives = [OCIO.NEGATIVE_CLAMP, OCIO.NEGATIVE_PASS_THRU]
```
michdolan, line 101: Recommend using lower case ``exception_negatives`` to avoid confusion as a type.

tests/python/FileTransformTest.py
```diff
@@ -0,0 +1,207 @@
...
+        for invalid in (None, 1):
+            with self.assertRaises(TypeError):
+                self.file_tr.setDirection(invalid)
+
+    def test_getformats(self):
+        """
+        Test the getFormats() method.
+        """
+
+        format_iterator = self.file_tr.getFormats()
```
michdolan, line 108: ``getFormats`` is static, so would be better to test like:
```suggestion
        format_iterator = OCIO.FileTransform.getFormats()
```

tests/python/ExponentTransformTest.py
michdolan: Note: This could be done in a later PR.

meimchu: I was under the impression that we want to avoid having the individual tests depend on something else, as such, I did not consider creating a base class or utility function. Given your suggestion, I am happy to look into one of those options to reduce code duplications. Thank you for your suggestion.

michdolan: I think this case justifies it to avoid having this same method across all the transform tests.

src/bindings/python/PyColorSpaceTransform.cpp
meimchu: Addressed in 2d14c6c.

tests/python/ExponentWithLinearTransformTest.py
meimchu: Addressed in 2d14c6c.

tests/python/FileTransformTest.py
meimchu: Addressed in 2d14c6c.

meimchu:
**ExposureContrastTransformTest**
```
======================================================================
FAIL: test_constructor_with_keywords (ExposureContrastTransformTest.ExposureContrastTransformTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/meichu/CodeProjects/OpenColorIO/tests/python/ExposureContrastTransformTest.py", line 216, in test_constructor_with_keywords
    self.assertEqual(exp_tr.getStyle(), self.TEST_STYLE)
AssertionError: ExposureContrastStyle.EXPOSURE_CONTRAST_LINEAR != ExposureContrastStyle.EXPOSURE_CONTRAST_LOGARITHMIC

======================================================================
FAIL: test_constructor_with_keywords (ExposureContrastTransformTest.ExposureContrastTransformTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/meichu/CodeProjects/OpenColorIO/tests/python/ExposureContrastTransformTest.py", line 247, in test_constructor_with_keywords
    self.TEST_LOGEXPOSURESTEP)
AssertionError: 0.088 != 0.1
```
I seem to have issue using constructor to set style to `OCIO.EXPOSURE_CONTRAST_LOGARITHMIC`. If I comment out that setting style in constructor, then I get the log exposure step error. I am not as familiar with this particular transform. I suspect perhaps those parameters work in conjunctions and so if one is not set, the other will also not set as well?

tests/python/ExponentTransformTest.py
meimchu: Also thinking now about whether to also include validate() of direction in this utility/base class for testing transform direction? The assumption is that transform direction must be forward or inverse and not unknown. Therefore, I think it makes sense to include a validate test every time transform direction is tested. So it would make sense to bundle those two together in a utility/base class, right?

michdolan:
> No CDLStyle parameter in constructor? I took a look at #1042 and didn't notice it added there either. Just curious.

Opened issue #1078

tests/python/ExponentTransformTest.py
michdolan: That sounds fine to me.

michdolan:
For the CI error, it's an issue with the binding. The constructor never actually sets style in `ExposureContrastTransform`. Would you mind adding it to the binding?
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/src/bindings/python/PyExposureContrastTransform.cpp
```
ExposureContrastTransformRcPtr p = ExposureContrastTransform::Create();
p->setStyle(style); // <-- Add this line
p->setExposure(exposure);
```
Another great example of why this test suite is so useful!

meimchu:
I have saw #1121 that could lead to a change in the default parameter for FileTransform. I have already done the unit test for FileTransform which noted the initial value as well. Should that default parameter changes, I could make the change now or wait.

hodoulp:
Thanks @meimchu but we are currently thinking about the best approach. Stay tuned :)

If that's a blocker you can change the interpolation default value to 'default' (instead of DEFAULT->getInterpolation()) and add a todo as a reminder.

meimchu:
I would be perfectly ok with some good code review and then closing this PR. I can wrap up this PR by addressing whatever review notes necessary. I know an outstanding item from this is about creating a base unit test for direction, but that can wait till the next PR.

tests/python/ExponentWithLinearTransformTest.py
```diff
@@ -0,0 +1,196 @@
...
+        Test the setNegativeStyle() and getNegativeStyle() methods.
+        """
+
+        # Default initialized negative style is clamp.
+        self.assertEqual(self.exp_tr.getNegativeStyle(), OCIO.NEGATIVE_LINEAR)
+
+        for negative_style in OCIO.NegativeStyle.__members__.values():
+            # These negative extrapolations are not valid for
+            # MonCurve exponent style.
+            exception_negatives = [OCIO.NEGATIVE_CLAMP, OCIO.NEGATIVE_PASS_THRU]
```
michdolan, line 101: This could be defined outside the loop

tests/python/ExposureContrastTransformTest.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # Default initialized gamma is 1.
+        self.assertEqual(self.exp_tr.getGamma(), 1)
+
+        for gamma in self.TEST_VALUES:
+            self.exp_tr.setGamma(gamma)
+            self.assertEqual(self.exp_tr.getGamma(), gamma)
+
+        # Wrong type tests.
+        for invalid in (None, 'hello', [1, 2, 3]):
```
michdolan, line 108: Could define this list of invalid values once and reference it throughout the code.

tests/python/FileTransformTest.py
```diff
@@ -0,0 +1,207 @@
...
+from UnitTestUtils import TEST_NAMES
+
+
+class FileTransformTest(unittest.TestCase):
+    TEST_DIRECTION = OCIO.TRANSFORM_DIR_FORWARD
+    TEST_INTERPOLATION = OCIO.Interpolation.INTERP_LINEAR
+    TEST_ID = 'sample file'
+    TEST_SRC = 'foo'
+    TEST_DST = 'bar'
+    DEFAULT_FORMATS = [('flame', '3dl'),
```
michdolan, line 16: Recommend not hard coding these value (`DEFAULT_FORMATS`)s in the Python test.

tests/python/ExposureContrastTransformTest.py
michdolan: This is probably true of most of the test modules.

tests/python/FileTransformTest.py
meimchu: If you don't mind elaborating on what you think would be a better way to check the default `getFormats()` function. I created that list based off of a loop and so I just want to ensure future `getFormats()` call equal that.

michdolan: Yeah I think you're right. I was hesitating to add an additional place where these are hard coded since they all must be maintained, but this is probably the most useful test regardless.

tests/python/ExponentWithLinearTransformTest.py
meimchu: Resolved with c60f458.

tests/python/ExposureContrastTransformTest.py
meimchu: Resolved with c60f458.

michdolan (approved):

hodoulp (approved):
Great job. Thanks @meimchu.

In following pull request(s) it would be great to add some 'serialization' and 'apply' validations.

Issue #1032, opened 2020-06-06
Allow applications to filter views based on color space


doug-walker:
OpenColorIO allows defining multiple views for each display.  In many cases, certain views will be intended for use with certain types of color spaces.  For example, a Raw view may be intended for data color spaces and a Film view may be intended for scene-linear or log color spaces.  However, there is presently no easy way for applications to filter the list of views based on the color space of the image being shown in a viewport.  The application typically must present all the views and require the user to manually switch them.  This can become tedious when switching between images in a variety of color spaces (e.g. switching between scene-linear images, video images, and alpha channels).  

Ideally, it would be possible for an application to know which view would be preferred for each color space in order to minimize the amount of manual view switching users must do and provide a viewport result which looks correct by default.  It would also help apps be able to automatically generate plausible thumbnails and preview images.

This proposal is to allow config authors to provide additional information that would allow applications to filter the list of views based on the color space being viewed.  Specifically it would allow a config to contain a viewing_rules section analogous to the file_rules section that was added in a recent PR.  A view would then be allowed to refer to one of these viewing rules.  Here is an example:

```
viewing_rules:
    - !<Rule> {name: video-spaces, colorspaces: [sRGB, Rec709]}
    - !<Rule> {name: data-spaces, colorspaces: [alpha, normals]}

displays:
  sRGB:
    - !<View> {name: Video, view_transform: colorimetry, display_colorspace: sRGB, rule: video-spaces}
    - !<View> {name: Raw, colorspace: nc10, rule: data-spaces}

```
This is helpful for situations where a given view is intended for use with just a few specific color spaces.  However in other situations, it would be helpful to be able to define rules to be used with a broader set of color spaces.

In issue #1031, it is proposed to add an "encoding" attribute to color spaces to allow grouping color spaces into groups such as "scene-linear", "log", "sdr-video", and "data".  This proposal therefore provisions for being able to define a rule based on the encoding attribute of a color space rather than a set of named color spaces.  For example:

```
viewing_rules:
    - !<Rule> {name: scene-linear-or-log, encodings: [scene-linear, log]}

displays:
  sRGB:
    - !<View> {name: ACES, view_transform: ACES-sdr-video, display_colorspace: sRGB, rule: scene-linear-or-log}

```
The colorspaces and encodings attributes may contain a single value or a list of values.  It will be illegal for a rule to define both a list of colorspaces and encodings simultaneously.  

[...]

Issue #1031, opened 2020-06-06
Add a color space property to describe the encoding


doug-walker:
The ColorSpace class in OpenColorIO currently has a number of descriptive properties such as bitdepth, equalitygroup, family, and isdata.

In addition, it would be very helpful for applications to be able to know the basic type of encoding of a color space.  For example, it is well known that the performance of various types of image processing algorithms varies based on the type of encoding.  Applying a spatial filter to a scene-linear image gives a different subjective result than if applied to the same image encoded in a log color space.  Likewise certain algorithms such as keying or tracking may assume that the color encoding is roughly perceptually uniform and thus may have difficulties with scene-linear images.

At first glance, the family attribute seemed like a potential way to specify this information.  In the OpenColorIO user guide, family is defined as "Allows for logical grouping of colorspaces within a UI" and the examples are "log" and "display".  

However, looking at example configs such as ACES, the family attribute is used to group color spaces by purpose and vendor, for example "Input/ARRI".  So the emphasis is on organizing the UI.  Furthermore, in a recent PR, the ColorSpaceHelpers classes were added and have a feature to facilitate building hierarchical menus for application UIs based on the family attribute.  

So it has become clear that the main benefit of the family is for applications to better organize their UI.  Furthermore, it is clear that the vocabulary of the family attribute is unconstrained and based on very config-specific details.  The family attribute is thus not well-suited to carry the color space encoding information.

Therefore this proposal is to add an "encoding" attribute to the ColorSpace class, including setEncoding and getEncoding methods (similar to family).  This would also be supported in the YAML serialization of a color space.  The attribute would be purely informative and would not affect the processing provided by OpenColorIO.

Furthermore, the proposal is that the attribute have (unlike family) a specifically defined terminology.  Similar to roles, there will be nothing to prevent config authors from using other strings.  However, having a clear set of predefined values for the common cases is what will make this attribute useful to application developers.

The proposed values and definitions are:

"scene-linear" -- A scene-referred encoding where the numeric representation is proportional to scene luminance.  Examples: ACES2065-1, ACEScg.

"display-linear" -- A display-referred encoding where the numeric representation is proportional to display luminance.  Example: CIE XYZ values measured off of a display or projection screen.

[...]

Issue #1030, opened 2020-06-06
Add the ability to define shared views


doug-walker:
When authoring configs with a large number of displays, the display/view section of the config can become quite lengthy and difficult to read.  In other words, with D displays, V views, there would be V x D definitions (assuming each view is used with each display, which is a common scenario).  Some configs will also define copies of the same view with and without looks, leading to even more definitions.  Reading and maintaining such a config may become difficult and error-prone.  

This proposal is to allow a config to use the notion of a "shared view" which is a view that is defined once but may be used with multiple displays.

To illustrate with an overly simple example from the Imageworks spi-vfx config, it would allow replacing this:

```
displays:
  DCIP3:
    - !<View> {name: Film, colorspace: p3dci8}
    - !<View> {name: Log, colorspace: lg10}
    - !<View> {name: Raw, colorspace: nc10}
  sRGB:
    - !<View> {name: Film, colorspace: srgb8}
    - !<View> {name: Log, colorspace: lg10}
    - !<View> {name: Raw, colorspace: nc10}

```
with this:

```
shared_views:
    - !<View> {name: Log, colorspace: lg10}
    - !<View> {name: Raw, colorspace: nc10}

displays:
  DCIP3:
    - !<View> {name: Film, colorspace: p3dci8}
    - !<Views> [ Log, Raw ]
  sRGB:
    - !<View> {name: Film, colorspace: srgb8}
    - !<Views> [ Log, Raw ]

```
Obviously with only two displays and three views there is not much benefit, but when the number of displays gets large there is a big benefit.

A previous PR added the notion of a display color space and a view transform.  This allows breaking down the conversion from a scene-linear color space into a view transform that converts scene-linear to display-linear colorimetry followed by a display color space.  An example of how that looks in config syntax is:

```
displays:
  sRGB:
    - !<View> {name: ACES, view_transform: ACES-sdr-video, display_colorspace: sRGB}
    - !<Views> [ Log, Raw ]
  Rec709:
    - !<View> {name: ACES, view_transform: ACES-sdr-video, display_colorspace: Rec709}
    - !<Views> [ Log, Raw ]

```
Such a config would also define display color spaces for sRGB and Rec709:

```
display_colorspaces:
    - !<ColorSpace> 
      name: sRGB
      from_display_reference: <omitted for brevity>
    - !<ColorSpace> 
      name: Rec709
      from_display_reference: <omitted for brevity>

```
[...]

Issue #1029, opened 2020-06-06
Refactoring DisplayTransform


doug-walker:
DisplayTransform, along with ColorSpaceTransform, form the heart of OCIO.  This is where all the business logic for connecting color spaces lives.  However, DisplayTransform is quite complicated, with many features that do not seem widely used (linearCC, colorTimingCC, channelView, displayCC, looksOverride).  Also, the implementation of colorTimingCC and channelView are not well suited to all apps.  More importantly, in OCIO v1 it thows if direction==inverse.  

It seems that there are two someone separate requirements here:

1)  Have a clear and robust transform that allows apps to easily request the result of the display/view section of an OCIO config.  Note that this includes the processing of any LookTransforms that are part of that view.  The code should be as simple as possible and should have a properly implemented inverse (which is not true today).

2)  Provide an example for how to implement a viewing pipeline that could be used as the basis for a DCC application's viewport.  This would include all the current capabilities of DisplayTransform, including channel swizzling, and the ability to do adjustable exposure and gamma controls.

Note that in previous PRs, we have added the apphelpers directory (outside of OCIO core) which is intended to serve as a set of templates for application developers.  The goal is to provide best practices that could help promote consistency across applications as well as provide good examples of OCIO usage.

The first part of the proposal is to simplify and refactor DisplayTransform so that it only implements what the display/view section of a config file does.  It would have a properly implemented inverse (that properly utilizes the from_ref and to_ref transform directions), along the lines of how ColorSpaceTransform works.

The new DisplayTransform would apply any LookTransforms that are specified in a view.  It would also offer a setLooksBypass and getLooksBypass method that would allow applications to offer a control to easily toggle a view's looks on and off.  Note that this simpler control would replace the looksOverride functionality for DisplayTransform.

Like ColorSpaceTransform, the new DisplayTransform will be a no-op if the source or destination is a data color space.  Note that this is a change from the current DisplayTransform, which has a fairly complicated set of logic around what gets applied and what does not.

The second part of the proposal is to add a ViewingPipelineTransform class under the apphelpers directory.  This transform would fully support all of the current functionality of DisplayTransform including LinearCC, ColorTimingCC, channelView, DisplayCC, and looksOverride.  It would call the refactored DisplayTransform to implement that part of the processing.

[...]

PR #1038, opened 2020-06-16
Adsk Contrib - Add ViewingRules, DisplayViewTransform, and ViewingPipeline

BernardLefebvre:
This PR implements the following Issues:
#1032 Allow applications to filter views based on color space
#1031 Add a color space property to describe the encoding
#1030 Add the ability to define shared views
#1029 Refactoring DisplayTransform
#1028 Option to process data channels
 
It also addresses the subject of these older PRs by others that were never approved:
#361 Get all displays from the config file
#357 Add description to Display/View combination
 
Note that the name of the Config method addDisplay was changed to addDisplayView. Also getDisplayColorSpaceName and getDisplayLooks  have been renamed getDisplayViewColorSpaceName and getDisplayViewLooks.

The ViewingPipeline binding has been deferred due to the uncertainty around how the apphelpers bindings will be done.  For example, should it be part of the main binding or a separate binding?  This should be discussed in the TSC or working group.


hodoulp:
Given that there has been no feedback, here is a friendly reminder that the branch can now be merged next week.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -557,6 +606,35 @@ class OCIOEXPORT Config
     //!cpp:function::
     const char * getActiveViews() const;
 
+    //!cpp:function:: Get all displays in the config, ignoring the active_displays list.
+    int getNumDisplaysAll() const;
```
michdolan, line 610: Just my two cents, but I have sometimes found it less than ideal that `get/setActiveDisplays` and `get/setActiveViiews` get/set the raw comma-delimited strings instead of iterating over active display/view names like the other display/view methods. The current case of needing a `getDisplayAll` feels a bit awkward, as opposed to `getDisplay` returning all displays, and `getActiveDisplays` returning active displays. We could even add `serializeActiveViews` and `deserializeActiveViews` methods to accomodate the previous `get/setActive* `behavior. My suggestion would be a big API change so I'm not expecting it to be done, but raising it as a discussion point while this work is happening.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -713,55 +791,55 @@ class OCIOEXPORT Config
...
+                                     TransformDirection direction) const;
 
     //!rst: Get a processor to convert between color spaces in two separate configs.
 
     //!cpp:function:: This relies on both configs having the aces_interchange role (when srcName
     // is scene-referred) or the role cie_xyz_d65_interchange (when srcName is display-referred)
     // defined.  An exception is thrown if that is not the case.
     static ConstProcessorRcPtr GetProcessor(const ConstConfigRcPtr & srcConfig,
-                                            const char * srcName,
+                                            const char * srcColorSpaceName,
```
michdolan, line 809: I don't disagree with this change, but had the impression that the previous use of `srcName`/`dstName` was calling out the fact that this can be a color space name or a role name.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -1037,23 +1232,45 @@ class OCIOEXPORT ColorSpace
...
+    // are digitally encoded.  For example, images in scene-linear, logarithmic, video, and data
+    // color spaces could have different default views.  Unlike the Family and EqualityGroup
+    // attributes of a color space, the list of Encodings is predefined in the OCIO documentation
+    // (rather than being config-specific) to make it easier for applications to utilize.
+    //
+    // Here is an example config entry that could appear under a ColorSpace:
+    // encoding: scene-linear
+    //
+    // Encoding strings are not case-sensitive. Although users may add their own encodings, the
+    // strings will typically come from a fixed set listed in the documentation (similar to roles).
```
michdolan, line 1249: We chould define constants for these in `OpenColorTypes.h` like was done for standard roles. Very useful for integration.

include/OpenColorIO/OpenColorTransforms.h
```diff
@@ -1115,9 +1082,21 @@ class OCIOEXPORT LookTransform : public Transform
     // look specification. (And forward is assumed in the absence of either)
     void setLooks(const char * looks);
 
+    bool getSkipColorSpaceConversion() const;
```
michdolan, line 1085: Worth a comment to explain what this does.

src/libutils/apphelpers/ViewingPipeline.cpp
```diff
@@ -0,0 +1,372 @@
...
+    {
+        std::ostringstream oss;
+        oss << "ViewingPipeline is not valid: "
+            << e.what();
+        throw Exception(oss.str().c_str());
+    }
+}
+
+ConstProcessorRcPtr ViewingPipeline::getProcessor(const ConstConfigRcPtr & configIn,
+                                                           const ConstContextRcPtr & context) const
```
michdolan, line 164: [minor] Indentation off?

michdolan:
Great work by the way @BernardLefebvre ! Looking forward to testing this.

include/OpenColorIO/OpenColorIO.h
michdolan: Though since these two names are interchangeable, and it is clearly documented, this LGTM

include/OpenColorIO/OpenColorIO.h
BernardLefebvre: as you explained, we wanted to preserve the behavior of getDisplay to only consider active displays. The all function is needed internally to be able to save all displays and we would not have exposed them to the API if the pimpl interface design would had allowed us to do so.

include/OpenColorIO/OpenColorTransforms.h
BernardLefebvre: adding comment

src/libutils/apphelpers/ViewingPipeline.cpp
BernardLefebvre: fixed

include/OpenColorIO/OpenColorIO.h
michdolan: Did you consider an overloaded `getDisplay` method with an argument to return all displays instead of creating additional interface? That was done with `getColorSpace` (reference space and visibility options).

michdolan: Actually I guess that won't work well with the index-based lookups... since index would then have variable behavior depending on display order. I concede.

include/OpenColorIO/OpenColorIO.h
michdolan: This can stand as a suggestion for a future PR, since it would not directly effect this work.

include/OpenColorIO/OpenColorTransforms.h
michdolan: Thanks!

PR #1042, opened 2020-06-23
PyOCIO: Fix AllocationTransform __str__, cleanup params

michdolan:
In response to #1036 discussion:

`AllocationTransform` was failing to cast properly when using the base `__str__` definition (from `defStr`), which was inherited from `Transform`. All other transforms worked correctly here, but something about `AllocationTransform` caused dynamic cast to fail on the C++ side. I _think_ the issue is related to the `Transform` subclass bindings having an additional `class_` template argument (specifying a base class), but I could not find an obvious difference between `AllocationTransform` and the other transforms that would indicate why. Extending `defStr` to be variadic and using it on each `Transform` class resolves the issue.

This PR also solves an issue where not specifying allocation vars with the convenience `AllocationTransform` constructor would throw due to an invalid vars size.

I also did a pass over all of the Python bindings to improve parameter name consistency. This is more critical for Python than C++ since in Python you can use the parameter name as a keyword arg. I only updated the surface-level param names, keeping any lambda functions consistent with the C++ API.


hodoulp:
> This PR also solves an issue where not specifying allocation vars with the convenience AllocationTransform constructor would throw due to an invalid vars size.

Found and fixed some issues around vars in the pull request #1048

PR #1043, opened 2020-06-23
Rename andriod-toolchain.cmake to android-toolchain.cmake

Simran-B:
I happened to see this typo. The file doesn't seem to be referenced by any files of this repo, thus no other changes in this PR.


hodoulp:
Thanks @Simran-B for the fix.
However, the status of the Android support is unknown. If you have interest in supporting OCIO on Android, we always welcome any kind of help (i.e. discussion/code/doc/python).

Simran-B:
Thanks for the offer. Unfortunately, I have no development experience with Android and the project I support, which depends on OCIO, doesn't target Android at all 😞

Issue #1039, opened 2020-06-17
Replace GLX dependency with EGL dependency


ChinYing-Li:
Quoting @jfpanisset : 
"" Currently OpenColorIO uses GLUT in tests/gpu/GPUUnitTest.cpp to create a 
window to render OpenGL tests, which in turn depends on GLX. I'm assuming 
these days it's linking against FreeGLUT, and it is apparently possible to use 
EGL instead of GLX for creating the rendering context. The advantage is that it 
might be possible to do away with having to stand up an X server for the CI 
pipeline, and whereas EGL rendering is supported inside NVIDIA containers
 (with built-in support starting in Docker 19.03), GLX is not. "

ChinYing-Li:
In fact, freeglut also [seems to support headless rendering](https://sourceforge.net/p/freeglut/mailman/message/29619660/) but has to go through a X server.

ChinYing-Li:
Hi, a bit of guidance needed. In order to support headless rendering,  I refactored```OglApp```, which is now a base class, from which derived ```ScreenApp``` and ```HeadlessApp```. 
This means that wherever ```OCIO::OglApp & app``` is passed as a function parameter, I have to substitute it with ```OCIO::OglAppRcPtr```. (```typedef OCIO_SHARED_PTR<OglApp> OglAppRcPtr```) 
So far, I don't see if there would be any drawbacks of this approach, and would like others' input on my approach. Thanks in advance.

hodoulp:
Note: Discussion started in #547

hodoulp:
Hi,

The approach seems to be the right one, and most of the code is already using a `OCIO::OglAppRcPtr` so the transition should be fine.

However, `EGL` is nVidia specific and OCIO must **not** stick to a specific vendor. Even if the CI build system is running on nVidia graphic cards that should not be mandatory for our users (e.g. using an app to debug or investigate a color transformation) & contributors (e.g. running GPU unit tests on their machine).  How do you plan to handle that?

Ideally the code should detect the graphic card vendor and auto adjust itself to the right 'headless implementation' (i.e. perhaps using `glGetString(GL_VENDOR)`?).

Here is a list of apps needing oglApp, with a transition proposal:
* ociochecklut -> `HeadlessApp`
* ocioconvert -> `HeadlessApp`
* ociodisplay -> `ScreenApp`
* ocioperf -> `HeadlessApp`
[...]

PR #1046, opened 2020-06-25
Add support to headless rendering on linux machine with EGL.

ChinYing-Li:
Note: this branch is not ready to be merged, as there are some debug prints. Also need to verify that this works for different vendors. 
Aims to solve issue #1039 .


hodoulp:
Thanks @ChinYing-Li for the contribution.  I will definitively have a look this afternoon or tomorrow.

Note: You can see build breaks with github actions by clicking on 'details' corresponding to the failing build.

PR #1074, opened 2020-07-23
Feature/python docstrings

scoopxyz:
I know this is broken for now, but wanted to make PR to facilitate discussion and have code to talk over.

This PR intends to enable the process of:

Public Header Doxygen Comments --> Python Docstring String-Literals --> Macro Expansion --> PyBind11 Python Bindings (with DocStrings)

The goal is that if you `help(foo)` in Python you get full docstrings, that are equivalent to the same C++ comments. This also facilitates IDE hints in VSCode etc...

It also allows docstrings to exist in one and only one place. The public C++ headers.


michdolan:
I looked at the CI failure. Linux and Windows are failing on configure because our VMs/images don't have LATEX installed. Is the LATEX dependency just for the Sphinx PDF target? If so what are your thoughts on removing that target and dependency. RTD can generate a PDF from the HTML docs, correct?

For macOS, which probably has LATEX installed already, the failure is due to the missing `clang` Python package, which `mkdoc` depends on. To add this dependency via CMake (with support for `OCIO_INSTALL_EXT_PACKAGES`) you can use this CMake macro, which acts like a `Find<package>.cmake` module for Python packages. It will check for the existence of the package by trying to import it into Python, and then fallback to installing it the ext/ dir with `pip`.
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/share/cmake/macros/FindPythonPackage.cmake

The find command should be put in this file, under the `OCIO_BUILD_DOCS` section (you'll see the find Sphinx command there):
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/share/cmake/modules/FindExtPackages.cmake

Finding a package in that module makes it globally accessible to the rest of the project, so you would add something like this:

```
include(FindPythonPackage)

find_python_package(clang, <min version>)
```

Each needed Python package can have a similar line.

Simran-B:
I can't find any specific information in RTD docs about how they generate the PDF format, but that suggests that they simply let Sphinx generate it from the documentation source files (not the HTML output) on their side.

KelSolaar:
> I can't find any specific information in RTD docs about how they generate the PDF format, but that suggests that they simply let Sphinx generate it from the documentation source files (not the HTML output) on their side.

Yeah it is generated from the source and from scratch by default:

python /home/docs/checkouts/readthedocs.org/user_builds/colour/envs/develop/bin/sphinx-build -b latex -D language=en -d _build/doctrees . _build/latex

scoopxyz:
Purely adding the above macro to find additional pypi packages doesn't appear to work, I get the following error:

```
CMake Error at share/cmake/modules/FindExtPackages.cmake:54 (find_package):
  By not providing "Findbreathe.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "breathe", but
  CMake did not find one.

  Could not find a package configuration file provided by "breathe"
  (requested version 4.19.0) with any of the following names:

    breatheConfig.cmake
    breathe-config.cmake

  Add the installation prefix of "breathe" to CMAKE_PREFIX_PATH or set
  "breathe_DIR" to a directory containing one of the above files.  If
  "breathe" provides a separate development package or SDK, be sure it has
  been installed.
  CMakeLists.txt:260 (include)


-- Configuring incomplete, errors occurred!
See also "/media/fawlty/Alpha/Dev/github/OpenColorIO/build/CMakeFiles/CMakeOutput.log".
```

The Sphinx package find/download appears to only work because the additional pypi download is specified in it's `FindSphinx.cmake` module.

michdolan:
@scoopxyz what are the nest steps with this work? Let me know if you want some help sorting out the API docs.

michdolan:
Will have a new PR coming soon to piggy back on this work. Going to close this. Thanks @scoopxyz !

Issue #1034, opened 2020-06-09
Create Python bindings for libutils


michdolan:
I think it could be worthwhile to create a optional, secondary Python lib which binds `libutils` for utilizing the convenience of the various helpers, and helping to interface with OpenImageIO's Python bindings, and PyOpenGL, etc. There are quite a few cases in a pipeline for implementing these concepts in Python tools (which primarily leverage C++ libs under the hood), so it would be great to offer the full suite of implementation tools in both languages.

ChinYing-Li:
Not so familiar with python bindings for cpp, but can help with this before the documentation related GH issues are out.

michdolan:
Hi @ChinYing-Li 

I've started some work on this but would be more than happy to have help. I'll create a PR this evening with my work so far to show you what I'm doing. There are three libraries to bind as part of this and we could collaborate if you're interested.

ChinYing-Li:
@michdolan 
An update on the progress: 
A `HeadlessApp` instance can't make EGL context current when using the `PyOpenColorIO_oglapphelpers` library, yet there's no such error during GPU unit testing. I'll open a draft PR once the problem is solved.

PR #1080, opened 2020-07-29
Py oglapphelper

ChinYing-Li:
This PR create python binding for `libutils/oglapphelpers`, and is part of work related to #1034 .

Note: I saw that
```
COMPILE_FLAGS` in `set_target_properties(library_name PROPERTIES
        COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS}")
``` 
is deprecated in CMake >=3.11. Might want to change it to `COMPILE_OPTIONS`.


doug-walker:
Thanks @ChinYing-Li !  Having the ability to use the GPU renderer from Python would be very helpful.  We will take a look at this soon.  

What unit tests do you think should be added?

ChinYing-Li:
@doug-walker 
If we are just testing whether the binding is working correctly, I think we can just write the python version of some cherry-picked unit tests from each of `gpu/*_test.cpp`. However, I don't have any idea what the criteria of cherry-picking should be. Any suggestions?
Python bindings already exist for the `Transform` classes used in gpu unit tests.

src/bindings/python/oglapphelpers/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright Contributors to the OpenColorIO Project.
+
+set(SOURCES
+        PyOpenColorIO_oglapphelpers.cpp
+        PyOpenColorIO_glsl.cpp
```
michdolan, line 6: Suggest naming the later two .cpp files just `Pyglsl.cpp` and `Pyoglapp.cpp`. Keeping PyOpenColorIO_oglapphelpers.cpp is ok since that is the entry point for the Python module of the same name.

src/bindings/python/oglapphelpers/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+	)
+endif()
+
+if(UNIX)
+	# A 'module' is a dynamic library on Linux (i.e. '-fPIC' needed),
+	# but a static library on Windows.
+
+	# If supported for the target machine, emit position-independent code
+	# suitable for dynamic linking.
+        set_property(TARGET PyOpenColorIO_oglapphelpers PROPERTY POSITION_INDEPENDENT_CODE ON)
```
michdolan, line 38: This file has some inconsistent indentation.

src/bindings/python/oglapphelpers/PyOpenColorIO_oglapp.cpp
```diff
@@ -0,0 +1,82 @@
...
+            .def("redisplay", &OglApp::redisplay)
+            .def("readImage", &OglApp::readImage)
+            .def("printGLInfo", &OglApp::printGLInfo)
+            .def_static("CreateOglApp", &OglApp::CreateOglApp,
+                        "winTitle"_a,
+                        "winWidth"_a,
+                        "winHeight"_a);
+
+    py::enum_<OglApp::Components>(cls, "Components")
+            .value("Components_RGB", OglApp::Components::COMPONENTS_RGB)
```
michdolan, line 41: `Components` should be `COMPONENTS` in both items.

src/bindings/python/oglapphelpers/PyOpenColorIO_oglapp.cpp
```diff
@@ -0,0 +1,82 @@
...
+            .export_values();
+}
+
+void bindPyScreenApp(py::module & m)
+{
+    auto cls = py::class_<ScreenApp,
+                          ScreenAppRcPtr /* holder */,
+                          OglApp /* base */>(m, "ScreenApp")
+            .def(py::init([](const char * winTitle,
+                          int winWidth,
```
michdolan, line 52: These two lines should be indented to align with the line above, like what was done for `HeadlessApp` and `OglApp`. You could also opt to put the arguments on the same line if they fit. My rule of thumb with pybind11 is generally to give function parameters their own lines only when one or more has a default value.

ChinYing-Li: Thank you for reviewing. I just realized the indentation of my code is different from that of existing PyOpenColorIO code. This is fixed along with the style changes you suggested.

ChinYing-Li:
I will then work on the python gpu unit tests. Please let me know in case there are certain tests to be prioritized!

Issue #1079, opened 2020-07-28
[diversity] Use inclusive language


Simran-B:
As discussed in the TSC meeting on 2020-07-27 and the [ASWF #diversity Slack channel](https://academysoftwarefdn.slack.com/archives/C0152KSAJSJ/p1594730786066000), the term "sanity check" should be avoided as it is considered [ableist language](https://www.selfdefined.app/definitions/sanity-check/). I checked the OCIO source code for other terms, but there are just a few ones that should potentially be substituted.

1. _sanity check_: validate (perform check), smoke test, validity / plausibility / soundness check
2. _sane_, _insane_, _sanity_: valid, invalid, validity (syntactic?)
3. _dumb_: basic? (draft implementation)
4. _lame-ass_: boring, unpleasant, crude? (support code)
5. _whitelist_: [approved list](https://github.com/communitybridge/docs/search?q=whitelist&unscoped_q=whitelist), allow-list (CLA)
6. _master_: main? (Git branch)

1 and 2 are urgent to be renamed in time for v2, as they are used in a public interface. 3, 4 and 5 are trivial to replace. 6 would be a huge task... related: https://github.com/github/roadmap/issues/63

I also found the terms _stupid_, _crazy_ and _crippled_ in third-party code (comments/test data), but changing that seems out of scope.

hodoulp:
Hi @Simran-B, Can we close the issue?

Simran-B:
Yes and no... it is kind of a reminder for me to also check other ASWF projects and add something about inclusive language to the documentation.

hodoulp:
If you want to keep track of other ASWF projects, the best would be to create a defect in their repo.
On the other hand having a disclaimer in the documentation is a great idea.

Sayantani-sunshine:
Can I work in this issue?

doug-walker:
@Simran-B , we have a volunteer for this issue.  I know a lot of this has already been implemented.  Please let us know if there is anything left to do or else close the issue.  Thx!

PR #1081, opened 2020-07-31
Substitute term "sanity check" with "validation" / "validate"

Simran-B:
Also removes/renames a few other terms. See #1079 for details.

This is a first pass, all potentially offending terms should be substituted. Needs testing however, I did not try to compile it locally.


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 <!-- SPDX-License-Identifier: CC-BY-4.0 -->
 <!-- Copyright Contributors to the OpenColorIO Project. -->
 
+**Version 2.0.0 (unreleased):**
+    * Renamed sanityCheck() method to validate() in the classes
+      Config, FileRule and ViewingRule
+    * Renamed Sanity enumeration to Validation in Config::Impl class,
+      and its members SANITY_UNKNOWN, SANITY_SANE and SANITY_INSANE to
+      VALIDATION_UNKNOWN, VALIDATION_VALID and VALIDATION_INVALID
```
Simran-B, line 9: The enum doesn't seem to be used outside the class, thus I'm unsure if it should be mentioned at all.

KevinJW:
Not a code comment per se, but is valid/validate the best choice for the name change, if sanity checking is considered poor choice, then I would expect valid could become so by negation ([2nd definition](https://dictionary.cambridge.org/dictionary/english/invalid)) Would "consistency" be more appropriate?

Simran-B:
_invalid_ may be insensitive too, that is right, but we don't refer to a person with a disability here. Anyway, if there's a better set of words then we can also go for that.

_Consistent_ doesn't really cut it in my opinion, because it makes it sound like it would be about uniformity. The validation goes much further however. Aside from a YAML syntax check, I think I saw tests for missing values, sensible value ranges and plausible combinations of things.

To cover both form and content checks, the following would work quite nicely:
- VALIDATION_UNKNOWN (maybe also UNTESTED or UNCHECKED)
- VALIDATION_PASSED
- VALIDATION_FAILED

In combination with _validation_ it should be fine to use _failed_ ([alexjs.com](https://alexjs.com/#demo) warns about it)

hodoulp:
@Simran-B Could you update the pull request? Thanks.

Simran-B:
@hodoulp Merge conflict resolved 👍 

In reaction to Kevin's concern, I asked in the [#diversity](https://academysoftwarefdn.slack.com/archives/C0152KSAJSJ/p1596540761074500) ASWF slack channel for more feedback:

> [...] In particular, do you consider "invalid" to be problematic as term in the context of configuration validation? Should the terminology be inclusive without context too?

> JT Nelson:
> As a programmer of many years, IMO and from common shared knowledge as I have experienced it in my diverse background, validation is a deep and across the board critical category in programming so context is of such high importance and relevancy I would assert no, it is not problematic.  Where changing it would be quite problematic as in any possible alternative would most likely cast unreasonable confusion on the meaning.   If "in' va lid" is the questionable pronunciation and inference, the actual pronunciation is "in val' id" which casts it as a different word.

src/OpenColorIO/Config.cpp
```diff
@@ -249,11 +249,11 @@ static constexpr unsigned LastSupportedMinorVersion = OCIO_VERSION_MINOR;
...
 public:
-    enum Sanity
+    enum Validation
     {
-        SANITY_UNKNOWN = 0,
-        SANITY_SANE,
-        SANITY_INSANE
+        VALIDATION_UNKNOWN = 0,
+        VALIDATION_VALID,
+        VALIDATION_INVALID
```
Simran-B, line 256: Possible alternatives that read slightly better:

- CONFIG_UNTESTED
- CONFIG_VALID
- CONFIG_INVALID

Alternative naming, avoiding the term "invalid":

- VALIDATION_UNKNOWN (maybe also UNTESTED or UNCHECKED)
- VALIDATION_PASSED
- VALIDATION_FAILED

src/OpenColorIO/Config.cpp
```diff
@@ -641,14 +641,14 @@ class Config::Impl
...
-                throw Exception(m_sanitytext.c_str());
+                m_validationtext = os.str();
+                throw Exception(m_validationtext.c_str());
             }
         }
     }
 
     // Check a shared view used in a display. View itself has already been checked.
-    void sanityCheckSharedView(const std::string & display, const ViewVec & viewsOfDisplay,
+    void validationSharedView(const std::string & display, const ViewVec & viewsOfDisplay,
```
doug-walker, line 651: validationSharedView --> validateSharedView

src/OpenColorIO/Config.cpp
doug-walker: Good idea, I prefer PASSED/FAILED.

hodoulp: That's better enum names.

src/OpenColorIO/Config.cpp
```diff
@@ -545,14 +545,14 @@ class Config::Impl
         }
     }
 
-    // Sanity check view object that can be a config defined shared view or a display-defined view.
-    void sanityCheckView(const std::string & display, const View & view) const
+    // Validation view object that can be a config defined shared view or a display-defined view.
```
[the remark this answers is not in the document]
hodoulp: Agree with @doug-walker a method must start with a verb i.e. it represents an action.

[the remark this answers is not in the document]
Simran-B: Whoops, my bad! Fixed by https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1081/commits/68f23ed5a69fceb1c22bf3410325cd66ec626dfa

src/OpenColorIO/Config.cpp
Simran-B: Renamed in https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1081/commits/6c1cceaa1992c4523b25f19e43cda692fa7235b6

CHANGELOG.md
Simran-B: Looking at the rest of the changelog, it seems to describe changes from a user perspective. Hence, I simplified it to a single changelog entry. Removed the mention of the C++ classes, as it equally applies to the Java and Python interfaces.

Simran-B:
@doug-walker Thanks, I addressed the review comments and it should be ready to merge now.

PR #1095, opened 2020-08-11
[digipro] Feature/temp api rst

scoopxyz:
Not complete, and still contains errors. But putting here to share progress.


michdolan:
@scoopxyz are you planning to make more pushes to this branch before DigiPro? Or should we merge this for now and revise later?

hodoulp:
@scoopxyz Here is a git diff [doc_api.txt](https://github.com/AcademySoftwareFoundation/OpenColorIO/files/5059970/doc_api.txt) fixing most of the 'doxygen warnings' (i.e. removing or rewriting the lines) from the rst files for the C++ API. The idea is to have an API doc ready for DigiPro.

In some cases, I was able to fix the OpenColorIO.h file, but other problematic cases are:
- Some Python methods are in the C++ parts
- C++ methods with the same name are 'sometime' problematic

Note: Not knowing how to regenerate these c_xx.rst files, I tried to update the C++ include files without being able to validate the changes. But we have anyway to revisit the C++ include files so it's not an issue for short-term.

scoopxyz:
I've pushed another PR over here (#1097) just to keep things clean. There shouldn't be any doxygen errors this time around Patrick.

PR #1099, opened 2020-08-13
Adsk Contrib - Improve cmake build

hodoulp:

The pull request improves the cmake files:

- to fix the pybind11 build which was always in debug i.e. the CMAKE_BUILD_TYPE was missing
- to fix the pybind11 & yaml-cpp which were not using the requested C++ version
- to fix the yaml-cpp find_package() which fails to correctly find an installed version when several repo clones exist
- to add the pkg-config file
- to improve the resource file for the DLL details.


share/cmake/modules/Findyaml-cpp.cmake
```diff
@@ -26,27 +26,50 @@
...
 
 if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
+    set(_yaml-cpp_REQUIRED_VARS yaml-cpp_LIBRARY)
+
     if(NOT DEFINED yaml-cpp_ROOT)
+
         # Search for yaml-cpp-config.cmake
-        find_package(yaml-cpp ${yaml-cpp_FIND_VERSION} CONFIG QUIET)
+
+        # TODO: The find_package() with yaml-cpp does not provide a good module
```
michdolan, line 35: I feel like the likelihood of multiple OCIO clones existing is probably lower than that of having a system-installed yaml-cpp package with a bundled CMake config. I would argue that your scenario is a valid case to pass `yaml-cpp_ROOT` as a CMake argument to tell OCIO where to find it. Also, users are a lot less likely to have pkg-config installed on non-Linux systems. 

Or perhaps a `find_package` argument can be used to avoid any OCIO clone directory, since `ext/dist`  would only be used if our build system were installing the package. If you have an example directory structure of where you are experiencing this problem, I'm happy to do some debugging and suggest other options if that's helpful.

hodoulp: Our use case is to have several clones of the OCIO repo (all compiled in release and/or debug) then compiling one brings the `yaml-cpp` library from another repo. The worst case is that it takes the debug version when compiling a release one (because the search is looking for the header file). That scenario is only hitting devs (i.e. no impact on CI builds and single repo clone) but it's causing us lot of headache. That behavior happens on Windows but also on MacOS.

The changes is only a short-term solution to guaranty that the `yaml-cpp` library used is the right one.

cmake `find_package()` provides several arguments but none was helping. Not being a cmake expert, I will appreciate any help to correctly solve that problem.

michdolan: Have you tried specifying `-Dyaml-cpp_ROOT=<path to yaml-cpp root dir>` in your cmake args to tell it where to find the install you want? That's the solution I would recommend, as there are many cases where that will be the only option, particularly for studios that maybe have libraries stored in a custom package management system.

hodoulp: Any 'mandatory' third-party libraries must fallback to building itself if not found on the system. So even studios should not have to worry about `yaml-cpp`.

If `yaml-cpp` is already installed in a conventional way then find_package() must find it; otherwise, the `-Dyaml-cpp_ROOT` is the only approach i.e. some custom installation for a custom library for example.

Using `-Dyaml-cpp_ROOT` here is another kind of hack which could only add complexity because cmake is supposed to correctly handle that case, and devs "not reading the doc" (i.e. almost all devs) could fall in the same problem.

hodoulp: `cmake ../.` (i.e. not extra option) must always work i.e. successfully & correctly compile the 'core' library.

BernardLefebvre: We had a case where a release build in a clone was finding a & using an existing debug build from another clone.

michdolan: Many libraries have strong requirements for a builder providing mandatory third-party libraries themselves. I typically only expect `cmake ../.` to work when building a library with no external dependencies.

I think OCIO is quite unique in that we will pull all dependencies from upstream for you automatically.  I agree that supporting a simple core library build is important, but I do feel strongly that the change proposed in this PR is a heavy handed fix. I think this change will break as many builds as it will fix since it prevents CMake from finding system installed yaml-cpp unless either pkg-config is available, or if an OCIO build was previously (and successfully) done in the current build directory with `OCIO_INSTALL_EXT_PACKAGES` set to `MISSING` or `ALL`. It will only discover the later on a second build.

I disagree that specifying a ROOT dir for a dependency is a hack. It's a CMake convention for assisting a find module when the build environment has unique circumstances, or the builder wants specific control of linkage. In the builds I maintain I often use ROOT for every dependency when supported since I then have full control over the version of the dependency which is used.

IMO we should assist a builder as much as possible to ease the build, but the builder is also responsible for reading the install documentation to understand the required build process. I think we should duplicate that information in a root-level INSTALL.md file to make it more discoverable. Whenevr I build a third-party library that's the first thing I check before trying to build.

If the yaml-cpp CMake config is not functioning properly I would rather revert to the previous behavior of the find module which utilizes native CMake functions for finding the install more explicitly. It may just be that that the yaml-cpp CMake config is not in an ideal state for use.

michdolan: I recommend omitting this specific change for now, so the other fixes can be merged, and we can address the yaml-cpp discovery issue in another PR. If you can provide me a listing of your top-level directory structure I will do some investigation into an alternative solution this weekend. ROOT can be used in the interim until we find a solution you are more comfortable with.

hodoulp: > Many libraries have strong requirements for a builder providing mandatory third-party libraries themselves. I typically only expect cmake ../. to work when building a library with no external dependencies (or use of common system dependencies).

A usual the Windows platform case is much more tricky (i.e. try to compile OpenImageIO on Windows) so that must be a strong commitment. In the future, the `vcpkg` tool will certainly decrease that need.

> I disagree that specifying a ROOT dir for a dependency is a hack. 

Let me better explain my point. The ROOT variable is perfect solution for custom installation. Used as a workaround to help finding the right yaml library (because find_package() for yaml-cpp is broken) is problematic.


@michdolan What about adding `yaml-cpp_ROOT` to the find_package() if it exists?


> IMO we should assist a builder as much as possible to ease the build, but the builder is also responsible for reading the install documentation to understand the required build process.

For a packager builder or a studio you are right.
For an occasional contributor or dev that extra requirement will generate much more questions e.g. we already have to explain (for all platforms) how to build OCIO!

> If the yaml-cpp CMake config is not functioning properly I would rather revert to the previous behavior of the find module which utilizes native CMake functions for finding the install more explicitly. It may just be that that the yaml-cpp CMake config is not in an ideal state for use.

@michdolan That could be a better solution with a TODO to explain that find_package() with yaml-cpp has limitations. I can revert the code to the former yaml-cpp search code. What do you think?

**Use case:**
Without any guaranty to reproduce the problem, our use case is:
1. git clone several times (e.g. three clones ocio_1, ocio_2 and ocio_3) and build them all 
2. Pick one clone, remove your build directory, generate again the build file using `cmake ../.` and finally check the file path of the discovered yaml-cpp

hodoulp:
But if you have time and want to investigate during the weekend, I can definitively remove that part only so we can move forward. However I do feel that we must have a solution by SIGGRAPH.

michdolan:
Let me investigate this afternoon and evening. I'll try to have a PR with a fix tomorrow (or revert to the old module, but keeping CMake 3.12 conveniences). I have a windows build env for OCIO setup already, so I can try to reproduce and investigate a solution. I can then either PR that work or provide you the module for this PR. Thanks @hodoulp !

hodoulp:
Let me now if you need some extra testing. I have two platforms with the faulty behavior (i.e. MacOS and Windows).

hodoulp:
@michdolan I have reverted the `yaml-cpp` change to merge the PR if you approve it.

Issue #1100, opened 2020-08-14
New transforms for building Looks


doug-walker:
Today in OCIO, looks are generally built from either an ASC CDL, or a 3d-LUT. An ASC CDL is easily understood due to its parametric nature but is limited to fairly basic adjustments. A 3d-LUT allows for very powerful adjustments, but it’s difficult to understand, at a glance, what it is modifying. It would be useful to have some additional options that would occupy the middle ground.

This issue proposes a new set of transforms that provide some of the power of a 3d-LUT but with the parametric nature of an ASC CDL. This should allow for the creation of Look Transforms that are powerful and yet easy to understand. 

The first new transform is for making what are often termed “primary” color corrections to adjust overall exposure, color balance, and saturation. A second “tone” transform allows for more targeted adjustments to specific parts of the tonal range. For example, adjusting the brightness or color balance of just a portion of the highlights or shadows. The range over which each control operates is also parametrically adjustable. A third “curves” transform may be used to define spline-based RGB tonal curves using control points. All of the transforms work in both the forward and inverse directions.

Traditional color grading tools are often intended to work on specific types of color space encodings. For example, applications will sometimes offer separate tools for video vs. logarithmic color space encodings. This is because the aesthetic impact of a given mathematical operator (e.g., a gain/multiply) on an image varies greatly with how that image is encoded. The new tools offer three modes intended for video, logarithmic, and scene-linear color encodings. The algorithms have been carefully designed to avoid clamping, which is particularly important when working with wide-range scene-linear color encodings.

They would also leverage the dynamic property infrastructure introduced in PR #743 so that they may be created in real-time using a GPU (in situations where it is appropriate to create them).  Note: When this proposal was discussed at the OCIO v2 working group meeting, there was a lot of discussion about the fact that the transforms would use the dynamic properties and concerns over workflow problems caused by editing of configs.  

Just to be clear, the present issue is neither pro or con on the subject of editing transforms referenced by configs -- that is a side issue that is separate from the subject of adding new transforms.  After all, there is nothing that prevents the existing ASC CDL or Matrix Transforms already available from being easily edited.  All of these look transforms do need to be created at some point, and it is obviously important that that happen in the context of the rest of the viewing pipeline.

KelSolaar:
Hi,

Isn't this a bit out of scope for OCIO? Where do we put the line between colour management and colour grading? This proposal and the related #1102 PR is very much stepping in the second domain.

[...]

PR #1102, opened 2020-08-14
Adsk Contrib - Add grading ops

BernardLefebvre:
The adds the grading transforms described in issue #1100 

- GradingPrimary
- GradingRGBCurve
- GradingTone

Inverse transform for GradingRGBCurve and GradingTone still have to be implemented.

Note that following the requirement introduced by the new grading transforms, GLSL fragment program generated by OCIO has now to support the static array intialization which requires the GLSL v1.2 at minimum (it means OpenGL 2.1 published in 2006). So, the minimal GLSL version is now 1.2. As side-effect any generated fragment shader program consuming OCIO generated code must now add the version directive with the appropriate version to avoid being in GLSL 1.1 (which is the behavior when no version directive is present).
 

hodoulp:
Found in the [GLSL version 1.20](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.1.20.pdf) that `uniform` can now be statically initialized: `Initializers may only be used in declarations of globals with no storage qualifier, a const qualifier, or a uniform qualifier` from section 4.3.
That could be used in OCIO when generating the fragment shader program which then simplifies the OpenGL helper library.

share/cmake/modules/Findyaml-cpp.cmake
```diff
@@ -29,8 +29,24 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
     set(_yaml-cpp_REQUIRED_VARS yaml-cpp_LIBRARY)
```
michdolan, line 29: This change seems like it slipped in from Patrick's CMake work which we resolved. Should likely revert this file.

src/bindings/python/PyGradingData.cpp
```diff
@@ -0,0 +1,252 @@
...
+    for (size_t pt = 0; pt < numPt; ++pt)
+    {
+        to->getControlPoint(pt) = from->getControlPoint(pt);
+    }
+}
+}
+
+void bindPyGradingData(py::module & m)
+{
+    py::class_<GradingRGBM>(m, "GradingRGBM")
```
michdolan, line 33: TODO: Add constructor with keyword arg support to all Grading classes (can be done in later PR).

share/cmake/modules/Findyaml-cpp.cmake
BernardLefebvre: Will do.

src/bindings/python/PyGradingData.cpp
BernardLefebvre: Ok

share/cmake/modules/Findyaml-cpp.cmake
michdolan: There's still changes being made to this file in the most recent commit. I think the state the file is in master currently is working well so would be good to leave it unchanged.

BernardLefebvre: I had tried to revert my original local commit but that might have interfere with the merge from master. Next commit will replace these files with their versions in master.

hodoulp:
Following the two-weeks rule, here is a friendly reminder that the branch can now be merged at the end of this week unless there are new comments.

KelSolaar:
Hello,

I wanted to report that following my [comment in issue #1100](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1100#issuecomment-674340278), a few people, that shall remain nameless for now, from the industry, stakeholders in their respective studios/companies, have reached to me and raised concern about the lack of transparency in the process that ending up having the 22K lines of code of that PR landing like a _fly in the ointment_ if I can say so.

The fact that the two-weeks rule is being invoked here makes it even more worry-some. Again, I would like to make an emphasis on the fact that I'm strictly talking about the process used to deliver the feature and not the content of the feature itself.

Cheers,

Thomas

PS: I will create a separate issue to centralise the discussions pertaining to the problem being discussed here.

doug-walker:
Hi @KelSolaar ,

As always, we will not merge if there are ongoing discussions.  It was simply not clear to us that there were still topics left to discuss about the PR itself.

Thank you for creating issue #1124, so that we can discuss the development process.  

If someone would like to actually discuss the new transforms for building looks, please make a comment (either here or in issue #1100).

Doug

KelSolaar:
Hi @doug-walker,

Your explanations about the scope in #1100 made sense to me and I'm certainly not opposed to merging the PR but it would have been disingenuous from my part to not mention the _private_ discussions that occurred lately.

> Thank you for creating issue #1124, so that we can discuss the development process.

You are very much welcome! I should have done it earlier but... time... you know :)

michdolan:
@KelSolaar and others. I know there are some reservations about this work in terms of OCIO's scope. Are there objections to merging this and working to clearly communicate the intent of these transforms in documentation?

KelSolaar:
@michdolan : As mentioned above, I'm fine. I'm still a bit concerned by the fact we have opened a Pandora Box here but fine nonetheless :)

doug-walker (approved):

doug-walker:
Thank you Michael for your review and Thomas for the interesting discussions.  Given that the PR has been available for review/comment for over three weeks and there are no ongoing active discussions, we will proceed with the merge.

Issue #1103, opened 2020-08-15
"OCIO_INSTALL_EXT_PACKAGES" value check is case sensitive and does not raise an exception if an invalid value is passed.


KelSolaar:
Hi,

Title says it all!

I was trying to figure out why the dependencies weren't pulled and I typed `-DOCIO_INSTALL_EXT_PACKAGES=all` instead of `-DOCIO_INSTALL_EXT_PACKAGES=ALL`.

The comparison should probably be case insensitive and raising an exception if an invalid value is passed.

Cheers,

Thomas

michdolan:
Fixed by #1104 

Comparison is now case insensitive and an invalid value will result in fatal configuration error.

PR #1104, opened 2020-08-15
Improve yaml-cpp config search and CMake messaging

michdolan:
- Fixes issue where yaml-cpp will be detected from a CMake config in a previous OCIO build `ext/dist` dir, which was caused by the config search checking the CMake User Package Registry.
- Cleans up Python package checking messaging. Previously appeared to be an error, and now the message is more concise and aligned with other package check messages since the traceback is now omitted.
- Make OCIO_INSTALL_EXT_PACKAGES case insensitive, and error for an invalid value


hodoulp:
Good catch.

Here is the link to understand the cmake feature: https://cmake.org/cmake/help/v3.12/manual/cmake-packages.7.html#user-package-registry

@michdolan Is there something to do for the 'system package registry' ?

https://cmake.org/cmake/help/v3.12/command/find_package.html
https://cmake.org/cmake/help/v3.12/manual/cmake-packages.7.html#system-package-registry

CMakeLists.txt
```diff
@@ -206,8 +206,24 @@ endif(NOT HAVE_SSE2)
...
+# Support options
+set(_EXT_OPTIONS NONE MISSING ALL)
+string(REPLACE ";" ", " _EXT_OPTIONS_STR "${_EXT_OPTIONS}")
+
+# Default option
+set(_OCIO_INSTALL_EXT_PACKAGES "MISSING")
+
+if(DEFINED OCIO_INSTALL_EXT_PACKAGES)
+    # Case insensitive match
+    string(TOUPPER "${OCIO_INSTALL_EXT_PACKAGES}" _OCIO_INSTALL_EXT_PACKAGES)
```
KelSolaar, line 218: Nice! 👍

michdolan:
It seems like using the system registry is more explicit since CMake provides no interface to do it. The build setup would have to manually add a registry entry on Windows for example. For now, I think we could leave it alone, but it is good to keep it in mind for the future.

CMakeLists.txt
michdolan: Was great timing for your feedback :)

KelSolaar: Yep!

hodoulp: Change fixing #1103

hodoulp:
Great. Monday, I will check all of that on macOS & Windows machines.

michdolan:
Good find @hodoulp . I think that sounds reasonable.

michdolan:
Swapped to disabling both package registries globally, which also resolves the issue in my tests.

Note: I noticed that in more recent CMake versions, the user package registry variable has been deprecated in favor of similar variables which doesn't contain `_NO_` and needs to be set to `OFF` instead of `ON`:

https://cmake.org/cmake/help/v3.18/manual/cmake-packages.7.html#disabling-the-package-registry

PR #1110, opened 2020-08-20
Adsk Contrib - Improve GLEW detection

hodoulp:

The GPU CI build investigations raise a `GLEW` library detection problem coming from a `cmake` bug. The pull request manages the `cmake` bug to always correctly detect the `GLEW`library.

Refer to https://github.com/AcademySoftwareFoundation/aswf-docker/issues/43 for details.
Refer to https://gitlab.kitware.com/cmake/cmake/-/issues/19662 for cmake details.

To correctly test the fix, you have to install a ASWF docker image. And I used: `aswf/ci-ocio:2021`.

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -205,10 +205,7 @@ jobs:
       CXX: ${{ matrix.cxx-compiler }}
       CC: ${{ matrix.cc-compiler }}
     steps:
-      # TODO: Remove this workaround following resolution of:
-      #       https://github.com/AcademySoftwareFoundation/aswf-docker/issues/43
       - name: Setup container
-        run: sudo rm -rf /usr/local/lib64/cmake/glew
         if: matrix.vfx-cy == 2020
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


michdolan:
Good find with the cmake bug! To test this fix in CI, you could also remove the workaround from the GH Actions workflow files:

```
      # TODO: Remove this workaround following resolution of:
      #       https://github.com/AcademySoftwareFoundation/aswf-docker/issues/43
      - name: Setup container
        run: sudo rm -rf /usr/local/lib64/cmake/glew
        if: matrix.vfx-cy == 2020
```

jfpanisset:
The CMake issue seems to prevent setting GLEW_INCLUDE_DIRS as well: this issue is "hidden" since it's just /usr/local/include and thus the GLEW includes are still found by the build (something else is adding -I/usr/local/include), but would it make sense to also explicitly set GLEW_INCLUDE DIRS to be explicit?

hodoulp:
@michdolan The first commit triggered the GitHub Action but missed the successful results. The last commit did not even trigger GitHub Action. I relaunched the previous build hoping it will bring the latest changes but it does not (as expected). 
Any idea?

PR #1120, opened 2020-08-24
Fix ociochecklut bugs

remia:
Fix for `ociochecklut`:
*  `ociochecklut` `--gpu` and `--gpulegacy` options were not available when compiling latest master, this was due to a duplicated flag in the option list.
*  `ociochecklut --gpu` was not working correctly for 1DLUT allocated as 2D texture (see commit message for more details)
*  Add `ociochecklut --invlut` to sample the LUT in inverse direction 

Fix on the cmake build:
*  On macOS 10.15 (with SIPS) and cmake 3.17.3, when trying to install OCIO at a custom location (`-DCMAKE_INSTALL_PREFIX`), cmake try to copy the pkconfig.pc file to /pkgconfig root folder. @hodoulp let me know if you manage to reproduce this, thanks.


src/apps/ociochecklut/main.cpp
```diff
@@ -97,6 +97,7 @@ class ProcessorWrapper
 
 #ifdef OCIO_GPU_ENABLED
     OCIO::OglAppRcPtr m_oglApp;
+    OCIO::GpuShaderDescRcPtr m_shaderDesc;
```
hodoulp, line 100: I do not seen the need to keep `shaderDesc` as a member variable. What is the problem you try to solve?

src/OpenColorIO/CMakeLists.txt
```diff
@@ -157,7 +157,7 @@ if(NOT WIN32)
     set(libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
     set(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
```
hodoulp, line 158: For consistency, could you change `\${prefix}` to `\${exec_prefix}`?

src/apps/ociochecklut/main.cpp
hodoulp: We only have the title of the commits. Anyway, line 91 is already done at line 69 so it recreates two times the complete GPU fragment shader and resources (`m_oglApp->setShader(m_shaderDesc)` called twice). That seems way overkill. But you found a problem..

remia: Sorry, I expected everyone could see the commit message, here it is:

> In OglApp::applyGPU, call to updateImage() update the currently bound GL_TEXTURE_2D
> target, by doing so, the latter sampling of a 1DLUT that was allocated as
> a GL_TEXTURE_2D (because of it's size), will sample the wrong texture.
> 
> Solution proposed is to call OglApp::setShader() again after image was updated.

I agree it's overkill, but ociocheck is not really about performance. I guess we could extend OglApp to have a more lightweight method to set the correct GL state.

remia: Fixed the problem at the source, ie. from `OglApp::redisplay()`

src/OpenColorIO/CMakeLists.txt
remia: I did that but noticed issues at the same time:
*  To use `CMAKE_INSTALL_LIBDIR` and similars, you need to include `include(GNUInstallDirs)`
*  There is an extra `\` at the start of each variable
*  We only really need `libdir` and don't use `includedir`, my take would be to remove all these and install straight at the relative path, like is done in other places, eg. OCIO apps
```
if(NOT WIN32)
    configure_file(res/OpenColorIO.pc.in ${CMAKE_CURRENT_BINARY_DIR}/OpenColorIO.pc @ONLY)
    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/OpenColorIO.pc DESTINATION lib/pkgconfig)
endif()
```

src/libutils/oglapphelpers/oglapp.cpp
```diff
@@ -111,6 +111,11 @@ void OglApp::redisplay()
         std::swap(pts[1], pts[3]);
     }
 
+    // Bind OpenGL resources before drawing
+
+    m_oglBuilder->useAllTextures();
```
BernardLefebvre, line 116: Should the calls at the end of setShader be removed?

BernardLefebvre: Maybe updateUniforms() should be removed.

remia: > Should the calls at the end of setShader be removed?

My guess is yes, assuming this behavior is not used somewhere else, but I'd say it would be an error.

> Maybe updateUniforms() should be removed.

Do you mean removing `::updateUniforms()` method but leaving the `m_oglBuilder->useAllUniforms();` call in `::redisplay()` ?

My first impression would be that ociochecklut needs test first, as in without this PR it won't produce any results with the --gpu or --gpulegacy flags I think. Then even with --gpu working, you would still get false results without the proper texture bindings. Looking at the code, there is no tests for OCIO apps currently no ?

hodoulp: > Looking at the code, there is no tests for OCIO apps currently no ?

That's true. But we are open to external contributors with some time to build unit tests for the command line tools :-)

src/OpenColorIO/CMakeLists.txt
```diff
@@ -152,12 +152,14 @@ if(NOT WIN32)
 
     # Install the pkg-config file.
 
+    include(GNUInstallDirs)
+
     set(prefix ${CMAKE_INSTALL_PREFIX})
-    set(exec_prefix "\${prefix}")
-    set(libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
-    set(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
+    set(exec_prefix "${prefix}")
```
hodoulp, line 158: Preserve the '\' so that the generated pkgconfig file still contains `${exec_prefix}` i.e. it does not expand all the file paths.

src/libutils/oglapphelpers/oglapp.cpp
hodoulp: The update of the uniforms is definitively missing from `OglApp::redisplay()`. But it should use `OglApp::updateUniforms()` to do it.

hodoulp:
I think that the latest refactoring to support headless machines (i.e. use of EGL) broke something by inverting some initialization steps.  The original initialization steps were 1) initialize the image texture and 2) create & initialize the color transformation. So, I suggest to change `src/apps/ociochecklut/main.cpp` by moving line 91 (i.e. `m_oglApp->updateImage(pixel.data());`) at line 48.

src/OpenColorIO/CMakeLists.txt
remia: Thanks @hodoulp, in the end the only change left is just the `include(GNUInstallDirs)`

src/libutils/oglapphelpers/oglapp.cpp
remia: Doesn't the texture content changes for each test values you want to check the LUT for (ig. `--t`) ?

`m_oglBuilder->useAllTextures();` seems to be doing both binding and uniform setup, agree that the naming is not obvious.

I'll update to `OglApp::updateUniforms()` then, thanks.

BernardLefebvre: My point is that if uniforms are always updated when a rediaply is done, why would another call to update the uniforms be needed.

hodoulp: > My point is that if uniforms are always updated when a redisplay() is done, why would another call to update the uniforms be needed?

That's a good point. The public method (i.e. `OglApp::updateUniforms()`) is then useless.

> Doesn't the texture content changes for each test values you want to check the LUT for (ig. --t) ?

Only the input texture (i.e. the image to process) changes. The textures related to the color transformation such as those created for the LUTs, do not. And `m_oglBuilder->useAllTextures();` only binds & updates the latter.

hodoulp:
Hi @remia 

Here are my tests highlighting the problem:
`./src/apps/ociochecklut/ociochecklut ./testdata/lut1d_halfdom.ctf 0.12 .9 1`
`./src/apps/ociochecklut/ociochecklut ./testdata/lut1d_halfdom.ctf --gpu 0.12 .9 1`
`./src/apps/ociochecklut/ociochecklut ./testdata/lut1d_halfdom.ctf --gpulegacy 0.12 .9 1`

Included is a patch to apply to your branch to fix the problem. Even if some method names are ambiguous, I tried to minimize the changes to speed-up the merge.

[texture_patch.diff.txt](https://github.com/AcademySoftwareFoundation/OpenColorIO/files/5130410/texture_patch.diff.txt)

remia:
Thanks @hodoulp, good catch, I missed the fact that binding target is linked with the currently activated texture unit.

I applied your changes.

Issue #1084, opened 2020-08-01
A cache for Processors


doug-walker:
Some applications would benefit from having a cache for Processor instances.  For example, some applications import hundreds of files and wind up requesting a conversion between the same pair of color spaces many times.  This is slow and could waste memory.  As another example, timeline-based applications have strict performance requirements and cannot wait to create a new Processor on each shot.

The OCIO context variable mechanism poses challenges in the development of a useful cache.  A naive solution would be to simply invalidate the entire cache whenever the context changes at all.  However, this would make it worthless for timeline-based applications since it is common for at least some part of the context to change on every shot.  Ideally, OCIO would be able to determine which context variables were used for the creation of a given Processor and use this information to invalidate the cache intelligently.

The generation of the cache key needs to be very fast, to make it suitable for timeline-based applications.

In addition to the original Processor, the optimized Processor, the CPUProcessor, and the GPUProcessor would be cached.

Unlike the existing file cache, which is global, the processor cache would be part of a config instance.

The API would provide a way to turn off the caching (e.g. for debugging purposes).

As discussed at a recent OCIO v2 working group meeting, we have been working on a cache and are creating this issue to collect thoughts and requirements.

zachlewis:
This would be great. I'd love to be able to "pre-cache" a processor for a set of set of contexts in applications with multi-shot timelines. 

Would the cache persist when toggling it on and off?

Could the caches be serialized? I.e., as a catalog of ctfs for different permutations of context variables? That could be really handy.

doug-walker:
@zachlewis , yes, an application should be able to build all of the Processors in a "pre cache" step in order to load the cache.  Requesting any of those Processors a second time should be very fast.

What we have will clear the cache when it is turned off.  We feel that this is an important part of turning it off, so we want to keep it that way.  However someone could add a third mode (beyond off and on) that would not flush the cache but also not add any new ones.

Our implementation of the cache itself will not support serialization.  You may already serialize arbitrary Processors as CTF, so maybe that helps for what you have in mind.  Although building the cache is not something you want to do while playing back a timeline, it's not like it takes a really long time to do either, so it's not clear that a feature for serializing the cache is worth the effort.

hodoulp:
@zachlewis, The work is now completed and merged if you see some improvements you can still log another issue.

PR #1130, opened 2020-09-08
Adsk Contrib - Add the support for processor caches

hodoulp:

The pull request implements the feature explained in the issue #1084.

The OpenColorIO library now supports several caches to speed-up the access to the processor instances (i.e. instances from the Processor, CPUProcessor and GPUProcessor classes). All these new caches are attached to an instance (i.e. of Config or Processor classes) to avoid keeping useless memory. For example, a config instance contains a cache of Processor so deleting the instance deletes the cache. At the same time, the cache can be enabled or disabled per instance to provide flexibility for debugging. As investigations could happen at any time, there is also an env. variable to globally disable the processor caches.

The challenge was to correctly handle the context variables which could change the processor content. So, lot of unit tests are actually checking the cache behavior while using context variables. Please refer to the unit test named `(Config, context_variables_typical_use_cases)` in the `src/tests/cpu/Config_tests.cpp` to see the most common use cases.


zachlewis:
Thanks @hodoulp -- I've yet to experiment with the actual PR, but this looks quite encouraging. I appreciate the attention given to context variables.

While exploring #1143, I used these unit tests as reference for the `environment`string var defaults section (even though that issue refers specifically to the undocumented OCIO-1 feature). I ran into a couple of issues I thought might be worthy of consideration:

- String variable value validity with respect to spaces and / or quotes: It may be an OCIO-1-only thing, but it seems quotation marks are needed around values with spaces in the name; i.e. `environment: { FOO: 'Foo - Bar Log - Baz Wide Gamut (CAT02)' }` -- I can't remember if double-quotes behaved differently than single-quotes. Without having tested much further than PyOpenColorIO.Config(...).validate(), it seems that OCIO-2 configs may not require quotes?

- Processor validity with respect to string variable expansion used in arguments for transforms other than FileTransforms; i.e. `!<ColorSpaceTransform> { src: scene_linear, dst: $CAMERA_ENCODING }` (Pertains to #1143 -- `ColorSpaceTransforms` defined in this fashion are behaving as NoOps regardless of the actual resolved string variable values, valid, missing, or not. )


Just in general, it would be really useful to be able to inspect and debug context variables used anywhere in a Processor. 



Lastly -- a little outside the scope of this PR, but it seems like the [current OCIO-2] Python API is missing `PyOpenColorIO.Context().setStringVar(...)` bindings to match the `getStringVar()` instance method -- is the omission accidental?

hodoulp:
@zachlewis Thanks for taking the time to think and investigate the context variables because that's a simple & powerful feature with lot of complex side-effects. If you want to start from my branch (please do so), here is the trick:
From a console, go in your ocio clone directory:
1. `git fetch origin pull/1130/head:pr_1130`
2. `git checkout pr_1130`

Now you have the pull request code that you can compile / test using Python or C++.

hodoulp:
Hi @zachlewis 
Following your comment about the context variable syntax I did some investigations on the accepted syntaxes in a config file. Yaml is extremely permissive so several syntaxes are usual authorized for a type (i.e. a map, a string, etc.).

Here are some successful syntaxes:

```
environment : {}
```

```
environment : { SHOT: file.cdl }
```

```
environment : { SHOT: 'file.cdl' }
```

```
environment : { SHOT: "file.cdl" }
```

```
environment : 
  SHOT: file.cdl
```

zachlewis:
Wow -- thanks @hodoulp! I appreciate that. 
Anecdotally, it seemed the following form was problematic (in OCIO-1):

`
environment :
  CS: colorspace name with spaces
`

whereas this seemed to worked okay:
`environment: { CS: 'colorspace name with spaces' }`

But it's possible -- perhaps likely -- that user error is responsible for the problems experienced with the first example. IIRC, the thrown exception complained about an invalid context variable mapping in the second line. I've yet to reproduce for myself, but I'll investigate further. (In any case, I imagine yaml-cpp is responsible for all the YAML syntax heavy-lifting, and it seems unlikely that OCIO would introduce discrepancies.)

As for the string-variables-in-transform-arguments stuff, I'll spend some time this weekend or early next week working with your branch. (I'm hardly a C++ developer, but I think I can infer what l need to do from surrounding code; at minimum, I can write Python unit tests to build and convert Processors to GroupTransforms, and check that any string variables initially used in transform parameters are resolved appropriately, given the context).

hodoulp:
`Yaml` is exclusively used to read a config file and `ocio` only consumes the parsed results (i.e. no tricks). However, we upgraded the `Yaml` library from 0.3.0 to 0.6.3. That could explain some differences between v1 and v2.

hodoulp:
Given that the pull request has been available for review for 10 days, here is a friendly reminder that we can proceed with the merge next week unless objections.

michdolan:
@zachlewis regarding the Python support for `PyOpenColorIO.Context.setStringVar(...)`, it is present, but utilizing a more Pythonic interface. `Context` now behaves like a Python dictionary, so to set a string var, you would do: `context["var"] = "value"`. There are various places in the new Python API where I attempted to make it act more like Python, which is one of the big needs to documenting the Python bindings.

src/OpenColorIO/transforms/DisplayViewTransform.cpp
```diff
@@ -408,4 +410,95 @@ void BuildDisplayOps(OpRcPtrVec & ops,
         }
     }
 }
+
+bool CollectContextVariables(const Config & config,
```
michdolan, line 414: Could some of this impl be shared with `ColorSpaceTransform` and others?

src/apps/ocioperf/main.cpp
```diff
@@ -237,59 +326,206 @@ int main(int argc, const char **argv)
...
-
             // Get the transform.
             OCIO::FileTransformRcPtr transform = OCIO::FileTransform::Create();
             transform->setSrc(transformFile.c_str());
 
-            // Get the processor
-            processor = config->getProcessor(transform);
+            {
+
+//// TO BE REMOVED BEFORE DOING THE PULL REQUEST
```
michdolan, line 379: Should these noted sections be removed?

src/bindings/python/PyTypes.cpp
```diff
@@ -188,6 +188,13 @@ void bindPyTypes(py::module & m)
...
         .export_values();
 
+    py::enum_<ProcessorCacheFlags>(m, "ProcessorCacheFlags")
+        .value("PROCESSOR_CACHE_OFF", PROCESSOR_CACHE_OFF)
+        .value("PROCESSOR_CACHE_ENABLED", PROCESSOR_CACHE_ENABLED)
+        .value("PROCESSOR_CACHE_SHARE_DYN_PROPERTIES", PROCESSOR_CACHE_SHARE_DYN_PROPERTIES)
+        .value("PROCESSOR_CACHE_DEFAULT", PROCESSOR_CACHE_DEFAULT)
+        .export_values();
+
     // Conversion
```
michdolan, line 198: Should also define the `OCIO_DISABLE_*` variables in here.

src/OpenColorIO/PathUtils.cpp
```diff
@@ -1,50 +1,56 @@
...
+
+// Here is the explanation of the stat() method:
+// https://pubs.opengroup.org/onlinepubs/009695299/basedefs/sys/stat.h.html
+// "The st_ino and st_dev fields taken together uniquely identify the file within the system."
+//
+// However there are limitations to the stat() support on some Windows file systems:
+// https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?redirectedfrom=MSDN&view=vs-2019
+// "The inode, and therefore st_ino, has no meaning in the FAT, HPFS, or NTFS file systems."
+
 std::string ComputeHash(const std::string & filename)
```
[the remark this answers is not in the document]
hodoulp: As discussed that's a great idea to increase customization. Being outside the scope of the pull request, I suggest to create an issue.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -311,14 +320,17 @@ class OCIOEXPORT Config
...
     void addEnvironmentVar(const char * name, const char * defaultValue);
     int getNumEnvironmentVars() const;
     const char * getEnvironmentVarNameByIndex(int index) const;
     const char * getEnvironmentVarDefault(const char * name) const;
     void clearEnvironmentVars();
-    void setEnvironmentMode(EnvironmentMode mode);
-    EnvironmentMode getEnvironmentMode() const;
-    void loadEnvironment();
+
+    void setEnvironmentMode(EnvironmentMode mode) noexcept;
```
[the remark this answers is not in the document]
hodoulp: Being outside the scope of the pull request, I suggest to create a documentation issue.

michdolan (approved):

zachlewis:
> Given that the pull request has been available for review for 10 days, here is a friendly reminder that we can proceed with the merge next week unless objections.

Whether I have something to contribute or not, there are no objections from me. Thanks again!

> @zachlewis regarding the Python support for `PyOpenColorIO.Context.setStringVar(...)`, it is present, but utilizing a more Pythonic interface. `Context` now behaves like a Python dictionary, so to set a string var, you would do: `context["var"] = "value"`. There are various places in the new Python API where I attempted to make it act more like Python, which is one of the big needs to documenting the Python bindings.

That's incredible -- saves me the trouble of writing my own wrapper in Python to get that behavior. Thank you for doing that. I've been really enjoying that Pythonic flavor you've been seasoning the bindings with.

lgritz:
As an aside, if you really like the

    object["attribute"] = value

notation, you may appreciate the trick I use in OIIO for allowing this notation (or at least, as close as we can get) on the C++ side as well: https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/attrdelegate.h#L37

doug-walker (approved):

PR #1134, opened 2020-09-11
Adsk Contrib - Add 2021 VFX Reference Platform requirements

hodoulp:

Following the Open Source Day, I realized that the OCIO build system was missing the C++17 support so I've added it. I also refactored the root CMakeLists.txt to limit its size and increase its readability.

Note: I also put some attention to limit the number of publicly available cmake variables (using `ccmake` or `cmake-gui` for Windows).

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -55,8 +55,7 @@ jobs:
       image: aswf/ci-ocio:${{ matrix.vfx-cy }}
     strategy:
       matrix:
-        # Builds 11 & 12 - pybind11 2.4.3 with clang 9 & C++17 fails but succeeds with clang 10.
-        build: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        build: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
         include:
           # -------------------------------------------------------------------
           # VFX CY2021
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -55,13 +55,40 @@ jobs:
       image: aswf/ci-ocio:${{ matrix.vfx-cy }}
     strategy:
       matrix:
-        build: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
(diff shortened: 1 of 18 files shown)


.github/workflows/ci_workflow.yml
```diff
@@ -55,13 +55,40 @@ jobs:
       image: aswf/ci-ocio:${{ matrix.vfx-cy }}
     strategy:
       matrix:
-        build: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        build: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
```
hodoulp, line 58: @michdolan I was not able to find a way to specific that all builds must be executed.

michdolan: Are you referring to the new builds not being `required` in the checks? Or just a more streamlined approach instead of listing them all like this?

hodoulp: Avoid to list them all i.e. something like ` build: all`

hodoulp: Forget my comment.

pybind11 2.4.3 fails to compile with clang 9 & C++17!
But it succeeds with clang 10 & C++17 so I disabled build 11 & 12 for now.

michdolan: Because of the way GH Action matrix is currently designed, this index listing workaround was my way of supporting a variety of builds without exponentially increasing jobs.  This essentially replicates the Azure Pipelines matrix behavior, instead of GHA's current logic, which does:

matrix:
  a: [1, 2]
  b: [1, 2]

will result in jobs:
a1b1
a2b1
a1b2
a2b2

So with our number of matrix variables the job count becomes unmanagemable. Correcting this is an active GHA request, so hopefully we can improve it later.

hodoulp:
Given that the pull request has been available for review for 10 days, here is a friendly reminder that we can proceed with the merge this week (Friday for example) unless objections.

michdolan (approved):
Nice work! I like the use of external CMake modules  to clean up the main `CMakeLists.txt` file.

doug-walker (approved):

Issue #1087, opened 2020-08-06
Version handling of configs for v2


doug-walker:
OpenColorIO has only had one major version so far.  As we prepare to release v2, we need to do some work related to version management.  This issue is to collect and discuss aspects of this work.

Here is the proposed behavior for your review:

**1. New config objects will default to the current version (v2)**

**2. There will be a helper function for upgrading an older config to the current version**
Note that v2 has certain requirements that v1 does not have (for example, there must be a default file rule).  Hence the need for this item to make it simple to convert a v1 config to v2.

**3. When loading a v1 config, it will remain v1 and would get saved as v1**
Applications will be responsible for updating a config to v2, if desired.

**4. When loading or saving a Yaml file, a check will be made to ensure v2 features are not used in a v1 config**
Currently, sanityCheck is not run when loading or saving Yaml, so this will be a separate function.

**5. SanityCheck will be enhanced to do more version related checking**
For example, it will run the function from item 4, among other checks.

Side Note (not directly related to this issue, but worth mentioning as it's something we ran into when working on this): A new config will *not* pass validation when it is first created.  For example, there must be at least one color space added, among other requirements.

doug-walker:
Closing as this has been implemented.

PR #1147, opened 2020-09-18
Adsk Contrib - Update config default version and optimization enhancements

BernardLefebvre:
Default config version is now the latest version (today version 2).  Unit tests have also been adjusted to take this into account.  This implements issue #1087.

New optimization flags have been added:
- OPTIMIZATION_SIMPLIFY_OPS: break down certain ops into simpler ops when possible. Today, only CDL make use of this, and is replaced by a matrix when possible. The goal is to speed processing and allow more ops to be combined. The flag is enabled with the default optimization.

- OPTIMIZATION_FAST_LOG_EXP_POW: SSE implementation of log, power, and exponent is faster but less precise. When this new flag is set the fast implement is used, else the non-SSE implementation is used. The flag is enabled by default. This applies to several transforms including CDL, Exponent (v2), ExposureContrast, and Log.


zachlewis:
Does this mean that the only way to "create" a v1 config via the API is to load an existing default v1 config?

BernardLefebvre:
> Does this mean that the only way to "create" a v1 config via the API is to load an existing default v1 config?

You can call Config::Create and then set the version. See tests/cpu/Config_tests.cpp line 5178

doug-walker:
As usual this is a notice that we will merge this on Monday, if there are no more discussions on this PR.

include/OpenColorIO/OpenColorTypes.h
```diff
@@ -537,11 +537,18 @@ enum OptimizationFlags : unsigned long
      */
     OPTIMIZATION_LUT_INV_FAST                    = 0x02000000,
 
+    // For CPU processor, in SSE mode, use a faster approximation for log, exp, and pow.
```
michdolan, line 540: These new flags should also be added to the Python bindings:
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/src/bindings/python/PyTypes.cpp

michdolan (changes requested):
One minor omissions. LGTM otherwise.

include/OpenColorIO/OpenColorTypes.h
BernardLefebvre: Added

michdolan: Thanks!

michdolan (approved):

hodoulp (approved):

PR #1151, opened 2020-09-21
Fix ctest -R test_python for Python 3.8

Simran-B:
It couldn't find OpenColorIO_2_0.dll, because PATH is longer used for discovery. Use the new os.add_dll_directory() insead:

> - DLL dependencies for extension modules and DLLs loaded with ctypes on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with add_dll_directory() are searched for load-time dependencies. Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for add_dll_directory() and if it exists, use it to add your DLLs directory while loading your library. [...]

https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-python-api


hodoulp (approved):
Good job.

doug-walker (approved):

PR #1154, opened 2020-09-23
Add id and description to CDLTransform YAML

michdolan:
This fixes the issue @KelSolaar found with ID and description not being serialized to the config YAML. `name` was used in the config IO and tests, which passed tests since it became legitimate metadata, but ID and description were bypassed.

### DIFF
--- a/src/OpenColorIO/OCIOYaml.cpp
+++ b/src/OpenColorIO/OCIOYaml.cpp
@@ -579,7 +579,7 @@ inline void load(const YAML::Node& node, CDLTransformRcPtr& t)
 
     CheckDuplicates(node);
 
-    std::string key;
+    std::string key, stringval;
     std::vector<double> floatvecval;
 
     for (Iterator iter = node.begin();
@@ -637,9 +637,8 @@ inline void load(const YAML::Node& node, CDLTransformRcPtr& t)
         }
         else if (key == "style")
         {
-            std::string style;
-            load(second, style);
-            t->setStyle(CDLStyleFromString(style.c_str()));
+            load(second, stringval);
+            t->setStyle(CDLStyleFromString(stringval.c_str()));
         }
         else if (key == "direction")
         {
@@ -647,11 +646,15 @@ inline void load(const YAML::Node& node, CDLTransformRcPtr& t)
             load(second, val);
             t->setDirection(val);
         }
-        else if (key == "name")
+        else if (key == "id")
         {
-            std::string name;
-            load(second, name);
-            t->getFormatMetadata().addAttribute(METADATA_NAME, name.c_str());
+            load(second, stringval);
+            t->setID(stringval.c_str());
+        }
+        else if (key == "description")
+        {
+            load(second, stringval);
+            t->setDescription(stringval.c_str());
         }
         else
         {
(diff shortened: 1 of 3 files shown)


KelSolaar:
Cool, thanks for the prompt patch @michdolan !

src/OpenColorIO/OCIOYaml.cpp
```diff
@@ -637,21 +637,24 @@ inline void load(const YAML::Node& node, CDLTransformRcPtr& t)
...
+            load(second, stringval);
+            t->setStyle(CDLStyleFromString(stringval.c_str()));
         }
         else if (key == "direction")
         {
             TransformDirection val;
             load(second, val);
             t->setDirection(val);
         }
-        else if (key == "name")
```
hodoulp, line 650: The 'name' refers to a text name usually for display purpose. All transforms now have optional 'id', 'name' & 'description' attributes. So, the change needs to add the 'id' and 'description' serialization & deserialization but to keep 'name'. Anyway, all these attributes are saved in the metadata attribute i.e. 'm_metadata' in the class OpData.

A solution could be to update 'EmitTransformName(.., const FormatMetadata & metadata)' in OCIOYaml.cpp to save 'id', 'name' and 'description' is not empty.

hodoulp (changes requested):

src/OpenColorIO/OCIOYaml.cpp
BernardLefebvre: Note that API functions for description are described as deprecated, we should understand why @KelSolaar needs this to be saved in YALM. Function related to ID should probably be described as deprecated.

These are meaning full when the CDL comes from cc or ccc files.

Several transforms do have a meta-data structure that is saved to CLF/CTF. That structure has been used to hold the data from cc & ccc files for CDL.

From metadata, we currently only serialize name to YALM.

michdolan: Thomas' interest in this feature was only because I suggested using it for a placeholder for `BuiltinTransform` in the ACES config under development. But the issue this is trying to fix is that `CDLTransform` does have public API for setting `id` and `description`, but neither are persistent right now. They work in the immediate transform but aren't serialized to YAML. This also breaks compatibility with OCIO v1.

BernardLefebvre: I don't understand what OCIO v1 compatibility is broken. V1 s not saving these in YALM.
If the goal is to serialize everything, then the complete metadata has to be serialized (for all transforms handling it).

michdolan: Yeah you're right. I was mistaken. Perhaps a better question then is what's the point of have `CDLTransform::setID` and `CDLTransform::setDescription` if they aren't serializable? If the only time these are set is when reading `cc` or `ccc`, the public setters are deceiving at best. 

In the current docs, this comment advertises these as serializable, and the `description` setter/getter is marked deprecated:
```
// These do not affect the image processing, but
// are often useful for pipeline purposes and are
// included in the serialization.
```

michdolan: To note though, I do think all metadata should be serialized if it can be set on a transform by a user. Otherwise it should be read-only and limited to reading external files IMO.

michdolan: I guess the other case for writable metadata is when serializing to CLF/CTF, but regardless, setters should be persistent, otherwise the ability to write the same CLF/CTF from a config is not repeatable without pipeline code.

michdolan: I'm going to close this PR and move the conversation to an issue.

PR #1140, opened 2020-09-14
Document OpenColorIO Development Process

michdolan:
Following discussion on transparency and openness of new feature development, and related TSC discussion, I've added a PROCESS.md document to outline a feature proposal process. The process is based on current practices adhered to by the project with two changes:

1. Push up proposal discussion and issue creation in the timeline where possible. In short, it's up to the contributor when a design is proposed, but accepting the risk of potential redesign requested by the community.
2. Adding accepted proposals to a new Roadmap GH Project for progress tracking and transparency.

All of this is up for discussion of course. All comments/suggestions are welcome and appreciated.

Resolves #1124 

Signed-off-by: Michael Dolan <michdolan@gmail.com>

PR #1158, opened 2020-09-30
Revert "Document OpenColorIO Development Process"

doug-walker:
Reverts AcademySoftwareFoundation/OpenColorIO#1140


michdolan:
Instead of reverting this, I propose any suggested changes are presented in a follow up issue or PR. The sections @KelSolaar flagged are present in `CONTRIBUTING.md` even with this reverted, so the policy is present in either case.

michdolan:
@KelSolaar on the question of sharing this document for community input. I raised it for discussion at the last three TSC meetings requesting input and additional reviewers:

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/ASWF/meetings/tsc/2020-09-14.md
https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1156/files
(this week's TSC meeting, which I have not submitted minutes for yet)

To be fair though, it was revised quite a bit in the last week from discussion between @doug-walker and I. That discussion primarily led to the removal of some conditions which weren't used in practice. I think this document is still open for discussion and revision though, so by all means let's continue to refine it to get it right.

KelSolaar:
To me, the document is in a state where the larger community should be asked for feedback but not just in TSC, wider, e.g. Slack and Mailing list as not everybody is attending the TSC or reading the notes. If no feedback is given, we can assume it is happy :)

michdolan:
I'm closing this to move discussion to a new issue for further discussion: #1161 

I will invite the community there for additional comment.

PR #1159, opened 2020-09-30
Adsk Contrib - Improve CMake files

hodoulp:

The pull request mainly improves `CMake` files to better handle various compilation contexts (mainly Windows platform but not only):

-  `CMAKE_INSTALL_MESSAGE` is to decrease the number of 'useless outputs' to better highlight 'important outputs'.
-  `CMAKE_OBJECT_PATH_MAX` is to increase the default path max size from 255 (the default) to 300. Unfortunately, that limit can be reached on Windows platform with Visual Studio. As discussed during a TSC meeting, the 'propagation' of CMake flags to external third-party libraries is still problematic but the CMake files are all updated.
- Clean and improve the output binary names to include the major and minor version.


hodoulp:
As usual this is a notice that we will merge this on Tuesday, if there are no comments on the pull request.

michdolan (approved):

doug-walker (approved):

share/cmake/modules/Findpystring.cmake
```diff
@@ -77,6 +77,10 @@ endif()
 if(NOT pystring_FOUND)
     include(ExternalProject)
 
+    if(APPLE)
+        set(CMAKE_OSX_DEPLOYMENT_TARGET ${CMAKE_OSX_DEPLOYMENT_TARGET})
+    endif()
+
```
z-aki, line 83: May I ask what's the intention behind this set of change ? This should've been appended to `pystring_CMAKE_ARGS` so that pystring is built for the same minimum macOS version as OCIO. 
Same applies to other find modules where the same patterns is applied.

PR #1169, opened 2020-10-17
Added custom callback for ComputeHash function

### LINKED ISSUE #1150: Add support for custom ComputeHash function/callback
The current `ComputeHash` function in `PathUtils.cpp` depends on stat info from a supported filesystem. There could however be cases where this data is not sufficient for invalidating a cache (the filesystem might just be a proxy for a VCS system, for example), or a DCC could embed a file and use that when a config is not present. In these alternative cases, a function could be passed to OCIO to override the builtin hashing mechanism.

I would propose following a similar pattern as used in the current global `SetLoggingFunction` in the public API:

```
/// Define the logging function signature.
using LoggingFunction = std::function<void(const char*)>;
...
extern OCIOEXPORT void SetLoggingFunction(LoggingFunction logFunction);
```

Like:

```
using ComputeHashFunction = std::function<void(const std::string &)>;
...
SetComputeHashFunction(LoggingFunction logFunction);
```

EmperorYP7:
fixes #1150 

Description: Declared a global container to store ComputeHash function and provided an endpoint to the (relevant) client for generating a custom callback for the same.

(Still figuring out to make appropriate unit tests for this.)

### DIFF
--- a/include/OpenColorIO/OpenColorIO.h
+++ b/include/OpenColorIO/OpenColorIO.h
@@ -161,6 +161,14 @@ extern OCIOEXPORT void ResetToDefaultLoggingFunction();
 /// Log a message using the library logging function.
 extern OCIOEXPORT void LogMessage(LoggingLevel level, const char * message);
 
+/**
+ * \brief Set the Compute Hash Function to use; otherwise, use the default.
+ * 
+ * \param ComputeHashFunction
+ */
+extern OCIOEXPORT void SetComputeHashFunction(ComputeHashFunction hashFunction);
+extern OCIOEXPORT void ResetComputeHashFunction();
+
 //
 // Note that the following environment variable access methods are not thread safe.
 //
--- a/src/OpenColorIO/PathUtils.cpp
+++ b/src/OpenColorIO/PathUtils.cpp
@@ -26,8 +26,6 @@ namespace OCIO_NAMESPACE
 {
 namespace
 {
-
-
 // Here is the explanation of the stat() method:
 // https://pubs.opengroup.org/onlinepubs/009695299/basedefs/sys/stat.h.html
 // "The st_ino and st_dev fields taken together uniquely identify the file within the system."
@@ -36,7 +34,8 @@ namespace
 // https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?redirectedfrom=MSDN&view=vs-2019
 // "The inode, and therefore st_ino, has no meaning in the FAT, HPFS, or NTFS file systems."
 
-std::string ComputeHash(const std::string & filename)
+// That's the default hash method implementation to compute a hash key based on a file content.
+std::string DefaultComputeHash(const std::string &filename)
 {
     struct stat fileInfo;
     if (stat(filename.c_str(), &fileInfo) == 0)
@@ -57,6 +56,10 @@ std::string ComputeHash(const std::string & filename)
     return "";
 }
 
+// The global variable holds the hash function to use.
+// It could be changed using SetComputeHashFunction() to customize the implementation.
+ComputeHashFunction g_hashFunction = DefaultComputeHash;
+
 // We mutex both the main map and each item individually, so that
 // the potentially slow stat calls dont block other lookups to already
 // existing items. (The stat calls will block other lookups on the
@@ -74,6 +77,16 @@ typedef std::map<std::string, FileHashResultPtr> FileCacheMap;
 
 FileCacheMap g_fastFileHashCache;
 Mutex g_fastFileHashCache_mutex;
+}
+
+void SetComputeHashFunction(ComputeHashFunction hashFunction)
+{
+    g_hashFunction = hashFunction;
+}
+
+void ResetComputeHashFunction()
+{
+    g_hashFunction = DefaultComputeHash;
 }
 
 std::string GetFastFileHash(const std::string & filename)
@@ -101,7 +114,7 @@ std::string GetFastFileHash(const std::string & filename)
             // NB: OCIO does not attempt to detect if files have changed and caused the cache to
             // become stale.
             fileHashResultPtr->ready = true;
-            fileHashResultPtr->hash = ComputeHash(filename);
+            fileHashResultPtr->hash = g_hashFunction(filename);
         }
 
         hash = fileHashResultPtr->hash;
--- a/src/OpenColorIO/PathUtils.h
+++ b/src/OpenColorIO/PathUtils.h
@@ -11,6 +11,7 @@
 
 namespace OCIO_NAMESPACE
 {
+
 // This is not currently included in pystring, but we need it
 // So let's define it locally for now
 
--- a/tests/cpu/CMakeLists.txt
+++ b/tests/cpu/CMakeLists.txt
@@ -5,254 +5,254 @@
 add_definitions("-DOCIO_SOURCE_DIR=${CMAKE_SOURCE_DIR}")
 
 function(add_ocio_test NAME SOURCES PRIVATE_INCLUDES)
-	set(TEST_BINARY "test_${NAME}_exec")
-	set(TEST_NAME "test_${NAME}")
-	add_executable(${TEST_BINARY} ${SOURCES})
-	target_compile_definitions(${TEST_BINARY}
-		PRIVATE
-			OpenColorIO_SKIP_IMPORTS
+    set(TEST_BINARY "test_${NAME}_exec")
+    set(TEST_NAME "test_${NAME}")
+    add_executable(${TEST_BINARY} ${SOURCES})
+    target_compile_definitions(${TEST_BINARY}
+        PRIVATE
+            OpenColorIO_SKIP_IMPORTS
 
-	)
-	target_link_libraries(${TEST_BINARY}
-		PUBLIC
-			public_api
-		PRIVATE
-			expat::expat
-			IlmBase::Half
-			pystring::pystring
-			sampleicc::sampleicc
-			unittest_data
-			utils::strings
-			yaml-cpp
-			testutils
-	)
+    )
+    target_link_libraries(${TEST_BINARY}
+        PUBLIC
+            public_api
+        PRIVATE
+            expat::expat
+            IlmBase::Half
+            pystring::pystring
+            sampleicc::sampleicc
+            unittest_data
+            utils::strings
+            yaml-cpp
+            testutils
+    )
 
-	if(APPLE)
-		# Frameworks needed to access the ICC monitor profile. 
-		target_link_libraries(${TEST_BINARY}
-			PRIVATE
-				"-framework Carbon"
-				"-framework IOKit"
-		)
-	endif(APPLE)
+    if(APPLE)
+        # Frameworks needed to access the ICC monitor profile. 
+        target_link_libraries(${TEST_BINARY}
+            PRIVATE
+                "-framework Carbon"
+                "-framework IOKit"
+        )
+    endif(APPLE)
 
-	if(PRIVATE_INCLUDES)
-		target_include_directories(${TEST_BINARY}
-			PRIVATE
-				"${CMAKE_SOURCE_DIR}/src/OpenColorIO"
-				"${CMAKE_SOURCE_DIR}/tests/cpu"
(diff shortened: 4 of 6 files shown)


src/OpenColorIO/PathUtils.cpp
```diff
@@ -36,7 +34,8 @@ namespace
 // https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?redirectedfrom=MSDN&view=vs-2019
 // "The inode, and therefore st_ino, has no meaning in the FAT, HPFS, or NTFS file systems."
 
-std::string ComputeHash(const std::string & filename)
```
michdolan, line 39: Instead of removing `ComputeHash`, only the implementation of `ComputeHash` could handle using an optionally provided function or the default algorithm. Callers of `ComputeHash` should not know or care about that implementation being the default or a provided function, but you could provide the `DefaultComputeHash` function in addition as a means of guaranteeing the standard behavior when a caller needs it.

michdolan (changes requested):

michdolan:
Thanks for working on a contribution @EmperorYP7

tests/cpu/UnitTestPathUtils.h
```diff
@@ -0,0 +1,31 @@
...
+    std::ostringstream testHasher;
+    testHasher << std::hash<std::string>{}(filename + "this is custom hash");
+    return testHasher.str();
+}
+
+} //anon.
+
+} // namespace OCIO_NAMESPACE
+
+#endif //INCLUDED_OCIO_UNITTEST_PATHUTILS_H
```
hodoulp, line 31: The unit test framework is a custom mechanism where we do not add include files. Please move the method definition in the corresponding cpp file (and then delete that new file).

tests/cpu/PathUtils_tests.cpp
```diff
@@ -15,7 +16,22 @@ OCIO_ADD_TEST(PathUtils, compute_hash)
...
+    OCIO_CHECK_EQUAL(OCIO::g_hashFunction(file1), OCIO::g_hashFunction(file1));
 
-    OCIO_CHECK_NE(OCIO::ComputeHash(file1), OCIO::ComputeHash(file2));
+    OCIO_CHECK_NE(OCIO::g_hashFunction(file1), OCIO::g_hashFunction(file2));
+
+    const std::string result1 = OCIO::g_hashFunction(file1);
+    const std::string result2 = OCIO::g_hashFunction(file2);
+
+    {
+        OCIO::g_hashFunction = OCIO::CustomComputeHash;
```
hodoulp, line 27: You should use the public method to change the method implementation .

tests/cpu/PathUtils_tests.cpp
```diff
@@ -15,7 +16,22 @@ OCIO_ADD_TEST(PathUtils, compute_hash)
...
+
+        OCIO_CHECK_NE(OCIO::g_hashFunction(file1), result1);
+
+        OCIO_CHECK_NE(OCIO::g_hashFunction(file2), result2);
+
+        OCIO_CHECK_EQUAL(OCIO::g_hashFunction(file1), OCIO::g_hashFunction(file1));
+
+        OCIO::g_hashFunction = OCIO::DefaultComputeHash;
```
hodoulp, line 35: The code must use the public method to reset the method implementation.

hodoulp: ... and then add another test to demonstrate that the has key is now different from the previously compute hash keys.

hodoulp (changes requested):
@EmperorYP7 I think the approach is a good one following what was done for the logger mechanism. Please look at []/src/OpenColorIO/Config.cpp and []/include/OpenColorIO/OpenColorIO.h for details. But the implementation needs some improvements.

src/OpenColorIO/PathUtils.cpp
michdolan: Yeah, you're right. I was thinking that `ComputeHash` was an entry point, but I see it's only `GetFastFileHash`.

EmperorYP7: So, should I let it be `DefaultComputeHash` or make the function available as `ComputeHash`?

michdolan: Keep `DefaultComputeHash`. Disregard my request.

src/OpenColorIO/PathUtils.h
```diff
@@ -11,6 +11,14 @@
...
 namespace OCIO_NAMESPACE
 {
+
+//define Compute Hash function signature.
+using ComputeHashFunction = std::function<std::string(const std::string &)>;
+
+void SetComputeHashFunction(ComputeHashFunction hashFunction);
+
+void ResetComputeHashFunction();
+
```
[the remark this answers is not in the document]
EmperorYP7: The build failed everytime I tried to include this to the public API. Maybe its something wrong with my implementation. I'll work on this. Thanks! :smile:

tests/cpu/UnitTestPathUtils.h
EmperorYP7: I'll surely keep that in mind! I was trying to make a `PathUtilsTester` class for invoking functions from the public API but it didn't work for me so well. Should I be implementing a class like this anyways?

tests/cpu/PathUtils_tests.cpp
EmperorYP7: Working on this sir.

tests/cpu/PathUtils_tests.cpp
EmperorYP7: On it!

tests/cpu/CMakeLists.txt
```diff
@@ -239,7 +239,8 @@ set(TESTS
 	transforms/Lut3DTransform_tests.cpp
 	transforms/MatrixTransform_tests.cpp
 	transforms/RangeTransform_tests.cpp
-	UnitTestLogUtils.cpp
+    UnitTestLogUtils.cpp
+    UnitTestPathUtils.cpp
```
EmperorYP7, line 243: The indentation in all `CMakeLists.txt` files (\t) is not consistent with the code. (4 whitespaces) Is it intensional?

hodoulp: Yes, some CMakeLists.txt are not consistent with the coding which mentions 4 whitespaces and no tabulation so you can fix that.

EmperorYP7: Ok sir. Fixed it.

tests/cpu/UnitTestPathUtils.h
hodoulp: If you need an helper class (like the `PathUtilsTester`) then the code must declare & define it in the corresponding test file. For example, the class `PathUtilsTester` include and cpp file content must be moved in PathUtils_tests.cpp.

Note: The class name should be `HashGuard` or `ComputeHashGuard` to better highlight that it only cares of the `ComputeHashFunction`, and it restores the default implementation when the test completes.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -161,6 +161,17 @@ extern OCIOEXPORT void ResetToDefaultLoggingFunction();
...
+/**
+ * @brief Set the Compute Hash Function object to use; otherwise, use the
+ * default.
+ * 
+ * @param hashFunction 
+ * @return OCIOEXPORT 
+ */
+extern OCIOEXPORT void SetComputeHashFunction(ComputeHashFunction hashFunction);
+extern OCIOEXPORT void ResetComputeHashFunction();
+extern OCIOEXPORT std::string ComputeHash(const std::string & filename);
```
hodoulp, line 173: There is no need for now to directly access to the `ComputeHash` method so that line must ne removed.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -161,6 +161,17 @@ extern OCIOEXPORT void ResetToDefaultLoggingFunction();
 /// Log a message using the library logging function.
 extern OCIOEXPORT void LogMessage(LoggingLevel level, const char * message);
 
+/**
+ * @brief Set the Compute Hash Function object to use; otherwise, use the
+ * default.
+ * 
+ * @param hashFunction
```
hodoulp, line 168: Please remove the two following lines. BTW the second line is in fact wrong. `OCIOEXPORT` is not the return type but a macro to export or not the function C++ symbol.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -161,6 +161,17 @@ extern OCIOEXPORT void ResetToDefaultLoggingFunction();
 /// Log a message using the library logging function.
 extern OCIOEXPORT void LogMessage(LoggingLevel level, const char * message);
 
+/**
+ * @brief Set the Compute Hash Function object to use; otherwise, use the
```
hodoulp, line 165: [Minor] For now, the header is using `\brief` instead of `@brief`.

src/OpenColorIO/PathUtils.cpp
```diff
@@ -36,7 +34,8 @@ namespace
...
+// Default Compute Hash Function
```
hodoulp, line 37: The convention is to have sentences for explanations.
// That's the default hash method implementation to compute a hash key based on a file content.

src/OpenColorIO/PathUtils.cpp
```diff
@@ -57,6 +56,9 @@ std::string ComputeHash(const std::string & filename)
     return "";
 }
 
+//Holding default hash function
```
hodoulp, line 59: As mentioned that should be a sentence:
// The global variable holds the hash function to use. It could be changed using SetComputeHashFunction() to customize the implementation.

tests/cpu/UnitTestPathUtils.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <OpenColorIO/OpenColorIO.h>
+
+#include "UnitTestPathUtils.h"
+
+namespace OCIO_NAMESPACE
+{
+namespace
+{
+
+std::string CustomComputeHash(const std::string & filename)
```
hodoulp, line 17: The content of this file should be moved in tests/cpu/PathUtils_tests.cpp

tests/cpu/UnitTestPathUtils.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <OpenColorIO/OpenColorIO.h>
+
+namespace OCIO_NAMESPACE
+{
+
+namespace
+{
+
+//A custom compute hash function for testing
```
hodoulp, line 15: The content of this file should be moved in tests/cpu/PathUtils_tests.cpp

hodoulp (changes requested):
Great improvement. There are still some changes to do but you are definitively in the right direction.

include/OpenColorIO/OpenColorIO.h
EmperorYP7: Most of this brief was generated by my IDE. Completely missed out on this! I'll surely fix it sir.

include/OpenColorIO/OpenColorIO.h
EmperorYP7: Removed it sir.

include/OpenColorIO/OpenColorIO.h
EmperorYP7: Fixed.

src/OpenColorIO/PathUtils.cpp
EmperorYP7: Updated it. 👍🏻

EmperorYP7:
I think commit history is messed up. I'll squash these commits into one to make it look cleaner.

hodoulp (approved):
Good job.

doug-walker (approved):
Nice work.

EmperorYP7:
Thank you for your guidance sirs! :smile:

hodoulp:
@michdolan The pull request is now waiting for your feedback. When you will have some time could you finalize your review?

michdolan (approved):
Excellent work @EmperorYP7 . LGTM.

PR #1172, opened 2020-10-19
Adsk Contrib - Improve the FileRules behaviour for some limit cases

hodoulp:

The `file rules` behaviour for some limit cases was not enough documented, and not consistent between the two types of file rules. The pull request fixes these two aspects. When creating or updating a file rule, a null or empty string for file name pattern, file extension pattern or regular expression now fail (by throwing an exception), and the public header is updated accordingly.


michdolan (approved):

hodoulp:
As usual this is a notice that we will merge next week (i.e. Monday or Tuesday), if there are no discussions on this pull request.

doug-walker (approved):

PR #1176, opened 2020-10-21
Adsk Contrib - Default the family separator to '/' to comply with existing v1 ACES configs

hodoulp:

The pull request changes the family separator behaviour (added in v2) to always default to '/' (instead of empty for v1 and '/' for v2) so it can effortlessly support existing v1 ACES configs.

When loading config from any version or creating a config instance, the family separator defaults to '/' (or it takes the value from v2 configs if the key is present) but only v2 config supports writing the family separator (It writes it only if different from the default value). The pull request also includes an updated unit test to clearly demonstrate the new behaviour, and the differences between v1 and v2.

### DIFF
--- a/include/OpenColorIO/OpenColorIO.h
+++ b/include/OpenColorIO/OpenColorIO.h
@@ -272,7 +272,7 @@ public:
     void setMinorVersion(unsigned int minor);
 
     /// Allows an older config to be serialized as the current version.
-    void upgradeToLatestVersion();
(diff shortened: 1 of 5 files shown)


KelSolaar:
Nice!

michdolan (approved):

hodoulp:
As usual this is a notice that we will merge this week, if there are no discussions on the pull request.

doug-walker (approved):

Issue #1178, opened 2020-10-22
LookTransform Inverse direction issue


remia:
We stumbled upon a rather unusual config using the `LookTransform` inside a `GroupTransform`, it seems that the config breaks when the `LookTransform` is used in the inverse direction. A simplified version of the problematic config:
```
  - !<ColorSpace>
    name: vd_graded
    to_reference: !<GroupTransform>
      children:
        - !<LookTransform> {src: raw, dst: vd, looks: look1, direction: inverse}
```

The `LookTransform` is probably missing a `CombineTransformDirections` step that prevent it from working correctly in a situation like this. We were under an old OCIO version but the issue appears to be in master too.

Here is a quick path and dirty unit test to reproduce: https://gist.github.com/remia/2a928bc32bb1e9b514dff30f98782b2a

I wonder if there would be a way to perform this `CombineTransformDirections` at a higher level to avoid this kind of issue ?

doug-walker:
Good catch @remia.  Surprisingly, LookTransform seems to ignore the direction of the LookTransform object itself (unlike GroupTransform, ColorSpaceTransform, and DisplayViewTransform, which could be considered its peer Transforms).

The problem seems independent of its usage inside a GroupTransform, isn't it?  GroupTransform seems to handle that correctly already.  Why are you thinking it would be better to apply the CombineDirs at a higher level?

At first glance, your test and fix look good.  Would you make a PR on master with your patch?  We will then review more carefully.  Thanks in advance!

remia:
What I meant was basically is there a reason that each respective transform OpBuilders need to have the two directions provided as input or would the combined be enough ? Seeing that it is easy to forget it like here. And you are right, it's probably not limited to usage via a GroupTransform.

Anyway, the fix is simple, I will roll a PR for that one.

PR #1180, opened 2020-10-23
Fix LookTransform inverse direction

remia:
This address the bug found in #1178

### DIFF
--- a/src/OpenColorIO/transforms/LookTransform.cpp
+++ b/src/OpenColorIO/transforms/LookTransform.cpp
@@ -326,8 +326,8 @@ void BuildLookOps(OpRcPtrVec & ops,
     LookParseResult looks;
     looks.parse(lookTransform.getLooks());
 
-    // We must handle the inverse src/dst colorspace transformation explicitly.
-    auto combinedDir = CombineTransformDirections(dir, lookTransform.getDirection());
+    // The code must handle the inverse src/dst colorspace transformation explicitly.
+    const auto combinedDir = CombineTransformDirections(dir, lookTransform.getDirection());
     if(combinedDir == TRANSFORM_DIR_INVERSE)
     {
         std::swap(src, dst);
--- a/src/OpenColorIO/transforms/LookTransform.cpp
+++ b/src/OpenColorIO/transforms/LookTransform.cpp
@@ -327,12 +327,13 @@ void BuildLookOps(OpRcPtrVec & ops,
     looks.parse(lookTransform.getLooks());
 
     // We must handle the inverse src/dst colorspace transformation explicitly.
-    if(dir == TRANSFORM_DIR_INVERSE)
+    auto combinedDir = CombineTransformDirections(dir, lookTransform.getDirection());
+    if(combinedDir == TRANSFORM_DIR_INVERSE)
     {
         std::swap(src, dst);
         looks.reverse();
     }
-    else if(dir == TRANSFORM_DIR_UNKNOWN)
+    else if(combinedDir == TRANSFORM_DIR_UNKNOWN)
     {
         std::ostringstream os;
         os << "BuildLookOps error. A valid transform direction must be specified.";
--- a/tests/cpu/transforms/LookTransform_tests.cpp
+++ b/tests/cpu/transforms/LookTransform_tests.cpp
@@ -714,7 +714,7 @@ colorspaces:
                                            dstColorSpaceInv, srcColorSpace, false));
     OCIO_REQUIRE_EQUAL(ops2.size(), ops.size());
     OCIO_CHECK_NO_THROW(ops2.validate());
-    for (int i = 0; i < ops2.size(); ++i)
+    for (std::size_t i = 0; i < ops2.size(); ++i)
     {
       OCIO::ConstOpRcPtr op = ops[i];
       OCIO::ConstOpRcPtr op2 = ops2[i];
--- a/tests/cpu/transforms/LookTransform_tests.cpp
+++ b/tests/cpu/transforms/LookTransform_tests.cpp
@@ -2,7 +2,10 @@
 // Copyright Contributors to the OpenColorIO Project.
 
 
+#include "ops/cdl/CDLOpData.h"
 #include "ops/fixedfunction/FixedFunctionOpData.h"
+#include "ops/gamma/GammaOpData.h"
+#include "ops/log/LogOpData.h"
 
 #include "transforms/LookTransform.cpp"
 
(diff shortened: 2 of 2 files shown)


src/OpenColorIO/transforms/LookTransform.cpp
```diff
@@ -327,12 +327,13 @@ void BuildLookOps(OpRcPtrVec & ops,
     looks.parse(lookTransform.getLooks());
 
     // We must handle the inverse src/dst colorspace transformation explicitly.
```
Dithermaster, line 329: Is this comment accurate for the new code? (What does "handle explicitly" mean?)

doug-walker: Good question Dennis.  The comment is from v1 and I'm not sure what it meant either.  In any case, I don't think Remi's change affects the validity of it.  The lookTransform direction was being ignored before Remi's change, which is clearly a bug.

tests/cpu/transforms/LookTransform_tests.cpp
```diff
@@ -559,3 +562,191 @@ environment: { FILE1: cdl_test1.cc, FILE2: cdl_test1.cc }
...
+    OCIO::ConstColorSpaceRcPtr dstColorSpace2;
+    OCIO_CHECK_NO_THROW(dstColorSpace2 = config->getColorSpace("vd_graded_2"));
+
+    OCIO::OpRcPtrVec ops2;
+    OCIO_CHECK_NO_THROW(BuildColorSpaceOps(ops2, *config, config->getCurrentContext(),
+                                           dstColorSpace2, srcColorSpace, false));
+    OCIO_REQUIRE_EQUAL(ops2.size(), 11);
+    OCIO_CHECK_NO_THROW(ops2.validate());
+    op = ops2[0];
+    OCIO_CHECK_ASSERT(op->isNoOpType());
```
doug-walker, line 719: Looks like OpRcPtrVec does not have an op==, which could shorten the test and make it easier to read.  Maybe Patrick will have a suggestion for how to write it more compactly.  But what you have here looks correct, so no objection to leaving it as is.

tests/cpu/transforms/LookTransform_tests.cpp
```diff
@@ -559,3 +562,191 @@ environment: { FILE1: cdl_test1.cc, FILE2: cdl_test1.cc }
...
+    OCIO_REQUIRE_EQUAL(op->data()->getType(), OCIO::OpData::LogType);
+    lg = OCIO_DYNAMIC_POINTER_CAST<const OCIO::LogOpData>(op->data());
+    OCIO_CHECK_ASSERT(lg);
+    OCIO_CHECK_EQUAL(lg->getDirection(), OCIO::TRANSFORM_DIR_INVERSE);
+    op = ops[9];
+    OCIO_CHECK_ASSERT(op->isNoOpType());
+    op = ops[10];
+    OCIO_CHECK_ASSERT(op->isNoOpType());
+
+    // Compare to vd_graded_2 that should be identical (only diference is that
```
doug-walker, line 708: diference --> difference

doug-walker (approved):
Thanks Remi!

tests/cpu/transforms/LookTransform_tests.cpp
remia: Fixed, thanks

tests/cpu/transforms/LookTransform_tests.cpp
remia: Waiting for Patrick feedback, here is a shorter version I came up with just in case:
```
OCIO_REQUIRE_EQUAL(ops2.size(), ops.size());
OCIO_CHECK_NO_THROW(ops2.validate());
for (int i = 0; i < ops2.size(); ++i)
{
   OCIO::ConstOpRcPtr op = ops[i];
   OCIO::ConstOpRcPtr op2 = ops2[i];
   OCIO_REQUIRE_ASSERT(*op->data() == *op2->data());
}
```

src/OpenColorIO/transforms/LookTransform.cpp
remia: I understood that as a just a reminder that src and dst colorspace needs to be swapped, but I may have misunderstood the purpose of this comment, I guess it could be removed if it introduce more confusion that anything.

hodoulp: I also understand the comment as a reminder. I would only change the beginning of the sentence to `The code must handle ...`

tests/cpu/transforms/LookTransform_tests.cpp
```diff
@@ -559,3 +562,191 @@ environment: { FILE1: cdl_test1.cc, FILE2: cdl_test1.cc }
...
+    name: log
+    to_reference: !<LogTransform> {base: 2, direction: inverse}
+
+  - !<ColorSpace>
+    name: vd
+    from_reference: !<ExponentTransform> {value: [2.4, 2.4, 2.4, 1], direction: inverse}
+
+  - !<ColorSpace>
+    name: vd_graded
+    from_reference: !<GroupTransform>
```
hodoulp, line 603: I think that the `GroupTransform` level does not add any useful information to the unit test so it can be removed to simplify the config. The same comment applies below.

tests/cpu/transforms/LookTransform_tests.cpp
```diff
@@ -559,3 +562,191 @@ environment: { FILE1: cdl_test1.cc, FILE2: cdl_test1.cc }
...
+      children:
+        - !<LookTransform> {src: raw, dst: vd, looks: look1}
+
+  - !<ColorSpace>
+    name: vd_graded_2
```
hodoulp, line 608: I would suggest to rename this color space to `vd_graded_inverse` to better highlight the difference with the previous color space.

src/OpenColorIO/transforms/LookTransform.cpp
```diff
@@ -327,12 +327,13 @@ void BuildLookOps(OpRcPtrVec & ops,
...
-    if(dir == TRANSFORM_DIR_INVERSE)
+    auto combinedDir = CombineTransformDirections(dir, lookTransform.getDirection());
```
hodoulp, line 330: The variable must be a `const`.

hodoulp (changes requested):
Good job. I only have few minor comments to improve the understanding of the unit test.

src/OpenColorIO/transforms/LookTransform.cpp
remia: Fixed

tests/cpu/transforms/LookTransform_tests.cpp
remia: Updated the test

tests/cpu/transforms/LookTransform_tests.cpp
```diff
@@ -559,3 +562,191 @@ environment: { FILE1: cdl_test1.cc, FILE2: cdl_test1.cc }
     OCIO_CHECK_EQUAL(std::string("FILE2"), usedContextVars->getStringVarNameByIndex(1));
     OCIO_CHECK_EQUAL(std::string("cdl_test1.cc"), usedContextVars->getStringVarByIndex(1));}
 
+
+OCIO_ADD_TEST(LookTransform, inline_look_transform)
+{
+    constexpr const char * OCIO_CONFIG{ R"(
```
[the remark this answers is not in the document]
remia: Done

tests/cpu/transforms/LookTransform_tests.cpp
remia: Good point, done

tests/cpu/transforms/LookTransform_tests.cpp
remia: Totally, previous name was not helpful, fixed

src/OpenColorIO/transforms/LookTransform.cpp
remia: Fixed

tests/cpu/transforms/LookTransform_tests.cpp
```diff
@@ -559,3 +562,161 @@ environment: { FILE1: cdl_test1.cc, FILE2: cdl_test1.cc }
     OCIO_CHECK_EQUAL(std::string("FILE2"), usedContextVars->getStringVarNameByIndex(1));
     OCIO_CHECK_EQUAL(std::string("cdl_test1.cc"), usedContextVars->getStringVarByIndex(1));}
 
+
+OCIO_ADD_TEST(LookTransform, inverse_look_transform)
+{
+    // Test inversion of the transform
```
hodoulp, line 568: The fix is around a transform containing a look so I suggest `// Test inversion of the transform containing a look.`

remia: Done

hodoulp (approved):

PR #1183, opened 2020-10-29
Adsk Contrib - Add operator<< for FormatMetadata

BernardLefebvre:
Add an operator<< for FormatMetadata so that it can be printed out (usefull in pyhton).
Adjust FormatMetadata::addChildElement so that it does not return.
Improve documentation.
Add python test for FormatMetadata.

### DIFF
--- a/src/bindings/python/PyFormatMetadata.cpp
+++ b/src/bindings/python/PyFormatMetadata.cpp
@@ -50,7 +50,7 @@ void bindPyFormatMetadata(py::module & m)
                 throw py::key_error(os.str());
             },
              "name"_a)
-        .def("__setitem__", &FormatMetadata::addAttribute, "name"_a, "value"_a)
+        .def("__setitem__", &FormatMetadata::addAttribute, "name"_a.none(false), "value"_a.none(false))
         .def("__contains__", [](const FormatMetadata & self, const std::string & name) -> bool
             {
                 for (int i = 0; i < self.getNumAttributes(); i++)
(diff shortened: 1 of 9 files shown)


hodoulp:
As usual this is a notice that we will merge this week (Thursday or Friday), if there are no discussions on the pull request.

src/bindings/python/PyFormatMetadata.cpp
```diff
@@ -73,7 +81,11 @@ void bindPyFormatMetadata(py::module & m)
                 return ConstChildElementIterator(self);
             })
         .def("getChildElements", [](FormatMetadata & self) { return ChildElementIterator(self); })
-        .def("addChildElement", &FormatMetadata::addChildElement, "name"_a, "value"_a)
```
michdolan, line 76: I don't think this change is necessary. pybind11 handles `const char *` from Python string automatically. I have primarily been using `std::string` args in the constructors just for ease of checking for empty input.

BernardLefebvre: I'll revert that, but I also need to remove the test that check TypeError is raised when function is used with None for a parameter (segfault is generated)

michdolan: If you add `.none(false)` to the args which are causing segfaults when passing `None`, that should fix the issue. You can search for that in the bindings to see exmaples of where it has been added already.

michdolan (approved):

michdolan:
Closing and reopening to trigger CI again, in case the error was VM-caused

hodoulp (approved):

PR #1197, opened 2020-11-06
API Docs Implementation

### LINKED ISSUE #1148: C++/Python API Documentation Generation
This issue is intended to host a discussion around the right approach for generating C++ and Python API docs for OCIO. @scoopxyz has done an excellent job investigating and starting to implement a method for automatic doc generation with `doxygen`, `breathe`, and a tool that ships with `pybind11`, but we have not yet decided if this is the approach to commit to.

The primary concern that's been raised in recent discussion is around complexity, both in terms of docstring formatting in OCIO headers, as well as Python-specific explanations in the Python bindings. OCIO v1 API doc generation was fairly simple, and leveraged some custom tooling (and Python stubs) to shoehorn both languages into Sphinx. The new approach adds complexity to this in hopes of unifying documentation between C++ and Python as much as possible, as well as leveraging tools that are readily available. I'll let @scoopxyz explain the current approach since he's the most knowledgeable.

We need to decide whether to:
1. Stick with the current approach and focus on completing it. As long as it's well documented, complexity can be managed.
2. Separate the C++ and Python docs. Embed Python docstrings in the pybind11 bindings (since the interface differs in various ways), and simplify the C++ docstring extraction. The idea here is that the Python docs are probably more critical for many OCIO users, since they are generally used by config authors, and C++ engineers may prefer using the headers as documentation (I do personally).
3. Do something else. Open to ideas, but hopefully suggestions leverage work that has already been done to some extent. The API is large, so the solution needs to scale well.

Whichever approach, I have some rare free cycles coming up (in one week) where I could probably complete this work. Let's discuss and come up with a resolution before the next Documentation working group meeting (09/25/2020) if possible. Thanks!

michdolan:
There's a lot here to review, and some important notes below, but the bulk of the API docs are implemented here.

What was done:
- Finalize Doxygen and breathe implementation for C++ API doc generation.
- Restructure `api` RST with both breathe and autodoc directives for C++ and Python tabs.
- Adapted `sphinxcontrib.prettyspecialmethods` Sphinx extension to transform the Python doc naming to be more intuitive, and to play nicely with pybind11 generated signatures.
- Wrote Python tool, based on `pybind11_mkdoc`, which leverages Doxygen XML instead of Clang to generate Python docstrings in a C++ header. As mentioned in previous discussions this will fail the build if it encounters a Doxygen XML element which it doesn't know how to render. I tried to make these failures clear by printing the offending XML and the resulting generated RST, and will note that the elements Doxygen uses are sometimes different on different platforms (which I learned while getting CI to pass).
- Restructured Python bindings to include generated docstrings, and to forward declare iterators, enums, etc. used in Python classes, which allows generated docs to reference Python objects rather than C++ objects. Note that module attributes don't support `__doc__` assignment in pybind11, and exception registration doesn't support `__doc__` assignment in Python 2 bindings. For the attributes, which includes all the global constants, I put the docs directly in the RST, which is a bit of a maintenance annoyance since they are effectively duplicated, but it's limited to just that section. There may be ways around this in pybind11, but I could not find them as I dug through the pybind11 docs and source.
- Remove automatic installation of documentation dependencies, but improve the detection and documentation around these dependencies.
- Rework CMake to improve doc generation, including hints to rebuild when doc generation infrastructure (like the new Python tools, or Doxygen XML) change. My goal was to be able to build docs repeatedly and have the system only build what was changed, even if that was some upstream tooling which effects downstream docs in a subtle way. There may still be room for improvement here, but it is working fairly well in my tests.
- Fixed some holes in the Python bindings from recent C++ changes.

TODO (in other PRs):
- There are still quite a few Sphinx warnings to account for, but many are tied to the mixed doc style currently in the public C++ headers. I started to fix some of these, but decided to wait since I wasn't 100% sure of @doug-walker @hodoulp  and @BernardLefebvre 's vision there. Once those are restructured to adhere more strictly to Doxygen I can make refinements to the docstring extraction code as needed.
- The Python docs are verbatim from the C++ right now, so code samples are wrong and unique bits of the API are undocumented. I (and others) can make smaller PRs to address these. I just wanted to get the mechanics in place, which you'll note if you build the docs are fairly good as-is.
- I have not got these updates working in RTD yet. I will do that next, and we may want to wait on a merge until that point. Building the docs DOES require building OCIO, and there's no way around this without omitting the Python API docs. The build chain is such that the Python bindings MUST be built to have complete docs. The alternative to this would be to put the docs into stub Python modules, or directly in RST, which we _can_ do, but was not my approach here. I would opt instead to leverage a CD pipeline if we have to, but will do some investigation into our options.

Thanks for having a look, and I highly encourage pulling this down and doing a local build to look at the HTML:
```
mkdir _build
cd _build
cmake ../. -DOCIO_BUILD_DOCS=ON
make -j<your thread count> docs
<your web browser> docs/build-html/index.html
```

Resolves #1148


michdolan:
@hodoulp , all of your feedback from #1160 has been addressed here as well.

michdolan:
Fixes #1148

hodoulp:
Great. As 'time flies', I will have a look over the weekend to provide you feedback asap.

zachlewis:
Thanks Michael -- I'll give it a spin this weekend too.

michdolan:
I've been investigating the RTD integration. We're good on the C++ end since RTD provides Doxygen in it's Docker images, but there's no support for building the Python bindings to utilize autodoc to build the Python API docs (without a matching Python binding binary, and thus an OCIO binary as well).

One option could be to stash those binaries somewhere and download them into the RTD VM as a sphinx build step, but that has some artifact storage requirements.

The solution I am leaning toward, and would like feedback on, is to store a complete set of API RST in the repo. By "complete" I mean a baked RST from both breathe and autodoc. That would mean that anytime the public headers, or Python bindings are modified, the PR would have to also submit an intermediate RST build for use by RTD (and others wanting to build the docs without building OCIO). To guarantee this generated RST and the OCIO source stay in sync, I would add a check to the CI which would build the RST and diff it with the same files in the repo, failing the build if they don't match. It's not an ideal solution, but would be manageable.

Certainly open to other ideas as well, like maintaining special CD-generated doc branches, but it' up for discussion. I'll put it on the TSC agenda, but feel free to discuss here if you have an opinion.

scoopxyz:
I started down this path for the original SIGGRAPH implementation, only to have it blow up in my face because of the sub-par quality of Autodoc -> RST generation (https://github.com/sphinx-contrib/restbuilder). Might still be worth having someone else validate my experience, but it produced terribly mangled RST that didn't look pretty or make sense in a lot of cases (esp. with Breathe sources)

Could be looked into of course, but adds to the growing mountain of custom solutions just to get docs out the door (which was a trend I tried to avoid, but feels inevitable at this point).

michdolan:
Thanks @scoopxyz . My approach to "baking" the RST was going to be to write a very lightweight extension which simply echo's the RST that breathe and autodoc generate themselves, with no other transformation. I found some hints at how to do this with very minimal code, but I have not yet tested the theory. In short, I would mokeypatch an autodoc method which appends RST lines during doc generation, which we would redirect to some output. My intent was to put this into some separate files and include them into the main API doc structure, as to not mess with the controlled portions of the docs.

As far as using custom solutions, I get the sense that other projects are just handwriting their Python docs in a similar way to what OCIO previously did. The downside to that approach is that we wouldn't be able to leverage pybind11's built-in type hinting, which produces excellent signatures for the Python docs without human input.

michdolan:
Heads up, this PR is going to be in flux for the next few days while I swap in Python API doc freezing functionality.

michdolan:
Ok, I think I'm done with this. I may push this to a test branch to do some RTD integration tests before we merge, but my simulated tests locally have worked so far. Here's a summary of how this all works:

- If a contributor makes changes to any part of OCIO which affects the Python API docs (so, public headers, Python bindings, any documentation process code, etc.) they MUST do a local build with the new CMake option `-DOCIO_BUILD_FROZEN_DOCS=ON`, and add the modified RST files under `docs/api/python/frozen` to their PR.
- The OCIO `conf.py` module has a switch that detects when docs are being built on GH Actions (`CI` env var == `true`) it will backup the `frozen` folder to a sibling `backup` folder on Sphinx init, and following Sphinx build completion will do a file-by-file comparison of the new `frozen` and the `backup` folders. If there are differences, the CI job will fail with an error explaining where the differences where found and with instructions on how to fix it.
- The `conf.py` also has a switch that detects when it is being run on RTD, and in that case will itself run Doxygen to generate the XML needed by breathe prior to building the docs, and will also facilitate a CMake `configure_file`-like process (via Python) to handle substitutions in headers and docs source files that CMake would usually handle, but can't in this case. One potential plus to all of this is that if someone wants to just build OCIO docs, they can technically do so by running `sphinx-build` in the docs directory, and nothing more. Right now that only works when the `READTHEDOCS` env var == `True`, but it could be easily exposed another way if needed.
- All of these required several custom Sphinx extensions tuned for our project specifically. They _could_ be made more generic for improved sharing in the future, but that was not my main focus for now. It wouldn't be a huge amount of additional work to do that though, but in a future PR.
- Reviewing all of this in GH is going to be a pain with ~400 modified files, but most of that is small, granular, RST files for the frozen docs support.

michdolan:
The CI failures here are from the frozen doc comparison check I mentioned. For example:

```
Extension error:
Frozen RST is out of date! Build OpenColorIO with CMake option '-DOCIO_BUILD_FROZEN_DOCS=ON' to update required frozen documentation source files in: /__w/OpenColorIO/OpenColorIO/docs/api/python/frozen

    Changed files: pyopencolorio_filetransform.rst, pyopencolorio_gpushaderdesc.rst, pyopencolorio_formatmetadata.rst, pyopencolorio_cpuprocessor.rst, pyopencolorio_setcomputehashfunction.rst, pyopencolorio_autostride.rst, pyopencolorio_config.rst, pyopencolorio_processor.rst, pyopencolorio_gradingrgbcurve.rst, pyopencolorio_gpushadercreator.rst, pyopencolorio_dynamicproperty.rst, pyopencolorio_gpuprocessor.rst

      Added files: 

    Removed files: 
```

I'm noticing the error varies a bit per platform, which could be problematic, but I'll troubleshoot until CI passes.

michdolan:
Now that CI is passing on all platforms, I'm starting RTD integration testing, using the newly created `rtd_test` branch.

michdolan:
Build succeeded on RTD: https://opencolorio.readthedocs.io/en/rtd_test/api/_index.html#api

This is ready for review.

michdolan:
One note to add to other explanations. When testing the frozen RST generation against the existing source in the repo, there were differences per platform. I implemented a difference check with an error tolerance which tests for a similarity ratio, and if it falls below that it fails CI. This does imply that the docs can vary slightly depending on who builds them, and that the comparison could miss small inconsistencies. An alternative approach would be to only do this check on Linux CI, and be stricter about it. Linux is the ideal choice there since RTD builds are done on Linux.  Up for discussion.

hodoulp (approved):

doug-walker (approved):

PR #1207, opened 2020-11-17
Add TSC Notes for 11-16-2020

carolalynn22:
Signed-off-by: Carol Payne carolp@netflix.com

### DIFF
--- a/ASWF/meetings/tsc/2020-11-16.md
+++ b/ASWF/meetings/tsc/2020-11-16.md
@@ -1,0 +1,100 @@
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
+<!-- Copyright Contributors to the OpenColorIO Project. -->
+
+November 16, 2020
+
+Host: Doug Walker
+
+Rotating Secretary: Carol Payne
+
+Attendees:
+  * [x] Mark Boorer (_TSC_) - Industrial Light & Magic
+  * [x] Mei Chu (_TSC_) - Sony Pictures Imageworks
+  * [ ] Sean Cooper (_TSC ACES TAC Rep_) - DNEG
+  * [ ] Michael Dolan (_TSC Chair_) - Epic Games
+  * [x] Patrick Hodoul (_TSC_) - Autodesk
+  * [ ] John Mertic - Academy Software Foundation / Linux Foundation
+  * [x] Carol Payne (_TSC_) - Netflix
+  * [x] Mark Titchener (_TSC_) - Foundry
+  * [x] Carl Rand (_TSC_) - Weta Digital
+  * [x] Doug Walker (_TSC Chief Architect_) - Autodesk
+  * [x] Kevin Wheatley (_TSC_) - Framestore
+  * [x] Bernard Lefebvre - Autodesk
+  * [x] J Schulte - ILM
+  * [x] Matthias Scharfenberg - ILM
+  * [x] Larry Gritz - SPI
+  * [x] Aymee Kennett (Yeh) - Netflix
+
+Apologies:
+  Michael Dolan
+
+# **OCIO TSC Meeting Notes**
+
+* Named Color Transforms - Continued discussion
+    - Doug: Patrick, Bernard and I have been discussing the feedback and how to respond. We have made it possible to turn off the NamedTransforms - similar to an environment variable. Config author/ dev would be able to control whether or not the feature is visible to the user. Bernard has pushed that commit to the PR so it should be available for viewing.
+    - Mark: on/off isn't a true solution - we don't want to fracture the community. The opposition was because we felt that NamedTransforms opposed the way we want OCIO to work.
+    - Doug: would be interested to hear alternate proposals.
+    - Mark: the solutions already exist - you just have to utilize your working space as your reference space. 
+    - Doug: Not everyone is going to be the working space of the config. 
+    - Mark: You design your config based on your requirements for your work. Having different in and out colorspaces is not how it's defined to work.
+    - Doug: We can't expect everyone to have a working space per config. Even a difference between rendering space vs. compositing space. Thinking ahead to the studio config as well, it would be ideal if this workflow was easier to implement. 
+    - Mark: But ACES is trying to be the single working space - and should be for the future. 
+    - Carl: WETA works in a different compositing colorspace than rendering space, for example.
+    - Matthias: verifying that we're talking about NamedTransforms that are basically a transfer funtion or something similar? So you don't get the matrix transform that you would get by using the colorspace which goes through the reference space as well? It sounds like a "trapdoor" in the color management system.
+    - Kevin: I didn't see it quite that way. Instead of performing a 2-sided transformation, it's a 1 sided.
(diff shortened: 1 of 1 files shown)


ASWF/meetings/tsc/2020-11-16.md
```diff
@@ -0,0 +1,100 @@
...
+    - Mark: So added functionality to the API to get a CUDA callable alternative would this solve the problem? 
+    - Larry: CPU based c++ app launching CUDA kernels. Can the CPU app ask for the correct CUDA version to pass for the shader - think this would be fine. Just want something - get a handle to a CUDA callable transform.
+    - Mark: only thing we have now is the GLSL "get me a shader". Another API call where you pass in CUDA instead of whatever, that would work? Either texture source or byte code?
+    - Larry: either is fine.
+    - Mark: Textural source would be much easier. 
+    - Doug: Issue with the GPU is how to get the texture onto the GPU.
+    - Patrick: Yeah it's a big blob of data - the texture that you have to get onto the GPU. The structure is there to be able to generate the CUDA coda. There are two APIs - one that's openGL oreniented, the other that just takes what it needs to do.
```
michdolan, line 83: coda > code

michdolan (commented):
Thanks for the great notes @carolalynn22 ! Much appreciated.

michdolan:
Regarding the `NamedTransform` discussion, I'll point out this section of `PROCESS.md`:

> If one or more Committers oppose a proposed change, then the change cannot be accepted unless:
> 
> Discussions and/or additional changes result in no Committers objecting to the change. Previously-objecting Committers do not necessarily have to sign-off on the change, but they should not be opposed to it.
> 
> The change is escalated to the TSC and the TSC votes to approve the change. This should only happen if disagreements between Committers cannot be resolved through discussion.
> 
> Committers may opt to elevate significant or controversial modifications to the TSC by assigning the tsc-review label to a pull request or issue. The TSC should serve as the final arbiter where required.

I think we're still in the discussion phase of this and am hopeful we will reach an agreement there, but a reminder that if we can't this can be escalated to a TSC vote.

zachlewis:
How about an `OCIO_NIMBY` environment variable or something we could set at a facility level that would actually cause OCIO configs with named transforms to throw...? Everyone wins!

ASWF/meetings/tsc/2020-11-16.md
```diff
@@ -0,0 +1,100 @@
...
+  Michael Dolan
+
+# **OCIO TSC Meeting Notes**
+
+* Named Color Transforms - Continued discussion
+    - Doug: Patrick, Bernard and I have been discussing the feedback and how to respond. We have made it possible to turn off the NamedTransforms - similar to an environment variable. Config author/ dev would be able to control whether or not the feature is visible to the user. Bernard has pushed that commit to the PR so it should be available for viewing.
+    - Mark: on/off isn't a true solution - we don't want to fracture the community. The opposition was because we felt that NamedTransforms opposed the way we want OCIO to work.
+    - Doug: would be interested to hear alternate proposals.
+    - Mark: the solutions already exist - you just have to utilize your working space as your reference space. 
+    - Doug: Not everyone is going to be the working space of the config.
```
doug-walker, line 38: be the working space of the config --> be solely using the working space of the config

ASWF/meetings/tsc/2020-11-16.md
```diff
@@ -0,0 +1,100 @@
...
+    - Doug: Patrick, Bernard and I have been discussing the feedback and how to respond. We have made it possible to turn off the NamedTransforms - similar to an environment variable. Config author/ dev would be able to control whether or not the feature is visible to the user. Bernard has pushed that commit to the PR so it should be available for viewing.
+    - Mark: on/off isn't a true solution - we don't want to fracture the community. The opposition was because we felt that NamedTransforms opposed the way we want OCIO to work.
+    - Doug: would be interested to hear alternate proposals.
+    - Mark: the solutions already exist - you just have to utilize your working space as your reference space. 
+    - Doug: Not everyone is going to be the working space of the config. 
+    - Mark: You design your config based on your requirements for your work. Having different in and out colorspaces is not how it's defined to work.
+    - Doug: We can't expect everyone to have a working space per config. Even a difference between rendering space vs. compositing space. Thinking ahead to the studio config as well, it would be ideal if this workflow was easier to implement.
```
doug-walker, line 40: working space per config --> have a separate config per working space

ASWF/meetings/tsc/2020-11-16.md
doug-walker: oreniented --> oriented

doug-walker (changes requested):
Thank you for the notes Carol!  This was a complicated discussion and you captured it well.

doug-walker:
> How about an `OCIO_NIMBY` environment variable or something we could set at a facility level that would actually cause OCIO configs with named transforms to throw...? Everyone wins!

Zach, thank you for trying to come up with a mutually acceptable solution.  We already added the ability to turn off Named Transforms via an env var (although it works differently than what you proposed) and ILM said it was not acceptable to them.  We are very open to your proposal but would like confirmation from ILM that it would resolve their objections before proceeding to implement it.

zachlewis:
> > How about an `OCIO_NIMBY` environment variable or something we could set at a facility level that would actually cause OCIO configs with named transforms to throw...? Everyone wins!
> 
> Zach, thank you for trying to come up with a mutually acceptable solution. We already added the ability to turn off Named Transforms via an env var (although it works differently than what you proposed) and ILM said it was not acceptable to them. We are very open to your proposal but would like confirmation from ILM that it would resolve their objections before proceeding to implement it.

Indeed. Just to be clear, what I'm proposing addresses a slightly different problem from user visibility. Matthias's "trapdoor" analogy really resonates with me; and whether named transforms are user-visible or not, just having them in a config hypothetically "poisons" the reliability of the ecosystem, in a sense (especially in light of the API's ability to construct processors "between" two configs). This doesn't sit comfortably with me.

`NamedTransforms` have the potential to cause some very insidious, difficult to diagnose problems in production, that might be too subtle for artists to notice right away; and a config that "works properly"  in some circumstances but "is broken" in others is quite a departure from what we've come to expect from OCIO-1. 

At the same time, I can't argue with user demand. Personally, the Foundry OCIO configs are the bane of my existence, and color-unmanaged shows and assets do not coexist well with color-managed shows and assets. If a supe wants to do something weird with color management, I can do my best to advise them otherwise, but at the end of the day, it's not my job to tell them "no." 

But $OCIO_NIMBY allows me to passive-aggressively say "no" per facility policy! Which is absolutely something I'd like to have the option of doing, should `NamedTransforms` come to be, visible to users or otherwise.

Also, there are certain occasions where something like `NamedTransforms` would be really useful for legitimate and semi-legitimate reasons; and a transfer function toolkit does seem kind of handy, and could potentially make configs more readable. They'd also be super useful for config-free LUT shaping, e.h. with ociobakelut. 

Maybe getting a little ahead of myself, but I thought it might be interesting if we could imbue [channel-independent] NamedTransforms with different lookup "styles", like hue_restore_dw3, max, other norms... it could be a nice way of wrapping single-channel versions of Exponent etc / Grading RGB Curves Transforms in LUT-like functionality without referencing an actual LUT.

doug-walker (approved):

hodoulp (approved):

PR #1210, opened 2020-11-17
Exposed Python Function added .none(false)

divshekhar:
Signed-off-by: divshekhar <imdshekhar@gmail.com>


BernardLefebvre (changes requested):
Please make sure that you do not change the current formating of the file (your editor did add some indentation that needs to be removed).

Look for other functions in that file that are using string parameters and do similar fix.

PR #1214, opened 2020-11-19
Adsk Contrib - Various file rule improvements

hodoulp:

The pull request makes public the 'default' file rule names to ease their usage, changes some public methods to return an empty string on failure (instead of nullptr) to unify the behaviour with other API methods, and add some extra checks when reading a file rule from a config file.


include/OpenColorIO/OpenColorIO.h
```diff
@@ -1123,6 +1123,12 @@ extern OCIOEXPORT std::ostream& operator<< (std::ostream&, const Config&);
 class FileRules
 {
 public:
+
+    /// Reserved rule name for the default rule.
+    static const char * DefaultRuleName;
+    /// Reserved rule name for the file path search rule \see FileRules::insertPathSearchRule.
+    static const char * FilePathSearchRuleName;
```
michdolan, line 1130: Wondering if these should go in the `extern OCIOEXPORT` section of `OpenColorTypes.h` instead for consistency with other static strings. Just a question for discussion - don't feel strongly about it.

hodoulp: Good point. I thought (and still think) that this is the right location as it clearly demonstrates it can only be used by `FileRules` rule name. And that cannot be explicitly changed, modified, etc.. It's only helpers when looping through the rule names to easily detect these two 'special' rules.

michdolan (approved):

doug-walker (approved):

PR #1215, opened 2020-11-19
Adsk Contrib - File transform interpolation v1 compatibility

BernardLefebvre:
This removes the validation of the interpolation for file transforms that was introduced during v2 development. If an interpolation (including unknown) does not apply to a transform of the file that requires a interpolation, default interpolation is used.

When saving a v1 config, an interpolation valid for v1 will be saved (for instance, default and unknown will be saved as linear).

Solves issue 1194.


doug-walker:
We feel that this is a critical bug fix and would appreciate it if people could review this so it could be merged before the usual two weeks.

doug-walker (approved):

hodoulp (approved):

doug-walker:
The TSC approved merging this bug fix immediately.

Issue #1155, opened 2020-09-24
OCIO metadata serialization


michdolan:
Continuing discussion from #1154 ...

It was discovered recently (and maybe this was already known) that `CDLTransform` `setID` and `setDescription` are not persistent as they are not currently serialized to config YAML, though they are advertised as such in the header:

```
    //!rst:: **Metadata**
    //
    // These do not affect the image processing, but
    // are often useful for pipeline purposes and are
    // included in the serialization.

    /// Unique Identifier for this correction.
    virtual const char * getID() const = 0;
    virtual void setID(const char * id) = 0;

    /**
     * Deprecated. Use `getFormatMetadata`.
     * First textual description of color correction (stored
     * on the SOP). If there is already a description, the setter will
     * replace it with the supplied text.
     */
    virtual const char * getDescription() const = 0;
    /// Deprecated. Use `getFormatMetadata`.
    virtual void setDescription(const char * desc) = 0;
```

This led to a discussion around metadata serialization across the library. Currently all transforms will serialize `METADATA_NAME` to the config, but no other key is supported.

I understand that format metadata is intended to support serialization to CLF/CTF specifically, but think the metadata not being persistent to a config is problematic since it limits reproducibility. Consider a case where you use a config to drive the creation of some critical CLF files that are handed off to a client; right now you would need a Python script which builds a config on the fly and bakes from that. You would not be able to just load the config and bake as any intended metadata would have been lost (except for `name`).

We should discuss whether this is an issue and how it could be resolved if so. It's possible others disagree here. I think the ability to store custom metadata in a config per-transform would be useful personally, rather than overloading a `ColorSpace` `description` attribute for this purpose.

Whatever the outcome of that discussion, I think we should remove, or very clearly deprecate, any public API setters that are not persistent in the config.

zachlewis:
Thanks for bringing this up.

> I think the ability to store custom metadata in a config per-transform would be useful personally, rather than overloading a `ColorSpace` `description` attribute for this purpose.

I wholeheartedly agree. I was hoping for such a feature, especially for the scarier looking, less easily-identifiable transforms, like Matrix or CameraLog. At minimum, it would really help to keep CLF/CTFs intelligible, and to reduce user-error when manually editing configs; and would be useful for generating filenames or identifying / swapping out specific transforms in arbitrary Processes. Also, I guess it would be one way to get a cache id --> description map "for free" sometimes.

[...]

PR #1221, opened 2020-11-27
Adsk Contrib - CDLTransform cleanup

BernardLefebvre:
Rework CDLTranform interface to implement the proposal made in issue #1155.
CDLTransform can be created from a file and an id.
Files containing several CDL (such as ccc and cdl) can be loaded as GroupTransform to give a way to check the content of such CDLs.
Functions to get/set ID should be replaced by calls on the transform's formatMetadata.
Functions to get/set description have been renamed and documention has been added to explain how formatMetadata can be used.
FormatMetadata has helper functions to get/set common attributes 'name' and 'id'.
Write related functions have been moved to GroupTransform so that a processor does not have to be create to write a transform.


BernardLefebvre:
Doumentation has not been updated.

hodoulp:
As usual this is a notice that we will merge this week, if there are no discussions on the pull request.

hodoulp (approved):

doug-walker (approved):

PR #1228, opened 2020-12-04
Fix build break: Bump pybind11 required version to 2.6.1

michdolan:
pybind11 versions < 2.6 have a submodule of clang to support the pybind11_mkdoc feature, which we don't use. In 2.6 this was moved to it's own project. By default CMake's `ExternalProject_Add` command, which we use to install pybind11 if `OCIO_INSTALL_EXT_PACKAGES` is enabled, clones all submodules, so this change explicitly tells it not to for pybind11.

### DIFF
--- a/share/cmake/modules/FindExtPackages.cmake
+++ b/share/cmake/modules/FindExtPackages.cmake
@@ -61,7 +61,7 @@ if(OCIO_BUILD_PYTHON)
 
     # pybind11
     # https://github.com/pybind/pybind11
-    find_package(pybind11 2.4.3 REQUIRED)
+    find_package(pybind11 2.6.1 REQUIRED)
 endif()
 
 if(OCIO_BUILD_DOCS)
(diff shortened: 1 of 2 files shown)


michdolan:
And to note, the reason this was failing the OCIO build seems to be related to a curl timeout while cloning clang. I'm not sure what is imposing that timeout, but since we don't need clang, this should improve build time regardless.

michdolan:
And it's still cloning clang... stand by...

michdolan:
After deeper digging, git submodules can't be cleanly disabled from CMake until 3.16/3.17. @BernardLefebvre proposed another solution, which prevents the timeout when cloning: `git config — global http.postBuffer 524288000`, which could be implemented via the `GIT_CONFIG` option of `ExternalProject_Add`, but looking through the release notes of the recent pybind11 updates, I think we would do well to just update, which removed the submodule, and has a lot of bug fixes and feature improvements for our python bindings, since we're still in beta.

BernardLefebvre (approved):

hodoulp (approved):

hodoulp:
Nice work guys.

doug-walker (approved):

PR #1229, opened 2020-12-04
Adsk Contrib - Data bypass fix

BernardLefebvre:
BuildColorSpaceOps was using false for the dataBypass parameter when when true should be used. This was causing data color spaces not to be bypassed when using Config::getProcessor(csa, csb).
Tests for this have been added.

Also add a nullptr check for file rules path resolution functions and fix a typo.


michdolan (approved):

doug-walker (approved):
This is a critical bug fix.  Therefore, I will proceed to merge this immediately and we will create a 'beta2' tag.

PR #1234, opened 2020-12-09
Adsk Contrib - Improve ACEScc and XYZ built-in transforms

doug-walker:
- Improved the accuracy of the ACEScc built-in transform.
- Adjust the normalization of CIE XYZ D65.

Signed-off-by: Doug Walker <Doug.Walker@autodesk.com>

### DIFF
--- a/src/OpenColorIO/transforms/builtins/ACES.cpp
+++ b/src/OpenColorIO/transforms/builtins/ACES.cpp
@@ -113,7 +113,7 @@ void GenerateOps(OpRcPtrVec & ops)
     CreateMatrixOp(ops, &EXP_TO_ACES[0], TRANSFORM_DIR_FORWARD);            
 }
 
-}  // namespace ADX
+}  // namespace ADX_to_ACES
 
 
 namespace ACES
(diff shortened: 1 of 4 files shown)


zachlewis (approved):

zachlewis (approved):

zachlewis:
@doug-walker, would you mind explaining the CIE XYZ D65 normalization changes?

From what I can tell, instead of computing matrices directly from XYZ with D65 primaries, you're starting with illuminant E and adopting D65 as a separate step before the vonkries matrix. I imagine the difference boils down to matrix precision when concatenating matrices?

doug-walker:
@zachlewis , it's clearer to use the conventional normalization of CIE XYZ.  So for example, a neutral in the CIE XYZ D65 space with a luminance of 1 would have a value of [0.9504..., 1., 1.089...] rather than [1., 1., 1.].  This is also more in line with the SMPTE/DCI specs for digital cinema.  

The implication of this is that when defining a Primary set for XYZ, it needs to use illuminant E as the fourth xy pair, since the "white point" of this structure is not really the adaptive white of the viewer but rather the chromaticity that you want equal RGB values to map to.  I then added a more flexible version of build_conversion_matrix that allows passing in the white points rather than inferring them from the Primaries object.  There should be no precision or matrix concatenation issue.

zachlewis:
Solid. Thanks Doug. I agree with you, it's more straightforward this way -- seeing the XYZ Primaries coupled with illuminant D65 instead of E had caught me off-guard previously, but I figured it made enough sense... which is why this PR grabbed my attention! 

Anyway, thanks. Equal energy for all.

hodoulp (approved):

michdolan (approved):

doug-walker:
This PR was not from the correct fork and could not be merged.  Closing this and created a new PR #1251.

Issue #1129, opened 2020-09-03
ColorSpace 'alias' mechanism


zachlewis:
*edit: See @doug-walker's proposal and discussion below
> [...] I'd like to propose an alternate solution which is to allow use of an "alias" key in a ColorSpace. This would allow the alternate name to be used anywhere the main name is used. This would include file rules but also command-line tools and other scenarios. Here is an example (the fourth line is the only difference from the current ACES config):
> 
> 
> ```
>   - !<ColorSpace>
>     name: ACES - ACEScct
>     family: ACES
>     alias: acescct
>     equalitygroup: ""
>     bitdepth: 32f
>     description: |
>       The ACEScct color space
>       
>       ACES Transform ID : ACEScsc.ACEScct_to_ACES
>     isdata: false
>     allocation: uniform
>     allocationvars: [-0.249136, 1.468]
>     to_reference: !<GroupTransform>
>       children:
>         - !<FileTransform> {src: ACEScct_to_linear.spi1d, interpolation: linear}
>         - !<MatrixTransform> {matrix: [0.695452, 0.140679, 0.163869, 0, 0.0447946, 0.859671, 0.0955343, 0, -0.00552588, 0.00402521, 1.0015, 0, 0, 0, 0, 1]}
> ```

Original post:
I would like to be able to repurpose the ColorSpace equalityGroup attribute to serve as ad-hoc colorspace 'aliases' where filename parsing is concerned. 

It's not uncommon for configs to contain copies of the same colorspace with alternate naming (and sometimes different families) --  E.g., "ACES - ACES2065-1", "aces", and "lin_ap0" are identical, and are found under different 'family paths'. 
Sometimes, a config will feature colorspaces with longer, human-friendly names prominently in interfaces and colorspace family hierarchy; and tucked away (or perhaps hidden in v2) are duplicate colorspaces with 'short names', often intended to serve as a filename 'color token', as is necessary for strict-parsing workflows. 

I propose either creating a new File Rule type for matching colorspace equalityGroups in a string, or updating the existing PathSearchRule to optionally search for equalityGroups as well. 

That's the primary proposal -- should be modest enough of a challenge for me to handle -- but it would be nice to reference colorspaces by their equality groups as shorthand for their actual names throughout OCIO; perhaps even have a place for them in menu interfaces, if wanted. 

Thoughts?

sobotka:
+1 if there is a way to reduce the stanzas that define the actual transforms.

[...]

PR #1241, opened 2020-12-18
Adsk Contrib - Add color space aliases

BernardLefebvre:
Aliases are added to color spaces as described in issue #1129 
Also adding a name to the config.


zachlewis (commented):
Nicely done, Autodesk. Multiple alases are a pleasant surprise, as is the config name attribute. Both seem insanely useful for reconciling bespoke conventions among configs / pipelines. I appreciate having a canonical method for retrieving canonical names, and similar pipeline-facilitating mechanisms built in to OCIO.

The test cases suggest that aliases work exactly how I'd like and expect. 


Thanks!

src/OpenColorIO/ColorSpaceSet.cpp
```diff
@@ -109,21 +119,53 @@ class ColorSpaceSet::Impl
 
     void add(const ConstColorSpaceRcPtr & cs)
     {
-        const std::string csName = StringUtils::Lower(cs->getName());
-        if (csName.empty())
+        const char * csName = cs->getName();
```
michdolan, line 122: Should this still be lowercase, for the below comparison?

BernardLefebvre: StringUtils::Compare is not case sensitive

BernardLefebvre: StringUtils::Compare is not case sensitive

include/OpenColorIO/OpenColorIO.h
```diff
@@ -291,6 +291,15 @@ class OCIOEXPORT Config
      */
     void validate() const;
 
+    /**
+     * Get/set a name string for the config.
+     *
+     * In combination with a color space name, this may be used to provide a more global reference
```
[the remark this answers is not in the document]
BernardLefebvre: discussed at TSC

src/OpenColorIO/ColorSpaceSet.cpp
michdolan: Ah, ok. Thanks for the clarification.

hodoulp (approved):

PR #1246, opened 2020-12-26
[CI build break] Fix Sphinx 3.4 AutoStride conflict

michdolan:
The CI build break was due to pip bumping the default Sphinx version to 3.4.0 (from 3.3.1), which we were using for doc builds during CI. There was an autodoc update which is now unable to document `PyOpenColorIO.AutoStride` for reasons I could not identify clearly. `AutoStride` was the only global attribute we were asking autodoc to document, so it seems to point to a conflict there, but worked fine in the previous version. I've changed the RST to not use autodoc for that attribute (which doesn't have a `__doc__` attribute anyway), which fixes it.

I've also linked the `OCIO_BUILD_FROZEN_DOCS` option to the `OCIO_BUILD_DOCS` option, so that if the former is defined the latter will be too to ease the freezing process somewhat.


doug-walker (approved):

doug-walker:
Thanks for fixing this Michael!

PR #1249, opened 2021-01-05
Improve PyCPUProcessor NumPy support

### LINKED ISSUE #1206: On improving the interface between "ImageDesc" and "ndarray".
Hello,

As explained on [Slack](https://opencolorio.slack.com/archives/C3Q43TN4Q/p1605321436095300), I could not find an elegant way to have image data flowing between OCIO and Numpy.

At the moment to basically creating a `PackedImageDesc` not only requires passing the `ndarray` but also specifying the _width_ and _height_ and _channel_ count which seems superfluous because that data is already available directly from the shape of the array likewise, to get the data out, I need to reshape it. Here is a snippet illustrating the current code:

```python
IMAGE_DESC = ocio.PackedImageDesc(
    DIGITAL_LAD_ACESCG.astype(np.float32),
    WIDTH, HEIGHT, 3)

PROCESSOR.apply(IMAGE_DESC)

DIGITAL_LAD_ODT_sRGB = IMAGE_DESC.getData().reshape([HEIGHT, WIDTH, 3])
```

where `WIDTH` and `HEIGHT` are defined as follows:

```python
WIDTH, HEIGHT = DIGITAL_LAD_ACES2065_1.shape[1], DIGITAL_LAD_ACES2065_1.shape[0]
```

Ideally, for a Pythonista, the above snippet should be:

```python
IMAGE_DESC = ocio.PackedImageDesc(DIGITAL_LAD_ACESCG.astype(np.float32))

PROCESSOR.apply(IMAGE_DESC)

DIGITAL_LAD_ODT_sRGB = IMAGE_DESC.getData()
```

Or even better:

```python
PROCESSOR.apply(DIGITAL_LAD_ACESCG)
```

Changing the array in-place is mostly fine assuming warnings are explicit in the docs.

michdolan:
Fixes #1206 

- Updates `CPUProcessor` `applyRGB` and `applyRGBA` methods to process the full array in place. Uses an `PackedImageDesc` internally.
- `applyRGB` and `applyRGBA` now support float, half, uint16, and uint8 NumPy arrays with a correctly configured processor (see accompanying unit tests for examples).
- New uint tests verify all supported buffer types.
- Note: adds Python test suite NumPy dependency.
- Additionally fixes bad INSTALL.md link to installation.rst


tests/python/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright Contributors to the OpenColorIO Project.
 
+include(FindPythonPackage)
+
+# NumPy
+# https://pypi.org/project/numpy/
+find_python_package(numpy REQUIRED)
```
doug-walker, line 8: Are we sure we really want to make this a requirement?  Could it be optional?

docs/api/python/frozen/pyopencolorio_cpuprocessor.rst
```diff
@@ -30,14 +30,14 @@
 
       Overloaded function.
 
-      1. applyRGB(self: PyOpenColorIO.CPUProcessor, pixel: buffer) -> buffer
+      1. applyRGB(self: PyOpenColorIO.CPUProcessor, data: buffer) -> None
 
       Apply to a single pixel respecting that the input and output bit-depths be 32-bit float and the image buffer be packed RGB/RGBA.
```
doug-walker, line 35: "Apply to a single pixel ..." needs to get updated now, could be more than that, right?

Also, does the note about efficiency still apply?  Maybe needs to be reworded to encourage passing in multiple pixels rather than looping in python.

doug-walker (changes requested):
Have not reviewed all the tests yet, but it's really awesome to have such an extensive set of tests!  My main question is whether we really want to make numpy a dependency.

tests/python/CMakeLists.txt
michdolan: Good point. I was on the fence about it. I have now made it optional and include warnings when running Python tests that test coverage is incomplete without it. The `CPUProcessor` tests will still run, but only the two apply methods that accept a Python list.

docs/api/python/frozen/pyopencolorio_cpuprocessor.rst
michdolan: I updated the documentation in the Python binding source to explain each of the apply methods and noting where they differ in functionality from the C++.

hodoulp (approved):

doug-walker (approved):

zachlewis:
You are the wind beneath my wings, Michael.

tests/python/CPUProcessorTest.py
```diff
@@ -0,0 +1,629 @@
...
+        # ---------------------------------------------------------------------
+        # Test image data
+        # ---------------------------------------------------------------------
+        if not np:
+            # NumPy not found. Only test lists.
+            log_range = [float(2**i) for i in range(-15, 15)]
+
+            cls.float_rgb_list = (
+                [-65504.0] + 
+                list(reversed(log_range)) +
```
remia, line 87: Did you meant `[-x for x in reversed(log_range)]` ?

Sorry for the late comment on this great PR.

michdolan: Yes, good catch yet again! Will fix on my next pass.

PR #1253, opened 2021-01-07
Adsk Contrib - Switch to rc1

BernardLefebvre:
Change from beta1 to rc1


hodoulp (approved):
All these changes in the `rst` files are clearly unfortunate. That's something to later investigate.

michdolan (approved):

michdolan:
As far the these RST changes, I _think_ they are simply an order of operations issue. If we reorder the Python binding definitions so that the binding for an object has been defined by the time another class references it, I believe it will show the python type instead.

PR #1260, opened 2021-01-12
Transform Python Bindings Fix

frankieeder:


frankieeder:
Another quick thought - I put the tests into a new file because the `TransformsTest.py` file didn't run successfully when I re-included it in the test suite I presumed this was because it was outdated, but curious what its history is and why it and other tests are being omitted. I think `TransformsTest.py` is the logical file to place this test in, however, so I'm wondering if we move it there and instead just `@unittest.skip()` the old failing test method.

zachlewis:
Thanks Frankie, this looks like solid work! Much appreciated.

frankieeder:
Thanks @zachlewis! Let me know if you have any thoughts regarding the ideal placement of the tests per my above comment as well.

doug-walker:
Thank you @frankieeder, it's great to have your participation on the project!

Regarding the TransformsTest.py file, it is largely redundant because the Transform children classes mostly already have more thorough tests in their own files.  We will clean up what's missing in another Python-related PR we have in progress.  If you want to delete the existing test in TransformsTest.py, put your test there, and enable it, that would be fine.

michdolan (approved):
Thanks for the contribution @frankieeder . Looks good!

doug-walker:
@frankieeder, we would need this finished by the end of the week in order to make it into the v2 release.  Thanks!

frankieeder:
Thanks for the bump @doug-walker. I was able to get the code updated, but having issues getting my commit signs to register here and trigger the checks correctly. Not sure why, but something I did since @michdolan's approval seems to have un-signed my first commit so the check wont run.

See my local fork properly signed [here](https://github.com/frankieeder/OpenColorIO/commit/5d0f8bda8d79fec235436947487aed04b806aae6) vs the main repo 
 unsigned [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1260/commits/5d0f8bda8d79fec235436947487aed04b806aae6):

Running `git rebase HEAD~26 --signoff` per the suggestion of the DCO check creates merge errors for every commit that isn't my own that I brought in using upstream pulls, even though this worked successfully previously.

Could use some help resolving this issue.

michdolan:
Sorry for the trouble @frankieeder . Might be easier to just start a fresh branch, pull your changes over, commit and sign them, and open a new PR. Detandling signing history can be a challenge.

Note that your tests are also failing in CI.

frankieeder:
Good call, and noted. Should have that up later this afternoon.

frankieeder:
Closing this pull request in preference for correctly signed and much cleaner PR #1277. Thanks for the suggestion @michdolan. For reference if anyone searches for a similar error in the future, the error seems to have come from accidentally running the signing rebase on master instead of my feature branch. The easiest way to fix ended up being to clean up the fork and restart from the upstream branch entirely (similar to [this](https://stackoverflow.com/questions/9646167/clean-up-a-fork-and-restart-it-from-the-upstream)).

PR #1267, opened 2021-01-19
Adsk Contrib - Add roles to API and fix file rules issues

BernardLefebvre:
Add role name declarations to the API for "rendering", "aces_interchange", and "cie_xyz_d65_interchange".
Adjust ociocheck to issue a warning rather than error if a role is missing (roles are not required for a config to be valid, but can be required for some features).

Fixes:
- Catch a std::regex exception (and throw an OCIO::Exception) in FileRules.
- ParseColorSpaceFromString uses active and inactive color spaces, make sure result is used accordingly.


michdolan (approved):

doug-walker (approved):
Received TSC approval to merge this small PR this week.

PR #1269, opened 2021-01-19
Add TSC meeting notes for 01-18-2021

michdolan:


doug-walker (approved):

doug-walker:
Great notes as always, thanks Michael!

michdolan:
@scoopxyz I added your change from DNEG to ARRI in applicable documents.

lgritz:
A couple minor notes of reply after reading the notes:

* Re the 2.x ABI compatibility: It's not that you *can't* break ABI compatibility for 9 months, it's just that if you do, you need to bump the version and those changes won't be adopted by DCC apps until the following year (but others apps may use it). It's hard to tell if that's anything to worry about, given how much advance time the DCC apps usually lock down versions, it's not likely that an ABI-breaking release in July will be picked up by DCC's during 2021, even if VFX Platform didn't have anything to say about it.

* If you're switching to a multi-stage "OCIO libs / OIIO / OCIO apps" build process, it may be worth observing that ocioconvert serves no purpose versus running oiiotool, and you could also modify OIIO's "iv" viewer to have OCIO support (I'm very embarrassed that it does not yet) to make ociodisplay also redundant. If elimination of those apps simplifies things for you. This is notwithstanding that the *source code* of ocioconvert and ociodisplay may be nice as a simple coding examples of how to use the OCIO libraries (oiiotool and iv are not good for that because they do so much more, it's not a good platform to highlight how to use the OCIO API), but executing ocioconvert and ociodisplay is not something anybody really needs to do for real work.

* If you wanted to go with just having those example programs use OpenEXR only (to cut OIIO out of the dependency) -- I know that the OpenEXR API is pretty complex in itself, but one route to take could be to only use the simplified RgbaInputFile and RgbaOutputFile interface and live with only allowing the subset of OpenEXR format that those support.

michdolan:
Thanks for the guidance @lgritz . In the end we decided to keep CY2021 at 2.0.x for similar reasoning to your points.

The main value in ocioconvert and ociodisplay are as code examples. We're booting up a UX working group and I am anticipating ociodisplay being a potential home for an example implementation. Totally agree on the OIIO tools being the choice for production.

When I was investigating swapping in OpenEXR to the apps, I was hoping to use RgbaInputFile and RgbaOutputFile, but noticed those are limited to HALF pixels (which is probably fine for the vast majority of color images no doubt). I ended up writing a similarly stripped down implementation that could handle FLOAT too, but only supported RGB/RGBA single-part scanline images. Doing so reminded me of how much I appreciate OIIO :)

lgritz:
Oh, yes, good point, I failed to consider the limitation to 'half'. Oh well.

hodoulp (approved):

PR #1272, opened 2021-01-20
Adsk Contrib - Avoid NaN values in certain Lut1D arrays

BernardLefebvre:
When creating an identity half-domain Lut1D as a starting point for sending through some shaper function, it is sometimes helpful to set the NaN values in the half-domain to 0.  Otherwise, when the LUT values are sent through further calculations, the NaN values persist and it is usually undesirable to cause image pixels to become NaNs.  The NaN values were also interfering with the determination of whether all three columns in a Lut1D were equal and causing three columns to be written into CLF files rather than one.


src/OpenColorIO/ops/OpArray.h
```diff
@@ -113,6 +115,12 @@ template<typename T> class ArrayT : public ArrayBase
             bool sameCoeff = true;
             for (unsigned long idx = 0; idx < m_length && sameCoeff; ++idx)
             {
+                if (IsNan(m_data[idx * 3]) &&
```
michdolan, line 118: Should this also handle the case where only one or two components are NaNs?

doug-walker: Good question, but no.  If one component is NaN and another is not, it means the channels are not all equal.  Therefore, m_numColorComponents should remain 3.  So we want to do the next if-statement, set sameCoeff=false and then break.

michdolan (approved):

doug-walker (approved):
We have approval from the TSC to merge this small improvement this week.

PR #1274, opened 2021-01-21
Adsk Contrib - More python tests and bindings fixes

BernardLefebvre:
- re-add commented tests.
- add new tests for FileRules.
- Fix issues outlined by tests.

Signed-off-by: Bernard Lefebvre <bernard.lefebvre@autodesk.com>

PR #1275, opened 2021-01-21
Restructure PyOpenColorIO, complete PyGpuShaderDesc

### LINKED ISSUE #1264: Python GpuShaderCreator/GpuShaderDesc missing uniform interface
The Python binding for `GpuShaderDesc` and `GpuShaderCreator` is currently missing `UniformData`, `getNumUniforms`, `getUniform`.

### LINKED ISSUE #1278: Add default constructor to all PyTransform subclasses
Following up from the discussion in #1277 , we should add a default constructor to these transform types:

- FixedFunctionTransform
- LogCameraTransform
- LookTransform

All other transforms support this. Part of this work would also involve revisiting the `TransformsTest.test_binding_group_polymorphism` test to remove the `default_args` handling.

michdolan:
Fixes #1264 
Fixes #1278

Add missing uniform interface to GpuShaderDesc. I have also removed the GpuShaderCreator constructor since its trampoline class was not able to support implementing the uniform interface without implementing some callback routing. That's not to say it can't be supported, but given the complexity and the low chance of it being implemented (subclassing GpuShaderCreator in Python), it can wait for a later update.

The bulk of this PR moves all core Python binding class declarations to PyTypes.cpp, which exposes the full bound interface to the other bindings, resolving references to C++ classes that were showing up in the Python documentation.

This will need some manual merging with #1274 and #1260


doug-walker (approved):

michdolan:
I'm done touching this now. Once this is in I believe all the known Python issues are now resolved for the v2 release.

docs/api/python/frozen/pyopencolorio_grouptransform.rst
```diff
@@ -69,7 +69,7 @@
 
       Overloaded function.
 
-      1. write(self: PyOpenColorIO.GroupTransform, config: PyOpenColorIO.Config = 0x1f17490 (version: 2), 0 active color spaces, 0 active displays., formatName: str, fileName: str) -> None
+      1. write(self: PyOpenColorIO.GroupTransform, config: PyOpenColorIO.Config = 0x26d9090 (version: 2), 0 active color spaces, 0 active displays., formatName: str, fileName: str) -> None
```
BernardLefebvre, line 72: Is the doc using Config.__repr__ with an empty shared pointer?
We should probably update Config.__repr__ to handle empty shared pointers.

michdolan: That's originating from the use of `EMPTY_CONFIG` being the default value in `GroupTransform.write`, yes.

michdolan: I'll update it.

BernardLefebvre: Maybe __repr__ should say "default" if self is empty (null pointer)

michdolan: It's currently printing the memory address, and in this case it's a valid, but empty, config shared ptr. I'm tweaking two things:

- Changing the config repr to print a simple `<Config name=, description=, etc.>` which will be consistent with other classes, but including the attributes currently present in the method.
- Changing GroupTransform so that the config arg defaults to `nullptr` which will be interpreted as `None` in Python, and will work with the current implementation without requiring an empty config be passed in. I do have to change the order of arguments to the write method though, since keyword args must follow positional args.

Commit coming soon.

BernardLefebvre: I tried it locally and it does not do what I expected.
This PR is great as it is.

BernardLefebvre:
Great!

hodoulp (approved):

PR #1276, opened 2021-01-21
Transform Python Bindings Fix - Clean

frankieeder:


frankieeder:
Closing this branch as signing errors still exist. See #1260.

Issue #1211, opened 2020-11-17
Getting the type of a transform returns the parent class.


KelSolaar:
Hi,

As discussed on Slack, `type(my_builtin_transform)` returns the parent class, i.e. `<class 'PyOpenColorIO.Transform'>` and not `BuiltinTransform`.

Currently using something like that for comparing, which is not super elegant: `transform.getTransformType() == ocio.TransformType.TRANSFORM_TYPE_BUILTIN`

Cheers,

Thomas

frankieeder:
Hey Thomas,

I've been meaning to try to start digging into the OCIO backend a bit and finally have some extra time in my schedule. I started taking a peek at this issue (that Colab notebook you put in Slack was super helpful), and on a first glance it seems to me like it might be related to some quirks with polymorphism in PyBind where it might be binding to an incorrect pointer type. Just to confirm, we would be looking to be able to do a more standard typing functionality like `type(transform) == ocio.TransformType.TRANSFORM_TYPE_BUILTIN`, like what seems to be possible for all other transforms, correct?

I'd be happy to try to give this a shot in the near future if someone could assign me to it.

Best,

Frankie

doug-walker:
Assigning to you @frankieeder , thanks for your help!

michdolan:
Thanks for your help @frankieeder . The goal would be to support this:

`type(transform) == ocio.BuiltinTransform`

Where it returns the `Transform` child class. Please note that there is some code in place in https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/src/bindings/python/PyOpenColorIO.h which supports `Transform` type downcasting in pybind11 via `polymorphic_type_hook`. This results in behavior like iterating a `GroupTransform` returning it's various child `*Transform` instances instead of all parent `Transform` instances.

frankieeder:
Got it! Thanks for the additional detail as well, looking into it now.

michdolan:
@frankieeder there are some additional clues in related issue #1254

zachlewis:
To consolidate issues, here's what I wrote for #1254:

check it:

>>> import PyOpenColorIO as ocio

>>> group_transform = ocio.Config().getProcessor(
...         BuiltinTransform("ARRI_ALEXA-LOGC-EI800-AWG_to_ACES2065-1")
...     ).createGroupTransform()

>>> print(f"Looks as expected: \n\n", group_transform)

Looks as expected: 

[...]

PR #1277, opened 2021-01-21
Transform Python Binding Fix

frankieeder:

### DIFF
--- a/src/bindings/python/PyOpenColorIO.h
+++ b/src/bindings/python/PyOpenColorIO.h
@@ -71,6 +71,10 @@ struct polymorphic_type_hook<OCIO::Transform> {
             {
                 type = &typeid(OCIO::AllocationTransform);
             }
+            if(dynamic_cast<const OCIO::BuiltinTransform*>(src))
+            {
+                type = &typeid(OCIO::BuiltinTransform);
+            }
             else if(dynamic_cast<const OCIO::CDLTransform*>(src))
             {
                 type = &typeid(OCIO::CDLTransform);
@@ -111,6 +115,10 @@ struct polymorphic_type_hook<OCIO::Transform> {
             {
                 type = &typeid(OCIO::GradingRGBCurveTransform);
             }
+            if(dynamic_cast<const OCIO::GradingToneTransform*>(src))
+            {
+                type = &typeid(OCIO::GradingToneTransform);
+            }
             else if(dynamic_cast<const OCIO::GroupTransform*>(src))
             {
                 type = &typeid(OCIO::GroupTransform);
@@ -118,6 +126,10 @@ struct polymorphic_type_hook<OCIO::Transform> {
             else if(dynamic_cast<const OCIO::LogAffineTransform*>(src))
             {
                 type = &typeid(OCIO::LogAffineTransform);
+            }
+            else if(dynamic_cast<const OCIO::LogCameraTransform*>(src))
+            {
+                type = &typeid(OCIO::LogCameraTransform);
             }
             else if(dynamic_cast<const OCIO::LogTransform*>(src))
             {
--- a/tests/python/OpenColorIOTestSuite.py
+++ b/tests/python/OpenColorIOTestSuite.py
@@ -73,13 +73,13 @@ import MixingHelpersTest
 import NamedTransformTest
 import OpenColorIOTest
 import ProcessorTest
+import TransformsTest
 import ViewingRulesTest
 #from MainTest import *
 #from ConstantsTest import *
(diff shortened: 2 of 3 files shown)


frankieeder:
@michdolan trying to merge with your changes from #1275 now as well. Seems to be why this PR is now failing.

michdolan:
My changes from #1275 have not been merged yet, so shouldn't affect your PR.

your added test is failing with:
```
======================================================================
3: FAIL: test_binding_polymorphism (TransformsTest.TransformsTest)
3: Tests polymorphism issue where transforms are cast as parent class when using
3: ----------------------------------------------------------------------
3: Traceback (most recent call last):
3:   File "/Users/runner/work/OpenColorIO/OpenColorIO/tests/python/TransformsTest.py", line 22, in test_binding_polymorphism
3:     allTransformsAsGroup.appendTransform(c())
3: TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
3:     1. PyOpenColorIO.FixedFunctionTransform(style: PyOpenColorIO.FixedFunctionStyle, params: List[float] = [], direction: PyOpenColorIO.TransformDirection = <TransformDirection.TRANSFORM_DIR_FORWARD: 0>)
3: 
3: Invoked with: 
3: 
3: During handling of the above exception, another exception occurred:
3: 
3: Traceback (most recent call last):
3:   File "/Users/runner/work/OpenColorIO/OpenColorIO/tests/python/TransformsTest.py", line 28, in test_binding_polymorphism
3:     'Unintended Error Raised: {0}'.format(e)
3: AssertionError: '__init__(): incompatible constructor argu[269 chars]th: ' != 'PyOpenColorIO.Transform: No constructor defined!'
3: + PyOpenColorIO.Transform: No constructor defined!- __init__(): incompatible constructor arguments. The following argument types are supported:
3: -     1. PyOpenColorIO.FixedFunctionTransform(style: PyOpenColorIO.FixedFunctionStyle, params: List[float] = [], direction: PyOpenColorIO.TransformDirection = <TransformDirection.TRANSFORM_DIR_FORWARD: 0>)
3: - 
3: - Invoked with:  : Unintended Error Raised: __init__(): incompatible constructor arguments. The following argument types are supported:
3:     1. PyOpenColorIO.FixedFunctionTransform(style: PyOpenColorIO.FixedFunctionStyle, params: List[float] = [], direction: PyOpenColorIO.TransformDirection = <TransformDirection.TRANSFORM_DIR_FORWARD: 0>)
```

michdolan:
The issue is that FixedFunction transform requires an argument, so you can't use an empty constructor. You can test this locally (after building OCIO) with:

```ctest -V```

frankieeder:
On it! Thanks for the pointers. My apologies, I misunderstood your PR.

michdolan:
Thanks for the contribution! It's very helpful!

frankieeder:
Looks like I got a working version. I could have sworn the checks passed in my earlier PR attempt, so not sure how I missed these args issues. Though this solution seems to work, it definitely doesn't feel ideal to typecheck during a typechecking test. I'm wondering:

- Is it fair to assume there's no issues with the direct bindings and we are only seeing issues (#1211) when using GroupTransforms? If so, this might be sufficient. Curious if anyone has ideas for how to (better) handle otherwise.
- If there's a better, perhaps more obvious, approach to handle these default arguments that I'm missing as someone newer to OCIO. I took a best guess at some simple args, but wasn't sure what the range for `linSideBreak` is.

Curious if there's any other thoughts or concerns as well. I'm more concerned with these tests catching possible errors in future Transform implementations than the existing ones.

michdolan (approved):

michdolan:
The issue can occur with GroupTransform, ColorSpace, Look, and any other class with a TransformRcPtr member and accessor. Even the GroupTransform itself needs to be downcasted when getting it from a ColorSpace for example. GroupTransform is a good test case though since it's pretty self contained, and it should reflect the behavior across the library since the implementation is a pybind11 hook, so applies to all casting.

I think your solution will accomplish the goal. If anything it suggests to me that we should add default constructors to those three transforms in a future release. If ColorSpaceTransform doesn't need a src and dst argument immediately, LookTransform shouldn't either. In some cases the issue is that we call Transform.validate() in the binding contructors, and some transforms are not valid without an attribute. But if we have a default constructor, it would not make sense to validate. I'll open an issue for this, and we can revisit this test at that time.

It's ok for your test to catch future errors IMO, and an excellent hint to a contributor that a new binding class doesn't support a default constructor. Excellent work.

doug-walker (approved):

PR #1285, opened 2021-01-25
Adsk Contrib - RB-2.0 release

BernardLefebvre:
All commits from master are good for V2 release.
Do not squash to keep commits.


BernardLefebvre:
Why are some incorrectly signed-off

PR #1288, opened 2021-01-26
Created a TransformsBaseTest module.

meimchu:
Moving some unit testing of Transforms module to a base class. Per feedback from: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1036#discussion_r456200810

Using CDLTransforms's direction as a starting test for feedback on formatting/structure.


zachlewis (approved):
Nice!

tests/python/TransformsBaseTest.py
```diff
@@ -0,0 +1,21 @@
...
+        """
+        Test the setDirection() and getDirection() methods.
+        """
+
+        # Default initialized direction is forward.
+        self.assertEqual(self.tr.getDirection(), OCIO.TRANSFORM_DIR_FORWARD)
+
+        for direction in OCIO.TransformDirection.__members__.values():
+            self.tr.setDirection(direction)
+            self.assertEqual(self.tr.getDirection(), direction)
```
zachlewis, line 21: [Minor] - add a newline at the end of the file)

```suggestion
            self.assertEqual(self.tr.getDirection(), direction)
            
```

zachlewis (approved):

tests/python/TransformsBaseTest.py
meimchu: Addressed with 474b0f6

tests/python/TransformsBaseTest.py
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright Contributors to the OpenColorIO Project.
+
+import PyOpenColorIO as OCIO
+
+
+class TransformsBaseTest(object):
```
hodoulp, line 7: That's a good idea @meimchu .

You could improve your idea by making this class the 'base' class for all unit tests:
1. Inherit from unittest.TestCase
2. Implement the tearDown() method to 'delete' the self.tr:
       def tearDown(self):
            self.tr = None
So you can remove the tearDown method from all the XTransformTest classes.

hodoulp (changes requested):

tests/python/TransformsBaseTest.py
meimchu: Please take a look when you have a moment. They should be addressed with https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1288/commits/e592a3d4937d2d4e48b281b02251c0f7b2d048ff and https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1288/commits/633d78178e2c6c83ac6e62bf0148e5e34bcdbc24

hodoulp: Without being a Python expert, I think that you could have created a `TransformsBaseTest::setUp` which sets `self.tr` to `None`. This would have 'created' the variable. Another aspect is that inheriting from a base class with some variable(s) a derived class must add `super(TransformsBaseTest, self).setUp()` in its `setUp` method if it needs one e.g. to initialize some members.
```
class TransformsBaseTest(unittest.TestCase):

    def setUp(self):
        super(unittest.TestCase, self).setUp()
        self.tr = None

    def tearDown(self):
        self.tr = None

class LookTransformTest(TransformsBaseTest):

    def setUp(self):
        super(TransformsBaseTest, self).setUp()
        self.tr = OCIO.LookTransform('src', 'dst')
```
If the derived class does not need some extra initializations then no setUp method is required but the base class setUp method is still called when needed.

meimchu: Upon some further research and also trying the example you've given, unfortunately it will not work. The reason is because `TransformsBaseTest(unittest.TestCase)` will trigger it to test in the base test. To give an example, I ran two unit test modules using your example above, `ExponentTransformTest` and `LogTransformTest` with this:

```
class TransformsBaseTest(unittest.TestCase):
    def setUp(self):
        self.tr = None
```
```
test_constructor_with_keyword (ExponentTransformTest.ExponentTransformTest) ... ok
test_constructor_with_positional (ExponentTransformTest.ExponentTransformTest) ... ok
test_constructor_wrong_parameter_type (ExponentTransformTest.ExponentTransformTest) ... ok
test_direction (ExponentTransformTest.ExponentTransformTest) ... ERROR
test_format_metadata (ExponentTransformTest.ExponentTransformTest) ... ok
test_negative_style (ExponentTransformTest.ExponentTransformTest) ... ok
test_transform_type (ExponentTransformTest.ExponentTransformTest) ... ok
test_validate (ExponentTransformTest.ExponentTransformTest) ... ok
test_values (ExponentTransformTest.ExponentTransformTest) ... ok
test_direction (TransformsBaseTest.TransformsBaseTest) ... ok
test_base (LogTransformTest.LogTransformTest) ... ok
test_constructor_with_keyword (LogTransformTest.LogTransformTest) ... ok
test_constructor_with_positional (LogTransformTest.LogTransformTest) ... ok
test_constructor_wrong_parameter_type (LogTransformTest.LogTransformTest) ... ok
test_direction (LogTransformTest.LogTransformTest) ... ok
test_format_metadata (LogTransformTest.LogTransformTest) ... ok
test_transform_type (LogTransformTest.LogTransformTest) ... ok
test_validate (LogTransformTest.LogTransformTest) ... ok
test_direction (TransformsBaseTest.TransformsBaseTest) ... ERROR
```
`test_direction` is called four times. Every time an individual transforms unit test is called, it is going to run the base class unit test. One of the solutions I saw online is basically what I have done before, having `TransformsBaseTest` inherit from `object` to bypass that (apparently called unit test mixin). Another method I am thinking of would be to include a skip in the `test_direction` if `self.tr == None` like this:
```
def test_direction(self):
    if self.tr == None:
        self.skipTest('Base class')
```

Let me know what you guys think or if there are any other suggestions! I appreciate your feedbacks and suggestions.

hodoulp: Thanks for investigations. You are right i.e. the best is to keep your approach.

hodoulp (approved):

meimchu:
I'm good with moving forward with another PR to build on top of this if you guys are ok with this PR.

hodoulp:
Please review the pull request so @meimchu can move forward. Thanks.

doug-walker (approved):

PR #1299, opened 2021-02-01
Fixes website overflow on mobile

Shrinks99:
- Fixes overflow caused by hero logo
- Fixes overflow caused by video iframe

Both of these caused a scroll bar to appear that went beyond the page content and now they do not!

Additionally there seems to have been a fix that was committed to solve the `object-fit` issue that caused this to happen on mobile but I don't think the site has been rebuilt to the gh-pages branch yet so this doesn't seem to be live, the code is there though!.

The grey bar on the side of this image is what this PR fixes.

Signed-off-by: Henry Wilkinson <henry@wilkinson.graphics>


hodoulp:
I do not have the problem on my smartphone (Android, Samsung S9) using Chrome & the Samsung Internet browser.
Can someone else check and validate the fix?

Shrinks99:
Interesting it doesn't happen on Chrome mobile for Android.  Tested with Firefox, Safari, and Chrome on  desktop (make the window really narrow) and only Firefox on iOS but at the end of the day they're _all_ WebKit anyways on iOS :P

Here's some better examples of the problem on desktop as well, the squished hero image is fixed in the code as mentioned but the website just needs to be rebuilt.  This PR does fix the margin issue that contributes to the overflow along with the video.

Existing on the left, this PR on the right.
OCIO Problems

hodoulp (approved):

carolalynn22 (approved):
Nice find, thanks! Good to go.

hodoulp:
Friendly reminder that the pull request can now be merged. @doug-walker or @michdolan, could you approve it?

doug-walker (approved):

Issue #1059, opened 2020-07-10
Update Public Header Documentation to Doxygen Style


scoopxyz:

doug-walker:
We discussed this at the 2020-11-13 documentation working group meeting.  The first deliverable for this issue should be to take one of the public headers (or even a section of it), put it into Doxygen style, and present the output (as it would appear in the documentation) for review by the community.

One issue to be decided is how verbose to make it.  The original OCIO headers had a very succinct style where many arguments, functions, and methods whose meaning was obvious were not documented.  Do we want to continue with this style or make it more verbose?

michdolan:
I think if a function is clearly named and has detailed type info it can remain succinct. If however there is any ambiguity about what a class or function does, or how it is intended to be used, we should document it with adequate verbosity to understand it without having to reach out to the community for explanation. Not that there's anything wrong with asking for help/clarification in the community, but it would be good to strive for understanding in isolation if possible.

zachlewis:
I volunteered at that 2020-11-13 documentation working group meeting to take a stab at this... and I've been procrastinating ever since :)

I have zero experience with doxygen, but I'm definitely willing to put the work in. Barring no objections, I'll try to have at least part of one of the headers converted and rendered for the next time we meet.

doug-walker:
Excellent, thanks Zach!

PR #1307, opened 2021-02-12
Convert public header docs to Doxygen style

zachlewis:
Addresses #1059 
(Work in progress)

This also includes @hodoulp's recent update to the `GpuShaderDesc` example code.


zachlewis:
Uh oh, let me create a new, cleaner PR -- this one has too much going on.

PR #1315, opened 2021-02-17
Convert public headers to doxygen style

zachlewis:
Addresses #1059

(Replaces #1307)


include/OpenColorIO/OpenColorIO.h
```diff
@@ -186,53 +185,49 @@ extern OCIOEXPORT void UnsetEnvVariable(const char * name);
...
+ * necessary to use customized ColorSpaceTransform and
+ * DisplayViewTransform operations.
+ *
+ * See the \ref user-guide for more information on
+ * selecting, creating, and working with custom color configurations.
+ *
+ * For applications interested in using only one color config at
+ * a time (this is the vast majority of apps), their API would
+ * traditionally get the global configuration and use that, as opposed to
+ * creating a new one. This simplifies the use case for
```
doug-walker, line 209: Would be nice to rewrap places where there are huge difference in adjacent lines.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -1000,7 +995,7 @@ class OCIOEXPORT Config
     /**
      * Most applications will use the preceding method, but this method may be
      * used for applications that want to know which was the highest priority rule to match
-     * filePath.  The :cpp:func:`FileRules::getNumCustomKeys` and custom keys methods
+     * filePath. \ref FileRules::getNumCustomKeys and custom keys methods
```
doug-walker, line 998: So is it correct that Doxygen finds the class references but it needs some extra help for this reference?  Could you share an HTML file that shows/confirms how the linking will work?

include/OpenColorIO/OpenColorIO.h
```diff
@@ -1033,9 +1028,9 @@ class OCIOEXPORT Config
     //
 
     // TODO: Move to .rst
-    // Create a :cpp:class:`Processor` to assemble a transformation between two
-    // color spaces.  It may then be used to create a :cpp:class:`CPUProcessor`
-    // or :cpp:class:`GPUProcessor` to process/convert pixels.
+    // Create a Processor to assemble a transformation between two
+    // color spaces.  It may then be used to create a CPUProcessor
+    // or GPUProcessor to process/convert pixels.
     // rst:: Get the processor to apply a ColorSpaceTransform from a source to a destination
```
doug-walker, line 1034: Is the "rst::" still needed?

doug-walker (commented):
Thanks for working on this Zach.

include/OpenColorIO/OpenColorIO.h
zachlewis: Good call. It looks like you guys are using roughly ~100 characters for documentation in other headers; I'll try to straighten things up to match that.

There are some documentation style inconsistencies throughout as well that I was gonna ask about; but first, I wanted to get to a point where everything was rendering properly. (My work on the docs has been a little haphazard, as I haven't figured out a great workflow for myself yet, and I don't totally know what I'm doing -- just rebuilding and trying to decipher error messages as they pop up. It's been a learning experience...)

include/OpenColorIO/OpenColorIO.h
zachlewis: No sir. 

There are some regions that differentiate between documentation directly associated with class and function definitions, and documentation that more generally explain concepts. I'm not 100% certain of this, but it seems like doxygen only knows what to do with the former, and I think I may have left some `rst::` bits in place either because I hadn't figured out what to do with the latter, or because I missed them. Or both.

In other words, if the comments aren't directly above a definition of some sort, they seem to get ignored when rendering. For instance, there might be a block of comments that introduces a concept worth documenting that applies to a subset of class functions. This makes sense visually and organizationally in the header itself, but I'm finding that these "floating comments" just get ignored by doxygen. In some cases, I've moved these comment blocks up to the class-level documentation so they get rendered, but I'm not sure that's the best solution either, since it undermines the author's original intention, in terms of making the header file itself more readable, which is of the upmost importance. On the other hand, it's not intuitive to authors that these comments will not get rendered at all. 

I'm not sure what the best solution is. I think maybe these comments should stay in place, as the author intends; but worthwhile concepts should be manually copied or promoted elsewhere for rendering. Considering how many gems are hidden in the comments in much of the cpp code, I wonder if its worth stubbing out separate hand-authored "concept" pages. 

(From a user perspective, I'd kind of expect the API docs to supplement Concept docs curated elsewhere)

include/OpenColorIO/OpenColorIO.h
zachlewis: To be honest, I'm still wrapping my head around references, and they're the most WIP-y, drafty part of this PR.

For certain, the `:cpp:func:` stuff is only getting in the way (it renders as text), and I'm pretty sure the `\ref` isn't actually doing anything useful itself. Doxygen itself knows to automatically link the text `FileRules::getNumCustomKeys` in this case, which is in fact what's happening in the [presently-live docs](https://opencolorio.readthedocs.io/en/latest/api/config.html#_CPPv4NK16OpenColorIO_v2_06Config25getColorSpaceFromFilepathEPKcR6size_t). Doxygen will likewise do its best to any other classes / enums / functions it recognizes. 

I'm pretty sure adding the `\ref` isn't helping anything, but it doesn't show up in the rendered HTML, and it seems benign (although I do need to check if need to check if causes warnings). I figured it might help Michael's python doc stuff differentiate between references and plain old words. And to be perfectly clear, this particular change doesn't change the linking behavior on the cpp side, it just takes out the `:cpp:func:` business, and removes the backtick-style  formatting (which we may want to keep).

(Sidebar: In retrospect, it might maybe it's better to just use backticks instead of using `\ref ` as a convention for indicating to Michael's Python docs generation stuff which words should be considered a link)

However, there's something else happening with this particular example -- it's an overloaded function, but the python side doesn't seem to recognize it as such; it's only rendering the documentation for the previously-defined function, despite having a different signature. I'm not sure why this is happening, because overloaded functions are handled fine [in other cases](https://opencolorio.readthedocs.io/en/latest/api/config.html#PyOpenColorIO.Config.GetProcessorFromConfigs).

include/OpenColorIO/OpenColorIO.h
```diff
@@ -1149,64 +1142,49 @@ class OCIOEXPORT Config
...
+ *   match and the next rule will be considered.
+ *   see \ref insertPathSearchRule.
+ *   It has the key:
+ *   * name: Must be "ColorSpaceNamePathSearch".
+ * * Default Rule: The file_rules must always end with this rule. If no prior rules match,
+ *   this rule specifies the color space applications will use.
+ *   see \ref setDefaultRuleColorSpace.
+ *   It has the keys:
+ *   * name: must be "Default".
+ *   * colorspace : Color space name to be returned.
```
zachlewis, line 1178: I need to take another stab at this. 
 - I'd like to restore the newlines if possible, to improve readability
 - The sub-bullets aren't rendering as sub-bullets, they're rendering as regular bullets, which isn't too helpful.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -2714,13 +2676,55 @@ class OCIOEXPORT PlanarImageDesc : public ImageDesc
...
+ *     vec4 OCIODisplay(in vec4 inColor)     //
+ *     {                                     // Function Header
+ *        vec4 outColor = inColor;           //
+ *     
+ *        outColor.rgb = texture3D(tex3, computePosition(inColor.rgb)).rgb;
+ *     
+ *        return outColor;                   // Function Footer
+ *     }                                     //
+ * 
+ * \endcode
```
zachlewis, line 2727: These code examples show up on the Python side too.  
This is probably gonna complicate things, but is there any way we can write Python code that only shows up on the Python side, and cpp code that only shows up on the cpp side?

include/OpenColorIO/OpenColorIO.h
```diff
@@ -2834,51 +2838,7 @@ class OCIOEXPORT GpuShaderCreator
...
-    //
-    // 1. A declaration part  e.g., uniform sampled3D tex3;
-    //
-    // 2. Some helper methods
-    //
-    // 3. The OCIO shader function may be broken down as:
-    //
-    //    1. The function header  e.g., void OCIODisplay(in vec4 inColor) {
-    //    2. The function body    e.g.,   vec4 outColor.rgb = texture3D(tex3, inColor.rgb).rgb;
-    //    3. The function footer  e.g.,   return outColor; }
```
zachlewis, line 2850: @michdolan, I think the Python stuff has trouble with numbered lists -- the build failed until I changed these to unnumbered lists. I'm sorry, I don't have the traceback available.

michdolan: Ok, I'll need to have look at the doxygen extraction code. Should be fairly easy to add.

include/OpenColorIO/OpenColorIO.h
michdolan: This will require overriding the docstrings in question directly in the python binding source.

michdolan:
Thanks for your work on this @zachlewis . Much appreciated.

hodoulp:
Hi @zachlewis, thanks for reviewing the documentation as that's important for the new release. What is the status of the pull request?

zachlewis:
Heya @hodoulp -- this PR should be good to merge -- the public headers have been converted to doxygen style.

include/OpenColorIO/OpenColorIO.h
zachlewis: This wasn't as large of an issue as I thought. I rearranged a couple of notes in order to make sure they were rendered to HTML, but in general, I erred on the side of developer-facing in-line readability.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -2132,15 +2094,15 @@ class OCIOEXPORT Processor
...
+     * Run the optimizer on a Processor to create a new Processor.
      * It is usually not necessary to call this since getting a CPUProcessor or GPUProcessor
      * will also optimize.  However if you need both, calling this method first makes getting
      * a CPU and GPU Processor faster since the optimization is effectively only done once.
      */
     ConstProcessorRcPtr getOptimizedProcessor(OptimizationFlags oFlags) const;
 
     /**
-     * Create a :cpp:class:`Processor` that is optimized for a specific in and out
+     * Create a Processor that is optimized for a specific in and out
```
doug-walker, line 2105: This could be wrapped better.  :)

doug-walker (approved):
Thank you Zach!!

hodoulp:
Hi @zachlewis, when you will have some time could you fix the merge conflict & fix the @doug-walker comment?

zachlewis:
> Hi @zachlewis, when you will have some time could you fix the merge conflict & fix the @doug-walker comment?

Done and done!

hodoulp (approved):

PR #1328, opened 2021-02-27
Update Lut1DTransformTest and Lut3DTransformTest

meimchu:
- Conform those tests with TransformsBaseTest.

Signed-off-by: Mei Chu <meimchu@gmail.com>

### DIFF
--- a/tests/python/Lut1DTransformTest.py
+++ b/tests/python/Lut1DTransformTest.py
@@ -16,50 +16,38 @@ except ImportError:
     np = None
 
 import PyOpenColorIO as OCIO
+from TransformsBaseTest import TransformsBaseTest
 
 
-class Lut1DTransformTest(unittest.TestCase):
+class Lut1DTransformTest(unittest.TestCase, TransformsBaseTest):
+
+    def setUp(self):
+        self.tr = OCIO.Lut1DTransform()
 
     def test_default_constructor(self):
         """
         Test the default constructor.
         """
-        lut = OCIO.Lut1DTransform()
-        self.assertEqual(lut.getLength(), 2)
-        self.assertEqual(lut.getDirection(), OCIO.TRANSFORM_DIR_FORWARD)
-        self.assertEqual(lut.getHueAdjust(), OCIO.HUE_NONE)
-        self.assertFalse(lut.getInputHalfDomain())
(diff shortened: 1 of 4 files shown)


doug-walker (approved):
Thanks Mei!

tests/python/Lut1DTransformTest.py
```diff
@@ -16,50 +16,38 @@
     np = None
 
 import PyOpenColorIO as OCIO
+from TransformsBaseTest import TransformsBaseTest
 
 
-class Lut1DTransformTest(unittest.TestCase):
+class Lut1DTransformTest(unittest.TestCase, TransformsBaseTest):
```
michdolan, line 22: Apologies for missing the previous review of `TransformBaseTest`. Have you considered making that base class inherit from `unittest.TestCase`, and then have the individual transform test cases inherit from only that, instead of using the base class as a mixin? I'm curious if the `tearDown` method from the base class gets called. It might, but would be worth confirming.

michdolan (commented):
Thanks for the contribution Mei! Just had one question/thought related to using multiple inheritance here.

tests/python/Lut1DTransformTest.py
meimchu: Hey Michael! So the reason for this is explained here: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1288#discussion_r572638482

I would love to know if there is a better way if you could show me. But the result of my testing is that `TransformBaseTest` base class would get call on if it inherits from `unittest.TestCase` on initialization. So without being able to initialize the `self.tr` to the individual transforms object, it is `None` which causes multiple call to `test_direction` to fail. Let me know what you think about this and thank you.

hodoulp: @michdolan @meimchu Is the discussion now completed?

hodoulp (approved):

tests/python/Lut1DTransformTest.py
michdolan: Sorry for the late reply. Thanks for sharing the link to that thread @meimchu . Sounds like you've already explored that option, so this approach should be fine. Thanks again for all your work on this!

PR #1333, opened 2021-03-04
[BakeLUT] Added --view option for display & view

### LINKED ISSUE #1323: Add "--display" and "--view" options to ociobakelut tool
Currently, `ociobakelut` makes it super easy to bake a LUT that converts from one colorspace to another. For example:
```
ociobakelut --inputspace 'logc' --outputspace 'acescg' --format cinespace logc_to_acesg.csp
```

It would be just **wonderful** if we could optionally specify a display + view instead of an output space. For instance:
```
ociobakelut --inputspace 'logc' --display 'DCI-P3' --view 'K1S1' --format cinespace logc_k1s1_p3.csp
```

EmperorYP7:
This PR fixes #1323


src/apps/ociobakelut/main.cpp
```diff
@@ -162,54 +183,90 @@ int main (int argc, const char* argv[])
...
+            OCIO::ConfigRcPtr editableConfig = OCIO::Config::Create();
+            OCIO::ColorSpaceRcPtr inputColorSpace = OCIO::ColorSpace::Create();
+            inputColorSpace->setName(inputspace.c_str());
+            editableConfig->addColorSpace(inputColorSpace);
+
+            OCIO::ColorSpaceRcPtr outputColorSpace = OCIO::ColorSpace::Create();
+            outputspace = "ProcessedOutput";
+            outputColorSpace->setName(outputspace.c_str());
+
+            auto t = OCIO::DisplayViewTransform::Create();
```
EmperorYP7, line 261: Here I wanted to get the `DisplayViewTransformRcPtr` from `groupTransform` using `groupTransform->getTransform(0)` but I wasn't able to cast it correctly. If someone is ready to help me with this, I'd be glad.

src/apps/ociobakelut/main.cpp
```diff
@@ -129,6 +138,13 @@ int main (int argc, const char* argv[])
         verbose = false;
     }
 
+    else if (useLut && useDisplayView)
+    {
+        std::cerr << "ERROR: Options --lut & --view can't be used at the same time" << std::endl;
+        ap.usage();
+        exit(1);
```
michdolan, line 145:
```suggestion
        return 1;
```

michdolan: For consistency with other exit conditions.

src/apps/ociobakelut/main.cpp
```diff
@@ -158,58 +174,111 @@ int main (int argc, const char* argv[])
...
-        outputColorSpace->setName(outputspace.c_str());
+            //Setting up the output ColorSpace
+            OCIO::ColorSpaceRcPtr outputColorSpace = OCIO::ColorSpace::Create();
+            outputspace = "ProcessedOutput";
+            outputColorSpace->setName(outputspace.c_str());
 
-        outputColorSpace->setTransform(groupTransform,
-            OCIO::COLORSPACE_DIR_FROM_REFERENCE);
+            // Creating the transform and applying to outputColorSpace
+            OCIO::DisplayViewTransformRcPtr t = OCIO::DisplayViewTransform::Create();
```
michdolan, line 273: This transform already exists in the group transform, albeit without a src color space. Recommend refactoring to prevent defining it twice. In general I think the `--display` option can be treated more like `--outputspace` than `--lut` in terms of matching implementation behavior. A display/view pair is in essence just another way of defining the output color space. It's more than that under the hood, but in terms of user experience it's equivalent.

michdolan (changes requested):
Thanks for your contribution @EmperorYP7

src/apps/ociobakelut/main.cpp
```diff
@@ -96,11 +101,15 @@ int main (int argc, const char* argv[])
                "--format %s", &format, formatstr.c_str(),
                "--shapersize %d", &shapersize, "size of the shaper (default: format specific)",
                "--cubesize %d", &cubesize, "size of the cube (default: format specific)",
+               "--view %s %s", &view, &display, "for OCIO view and display instead of outputspace",
```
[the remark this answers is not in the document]
zachlewis: Agreed that we should use the `--display <displayname> <viewname>` syntax. 
Maybe we should call the option `--displayview`? 

We should use the same convention across apps, whatever we do

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/c4a4b2bcd68d28c483bd13d3e9ffed265391db8f/src/apps/ocioconvert/main.cpp#L81-L82

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/c4a4b2bcd68d28c483bd13d3e9ffed265391db8f/src/apps/ociowrite/main.cpp#L47-L48

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/c4a4b2bcd68d28c483bd13d3e9ffed265391db8f/src/apps/ocioperf/main.cpp#L287-L288

hodoulp:
Hi @EmperorYP7, 
Do you think you will have some time to finalize the work?
If you still have any question(s) concerning the comments, please comment in the pull request so we can help you.

EmperorYP7:
> Do you think you will have some time to finalize the work?

I did just now. Kindly review this again and let me know if there's something to change.

zachlewis:
Thanks Yash, much appreciated!

Per https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1333#discussion_r597731073, before we merge, let's reconcile the `"--display"` / `"--view"` / `"--displayview"` /  option name across  `ociobakelut` / `ocioconvert` / `ocioperf` & `ociowrite` respectively. 

I'm happy with using `--displayview` everywhere.

This would mean the `ocioconvert` tool usage would mildly diverge from the OCIO-1 version the tool, but I don't think the change is going to rock anyone's world too hard.

zachlewis (changes requested):
The `vendor/openfx/openfx` submodule doesn't resolve for me; let's nuke this suker.

EmperorYP7:
> The `vendor/openfx/openfx` submodule doesn't resolve for me; let's nuke this suker.

Nuke detonated successfully! 😆

src/apps/ociobakelut/main.cpp
```diff
@@ -158,58 +174,109 @@ int main (int argc, const char* argv[])
...
+            OCIO::ColorSpaceRcPtr outputColorSpace = OCIO::ColorSpace::Create();
+            outputspace = "ProcessedOutput";
+            outputColorSpace->setName(outputspace.c_str());
 
-        outputColorSpace->setTransform(groupTransform,
-            OCIO::COLORSPACE_DIR_FROM_REFERENCE);
+            // Obtaining the transform and applying to outputColorSpace
+            OCIO::DisplayViewTransformRcPtr t = std::dynamic_pointer_cast<OCIO::DisplayViewTransform>(groupTransform->getTransform(0));
+            t->setSrc(inputspace.c_str());
+            outputColorSpace->setTransform(t, OCIO::COLORSPACE_DIR_FROM_REFERENCE);
```
doug-walker, line 275: Huh?  It doesn't seem like this would work.  The transform is in the user's config.

doug-walker (changes requested):
@EmperorYP7 , thanks for attempting this, but it doesn't seem like the code would actually work.  Did you test it somehow?

I think the real problem with implementing the feature is that the Baker class does not support using a display and view.  IMHO, the proper way of implementing this would be to upgrade the Baker class.  Then the ociobakelut changes are pretty trivial.

remia:
What's the status of this PR ? Just noticed it was not included in v2.0.1, would that be a good candidate for the v2.1.0 or do you feel the changes required would be too important given the time constraint ?

hodoulp:
If @EmperorYP7 does not object to, I think you can jump on it, and I do think the work is not complex. If you have time there is also #1423 .

remia:
Ok @hodoulp, if I understand correctly an approach to solving this with minimal refactor to existing code would be to add `setDisplay()` and `setView()` methods to the `Baker` class and use this information to fill the internal `inputspace`, `looks` and `outputspace` member based on the display transform decomposition. This way all existing FileFormat baking method do not have to change.

Is that what you were thinking @doug-walker ?

doug-walker:
Remi, I'm not sure how to do it without making some changes to all the fileFormats.  I'm also not sure how it's possible to set the outputspace (aka targetspace) from the display and view.  In other words, there may not always be an outputspace that corresponds to a display/view in a v2 config.  So you won't be able to create the Processor via getProcessor(src,dst), you'll need to use getProcessor(src, display, view).

It seems the fileFormat bake methods have a clause "if (!looks.empty()) { ... build look ... }  else { getProcessor( src, dst }" and you would probably need to add a third clause which calls getProcessor(src, display, view).  I have not looked at it carefully, but I would be curious why this code is in the fileFormats and not in the Baker class itself, so there may be a useful refactor to consider.

remia:
Thanks @doug-walker, you are totally right I was still in the OCIO v1 mindset!

remia:
I believe we can use some of the changes here but like @doug-walker suggested, this would be an opportunity to refactor how the baker works. I believe @zachlewis is also investigating this from an other angle in Python. In the end we will probably have to drop this particular PR and spin up a new one with said refactor, do you agree ?

doug-walker:
Yes, I agree with Remi.

hodoulp:
Unless there are some objections, I will definitively close this pull request next week as the PR [#1598](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1598) from @remia resumes that work.

Issue #1327, opened 2021-02-26
Manage the RPATH on all platforms


hodoulp:
When running some tests, I discovered that the RPATH management is completely missing in OpenColorIO.

On MacOS, the only guaranty to use a specific version is to set the `DYLD_LIBRARY_PATH` env. variable to the directory containing the 'modified` library before to run a command line tools for example.

At least the install step should handle the RPATH:
```
# Setup osx rpathing
SET (CMAKE_MACOSX_RPATH 1)
SET (BUILD_WITH_INSTALL_RPATH 1)
```
or a more elaborate management like the OpenImageIO example:
https://github.com/OpenImageIO/oiio/blob/00174059858774bc8927adb314ecb56725fb7bfe/src/cmake/compiler.cmake#L548

hodoulp:
`otool -L` for MacOS or `ldd` for Linux output the library/binary dependencies.

remia:
I believe `CMAKE_MACOSX_RPATH` is the default behaviour but yeah, install RPATH are empty currently.

For what it's worth, I have been using those flags (`-DCMAKE_INSTALL_RPATH="@loader_path/../lib;@loader_path/../.."`) when generating the project on macOS (similar to what OpenEXR do) to avoid fiddling with `DYLD_LIBRARY_PATH`. This handle both cmd line tools and python bindings (provided you don't move it `PyOpenColorIO.so`).

`otool -l /path/exe | grep RPATH -A4` is also helpful to debug the rpath on macOS.

hodoulp:
Something to read before doing anything:
https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling

hodoulp:
To be tested but my proposal is:
```
if (UNIX)
    # Do not use the install full path i.e. avoid machine specific paths.
    set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)

    # With the 'usual' install path structure in mind, search from the bin directory
    # (i.e. a binary loading a dynamic property) and then from the current directory
    # (i.e. dynamic library loading another dynamic library).  
    if (APPLE)
        set(CMAKE_INSTALL_RPATH "@executable_path/../lib;@loader_path")
    else()
        set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib;$ORIGIN")
    endif()
endif()
```

PR #1338, opened 2021-03-15
Adsk Contrib - Add the RPATH support

hodoulp:

The pull request adds the `RPATH` to correctly find & load the library.

The details are:
- `cmake` -> the RPATH is the compilation directory build.
- `cmake install` -> the RPATH is a relative path based on the installation directory structure i.e. usual value for *nix platforms.

Refer to #1327 for details.


michdolan (approved):

remia:
Works fine on macOS, thanks @hodoulp.

The generated python bindings will still fallback to a system wide location as the relatives path will not work so user might still want to manually edit these install RPATH. But that's probably a too specific specific situation to do anything generic about it, just need to be aware of it I guess.

hodoulp:
@remia, if the way ocio installs the Python binding dynamic library is the usual installation directory structure for *nix platforms (i.e. `lib/pythonX.Y/site-packages` ), I can definitively add `@loader_path/../..` to the rpath.

remia:
Thanks @hodoulp, great to use `set_target_properties` to limit the scope.

It works for me, and I don't see reason not to do it so far :)

hodoulp:
Tested on:
- ubuntu 20:4, cmake 3.16, gcc 9.3
- macOS 10.15, cmake 3.18, clang 12.0

michdolan (approved):

hodoulp:
Friendly reminder that the pull request can now be merged next week.

doug-walker (approved):

PR #1350, opened 2021-03-19
Adsk Contrib - Fix 'setup_ocio' script support

hodoulp:

The pull request re-enables the `setup_ocio` script for all the three platforms. The script helps user to set all env. variables to make the installed library available. Refer to #1341 for details.

Tests were done on MacOS, Linux (ubuntu) and Windows:
- `cmake -DCMAKE_INSTALL_PREFIX=../install ../.`
- `cmake all -jX`
- `cmake install`
- `cd ~`
- `source [xx]/install/share/ocio/setup_ocio.sh`
- `ociochecklut`

As mentioned in the issue, there are two `Nuke` related env. variables which do not seem needed anymore. Could we remove them?


zachlewis:
> As mentioned in the issue, there are two `Nuke` related env. variables which do not seem needed anymore. Could we remove them?

Yes, please remove them.

mjtitchener-fn:
I agree with Zach we should remove them. My feeling is if someone is building the Nuke plugins it's quite likely they'll move them to an existing location in their NUKE_PATH and not load them directly from the OCIO build/install dir.

hodoulp:
@mjtitchener-fn highlights a valid point about the `System Integrity Protection` of Apple which restricts the use of `DYLD_LIBRARY_PATH`.

Below his the thread:

>Hi @patrick I think we can remove them. I’ve added a comment to the PR.
Something related (but not Nuke) that I’d like to ask. In this file we define the following for macOS
export DYLD_LIBRARY_PATH="@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_LIBDIR@:${DYLD_LIBRARY_PATH}"
I was having a conversation about this with some Foundry engineers and they suggested that we shouldn’t be doing this on macOS anymore due to the increased security measures Apple introduced (SIP?).
I removed this variable from my own macOS system and I’m still able to use the ocio apps and the OCIO Python API with just the following set
export PATH="@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_BINDIR@:${PATH}"
export PYTHONPATH="@CMAKE_INSTALL_PREFIX@/@PYTHON_VARIANT_PATH@:${PYTHONPATH}"
So I was wondering if DYLD_LIBRARY_PATH is even being used on mac and if it’s something we could remove from this file? Is this something you have any knowledge of?
>> Hi @Mark Titchener, Thanks to bring that concern to my attention.
With my current understanding, the use of DYLD_LIBRARY_PATH has to be carefully balanced against Apple System Integrity Protection (which protects some system directories). However, a dev could still install OpenColorIO in his/her private space (i.e. /Users/username/*) where the script is then still useful. For the ‘normal’ installation (e.g. /usr/*) the env. variable could not be useful. But the pull request 1338 will help in that scenario because it now correctly sets the rpaths.
There is also the code signing to consider on MacOS which by design, cannot be done by the OpenColorIO cmake.
Not being a MacOs expert, I will be happy to hear more about improvement/workaround/solution to that potential problem.

michdolan (approved):

doug-walker (approved):

Issue #406, opened 2016-04-26
Issues with non-ascii folders on Windows


sergeyvfx:
It is not possible to use configuration file which is located in a folder with non-ASCII characters on Windows. This is because simple `char*` is used to communicate filepath to a config file in `Config::CreateFromFile()`.

In order to have full Unicode support it seems adding similar `Config::CreateFromFile(wchar_t*)` will be needed. Or at least will need to be done some ASCII->UTF conversion in the function itself.

In order to reproduce the issue it should be enough to simply call `Config::CreateFromFile()` with a filepath which has non-ascii characters, such as German umlauts for example.

hsdk123:
Hi, would it be possible to request an update to this? This has major impact on loading from non-english environments on Windows. The fix doesn't seem to be too difficult either.

EvanAW:
I am looking to get a pull request in for this hopefully later this week.

Dithermaster:
BTW, the way we solved this host-side was to call `GetShortPathName` Windows API to convert paths to something that OCIOv1 can eat.

PR #1351, opened 2021-03-19
Adsk Contrib - Improve the C++ version detection

hodoulp:

The pull request sets the default C++ version to be 17 (instead of the former 11) in order to follow the VFX Reference Platform recommendation for CY2021.

Notes:

All recent versions of the compilers we generally support (i.e. gcc, clang, msvc) support C++17.
But that's not the case for 11 and 14. So, if 11 is requested but not supported, the code automatically jumps to 14 and so on e.g. recent Visual Studio versions dropped the C++11 support.

If for any reason, an older compiler version must be used then `cmake -DCMAKE_CXX_STANDARD=11 ../.` is the command to select the appropriate C++ version. So any C++ version (i.e. 11, 14, and 17 for now) can be selected. The only hard stop is the use C++ versions 98 or 03 which are not anymore supported.

However on long term we will need a guideline / policy to define which versions to support. The challenge is that dev could at any time start using C++17 specific language features (e.g. not present in 11 or 14). For short-term, we have CI builds to trap that if it happens.


lgritz:
Data point: At our studio, we're just now phasing out C++11 and gcc4.8 and only building on gcc6/C++14. We have not yet started building anything for C++17 for production deployment. I think a lot of other places are in a similar situation.

doug-walker:
Thanks Larry, that's very helpful info.  Currently a number of the CI setups are building with C++ 11, so that effectively places an upper limit on what developers may currently use.

hodoulp:
@lgritz Thanks for the information.

There is also the public API versus implementation aspect to consider i.e. decoupling or not the public API from the implementation. The OpenColorIO public API imposes severe restrictions to avoid STL stuffs for example. So, building a dynamic library could potentially authorize C++17 implementation with C++11/14 public API. On the other hand, building static library definitively imposes same C++ version for OpenColorIO and its host.

Note: Once again a C API would have nicely decouple internal from external for the dynamic library case as well as ease the writing of various bindings e.g. Python, Rust, etc.

Anyway, OpenColorIO currently supports the three versions, and the CI builds enforce that status.

lgritz:
I think that realistically, it's fine to drop C++11 support from the current master (i.e. your next major release after 2.0). But I think it's early for the default build mode to be C++17. That's going to cause a lot of studio internal builds and other projects' CI to stumble and break until they adjust the build flags to lock it down explicitly to 11 or 14.

I think there is a discussion to be had in the TAC for how/whether we issue guidance about how far back in VFX Platform years is prudent for foundational libraries to support. Lots of projects are all asking the same questions.

@hodoulp From a linkage point of view, and for users whose OCIO comes from a pre-built system package manager, you are correct that you could adopt a policy of internal code and build being C++17 and external API being C++14. But that is extremely awkward for studios building from source, because a lot of the build and deployment systems they use, like [Rez](https://github.com/nerdvegas/rez), usually describe the compiler and/or C++ standard as as a dependency with the standard levels being incompatible "variants" of that package, so it's very awkward to describe a package like OCIO having a different set of dependencies variants internally versus externally. I think Rez would just throw up its hands and declare it impossible to resolve if the "requires" and "build_requires" lists held mutually exclusive versions of the compiler.

hodoulp:
@lgritz The pull request is here with the idea to talk about the next step (which C++ version?) at the OCIO TSC meeting this Monday. I selected C++17 to follow the VFX Reference Platform recommendation. But you raise valid concerns so I will certainly propose C++ 14.

Agree that the decoupling between the public API and the implementation is challenging even for the library itself. But the important point is to benefit or not from the latest C++17 features (e.g. filesystem, new algorithms, etc.). For short-term, there is no 'urgent' or 'pending' need to use some C++17 feature in OpenColorIO but it will pop at some point. We need to have some kind of policy/rule before it happens.

What is the approach for OIIO or OSL?

The two C++17 features with potential huge benefit for OpenColorIO are the `std::pmr` and `std::to_chars & std::from_chars` which both could really speedup the library read of config & lut files. But there are C++17 only.

lgritz:
I know, I've been lusting after C++17 for years, and now I'm starting to want some C++20 things. :-)

Currently, OIIO and OSL build for any of C++11, 14, or 17 (and I just added C++20 for OIIO, OSL is a little trickier). But the default mode is 11. This is strictly a function of studios until very recently still using some DCC's that needed C++11/gcc4.8 compatibility. Needless to say, this means that although we can build for 14/17/20, we actually use very few features beyond 11. (We mark more things as constexpr the higher you go, or we use std instead of boost equivalents, that kind of thing, but it's hard to do anything more aggressive if you still need to build across all those versions.)

But I think we're right now wrapping up our last show that's dependent on the gcc4.8-based toolchain, they are locked down and I don't need to build new software for them, so I'm just days from bumping the minimum for both (in their master branches, i.e., the next major release) to C++14/gcc6.3. I think I'll keep the default at 14 to minimize complaints.

I'm interested in some combination of the TAC and VFX Platform giving sunset dates for VFX Platform years. 

In general, I'm in favor of all of us staying much closer to current standards at all times, but we all need to do it together or it's just a support nightmare.

Simran-B:
> C++17 features (e.g. filesystem, […] no 'urgent' or 'pending' need to use some C++17 feature in OpenColorIO but it will pop at some point.

I would consider the issues with [non-ASCII characters in file paths on Windows](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/406) as pending at least (suspected to cause crashes of the host application). My understanding is that std::filesystem would resolve it without Windows-specific code path on OCIO side. Using C++17 isn't the only way however, there are also boost::nowide, native APIs etc.

Similarly, the issues with [non-C locales](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/297) would be resolved by charconv and allow to serialize numbers to their shortest round-trip safe representation throughout (aside from performance gains). I suppose this could be done in user land somehow to not rely on standards, but it would probably be a huge effort to implement and test.

Would different standards per platform be an option? It would certainly complicate the code though and maintaining compatibility between standards and platforms is a chore.

lgritz:
I have some utility functions in OIIO for dealing with UTF-8 filenames and also for locale-independent formatting that you are welcome to steal, to avoid adding any new Boost dependencies.

hodoulp:
Following your feedback, the TSC members agreed to stay with C++ 11 for now.
The pull request now only improves the C++ version detection.

EvanAW:
> 
> 
> I have some utility functions in OIIO for dealing with UTF-8 filenames and also for locale-independent formatting that you are welcome to steal, to avoid adding any new Boost dependencies.

Are you referring to what you said here @lgritz?
https://github.com/AcademySoftwareFoundation/openexr/issues/292#issuecomment-391081927

lgritz:
Yes, and more here: https://github.com/OpenImageIO/oiio/blob/master/src/libutil/filesystem.cpp#L447
Wrappers for open/fopen and various other common file operations that, on windows, will automatically do the conversion to wide chars and call the appropriate windows functions.

As far as the locales, we also use wrappers around the awesome https://github.com/fmtlib/fmt which is locale-independent and also much faster than most of the system conversions into and out of strings. And it supports the C++20 std::format notation, which is like Python string formatting (though also has a printf notation version).

hodoulp:
Thanks to Maya, we found an issue when using custom CMAKE_CXX_FLAGS because it could trigger unrelated warnings during the C++ version detection disabling the requested C++ version.

hodoulp:
Now that the build is fixed, I compiled & ran CPU & GPU unit tests with a sanitization.

Note: Unfortunately the Python binding unit tests fail for now.

doug-walker (approved):

hodoulp:

michdolan (approved):

PR #1363, opened 2021-04-09
Fixes Unicode paths on Windows

### LINKED ISSUE #406: Issues with non-ascii folders on Windows

itsmattkc:
Fixes #406.

Unlike most other platforms, Windows' Unicode is standardized around UTF-16, an encoding not compatible with `char *` arrays common in C/C++. As such, to support Unicode correctly when using Win32 APIs, strings must be converted to and from UTF-16 and the Unicode versions of the APIs must be used over the ANSI versions. 

This commit introduces the following:

- Utility functions for converting between UTF-8 and UTF-16LE on all platforms:
  - `Platform::Utf8ToUtf16`
  - `Platform::Utf16ToUtf8`
  - NOTE: These functions use `std::codecvt_utf8_utf16` on macOS and Linux but `MultiByteToWideChar` and `WideCharToMultiByte` on Windows because MSVC gave deprecation warnings recommending the Win32 functions with the former.

- Test for these conversion functions to ensure the conversion to and from UTF-8 and UTF-16LE is correct.

- Utility wrappers for "std::ifstream" that automatically convert to and from UTF-16 on Windows so that filenames requiring Unicode encoding function correctly:
  - `Platform::CreateInputFileStream`
  - `Platform::OpenInputFileStream`

- Moves the file default compute hash function to the Platform class (`Platform::CreateFileContentHash`) and switches to the Win32 UTF-16 variant on Windows.

- Adds the `UNICODE` macro before including "Windows.h" on Windows which ensures all functions called are the Unicode variants instead of the default ANSI variants.
  - Implicitly changes functions such as `GetEnvironmentVariable` and `SetEnvironmentVariable` to their Unicode variants.

- Changes the following environment variable related functions to their Win32 Unicode variants on Windows:
  - `environ` -> `_wenviron`
  - `_putenv_s` -> `_wputenv_s`

- Updates tests using `SetEnvironmentVariable` to use wide string literals since that function has been switched to the Unicode variant.

In my testing, this appears to function correctly, loading configs, LUTs, and looks with non-ASCII paths, however I have not tested exhaustively. There may be more file-related functions I haven't found/wrapped.

Happy to take any suggestions or make any changes to the implementation upon request.


src/OpenColorIO/Platform.cpp
```diff
@@ -197,6 +197,42 @@ std::string CreateTempFilename(const std::string & filenameExt)
...
+    stream->open(Utf8ToUtf16(filename).c_str(), mode);
+#else
+    stream->open(filename, mode);
+#endif
+}
+
+#ifdef _WIN32
+std::wstring Utf8ToUtf16(std::string str)
+{
+    int sz = MultiByteToWideChar(CP_UTF8, 0, &str[0], (int)str.size(), NULL, 0);
```
hodoulp, line 221: [Question] As the minimal C++ version for the project is now C++11, could the code use some new C++ feature? 

I found `std::codecvt_utf8_utf16` which provides a platform agnostic implementation. Unfortunately it's deprecated from C++17.

I also check OpenImageIO library which uses the proposed implementation with the same defines everywhere is the code.

hodoulp (changes requested):
Thanks @itsmattkc for this pull request to tackle the challenging file path character encoding difference between Windows and Linux platforms.

Even If I made several comments/suggestions, I think that you are on the right track.

Any changes in the library must now provide a unit test (a new one or updating an existing one) to demonstrate that the change works on all the supported platforms. Even if it could end-up to be a no-op for Linux platform, a unit test must validate it also.

Note: if the method `DefaultComputeHash` can be moved in `platform` then the methods `Utf16ToUtf8()` and `Utf8ToUtf16` could then be completely private i.e. no more define in an header file.

src/OpenColorIO/Platform.cpp
hodoulp: I used [godbolt](https://godbolt.org/) for investigation, and no recent major compilers flag the feature as deprecated!.

hodoulp:
Thanks @itsmattkc for all the efforts to fix the file path on the Windows platform.

I do not understand the rational for the latest commit. Could you explain it?

itsmattkc:
Hey, yes I've made a few more changes the PR overall and am still in the process of finalizing it. I implemented all of your suggestions but then thought it might be good to support Unicode environment variables too (such as the `OCIO` variable) which has led to having to change some other things too (notably the Windows-based environment variable tests).

As for the latest commit, I assume you're referring to enforcing UTF-16LE over UTF-16BE? That's just because Windows specifically uses UTF-16LE as its Unicode encoding. When I implemented a test of the UTF-8/UTF-16 conversion functions (using pre-existing literals as "control variables"), I found it would fail in some environments because some platforms (notably the CentOS 7 CI set up here) defaulted to UTF-16BE which would lead to the strings failing to compare. Since these functions are specifically intended for Windows anyway, it seemed appropriate to just enforce UTF-16LE in the conversion.

itsmattkc:
Okay, sorry for the delay, this PR is now ready to be re-reviewed. I've squashed my commits for neatness and updated the OP to better reflect the changes made.

Simran-B:
- The PR covers `std::ifstream`, but `std::ofstream` is also used in a few places (not in the core code however as far as I see). I suppose output file streams should also be handled?
  - apps (ociobakelut, ociocheck, ...)
  - bindings (Python)
  - tests (CPU, GPU)
  - vendor (Photoshop)
- I wonder if there is any Big Endian Windows where UTF-16BE might be used, such as on ARM, but it appears to be Little Endian even on AArch: https://reverseengineering.stackexchange.com/questions/17922/determining-endianness-of-pe-files-windows-on-arm
- Not sure if it's worth mentioning in the commit message, but Microsoft calls UTF-8 "multi-byte" and UTF-16 "wide-character" in their APIs. Some functions have a suffix "A" for ANSI and "W" for wide-character. This isn't really visible in the PR because the `UNICODE` macro takes care of using the right functions.

itsmattkc:
- Ah yeah I missed `ofstream`, I can add support for that too for completeness.
- Yeah, I believe for consistency, Windows is always UTF-16LE.
- I kind of allude in the commit message to the `UNICODE` macro switching the functions to "Unicode variants", but I can clarify a little further that functions like SetEnvironmentVariable, which by default resolves to the ANSI SetEnvironmentVariableA, now resolves to the Unicode/wide char SetEnvironmentVariableW thanks to the `UNICODE` macro.

vendor/aftereffects/OpenColorIO_AE_Context.cpp
```diff
@@ -153,7 +153,7 @@ bool Path::exists() const
     if(path.empty())
         return false;
     
-    std::ifstream f( path.c_str() );
+    std::ifstream f = Platform::CreateIfstream( path.c_str() );
```
hodoulp, line 156: The method name changed.

src/OpenColorIO/Platform.h
```diff
@@ -15,11 +15,14 @@
 #define WIN32_LEAN_AND_MEAN
 #endif
 
+#define UNICODE
```
hodoulp, line 18: Could you add an explanation of the define?

src/OpenColorIO/Platform.h
```diff
@@ -73,6 +76,21 @@ void AlignedFree(void * memBlock);
 //       the file if created.
 std::string CreateTempFilename(const std::string & filenameExt);
 
+// Platform-independent ifstream constructor based on UTF-8 input
```
hodoulp, line 79: We try now to have method explanation starting with a verb and ending with a '.'

src/OpenColorIO/Platform.cpp
```diff
@@ -50,11 +53,17 @@ bool Getenv(const char * name, std::string & value)
...
+    if(uint32_t size = GetEnvironmentVariable(name_u16.c_str(), nullptr, 0))
     {
-        std::vector<char> buffer(size);
-        GetEnvironmentVariable(name, buffer.data(), size);
-        value = std::string(buffer.data());
+        std::wstring value_u16(size, 0);
+        GetEnvironmentVariable(name_u16.c_str(), &value_u16[0], size);
+
+        // GetEnvironmentVariable always returns a null terminated string which we don't need
+        // since we're using std::strings, so we pop it off here.
```
hodoulp, line 63: The comment does not clearly explain the 'size+1' I think.
"The GetEnvironmentVariable always returns the string size including the null terminated character which ..."

tests/cpu/Platform_tests.cpp
```diff
@@ -160,3 +160,17 @@ OCIO_ADD_TEST(Platform, create_temp_filename)
     // Check that it only generates unique random strings.
     OCIO_CHECK_EQUAL(uids.size(), TestMax);
 }
+
+OCIO_ADD_TEST(Platform, utf8_utf16_convert)
+{
+    std::string utf8_str = "\xe3\x81\x93\xe3\x82\x93\xe3\x81\xab\xe3\x81\xa1\xe3\x81\xaf";
```
hodoulp, line 166: Could you add (as a comment) the letters used to ease the reading?
ex: HIRAGANA LETTER KO UTF-16: x3053, UTF-8: xe3, x81, x93

tests/cpu/Platform_tests.cpp
```diff
@@ -93,13 +93,13 @@ OCIO_ADD_TEST(Platform, setenv)
     }
 #ifdef _WIN32
     {
-        SetEnvironmentVariable("MY_WINDOWS_DUMMY_ENV", "1");
+        SetEnvironmentVariable(L"MY_WINDOWS_DUMMY_ENV", L"1");
         std::string env;
         OCIO_CHECK_ASSERT(OCIO::Platform::Getenv("MY_WINDOWS_DUMMY_ENV", env));
```
hodoulp, line 98: I think that the code should highlight that OpenColorIO only works with UTF-8 strings, and unit tests are examples on how to use a method, class, etc. Here the Windows API call uses the 'L' to mention that the strings are UTF-16 strings, so the Getenv() arguments must be prefix with 'u8' to highlight that they are UTF-8 strings.

hodoulp: To be consistent, the lines above must also be updated to 'u8' strings.

hodoulp (changes requested):
Great work.
I only made comments to improve the readability of the code.

tests/cpu/Platform_tests.cpp
hodoulp: Another comment: Based on the code changes, I think that the two previous unit tests (i.e. `(Platform, envVariable)`and `(Platform, getenv)`) must also add some tests around Windows & UTF-16.

itsmattkc:
Thanks for the review, I believe I have resolved all of your suggestions. Let me know if anything else should be changed.

fnordware:
Windows projects usually define UNICODE in the build process, usually a Visual Studio project. By defining it in a header like we have here, we may force UNICODE on a Windows project that doesn't want it.

As a library, I think we should respond to UNICODE, not set it. If we have something that explicitly needs the Unicode version, just call it directly, i.e. use SetWindowTextW() instead of SetWindowText().

UNICODE might need to be added to the CMake stuff somewhere. For Windows it might be on by default, but there should be a way to switch it off and everything should still work.

hodoulp:
Good catch @fnordware

hodoulp (approved):

fnordware:
I still think defining UNICODE in our own header is a mistake. It should be defined by the build system. It would be like defining __GNUC__ in a header.

itsmattkc:
Hmm you might be right, I can take a look and see if it can be made agnostic without much difficulty

hodoulp:
I googled on that `UNICODE` topic and almost all Windows UNICODE examples define `UNICODE` in the build system (i.e. not in the source code).

itsmattkc:
Yes I think that is the more correct way to do it. I'll work on amending this tomorrow.

itsmattkc:
Apologies for the long wait on this. I ran into some challenges and then shortly after underwent a big international move which has taken up a large chunk of my time. I should have time from now on to tackle this again if there's still interest in this being done.

itsmattkc:
Okay I have updated the PR. The changes made are:

- `UNICODE` and `_UNICODE` macros have been moved to CMake and code has been modified to ensure it can be compiled in both Unicode and ANSI modes. It may even be useful to compile both in the CI, though there isn't much point to an ANSI only build unless you're targeting Windows 9x. This mostly affects Platform.cpp and Platform_tests.cpp.
- SystemMonitor_windows.cpp has been updated to support both Unicode and ANSI.
- Platform.h creates definitions on Windows for types that resolve to the correct type depending on the Unicode declaration (e.g. `std::string` for ANSI and `std::wstring` for Unicode). These are primarily used in Platform_tests.cpp and SystemMonitor_windows.cpp.

Let me know if any further changes should be made.

hodoulp:
Thanks @itsmattkc to continue the work on this pull request. On my side I will try to have a look this week.

hodoulp:
@itsmattkc There is a missing `ifdef` somewhere for `Linux` and `macOS` platforms:
```
2: [ 902/1000] [Platform / create_temp_filename                             ] - PASSED
2: 
2: FAILED: Only supported by the Windows platform..
2: [ 903/1000] [Platform / utf8_utf16_convert                               ] - FAILED
```

itsmattkc:
Looks like I made a mistake in the CMake files, it should compile now

PR #1370, opened 2021-04-13
Adsk Contrib - Fix CDL identity optimization issue

hodoulp:

The pull request fixes a bug related to optimization of CDL Ops.   When an identity `CDL_V1_2` style CDL Op was replaced in the optimizer, it was only clamping negative values.  However, since the full op clamps on both sides (0 and 1), per the ASC specification, the identity replacement should also clamp on both sides. 

This bug only impacts v2 configs since v1 configs use the previous method of applying CDLs.  Also, it does not affect 'no clamp' CDL style in v2, the optimizer replaces those identities in a way that also does not clamp.


michdolan (approved):

hodoulp:
Friendly reminder that the pull request could be merged next Tuesday if no objections.

doug-walker (approved):

PR #1352, opened 2021-03-22
Adsk Contrib - Fix the legacy GPU path

hodoulp:
Signed-off-by: Patrick Hodoul <Patrick.Hodoul@autodesk.com>

The pull request fixes the 'legacy' GPU Path (i.e. the OCIO v1 GPU path) which is still present for backward compatibility only. But we strongly suggest to transition to the new GPU path in order to fully benefit of the CPU & GPU color processing parity.

The new legacy GPU path is enabled by:
- auto gpu = processor->getOptimizedLegacyGPUProcessor(oFlags, edgelen);
- auto shaderDesc = GpuShaderDesc::CreateShaderDesc();
- gpu->extractGpuShaderInfo(shaderDesc);

The new method `Processor::getOptimizedLegacyGPUProcessor()` creates a 'normal' GPU processor except that it only uses at most a single 3D LUT. The former method `GpuShaderDesc::CreateLegacyShaderDesc()` is now deprecated (i.e. it throws if used).

Internally, the new design is to preserve the no-ops (i.e. Allocation, etc.) from optimization when creating the processor (i.e. `Config::getProcessor()`). The processor instance is not different from the former implementation at the caller point of view. But it keeps the needed information to correctly build the 'legacy' GPU path when using `Processor::getOptimizedLegacyGPUProcessor()`. If the latter method is not used then the optimization (even the NONE one) removes the no-ops so that change is then transparent to the normal code path. 

Refer to the issue #1294 for test details.

One way to test the result is to compare the two resulting images:
- `export OCIO=~/dev/OpenColorIO-Configs/aces_1.0.3/config.ocio`
- `./src/apps/ocioconvert/ocioconvert --gpulegacy --gpuinfo -v ~/dev/aces/images/ACES/syntheticChart.01.exr acescg ./legacy_res.exr out_srgb`
- `./src/apps/ocioconvert/ocioconvert --gpu --gpuinfo -v ~/dev/aces/images/ACES/syntheticChart.01.exr acescg ./res.exr out_srgb`

Note:
The pull request adds a new method to the public API. In fact it adds a not virtual method to a not virtual class so it preserves the ABI backward compatibility.

PR #1372, opened 2021-04-17
Adsk Contrib - Add 'deprecated' C++ mechanism & cleanup code

hodoulp:

Based on the known limitations in RB-2.0, a C++ deprecated mechanism must now be in place to highlight methods/classes that should not be used anymore (e.g. `Config::parseColorSpaceFromString()` for now) starting with versions 2.x. At the same time, the Legacy GPU path bug (fixed by #1352) imposes the removal of a 'faulty' method from the public API asap to force a transition to the 'new' Legacy GPU path.

The pull request adds deprecated macro (code from OpenEXR) and removes all the code associated to the 'old' Legacy GPU Path.


include/OpenColorIO/OpenColorIO.h
```diff
@@ -2976,27 +2977,17 @@ class OCIOEXPORT GpuShaderCreator
...
- *    // The three potential scenarios are:
+ *    // The two potential scenarios are:
  *    //
  *    //   1. Instantiate the legacy shader description.  The color processor
  *    //      is baked down to contain at most one 3D LUT and no 1D LUTs.
  *    //
  *    //      This is the v1 behavior and will remain part of OCIO v2
  *    //      for backward compatibility.
  *    //
- *    OCIO::GpuShaderDescRcPtr shaderDesc
```
michdolan, line 2987: Is this the intended change here? Currently this is removing the generic shader desc from docs and leaving legacy and custom only. Sould it be removing legacy and leaving gneric and custom instead?

And with this new implementation, the new legacy render path is just a generic shader desc with special optimization to replicate OCIO v1 behavior correct?

hodoulp: The wrong description was preserved. The 'legacy' shader desc is now gone for v2.1.

> And with this new implementation, the new legacy render path is just a generic shader desc with special optimization to replicate OCIO v1 behavior correct?

Yes.

hodoulp: Fixed

michdolan (approved):

hodoulp:
Friendly reminder that the pull request could now be merged next Monday.

doug-walker (approved):

Issue #1187, opened 2020-11-01
Have OCIO build export a proper OpenColorIOConfig.cmake


lgritz:
The modern CMake way is that packages should produce and distribute a PackageConfig.cmake, eliminating the need for all the other CMake-based packages that depend on OCIO from each, separately, having to maintain some kind of FindOpenColorIO.cmake module that contains all the searching logic.

Yash-R:
Hey, i would like to help you with this issue can you guide me ??

hodoulp:
Hi @Yash-R,
Happy to see a volunteer to help the OpenColorIO development.

The task requires some knowledge of `C++` and `CMake` because it involves developing a CMake utility to find a C++ library in a multi-platform project. Below are some examples:
- https://github.com/OpenImageIO/oiio/blob/master/src/cmake/modules/FindOpenColorIO.cmake
- https://github.com/PixarAnimationStudios/USD/blob/release/cmake/modules/FindOpenColorIO.cmake
 
If you still have interest in the task then start the work and the community will help you.

Yash-R:
yes I'm very much interested to work, Can you please explain me the work briefly please ? @hodoulp

hodoulp:
The CMake utility helps other projects to discover an installed OpenColorIO library (i.e which means where are the header files, where are the binary file(s), and what is the version). In the previous post I pasted some examples. So, OpenColorIO must have a `findOpenColorIO.cmake` file and install it.
Keep in mind that the file detection is, most of the time, platform dependant (i.e OpenColorIO supports Windows, MacOS and Linux).

Before starting anything I suggest you to start reading the documentation, and to check on how to contribute to the project.

stilllman:
> So, OpenColorIO must have a `findOpenColorIO.cmake` file and install it.

The proper solution would be to install a config-file package (`OpenColorIOConfig.cmake`), not a find-module package. Find-modules are useful for finding libraries that are _not_ CMake-friendly, if your goal is to make OpenColorIO easily consumable through CMake then you should export your targets in a config-file. A good start would be to read the documentation here: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html

For the record, OpenColorIO 1.1 was actually already installing a config-file package, but this feature was removed by this commit apparently: https://github.com/AcademySoftwareFoundation/OpenColorIO/commit/2f5fbfc80218d5c31bdc5045df7b68e0aa149b6b

lgritz:
[...]

PR #1397, opened 2021-05-04
Add OpenColorIOConfig.cmake generation

remia:
This address issue #1187, the proposed implementation is based heavily on OpenImageIO.

Current implementation have a very simple Config.cmake.in template, as opposed to OpenImageIO which take cares of dependencies and set variables. According to some quick test I did, it doesn't seems needed here but happy to be proved wrong.

More anecdotally, I noticed that when I used the `find_package(OpenColorIO CONFIG)` directive, while it works, it doesn't print anything like the version found and path you usually see. Any ideas ?


lgritz (approved):
LGTM. After it's merged and I can try using it in OIIO's CI builds, I'll let you know if anything needs revision. But it seems reasonable to me.

There's some extra complexity in OIIO's config because there are dependencies on Imath/OpenEXR whose types get exposed in OIIO's exported APIs. But I don't think OCIO has anything like that where dependencies are directly exposed in your APIs, so you don't need to worry about that.

share/cmake/macros/TargetUtils.cmake
```diff
@@ -0,0 +1,22 @@
...
+###########################################################################
+# Macro to install targets to the appropriate locations.  Use this instead
+# of the install(TARGETS ...) signature. Note that it adds it to the
+# export targets list for when we generate config files.
+#
+# Usage:
+#
+#    install_targets (target1 [target2 ...])
+#
+macro (install_targets)
```
remia, line 16: Reading this again, I wonder if we should drop the COMPONENT here as we don't use any. Also on a related note, the OpenColorIO library target seems to be installed twice, one in the CMakeLists.txt just before the macro, the other here. Doesn't look like it's causing any issues but it maybe could be improved.

We could also drop the macro as it could be confusing, we only use that in 2 places for the public targets, all others targets are installed the usual way.

I can push some changes to address that.

Any thoughts @lgritz ?

lgritz: This is just a convenience macro. If it's not exactly what you need or not used in enough places to aid in simplifying your cmake scripts, just skip it and use `install()` directly. I don't remember/understand exactly what the COMPONENT bits do, I think maybe I wasn't even the person who wrote this macro or added that argument.

remia: Thanks for the feedback, I implemented the change, hopefully this simplify things.

hodoulp:
LGTM based on my curent understanding and the `cmake` documentation.

remia:
> All third-party dependencies are statically linked so the library itself and its targets must be hidden.

Is that always true? I just looked at dependent library on my macOS build and found that it link expat dynamically, I was not expecting that, could there that be the same for you with Half ? 

```otool -L myinstall/lib/libOpenColorIO.2.1.0.dylib 
myinstall/lib/libOpenColorIO.2.1.0.dylib:
  @rpath/libOpenColorIO.2.1.dylib (compatibility version 2.1.0, current version 2.1.0)
  /usr/lib/libexpat.1.dylib (compatibility version 7.0.0, current version 8.0.0)
  /System/Library/Frameworks/Carbon.framework/Versions/A/Carbon (compatibility version 2.0.0, current version 164.0.0)
  /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
  /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 905.6.0)
  /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
  /System/Library/Frameworks/ColorSync.framework/Versions/A/ColorSync (compatibility version 1.0.0, current version 4.7.0)
  /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1775.118.101)
  /System/Library/Frameworks/CoreGraphics.framework/Versions/A/CoreGraphics (compatibility version 64.0.0, current version 1463.4.2)
```

Currently I can't reproduce your issue though, my cmake target file doesn't contains the IlmBase::Half link import. Can you share more details on how you built it?

```
set_target_properties(OpenColorIO::OpenColorIO PROPERTIES
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libOpenColorIO.2.1.0.dylib"
  IMPORTED_SONAME_RELEASE "@rpath/libOpenColorIO.2.1.dylib"
  )
```

hodoulp:
> > All third-party dependencies are statically linked so the library itself and its targets must be hidden.

> Is that always true ? I just looked at dependent library on my macOS build and found that it link expat dynamically, I was not expecting that, could that be the same for you with Half ?

Finally, I have time to double-check, and the `half` unexpected dependency is gone!

hodoulp:
Without being a `cmake` expert, everything seems fine.

hodoulp (approved):

lgritz:
I didn't catch this originally (despite having approved it, sorry), but is there a reason why you made a separate target for the headers than for the library? It's not wrong per se, but it's very unusual, and I don't think I've seen it in any other projects (except those who have an option to do something useful in a header-only way and want to be able to avoid pulling in the library). Usually the main library target also knows about the header files, too, so you only need to mention that one target.

remia:
@lgritz Agreed, I don't believe there is a strong reason to have these two different targets either, and this shows up in the generated config file (although it's probably harmless). I'll look if I can remove it to simplify things.

Issue #1362, opened 2021-04-05
About a Pypi Wheel


KelSolaar:
Hi,

Just kicking out the discussion for this one! With OpenColorIO 2.x.x released, it would be great to have a Pypi wheel available, that would make integration with the Python ecosystem orders-of-magnitude easier. The OpenColorIO-Config-ACES repo could use it right-of-the bat.

@michdolan has started some work on that, so maybe we can continue from there?

Cheers,

Thomas

remia:
Hey @michdolan,

Was curious what was your take on this ?

I started looking into it and it seems to me that using an external builder like [cibuildwheel](https://github.com/joerick/cibuildwheel) would be a good idea to handle the build environment for us, it's used in many project, as listed in the Readme, including[ pybind11's example repository](https://github.com/pybind/cmake_example).

michdolan:
I think that's a great idea. Looks like an awesome tool! I'm definitely on board with not reinventing the wheel (no pun intended) and we use GitHub Actions, so should be doable.

remia:
Testing this workflow, I uploaded some wheel here https://test.pypi.org/project/opencolorio/ and the code can be seen here https://github.com/remia/OpenColorIO/pull/8.

Would be interested in any feedback and/or tests, thanks.

zachlewis:
Thanks @remia  -- works like a charm on CentOS-7.4 (python-3.7.4) / MacOS-11.2 (python-3.8.2) / whatever Google Colab uses (python-3.7). Nice work!

remia:
Thanks for testing @zachlewis. I realised that the version number is incorrect as it's built from the master branch, I will fix that, among other improvements.

remia:
Should I open a PR here to start review / discussion now that the basis workflow is in place ? The upload to https://test.pypi.org/ is obviously relying on GitHub repository secret but I guess it doesn't really matter at first when reviewing the code.

zachlewis:
Please do! I think this will be really well-received. Much appreciated!

remia:
Just wanted to note there was mention about adding a conda-forge package on slack: https://conda-forge.org/#about

It could be done after the PyPI release and would the package recipe would live in it's own repo within the conda-forge github org from what I understand.

KelSolaar:
Yes correct, it is almost automatic and rather low effort to maintain.

PR #1401, opened 2021-05-06
Add Python wheel generation

remia:
This address #1362 , adding Python wheel generation using `cibuildwheel`, demo wheels available here https://test.pypi.org/project/opencolorio/, this surely is not the only way, but it looks promising to me.

Current status:
* Build process work on Github Action for all ubuntu, macos, windows platforms, each job building for a specific platform / python interpreter version / architecture. Results are then uploaded to https://test.pypi.org/.
* OpenColorIO is linked statically to the PyOpenColorIO module, to avoid some issues in the wheel repair process seen below.
* The wheels are currently generated only for Python 3.6+, see below for details.

Questions:
* How do we decide when to build and when to publish those wheels. Should those two be invoked on different times, eg. building more or less regularly and publishing only on new tags to master ? Optional pip workflow that just install and test the wheels on supported platforms.
* How do we test the built wheel easily without needing the build tree and ctest ?

Notes:
* Wheel repair does not work for dynamically linked OpenColorIO. The build process generate PyOpenColorIO.so with RPATH that point to the temporary build folder (used by `pip wheel`), `auditwheel` produce wheels with OpenColorIO lib copied into the wheel but don't update the RPATH on the Python module (this fails for both linux and macos). When using `pip wheel --no-clean` the repair process works better fixing the RPATH and point to a subfolder within the wheel where OpenColorIO.so is copied. It looks like there is some lib duplication as it's also on the root wheel folder, and overall size is 3-4 times bigger than statically linked wheel. Why does `pip wheel` override the Python module RPATH to a temporary folder that's deleted at the end of the build process, preventing `auditwheel` to work properly ? What do we do differently from other projects here ? Do we care about dynamic link here at all ?
* manylinux2010 images doesn't build the documentation, failing could be due to the older `doxygen` version, see comment in wheel_workflow.yml. manylinux2014 images don't support Python 2.
* Issue correctly building pybind11 on Windows (CMake ExternalProject), it does not pick the correct Python interpreter and trying to hint it toward the solution didn't work so far. An alternative solution currently used is to simply install it from the setuptools requirements so it will link to the correct python interpreter automatically.
* (https://test.pypi.org/ only) The version number 2.0.0 was used for the first test by mistake, and it seems even removing the project don't free it up again, now using 2.0.0devN for tests. This was also wrong as the wheel are built on the current master branch, so the latest version is now named 2.1.0dev1.

### DIFF
--- a/src/bindings/python/CMakeLists.txt
+++ b/src/bindings/python/CMakeLists.txt
@@ -108,6 +108,8 @@ set(SOURCES
 	PyViewTransform.cpp
 )
 
+# Reason not to use pybind11_add_module ?
+# https://pybind11.readthedocs.io/en/stable/compiling.html#pybind11-add-module
 add_library(PyOpenColorIO MODULE ${SOURCES})
 
 if(OCIO_BUILD_DOCS)
--- a/.github/workflows/wheel_workflow.yml
+++ b/.github/workflows/wheel_workflow.yml
@@ -1,0 +1,221 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright Contributors to the OpenColorIO Project.
+#
+
+name: Wheel
+
+on: [pull_request]
+
+jobs:
+  # Linux jobs run in Docker containers (manylinux), so the latest OS version
+  # is OK. macOS and Windows jobs need to be locked to specific virtual
+  # environment versions to mitigate issues from OS updates, and will require
+  # maintenance as OS versions are retired.
+  #
+  # Notes:
+  #
+  # Documentation fails to build on manylinux2010, maybe too old
+  # doxygen ? First fail in PyColorSpaceSet for documentation
+  # referencing PyOpenColorIO (eg. __sub__). Consequence is we don't
+  # support python 2.7, alternatively build without docstrings, or
+  # fix the doxygen issue.
+  #
+  # When installing doc environment, Python package are not used as the wheel
+  # is built in it's own isolated environment, hence package are listed again
+  # in pyproject.toml.
+  #
+  # Wheels are currently uploaded to the test package index for testing
+  # https://test.pypi.org/
+  #
+  # Filters on event to decide when to build and upload the wheels are up for
+  # dicussions.
+  #
+
+  # ---------------------------------------------------------------------------
+  # Linux
+  # ---------------------------------------------------------------------------
+
+  linux:
+    name: Build wheels on Linux
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          # -------------------------------------------------------------------
+          # CPython 32 bits
+          # -------------------------------------------------------------------
+          - build: CPython 3.6 32 bits
+            python: cp36-*
+            arch: i686
+          - build: CPython 3.7 32 bits
+            python: cp37-*
+            arch: i686
+          - build: CPython 3.8 32 bits
+            python: cp38-*
+            arch: i686
+          - build: CPython 3.9 32 bits
+            python: cp39-*
+            arch: i686
+          # -------------------------------------------------------------------
+          # CPython 64 bits
+          # -------------------------------------------------------------------
+          - build: CPython 3.6 64 bits
+            python: cp36-*
+            arch: x86_64
+          - build: CPython 3.7 64 bits
+            python: cp37-*
+            arch: x86_64
+          - build: CPython 3.8 64 bits
+            python: cp38-*
+            arch: x86_64
+          - build: CPython 3.9 64 bits
+            python: cp39-*
+            arch: x86_64
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.8'
+
+      - name: Build wheels
+        uses: joerick/cibuildwheel@v1.11.0
+        env:
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
+          CIBW_MANYLINUX_I686_IMAGE: manylinux2014
+          CIBW_BUILD_VERBOSITY: 1
+          CIBW_BUILD: ${{ matrix.python }}
+          CIBW_ARCHS: ${{ matrix.arch }}
+          CIBW_BEFORE_BUILD: share/ci/scripts/linux/yum/install_docs_env.sh
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ./wheelhouse/*.whl
+
+  # ---------------------------------------------------------------------------
+  # macOS
+  # ---------------------------------------------------------------------------
+
+  macos:
+    name: Build wheels on macOS
+    runs-on: macos-10.15
+    strategy:
+      matrix:
+        include:
+          # -------------------------------------------------------------------
+          # CPython 64 bits
+          # -------------------------------------------------------------------
+          - build: CPython 3.6 64 bits
+            python: cp36-*
+            arch: x86_64
+          - build: CPython 3.7 64 bits
+            python: cp37-*
+            arch: x86_64
+          - build: CPython 3.8 64 bits
+            python: cp38-*
+            arch: x86_64
+          - build: CPython 3.9 64 bits
+            python: cp39-*
+            arch: x86_64
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.8'
+
+      - name: Build wheels
+        uses: joerick/cibuildwheel@v1.11.0
+        env:
+          CIBW_BUILD_VERBOSITY: 1
+          CIBW_BUILD: ${{ matrix.python }}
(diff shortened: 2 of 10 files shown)


hodoulp:
> The unit test data files are duplicated in the tests/python folder for two main reasons: setuptools do not support including data files outside the installed package (in this case tests/python) and it make it easier to include only what we need instead of 5Mo+ worth of un-needed file (compressed size).

This could become problematic on the long term. If something change without updating the duplicated files then Python regression tests will not use the right files. All the test files must in the the same directory and ideally used by all the unit tests (CPU, GPU & Python).

src/bindings/python/CMakeLists.txt
```diff
@@ -108,6 +108,8 @@ set(SOURCES
 	PyViewTransform.cpp
 )
 
+# Reason not to use pybind11_add_module ?
+# https://pybind11.readthedocs.io/en/stable/compiling.html#pybind11-add-module
```
hodoulp, line 112: If that's the right way to do it, I will support & suggest to make the change.

tests/python/CPUProcessorTest.py
```diff
@@ -84,7 +84,7 @@ def setUpClass(cls):
 
             cls.float_rgb_list = (
                 [-65504.0] + 
-                list(reversed(log_range)) + 
+                [-x for x in reversed(log_range)] +
```
hodoulp, line 87: Good catch

remia:
> This could become problematic on the long term. If something change without updating the duplicated files then Python regression tests will not use the right files. All the test files must in the the same directory and ideally used by all the unit tests (CPU, GPU & Python).

It seems like we should then have an explicit listing of what test files are used from Python unit tests and only include that to avoid bloat in the Python package size. This would require manually updating the list when needed with the risk or forgetting things, but we could maybe have a design that check for this automatically.

Do you see another way ?

src/bindings/python/CMakeLists.txt
remia: I was mostly curious if this was a conscious choice or not, @michdolan do know more about this ?

I can make some investigations if needed.

remia:
Removed the test files duplication as per @hodoulp suggestion. I uploaded a new release on test PyPI if you want to compare the resulting size, https://test.pypi.org/project/opencolorio/2.1.0.dev3/. Hopefully it's still small enough to not exceed any limitations.

Simplest way to do this was to add a `__init__.py` file in the tests/data folder and effectively make it a python package, hopefully it's not an issue. Alternative looks way more complex and installed site-package file hierarchy is harder to control.

src/bindings/python/CMakeLists.txt
michdolan: I was not aware of `pybind11_add_module`. I agree it seems like a good idea to use it. Thanks for pointing it out @remia

tests/python/CPUProcessorTest.py
michdolan: Yes, thanks for catching this!

michdolan:
> How do we decide when to build and when to publish those wheels. Should those two be invoked on different times, eg. building more or less regularly and publishing only on new tags to master ? Optional pip workflow that just install and test the wheels on supported platforms.

I think we can test that the wheels build as part of CI, and then publish as part of a new CD workflow, which would trigger on tag creation ( https://docs.github.com/en/actions/reference/events-that-trigger-workflows#create ), when tags match release tag formatting.

michdolan:
I think it would be ok to only distribute packages for Python 3, and that including the documentation is important so that the library is self-documenting for developers. Supporting Python 2 would be nice, but technically OCIO v2 is part of VFX ref platform 2021, which is 3.7, and 2022 will be 3.9. We could perhaps ship wheels for Python 2 without documentation if the doxygen issue doesn't have a clear solution.

michdolan:
Awesome work @remia ! Thanks so much for putting this together. This will be super beneficial for the OCIO community. What else is needed with this PR before being ready to merge? Implementing a CD deployment?

remia:
Thanks @michdolan!

> I think we can test that the wheels build as part of CI, and then publish as part of a new CD workflow, which would trigger on tag creation ( https://docs.github.com/en/actions/reference/events-that-trigger-workflows#create ), when tags match release tag formatting.

Yes that sounds good, currently it's set to run on tags push (`if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')`) but I can look into changing this to create, would have to read documentation as I'm not familiar with that one.

Just need to bear in mind that the wheel building on CI will add some times and a number of checks but we can probably give it a try at least. It also has the advantage of testing a number of different python on multiple OSes, maybe more than the existing CI as I had to fix some tests.

> I think it would be ok to only distribute packages for Python 3, and that including the documentation is important so that the library is self-documenting for developers. Supporting Python 2 would be nice, but technically OCIO v2 is part of VFX ref platform 2021, which is 3.7, and 2022 will be 3.9. We could perhaps ship wheels for Python 2 without documentation if the doxygen issue doesn't have a clear solution.

Also worth noting that `cibuildwheel` is dropping support for Python 2.7 and 3.5 in the upcoming major release 2 (https://github.com/pypa/cibuildwheel/releases/tag/v2.0.0a4). At least for now I think it may not be worth the effort ? If someone wants to still support python 2, there is still the possibility of building the library manually.

> Awesome work @remia ! Thanks so much for putting this together. This will be super beneficial for the OCIO community. What else is needed with this PR before being ready to merge? Implementing a CD deployment?

Yeah I think we need to agree on the deployment like discussed in the first point, once that's done we also need to setup the opencolorio PyPI account and add the credentials as GitHub secret for the repository.

I'll also take a look at the `pybind11_add_module` stuff but that might go into another PR.

hodoulp:
@remia & @michdolan, what is the status of the pull request?

remia:
@hodoulp I believe I need to check the rule to trigger the push to PyPI on tag creation like @michdolan, currently it's on tag push no sure what's the difference. I'll have a look early this week, then it should be ready to merge I believe.

remia:
Had a look at the `create` GH event and I don't feel it would be easy to use this because there is no easy way to filter and it will be triggered both on branch and tags creation, meaning all the build jobs will run when they don't always need to. The way we use currently is the one presented in https://github.com/pypa/gh-action-pypi-publish I believe it should work, what do you think @michdolan ?

Can we make sure the PyPI opencolorio project is created and that the GH repository secrets are correctly setup ?

Some follow up changes that should be addressed in a later PR includes:
* Migrate to `cibuildwheel 2.0` when release, should bring support for Python 3.10 and Apple Silicon
* Investigate `pybind11_add_module` usage
* Better handling of versioning, currently manually updating `setup.cfg`

remia:
Upgraded `cibuildwheel` to 1.12.0 which already support Apple Silicon for Py 3.8 and 3.9, note that I build one macOS wheel for each x86_64 and ARM64 architecture and not a single universal one, might be changed if that would be preferred for any reasons.

Uploaded to dev4 version here for testing https://test.pypi.org/project/opencolorio/, I can't myself unfortunately.

michdolan:
@remia I tested this out on Windows and Linux and both worked great. I do think it would be good to figure out how to not install the test suite in the future, or make it optional, but it can be solved later. Awesome work! Let's talk in the TSC meeting about steps for rolling this out to the main PyPI instance.

doug-walker:
Thanks for doing this Remi! 

Minor, but we should update the project description at  https://test.pypi.org/project/opencolorio/ to highlight that the wheel is for OCIO v2 since the text currently implies that it is for v1.

remia:
Good point @doug-walker, I updated the global Readme slightly to reflect the 2.0 release but I think there should be a larger rewrite since 2.0 has been published. Links to reference configurations could probably use a refresh as well as other things.

I can look at the toggle to publish the test suite in a separate PR, or maybe something better would be to play the test suite as done currently in the wheel workflow (ie. after it has been built & installed on the target architecture) as it allows to catch some issues that could be missed by the ci workflow, but simply not ship the data in the wheel.

remia:
Added a version lookup in `setup.py` to avoid maintaining the project version in two different places.

michdolan:
I just ran a test deploy with this code from the `pypi_test` branch and it worked well. We now have opencolorio in PyPI ! Thanks again for your work on this @remia !

Only question I have is that OCIO `v2.1.0dev` became `v2.1.0.dev0` in PyPI. Is that additional `0` expected?

michdolan (approved):

remia:
Thanks for the test @michdolan !

As discussed, the versioning scheme from Python do not accept the `2.1.0dev` syntax and the wheel build process is automatically normalizing that to `2.1.0.dev0`. There is a warning buried into the log saying just that: https://github.com/AcademySoftwareFoundation/OpenColorIO/runs/3422953874?check_suite_focus=true#step:4:6696 (`/tmp/pip-build-env-cwq8pk9a/overlay/lib/python3.6/site-packages/setuptools/dist.py:484: UserWarning: Normalizing '2.1.0dev' to '2.1.0.dev0'`).

This shouldn't affect main release following the X.X.X convention. More information here: https://www.python.org/dev/peps/pep-0440/#public-version-identifiers

Issue #1402, opened 2021-05-07
Implement the ACES 1.3 gamut compression transform


doug-walker:
AMPAS has just released ACES 1.3 and the main feature of this release is a new gamut compression transform.  Here are some key links:
https://github.com/ampas/aces-dev/blob/dev/transforms/ctl/lmt/LMT.Academy.GamutCompress.ctl
https://github.com/ampas/aces-vwg-gamut-mapping-2020/blob/master/model/docs/gamut-compress-documentation.md
https://paper.dropbox.com/doc/1BByr3dSQvpjlYcOJlVgR
https://community.acescentral.com/t/gamut-compression-fka-mapping-documentation-deliverable/3539

It would be great to have someone make a PR to implement this in OCIO. 

Here are some suggestions for how to approach this.

1)  Implement a new style for the FixedFunctionOp.  Based on discussion at the Gamut Mapping Implementation VWG, it sounds like this should expose the parameters.  However, we should be clear that this is not technically the ACES transform since that has a fixed set of parameters.

2)  Implement a new style for the BuiltinTransform that hardcodes the official parameters for the ACES transform.  This one should have something like "ACES 1.3" in the name.

I will recommend that the matrix from AP0 to AP1 before the actual compression and the matrix from AP1 to AP0 afterwards are placed in the BuiltinTransform rather than in the FixedFunction so that the optimizer is able to drop these or combine them with other matrices for better performance.

KelSolaar:
The actual CTL reference code is here: https://github.com/ampas/aces-dev/blob/dev/transforms/ctl/lmt/LMT.Academy.GamutCompress.ctl

zachlewis:
I'd like to give this a shot. If I can do it, pretty much anyone can.

remia:
Great @zachlewis! I was also interested on getting involved, so feel free to ask if you need a hand.

nick-shaw:
Great to see people are keen to jump on this. I’m sure it won’t be necessary, but feel free to give me a should if anything is unclear. And do reference the CTL @KelSolaar linked, rather than Jed’s repo, as his version diverges slightly from the AMPAS one.

KelSolaar:
> And do reference the CTL @KelSolaar linked, rather than Jed’s repo, as his version diverges slightly from the AMPAS one. 

[...]

PR #1404, opened 2021-05-14
Add Gamut Compress implementation

remia:
Gamut Compression algorithm addressing issue #1402.

* I didn't update the name of the `FixedFunction` (`FIXED_FUNCTION_ACES_GAMUTMAP_13`), I'm open to suggestions
* This is a straight port of the reference implementation to C++ / Shader, I didn't look into optimisation
* The current allowed range for parameters is from the DCTL implementation, also open to suggestions (https://github.com/ampas/aces-vwg-gamut-mapping-2020/blob/master/model/GamutCompress.dctl) 
* (Edit) Here is a comparison to `ctlrender`, the accuracy seems to be on the same scale as what you get without the compression applied (https://gist.github.com/remia/6ce318f3d78987ab6e3a85b26086f592)
```
Mismatched elements: 12 / 120 (10%)
Max absolute difference: 7.733702659606934e-06
Max relative difference: 7.584445847896859e-05
```

Thanks @zachlewis for doing an early review of this.


KelSolaar:
Awesome @remia!

> * I didn't update the name of the `FixedFunction` (`FIXED_FUNCTION_ACES_GAMUTMAP_13`), I'm open to suggestions

`FIXED_FUNCTION_ACES_GAMUTCOMPRESS_13` to match _CTL_ better?

nick-shaw:
Great work @remia! A couple of comments.

What was discussed in the VWG was that the Fixed Function should not have ACES in the name. Only the Built In Transform which applies the fixed function with the specific constant parameters is “ACES badged”. 

Also when I wrote the CTL I removed some double-checks for NaN (if lim ==1) because it wouldn’t happen with constant parameters. Those should probably go back for the Fixed Function. Have a look at the DCTL version.

remia:
Thanks @nick-shaw for the catch, I added the checks back, hopefully I didn't miss anything ? Do you have a set of parameters / values to reproduce those issues easily, we could add that to the unit tests ?

I think the `GamutCompress` name is already better yes, I'm not exactly sure what the name should be so I'll wait until an agreement has been reached to update the code. Here are a few names:
```
FIXED_FUNCTION_ACES_GAMUTCOMPRESS_13
FIXED_FUNCTION_AMPAS_GAMUTCOMPRESS_13
FIXED_FUNCTION_SMITH_GAMUTCOMPRESS_13
FIXED_FUNCTION_GAMUTCOMPRESS_RGBSAT_13
```

carolalynn22:
This is great, thanks @remia !!

My vote would be for something like: FIXED_FUNCTION_GAMUTCOMPRESS_SCENERGB.

In the fixed function version, we want to keep ACES naming out of it, including the version. I think adding scene is a good addition for context, as it is meant for scene-referred application. We can bring the naming options up at the next working group meeting for consensus, though.

KevinJW:
I agree, we want to consider what are reasonable bounds for the parameters (this impacts the use of the algorithm in general not just in OCIO), I always get annoyed when there are arbitrary limits imposed when there isn't a good reason.

I also mentioned the need for tests in the implementation group last week, so it would probably be of benefit to share the test cases between the groups too.

Kevin

remia:
Pushed a couple of small changes:
* Precompute the y intersect scale in the CPU Op, this slightly improve performance, see below. Left as is for GPU because I didn't have the profiling setup readily available and thought maybe it could be dynamic in the future (even if FixedFunction do not support DynamicProperty at this time, as far as I'm aware)
* Reduce code duplication in parameter bound check
* Fix an issue in `ocioperf` in the GPU Shader creation benchmark phase

I couldn't attend the last few meetings, has anything been decided concerning the outstanding topics ? It would be great to have standardised testing between the implementations. About the naming, I like the mention of "scene" to differentiate with potential future display state mapping. However I feel like FixedFunction are already an advanced tool (eg. include styles like "RedMod03Fwd",  "Glow03Fwd", "DarkToDim10"), it's maybe not worth hiding too much details and having at least the ACES version in the name would be helpful.

So far GPU tests were run on macOS / GL and CentOS / GL with success, hopefully I'll be able to test on Windows, but probably not on DirectX.


```
# Without Gamut Compress
Process the complete image (in place):          For 50 iterations, it took: [104.899, 101.779, 101.842] ms
# Initial PR
Process the complete image (in place):          For 50 iterations, it took: [222.699, 217.072, 217.184] ms
# New
Process the complete image (in place):          For 50 iterations, it took: [162.016, 155.873, 155.996] ms
```

doug-walker:
Thanks for the updates Remi!  Carol and Matthias have an action-item on the overall gamut compress naming and then we can decide how to map that into a FixedFunction enum name.  I agree with you that it should indicate the ACES release, like the other ACES related enums.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +686,147 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+    }
+
+    return comprDist;
+}
+
+__inline float Renderer_ACES_GamutMap13_Fwd::gamut_map(float val, float ach, float lim, float thr, float scale, bool invert) const
+{
+    // Distance from the achromatic axis, aka inverse RGB ratios
+    float dist;
+    if (ach == 0.0f)
```
hodoulp, line 763: The code formatting is to always have `{` and `}` even for a single line statement.

hodoulp, line 763: Is that strict equality or an equality within a tolerance?

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +686,147 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+        else
+        {
+            // Avoid singularity
+            if (dist > (thr + scl))
+            {
+                comprDist = dist;
+            }
+            else
+            {
+                comprDist = thr + scl * powf(-(p / (p - 1.0f)), 1.0f / m_power);
```
hodoulp, line 751: As `m_power` seems constant when processing an image, it means that `1.0f / m_power` can be precomputed upfront.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
```diff
@@ -341,7 +362,51 @@ FixedFunctionOpDataRcPtr FixedFunctionOpData::clone() const
...
+        const double thr_low_bound = 0.4;
+        const double thr_hi_bound = 1.0;
+        check_bounds("thr_cyan", thr_cyan, thr_low_bound, thr_hi_bound);
+        check_bounds("thr_magenta", thr_magenta, thr_low_bound, thr_hi_bound);
+        check_bounds("thr_yellow", thr_yellow, thr_low_bound, thr_hi_bound);
+
+        const double pwr_low_bound = 1.0;
+        const double pwr_hi_bound = 3.0;
+        check_bounds("power", power, pwr_low_bound, pwr_hi_bound);
+    }
```
hodoulp, line 408: The code could align all the method's parameters, and it can declare `static constexpr` all the hard-coded values.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
```diff
@@ -341,7 +362,51 @@ FixedFunctionOpDataRcPtr FixedFunctionOpData::clone() const
...
+
+        const double lim_cyan = m_params[0];
+        const double lim_magenta = m_params[1];
+        const double lim_yellow = m_params[2];
+        const double thr_cyan = m_params[3];
+        const double thr_magenta = m_params[4];
+        const double thr_yellow = m_params[5];
+        const double power = m_params[6];
+
+        auto check_bounds = [](const std::string & name, double val, double low, double high) {
```
hodoulp, line 384: The coding style puts curly brackets on a line.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -192,6 +192,179 @@ void Add_Glow_03_Inv_Shader(GpuShaderText & ss, float glowGain, float glowMid)
...
+                              float thr,
+                              float power,
+                              float invert)
+{
+    ss.newLine() << "{";
+    ss.indent();
+
+    ss.newLine() << "float lim = " << lim << ";";
+    ss.newLine() << "float thr = " << thr << ";";
+    ss.newLine() << "float pwr = " << power << ";";
```
hodoulp, line 208: Are these three local variables needed?  The method argument can be used directly in the code.

hodoulp (changes requested):
Without validating the algorithm itself, the integration is almost done except some code formatting and other minor improvements.
The code should update the unit test `OCIO_ADD_TEST(Config, fixed_function_serialization)` to test the serialization/deserialization of this new fix function style.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -103,6 +103,39 @@ class Renderer_ACES_DarkToDim10_Fwd : public OpCPU
     float m_gamma;
 };
 
+class Renderer_ACES_GamutMap13_Fwd : public OpCPU
+{
+public:
+    Renderer_ACES_GamutMap13_Fwd(ConstFixedFunctionOpDataRcPtr & data);
```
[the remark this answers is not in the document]
remia: I probably c/c from `Renderer_ACES_Glow03_Fwd` so I applied the same recommendation there. Is there a benefit of deleting the default constructor by the way, shouldn't it be already not emitted by the compiler because we define a user constructor ?

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +686,147 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+Renderer_ACES_GamutMap13_Fwd::Renderer_ACES_GamutMap13_Fwd(ConstFixedFunctionOpDataRcPtr & data)
+    :   OpCPU()
+{
+    m_limCyan = (float)data->getParams()[0];
+    m_limMagenta = (float)data->getParams()[1];
+    m_limYellow = (float)data->getParams()[2];
+    m_thrCyan = (float)data->getParams()[3];
+    m_thrMagenta = (float)data->getParams()[4];
+    m_thrYellow = (float)data->getParams()[5];
+    m_power = (float)data->getParams()[6];
```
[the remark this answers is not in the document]
remia: Fixed.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +686,147 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+
+    m_thrCyan = std::min(0.9999f, m_thrCyan);
+    m_thrMagenta = std::min(0.9999f, m_thrMagenta);
+    m_thrYellow = std::min(0.9999f, m_thrYellow);
```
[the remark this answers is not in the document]
remia: Fixed.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +686,147 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+
+    // Precompute scale factor for y = 1 intersect
+    auto f_scale = [this](float lim, float thr) {
+        return (lim - thr) / powf(powf((1.0f - thr) / (lim - thr), -m_power) - 1.0f, 1.0f / m_power);
+    };
+    m_scaleCyan = f_scale(m_limCyan, m_thrCyan);
+    m_scaleMagenta = f_scale(m_limMagenta, m_thrMagenta);
+    m_scaleYellow = f_scale(m_limYellow, m_thrYellow);
```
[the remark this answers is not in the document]
remia: Fixed.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +686,147 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+    else
+    {
+        // Disable compression, avoid nan
+        if (lim < 1.0001f)
+        {
+            return dist;
+        }
+
+        // Scale factor for y = 1 intersect
+        const float scl = scale;
```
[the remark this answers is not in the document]
remia: Fixed.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +686,147 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+        const float red = in[0];
+        const float grn = in[1];
+        const float blu = in[2];
+
+        // Achromatic axis
+        const float ach = std::max(red, std::max(grn, blu));
+
+        out[0] = gamut_map(red, ach, m_limCyan, m_thrCyan, m_scaleCyan, false);
+        out[1] = gamut_map(grn, ach, m_limMagenta, m_thrMagenta, m_scaleMagenta, false);
+        out[2] = gamut_map(blu, ach, m_limYellow, m_thrYellow, m_scaleYellow, false);
```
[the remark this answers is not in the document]
remia: Fixed.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
remia: Fixed.

remia: I believe strict to avoid divide by zero but I just followed the reference implementation here: https://github.com/ampas/aces-dev/blob/dev/transforms/ctl/lmt/LMT.Academy.GamutCompress.ctl#L113
Maybe @nick-shaw can confirm.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
remia: Yes, I did that at first but didn't saw any difference in benchmark so I removed it. Added back as this raises remark.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
remia: Fixed.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
remia: Fixed.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
remia: I used that to ease the writing of the function but I agree it make no sense for the code generation.
Fixed.

remia:
Thanks for the review @hodoulp really appreciated.

I think I addressed your points in the latest commit.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
nick-shaw: The `ach == 0.0` vs `fabs(ach) < TINY` is something that has been brought up, and we will probably change.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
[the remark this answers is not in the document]
hodoulp: In that case, you are right the default empty constructor is not generated but the default copy constructor & assignation operator are. So, the code forbids anything not needed by default and that also provides some code consistency i.e. if one of these four is needed then define the four (thanks to C++11 and higher `default` keyword can now be used).  That's an home-made extension of the [rule of three](https://en.cppreference.com/w/cpp/language/rule_of_three). The other side-effect is to disable the move constructor and assignation operator.

hodoulp:
@remia , As discussed, I planned to test the DX11 implementation. But there is a bug related to `FixedTransform` (not related to your work) forbidding me to test. Stay tuned...

src/OpenColorIO/transforms/builtins/ACES.cpp
```diff
@@ -708,6 +708,24 @@ void RegisterAll(BuiltinTransformRegistryImpl & registry) noexcept
...
+
+            CreateMatrixOp(ops, matrix, TRANSFORM_DIR_FORWARD);
+
+            CreateFixedFunctionOp(ops, FixedFunctionOpData::ACES_GAMUTMAP_13_FWD, {
+                1.147, 1.264, 1.312, 0.815, 0.803, 0.880, 1.2});
+
+            CreateMatrixOp(ops, matrix, TRANSFORM_DIR_INVERSE);
+        };
+
+        registry.addBuiltin("ACES-LMT - GAMUT_MAP_13",
```
doug-walker, line 725: Based on the new naming, propose: 
"ACES-LMT - ACES 1.3 Reference Gamut Compression",
"LMT (applied in ACES2065-1) to compress scene-referred values from common cameras into the AP1 gamut"

tests/cpu/fileformats/FileFormatCTF_tests.cpp
```diff
@@ -6229,6 +6229,30 @@ OCIO_ADD_TEST(CTFTransform, gamma6_ctf)
     OCIO_CHECK_EQUAL(expected, outputTransform.str());
 }
 
+OCIO_ADD_TEST(CTFTransform, fixed_function_aces_gamutmap_13)
```
doug-walker, line 6232: Would be nice to check the inverse too (see the Rec2100 surround test, although could also be done within the same test, if you prefer).

tests/cpu/ops/fixedfunction/FixedFunctionOpCPU_tests.cpp
```diff
@@ -265,6 +265,138 @@ OCIO_ADD_TEST(FixedFunctionOpCPU, aces_dark_to_dim_10)
                            1e-7f,
                            __LINE__);
     }
+}
+
+OCIO_ADD_TEST(FixedFunctionOpCPU, aces_gamut_map_13)
```
doug-walker, line 270: This is awesome!!!

Is there a URL you could add as a comment regarding the source of the data?

tests/cpu/ops/fixedfunction/FixedFunctionOpData_tests.cpp
```diff
@@ -66,6 +66,104 @@ OCIO_ADD_TEST(FixedFunctionOpData, aces_dark_to_dim10_style)
...
+    OCIO_CHECK_ASSERT(func.getParams() == params);
+
+    OCIO::FixedFunctionOpDataRcPtr inv = func.inverse();
+    OCIO_CHECK_EQUAL(inv->getParams()[0], func.getParams()[0]);
+    OCIO_CHECK_EQUAL(inv->getStyle(), OCIO::FixedFunctionOpData::ACES_GAMUTMAP_13_INV);
+    std::string cacheIDUpdated;
+    OCIO_CHECK_NO_THROW(cacheIDUpdated = inv->getCacheID());
+    OCIO_CHECK_ASSERT(cacheID != cacheIDUpdated);
+
+    OCIO_CHECK_ASSERT(func == func);
```
doug-walker, line 85: Intended?

tests/cpu/ops/fixedfunction/FixedFunctionOpData_tests.cpp
```diff
@@ -155,4 +253,15 @@ OCIO_ADD_TEST(FixedFunctionOpData, is_inverse)
...
+    OCIO::FixedFunctionOpData::Params p7 = { 1.147, 1.264, 1.312, 0.815, 0.803, 0.880, 1.2 };
+    auto f_gm = std::make_shared<const OCIO::FixedFunctionOpData>(
+        OCIO::FixedFunctionOpData::ACES_GAMUTMAP_13_FWD, p7);
+    auto f_gm_inv = std::make_shared<const OCIO::FixedFunctionOpData>(
+        OCIO::FixedFunctionOpData::ACES_GAMUTMAP_13_INV, p7);
+    OCIO_CHECK_ASSERT(f_gm->isInverse(f_gm_inv));
+    OCIO_CHECK_ASSERT(f_gm_inv->isInverse(f_gm));
+    OCIO_CHECK_ASSERT(!f_gm->isInverse(f_gm));
+    OCIO_CHECK_ASSERT(!f_gm_inv->isInverse(f_gm_inv));
+    OCIO_CHECK_ASSERT(!f_gm->isInverse(f_r));
```
doug-walker, line 266: Would be nice to change f_gm_inv pwr to 1.21 and verify that it is no longer an inverse of f_gm.

tests/cpu/transforms/BuiltinTransform_tests.cpp
```diff
@@ -392,6 +392,8 @@ AllValues UnitTestValues
         { { 0.125f, 0.1f, 0.075f }, { 0.211320835792f, 0.149169650771f, 0.085452970479f } } },
     { "ACES-LMT - BLUE_LIGHT_ARTIFACT_FIX",
         { { 0.5f, 0.4f, 0.3f }, { 0.48625676579f,  0.38454173877f,  0.30002108779f } } },
+    { "ACES-LMT - GAMUT_MAP_13",
+        { { 0.5f, 0.4f, 0.3f }, { 0.50000000000f,  0.40000000596f,  0.30000001192f } } },
```
doug-walker, line 396: The {0.5,0.4,0.3} is not a good test value.  The goal of these tests is just a quick test that it's working, but an identity op would pass this test.  Suggest using a value such as { 0.5, 0.4, -0.3 }, or some other color that would actually get compressed.

tests/gpu/FixedFunctionOp_test.cpp
```diff
@@ -173,6 +173,136 @@ OCIO_ADD_GPU_TEST(FixedFunction, style_aces_glow10_inv)
...
+{
+    const double data[7] = { 1.147, 1.264, 1.312, 0.815, 0.803, 0.880, 1.2 };
+    OCIO::FixedFunctionTransformRcPtr func =
+        OCIO::FixedFunctionTransform::Create(OCIO::FIXED_FUNCTION_ACES_GAMUTMAP_13, &data[0], 7);
+    func->setDirection(OCIO::TRANSFORM_DIR_FORWARD);
+
+    test.setProcessor(func);
+
+    OCIOGPUTest::CustomValues values;
+    values.m_inputValues =
```
doug-walker, line 186: Again, this is awesome but would be nice to put a comment about the source of the data.

tests/python/FixedFunctionTransformTest.py
```diff
@@ -60,8 +60,7 @@ def test_style(self):
 
         for style in OCIO.FixedFunctionStyle.__members__.values():
             if style not in [OCIO.FIXED_FUNCTION_ACES_GAMUTMAP_02,
-                             OCIO.FIXED_FUNCTION_ACES_GAMUTMAP_07,
-                             OCIO.FIXED_FUNCTION_ACES_GAMUTMAP_13]:
+                             OCIO.FIXED_FUNCTION_ACES_GAMUTMAP_07]:
                 self.tr.setStyle(style)
                 self.assertEqual(self.tr.getStyle(), style)
```
doug-walker, line 66: Would be nice to do a bit more Python testing, for example that it's possible to get and set the 7 parameters and that the validation works.

doug-walker (changes requested):
This is really excellent work Remi, thank you so much!

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
remia: Thanks @doug-walker, well made point.

I wonder if this paper from Gary will be available in Dropbox later on ?

nick-shaw: Gary's comment about the comparison to zero was in email feedback to the gamut compression group leadership, rather than in a paper. He does make mention of avoiding division by zero on page 13 of [this](https://www.dropbox.com/s/y35glx6vsd4xhei/Core_rendering_algorithms_v10_final.pdf) related paper.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +694,151 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+        // Compress
+        if (!invert)
+        {
+            comprDist = thr + scale * nd / (powf(1.0f + p, m_one_over_power));
+        }
+        // Uncompress
+        else
+        {
+            // Avoid singularity
+            if (dist > (thr + scale))
```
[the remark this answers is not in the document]
remia: Updated, should this be addressed in the CTL implementation ?

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +694,151 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+
+    // No compression below threshold
+    if (dist < thr)
+    {
+        comprDist = dist;
+    }
+    else
+    {
+        // Disable compression, avoid nan
+        if (lim < 1.0001f)
```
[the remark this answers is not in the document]
remia: I moved it a level above in the call stack but it's still in the loop. I didn't found a simple design to move it out the loop completely yet.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +694,151 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+        {
+            return dist;
+        }
+
+        // Normalize distance outside threshold by scale factor
+        const float nd = (dist - thr) / scale;
+        const float p = powf(nd, m_power);
+
+        // Compress
+        if (!invert)
```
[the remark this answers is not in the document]
remia: Split into `compress` and `uncompress`.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +694,151 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+    return comprDist;
+}
+
+__inline float Renderer_ACES_GamutMap13_Fwd::gamut_map(float val, float ach, float lim, float thr, float scale, bool invert) const
+{
+    // Distance from the achromatic axis, aka inverse RGB ratios
+    float dist;
+    if (ach == 0.0f)
+    {
+        dist = 0.0f;
```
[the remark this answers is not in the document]
remia: Added an early exit for this case.

tests/cpu/ops/fixedfunction/FixedFunctionOpData_tests.cpp
remia: Not sure about it, it's from `rec2100_surround_style` test below, probably make not much sense to test that for two different style.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
```diff
@@ -341,7 +362,52 @@ FixedFunctionOpDataRcPtr FixedFunctionOpData::clone() const
...
+        {
+            if (val < low || val > high)
+            {
+                std::stringstream ss;
+                ss << "Parameter " << val << " (" << name << ") is outside valid range [" << low << "," << high << "]";
+                throw Exception(ss.str().c_str());
+            }
+        };
+
+        static constexpr double lim_low_bound = 1.0;
```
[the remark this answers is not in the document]
remia: Added a comment as suggested.

The limits come in part from the DCTL implementation from the VWG. I agree these have no real reason to be there, should we extend them to arbitrary high values or just remove them ?

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
```diff
@@ -26,6 +26,8 @@ constexpr char ACES_GLOW_10_FWD_STR[]      = "Glow10Fwd";
 constexpr char ACES_GLOW_10_REV_STR[]      = "Glow10Rev";
 constexpr char ACES_DARK_TO_DIM_10_STR[]   = "DarkToDim10";
 constexpr char ACES_DIM_TO_DARK_10_STR[]   = "DimToDark10";
+constexpr char ACES_GAMUTMAP_13_FWD_STR[]  = "GamutMap13Fwd";
```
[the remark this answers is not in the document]
remia: Thanks for the update, I renamed everything accordingly.

include/OpenColorIO/OpenColorTypes.h
```diff
@@ -477,7 +477,7 @@ enum FixedFunctionStyle
     FIXED_FUNCTION_XYZ_TO_LUV,          ///< CIE XYZ to 1976 CIELUV colour space (D65 white)
     FIXED_FUNCTION_ACES_GAMUTMAP_02,    ///< ACES 0.2 Gamut clamping algorithm -- NOT IMPLEMENTED YET
     FIXED_FUNCTION_ACES_GAMUTMAP_07,    ///< ACES 0.7 Gamut clamping algorithm -- NOT IMPLEMENTED YET
-    FIXED_FUNCTION_ACES_GAMUTMAP_13     ///< ACES 1.3 Gamut mapping algorithm -- NOT IMPLEMENTED YET
+    FIXED_FUNCTION_ACES_GAMUTMAP_13     ///< ACES 1.3 Gamut mapping algorithm
```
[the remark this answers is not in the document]
remia: Make sense to me, updated that as you suggested.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -192,6 +192,177 @@ void Add_Glow_03_Inv_Shader(GpuShaderText & ss, float glowGain, float glowMid)
     ss.newLine() << "outColor.rgb = outColor.rgb * glowGainOut + outColor.rgb;";
 }
 
+void Add_GamutMap_13_Compress(GpuShaderText & ss,
```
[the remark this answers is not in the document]
remia: Didn't update the GPU renderer yet, wanted to have an agreement on the CPU side first.

Side note is that we may want to have a version supporting dynamic properties later on so we might still want to have more flexibility in the GPU path ? But if that comes at the price of performance I agree we should probably not consider it.

tests/cpu/fileformats/FileFormatCTF_tests.cpp
remia: Added a separate test, thanks for the suggestion.

tests/cpu/ops/fixedfunction/FixedFunctionOpData_tests.cpp
remia: Added as suggested.

tests/cpu/transforms/BuiltinTransform_tests.cpp
remia: Yes agreed, updated.

tests/python/FixedFunctionTransformTest.py
remia: Added a few test cases as suggested.

tests/gpu/FixedFunctionOp_test.cpp
remia: Added a mention pointing to `FixedFunctionOpCPU_tests.cpp`.

remia:
> This is really excellent work Remi, thank you so much!

Thanks for the thorough review @doug-walker ! I addressed some of the points, I'm waiting for feedback on the CPU renderer to then update the GPU path as mentioned.

tests/cpu/ops/fixedfunction/FixedFunctionOpData_tests.cpp
doug-walker: It's fine to leave it.  I suppose it's worth verifying that the op== method handles the extra parameters correctly.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
doug-walker: Perhaps for the benefit of future readers a comment would be helpful, such as:
// Note: Strict equality is fine here.  For example, consider the RGB { 1e-7, 0, -1e-5 }. This will become a dist = (1e-7 - -1e-5) / 1e-7 = 101.0. So, there will definitely be very large dist values. But the compression function is able to handle those since they approach the asymptote. So 101 will become something like 1.12. Then at the other end the B values is reconstructed as 1e-7 - 1.12 * 1e-7 = -1.2e-8. So it went from -1e-5 to -1.2e-8, but it caused no numerical instability.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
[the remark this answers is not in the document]
doug-walker: Yes, it should be in the CTL too.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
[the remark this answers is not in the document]
doug-walker: I guess the question is whether we want to allow lim == 1. (that is, no compression) or whether we simply want to clamp it to 1.0001, which would allow for slightly better performance by moving it out of the pixel loop.  Has there been any guidance from the implementation working group on that topic?  If not, I'm fine with whichever way you prefer.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +690,145 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+    const float dist = (ach - val) / std::fabs(ach);
+
+    // No compression below threshold
+    if (dist < thr)
+    {
+        return val;
+    }
+
+    // Compress / Uncompress distance with parameterized shaper function
+    const float comprDist = f(dist, thr, scale, power, one_over_power);
```
doug-walker, line 745: How will you do a function pointer on the GPU?  (There is no requirement that the CPU and GPU code are exactly the same of course, but it does help the readability and maintenance slightly.)  Also, does this prevent declaring compress an inline function?  Given that it's in the pixel loop, it might be best to avoid an actual function call.

Perhaps it would be clearer to split your gamut_comp function into separate forward and inverse versions and this would avoid the need for the template and the function pointer?  And then the GPU could look the same as well.  Just a suggestion, I'm fine if you prefer to leave it as is.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +690,145 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
...
+{
+    m_limCyan        = (float)data->getParams()[0];
+    m_limMagenta     = (float)data->getParams()[1];
+    m_limYellow      = (float)data->getParams()[2];
+    m_thrCyan        = (float)data->getParams()[3];
+    m_thrMagenta     = (float)data->getParams()[4];
+    m_thrYellow      = (float)data->getParams()[5];
+    m_power          = (float)data->getParams()[6];
+
+    m_one_over_power = 1.f / m_power;
```
doug-walker, line 764: Minor: This seems like a lot of work to avoid a single divide, given all the other math that is happening.  I know Patrick commented on it, but given that you already tested it didn't help performance to precompute it, perhaps it's cleaner to just do the divide?  (Again, feel free to ignore this comment if you prefer to leave it as is.)

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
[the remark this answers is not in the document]
doug-walker: If the values are recommended by the Implementation VWG, then we should probably follow their lead.

The comment about lim being clamped to 1.0001 is not currently accurate since the code allows lim==1., but if we clamp it in the constructor, based on my other question, then obviously the comment should stay.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -192,6 +192,177 @@ void Add_Glow_03_Inv_Shader(GpuShaderText & ss, float glowGain, float glowMid)
...
+ 
+    // No compression below threshold.
+    ss.newLine() << "if (" << dist << " < " << thr << ")";
+    ss.newLine() << "{";
+    ss.indent();
+    ss.newLine() << comprDist << " = " << dist << ";";
+    ss.dedent();
+    ss.newLine() << "}"; // if (dist < thr)
+    
+    // Compress.
```
doug-walker, line 216: Minor: Regarding my other comment about using a period at the end of sentences, I only meant that this is just the general convention of the code, but please don't feel like you have to do that even for single words.  Just use your best judgement about what looks good to you.  :)

doug-walker (changes requested):
Thanks for the updates Remi!  Added a few more comments/questions.

hodoulp:
Hi @remia , I successfully compiled & run the pull request on macOS and Windows (VS 2019).

Then I did a test with `ociodisplay` to have the GPU code, and I noticed several improvements. Below is part of the  GPU fragment shader.

The first lines could reduce the number of ifs
```
  {
    float ach = max( outColor.x, max( outColor.y, outColor.z ) );
    vec3 dist;
    dist.x = ach == 0.0 ? 0.0 : (ach-outColor.x)/abs(ach);
    dist.y = ach == 0.0 ? 0.0 : (ach-outColor.y)/abs(ach);
    dist.z = ach == 0.0 ? 0.0 : (ach-outColor.z)/abs(ach);
```
Even if much more changes could be done to improve the code, the code below at least removes two ifs:
```
    vec3 dist = vec3(0.0, 0.0, 0.0);
    if (ach != 0.0)
    {
       float ach_tmp = 1.0 / abs(ach);
       dist = vec3((ach-outColor.x) * ach_tmp, (ach-outColor.y) * ach_tmp, (ach-outColor.z) * ach_tmp);
    }
```

```
    vec3 cdist;
    {
      if (dist.x < 0.814999998)
      {
        cdist.x = dist.x;
      }
      else
      {
```
The line below is useless.
```
        if (1.14699996 < 1.0001f)
        {
          cdist.x = dist.x;
        }
        else
        {
          float scl;
          float nd;
          float p;
```
The line below introduces an important performance hit to compute a constant.
Note: I checked the CPU implementation and the line is precomputed so it must be the same in GPU.
```
          scl = (1.14699996 - 0.814999998) / pow(pow((1.0 - 0.814999998) / (1.14699996 - 0.814999998), - 1.20000005) - 1.0, 0.833333313);
          nd = (dist.x - 0.814999998) / scl;
          p = pow(nd, 1.20000005);
```
The line below is useless.
```
          if (1. == 0.0f)
          {
            cdist.x = 0.814999998 + scl * nd / (pow(1.0 + p, 0.833333313));
          }
          else
          {
            if (dist.x > (0.814999998 + scl))
            {
              cdist.x = dist.x;
            }
            else
            {
              cdist.x = 0.814999998 + scl * pow(-(p / (p - 1.0)), 0.833333313);
            }
          }
        }
      }
    }
[...]
```

Global comments for the GPU code path:
- Do not use `outColor` directly, use `shaderCreator->getPixelName()`
- Do not use `float` directly, use `st.float3Keyword()` or a more appropriate method.


@remia As the time is flying (i.e. 2.1 is planned for end of August), I would suggest to only change the `(pow(pow(..` and keep the other comments for a followup pull request if you think that the pull request is ready to merge.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -653,6 +690,145 @@ void Renderer_ACES_DarkToDim10_Fwd::apply(const void * inImg, void * outImg, lon
     }
 }
 
+float compress(float dist, float thr, float scale, float power, float one_over_power)
+{
+    // Normalize distance outside threshold by scale factor
+    const float nd = (dist - thr) / scale;
+    const float p = powf(nd, power);
```
hodoulp, line 697: There are sometime slight differences between the C and C++ version of these methods (powf in this case) so the code code must now only use the std version.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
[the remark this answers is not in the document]
remia: Yes exactly that was my concern about adding a hard clip at 1.0001 in the initialisation phase, tagging @nick for visibility.

Given the time constraint I propose leaving this topic aside and having a follow up performance improvements PR as @hodoulp mentioned. I believe this kind of optimisation is not obviously linked to performance improvements across the board given how good some processors are at branch prediction, this needs through profiling.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
remia: Thanks @doug-walker, it make sense to save future reader some headache indeed. Adding this and resolving the comment.

tests/cpu/ops/fixedfunction/FixedFunctionOpCPU_tests.cpp
remia: Sure thing, will do now.

remia: Sure thing, will do now.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
remia: Yes agreed, it feels a bit like premature optimisation and it could very well be that most compilers are able to fold this into a constant expression. Will revert for the time being.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
remia: That was a bit extreme, point noted :)

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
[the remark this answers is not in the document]
remia: Would it make sense to use arbitrary larger bounds for the time being ? I don't believe the current bounds from the DCTL implementation are anywhere near the official recommended settings to use. Could have something like `lim_hi_bound <= 100`, `thr_low_bound >= 0.1`and `pwr_hi_bound <= 10` although the last one might be too much.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
remia: Looks like the whole OCIO code base is using the C version at the moment, it would probably make little sense to just update the code in this PR and not the rest, should we address in a follow up PR ?

hodoulp: You pull request is already old enough to avoid adding more changes so I would suggest to change the new code only. The goal is only to avoid 'increasing the problem' for short-term.

remia: Sure, will do that now then.

hodoulp:
@remia As discussed together, some GPU Ops are perhaps not following all the rules to have the 'right' GPU code. But the GPU code for all Ops will soon greatly change following all my investigations around the OSL support. So, that's enough for short-term to only fix the new GPU code (i.e. code related to the Gamut Mapping), and let me 'adjust' the other Ops later.

Following my review, only the `pow(pow` is really something to fix for short-term (if that's not too complex) because it introduces a major performance hit for nothing i.e. it's a constant variable as the value can be computed when generating the GPU code.

@michdolan, @doug-walker, I think that we should have a merge as soon as possible so we can announce the improvement at SIGGRAPH (before on the forum) to let enough time to the community for testing before the 2.1 release. And this new Op imposes other work(s) related to the config and CLF file format version managements that must be done before the release. Time is getting short.

remia:
> @remia As discussed together, some GPU Ops are perhaps not following all the rules to have the 'right' GPU code. But the GPU code for all Ops will soon greatly change following all my investigations around the OSL support. So, that's enough for short-term to only fix the new GPU code (i.e. code related to the Gamut Mapping), and let me 'adjust' the other Ops later.
> 
> Following my review, only the `pow(pow` is really something to fix for short-term (if that's not too complex) because it introduces a major performance hit for nothing i.e. it's a constant variable as the value can be computed when generating the GPU code.
> 
> @michdolan, @doug-walker, I think that we should have a merge as soon as possible so we can announce the improvement at SIGGRAPH (before on the forum) to let enough time to the community for testing before the 2.1 release. And this new Op imposes other work(s) related to the config and CLF file format version managements that must be done before the release. Time is getting short.

I addressed the most urgent concerns and cleaned the GPU code output, there is still some issues like `floatKeyword()` being private I believe and `getPixelName()` usage that might clutter the code but I guess this is less urgent stuff. Hopefully the GPU shader generation should be more straightforward now. I'll address any remaining concerns you might have tomorrow.

Would be really interested about hearing more about this OSL integration, is there some place to check for discussions about this ?

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
remia: The GPU code path doesn't currently rely on this templated function argument but should be fairly close, I guess I followed the split approach you mentioned at a slight cost in code duplication.

But the rest probably belongs to a follow optimisation PR, again, I could be wrong but I would avoid making assumption without profiling, compiler should be pretty good about inlining and I don't believe the `__inline` used elsewhere in the code actually provide any guarantee the compiler will do such a thing. But only profiling will tell :)

doug-walker (approved):
Thanks for the additional work Remi, it's looking good!  The only remaining issue from my point of view is the allowed parameter bounds, particularly for lim.  My preference, as expressed below, would be to limit a bit more tightly than you currently are.  But I'm approving now to indicate that I'm ok with wherever the group comes out on this topic.

KelSolaar:
Nice!

remia:
> Thanks for the additional work Remi, it's looking good! The only remaining issue from my point of view is the allowed parameter bounds, particularly for lim. My preference, as expressed below, would be to limit a bit more tightly than you currently are. But I'm approving now to indicate that I'm ok with wherever the group comes out on this topic.

Your points make sense to me, I can push an update so that we clamp the limit to `1.0001` as discussed. I 'm considering updating directly the validate method and not authorise users to specify `threshold > 0.9999` and `lim < 1.0001`, I feel it might be better so that two CLFs (or anything using the FixedFunction) files can't produce different results depending on OCIO version if we were to later allow `lim == 1.0000`. And could always revert if the VWG decide otherwise before we merge the PR. Let me know if that sounds good.

doug-walker:
My understanding is that no one will be using this with values really close to 1 anyway, so I think it would be somewhat better to proceed with letting 1.0 pass validation and then clamp to the slightly different value in the constructor.  That approach is both simplest for the code and simplest for apps.  But I also don't object to having the validation use the precise values, if you or the Gamut WG prefer that approach.

remia:
Thanks @doug-walker, I pushed the update to clamp the limit parameter as was discussed and some last minor refactor to the GPU Op that remove all code duplication and bring it in line with the CPU Op, hopefully that's ok.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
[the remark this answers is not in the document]
Tristimulus: I have spoken with my co-chairs of the ACES Gamut VWG (@nick-shaw & @carolalynn22) and we concluded that for the FixedFunctionOp we would prefer not to impose any limits here unless there is a numerical or performance reason for it.
Sensible value limits for users of the Parametric Gamut Compression could be set in a UI via slider ranges instead (the VWG has yet to cover those) but we did not think that the underlying op should need to impose limits unless they are needed for numerical or application stability.

So I think the bounds could be something like this? :

`lim_low_bound  = 1.0000001f`
`lim_hi_bound = numeric_limits<float>::max()`

`thr_low_bound = 0.0f` (Where did the current value of 0.4 come from? From the VWG?)
`thr_hi_bound = 0.99999994`

`pwr_low_bound = 1.0`
`pwr_hi_bound  = numeric_limits<float>::max()` (Does this need be lower for numeric stability? Not sure.)

Since the check_bounds() and clamping functions treat the bounds values as inclusive/acceptable values I've chosen the next float32 (IEEE 754) increment / decrement from 1.0 where we really just mean `distance_limit > 1.0` and `threshold < 1.0`.
So if there are concerns about this running on machines with a different, lower precision floating point value representation these in-/decrements could be adjusted accordingly.

I would also think that the values given to the check_bounds() function and the values used for clamping in the op code itself should be identical, right? Otherwise the behavior could end up being counterintuitive.
For instance `FixedFunctionOpData::validate()` would accept a threshold value of 0.9999999 without throwing the parameter exception but the op then quietly clamp it to 0.9999.
(I know it's a bit pedantic but I wanted to mention it anyway...sorry.)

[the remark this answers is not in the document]
remia: Thanks @Tristimulus for the clarification of the VWG position.

The current threshold low bound is from https://github.com/ampas/aces-vwg-gamut-mapping-2020/blob/master/model/GamutCompress.dctl but that was just a placeholder waiting for the group recommendation. 

About the precision concerns, I guess it's hard to say, one potential limiting factor could be OpenGL ES and mobile GPU if we support them, we could lower that to float16 or alternatively just keep the current values ?

Currently the `validate()` allows for threshold and limit == 1 even though they get clamped to different values in the Op code, there was discussion yesterday about that below in the PR, conclusion was to leave it that way but nobody was opposed to your suggested approach, that is everything happening in `validate()` if I understood your point correctly. That might be slightly better to not have hidden clamping happening inside OCIO as well as remove some duplication and risk for errors.

hodoulp (approved):
Thanks @remia for the major improvements around the GPU code path. As discussed, I double-checked the DX11 support using Maya and that's a success.

remia:
> Thanks @remia for the major improvements around the GPU code path. As discussed, I double-checked the DX11 support using Maya and that's a success.

Thanks @hodoulp, I think we will just have on last commit to adjust the parameters limits as discussed above but the Op themselves should not change anymore. Thanks again for the testing on Windows.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpData.cpp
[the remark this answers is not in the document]
remia: Implemented the suggestions from above, only difference is I'm using limits based on float16 instead of float32 but that can be adjusted if wanted.

Issue #1235, opened 2020-12-10
DebugLock means that Mutex isn't a mutex at all when NDEBUG is defined


karlhendrikse:
This is the code of `DebugLock`:
```
// In debug mode, try to trap recursive cases and lock/unlock debalancing cases.
class DebugLock
{
public:
    DebugLock() = default;
    DebugLock(const DebugLock &) = delete;
    DebugLock& operator=(const DebugLock &) = delete;
    ~DebugLock() { assert(!m_locked); }

    void lock()     { assert(!m_locked); m_mutex.lock();   m_locked = true;  }
    void unlock()   { assert(m_locked);  m_mutex.unlock(); m_locked = false; }
    void try_lock() { assert(!m_locked); m_locked = m_mutex.try_lock(); }

    bool locked() const { return m_locked; }

private:
    // An exclusive and non-recursive ownership lock.
    std::mutex m_mutex;
    bool       m_locked = false;
};
```
Note that if two threads try to lock the mutex at the same time, one of them will assert. That defeats the entire purpose of a mutex. I assume the code is trying to prevent the same thread from recursively locking the mutex multiple times, but for that `m_locked` would need to be thread-local.

DarkDefender:
I just want to chime in and say that I ran into this issue when running Blender.
It would trigger this assert when trying to render certain scenes.

doug-walker:
This was apparently fixed by PR #1407.

PR #1407, opened 2021-05-25
Fix errors in mutex debugging (#1235)

brechtvl:
* Store the lock owner thread rather than `m_locked`, so that multiple threads
  accessing the same mutex do not give false positives, but recursive locks on
  the same thread are still caught.

* Match the `std::mutex` API: return bool from `try_lock()` and remove `locked()`.
  No code is using this, but any new code that did would only build in either
  release or debug configuration.

### DIFF
--- a/src/OpenColorIO/Mutex.h
+++ b/src/OpenColorIO/Mutex.h
@@ -7,6 +7,7 @@
 
 
 #include <mutex> 
+#include <thread>
 #include <assert.h>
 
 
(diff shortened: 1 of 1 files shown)


michdolan (approved):
Thanks for the contribution!

michdolan:
@brechtvl let us know if you need help resolving the CLA issue.

brechtvl:
The CLA issue should be resolved now.

src/OpenColorIO/Mutex.h
```diff
@@ -24,18 +25,26 @@ class DebugLock
...
+
+    void lock()
+    {
+        assert(m_owner != std::this_thread::get_id());
+        m_mutex.lock();
+        m_owner = std::this_thread::get_id();
+    }
+    void unlock()
+    {
+        assert(m_owner == std::this_thread::get_id());
```
hodoulp, line 38: The following line seems useless.

src/OpenColorIO/Mutex.h
```diff
@@ -24,18 +25,26 @@ class DebugLock
...
+        m_owner = std::thread::id();
+        m_mutex.unlock();
+    }
+    bool try_lock() { return m_mutex.try_lock(); }
```
hodoulp, line 42: Should the `try_lock()` implementation also have the `assert(m_owner != std::this_thread::get_id());` line?
I think it should.

hodoulp (changes requested):
Thanks a lot for the contribution!

src/OpenColorIO/Mutex.h
brechtvl: It checks that the same thread is holding the lock before unlocking. It seems useful to me, but I might be missing something?

src/OpenColorIO/Mutex.h
brechtvl: Yes, seems it's undefined behavior to `try_lock()` when already holding the lock.

brechtvl:
Apologies for the late reply, I lost of track of this pull request.

src/OpenColorIO/Mutex.h
hodoulp: I refer to line 39 (and not line 38 which is fine).
In normal use case, `m_owner` variable already owns the right thread id.

brechtvl: Ah, that's to indicate there is no owner. `std::thread::id()` sounds like it might return the current thread id, but it's just the default constructor with an empty thread id.

hodoulp: Now, I understand (Refer to https://en.cppreference.com/w/cpp/thread/thread/id)

hodoulp (approved):

Issue #1261, opened 2021-01-12
OCIO/OIIO CMake conflict when using OpenEXR >= 2.5 for Half dependency


michdolan:
I've only tested this on Windows so far, but while updating my Windows builds of OCIO v2.0.0-rc1 and OIIO 2.2.10.1, both linking against OpenEXR 2.5.4, this fails the CMake configuration.

The cause is that we are relying on OIIO's CMake config to find OIIO, which has a condition which relies on and calls OpenEXR's CMake config when using OpenEXR >=2.5. Since OCIO already has a target for Half, but not for the other OpenEXR components it creates a conflict and fails to setup the OIIO dependency.

I'm actively working on removing the OIIO dependency from OCIO, which will solve this issue, but I wanted to mention it. We could resolve it on our end by either building all of OpenEXR (instead of just Half), or by reverting to using our old OIIO find module instead of the config.

michdolan:
```
CMake Error at .../openexr-2.5.4/lib/cmake/IlmBase/IlmBaseTargets.cmake:37 (message):
  Some (but not all) targets in this export set were already defined.

  Targets Defined: IlmBase::Half

  Targets not yet defined:
  IlmBase::IlmBaseConfig;IlmBase::Iex;IlmBase::IexMath;IlmBase::Imath;IlmBase::IlmThread


Call Stack (most recent call first):
  .../openexr-2.5.4/lib/cmake/IlmBase/IlmBaseConfig.cmake:27 (include)
  C:/Program Files/CMake/share/cmake-3.16/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  .../openexr-2.5.4/lib/cmake/OpenEXR/OpenEXRConfig.cmake:30 (find_dependency)
  C:/Program Files/CMake/share/cmake-3.16/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  .../oiio-2.2.10.1/lib/cmake/OpenImageIO/OpenImageIOConfig.cmake:31 (find_dependency)
  src/CMakeLists.txt:11 (find_package)


-- Configuring incomplete, errors occurred!
```

remia:
I think I stumbled on the same issue, on macOS using `OIIO 2.2.10.1` and `OpenEXR 2.5.4`, here it complains about the Threads target. This is repeated for `ociolutimage` `ocioconvert` `ociodisplay` `ocioperf` `oiiohelpers`.

```
[...]

PR #1408, opened 2021-06-05
Remove FindOpenImageIO.cmake

lgritz:
OCIO wants a minimum OIIO of 2.1, and coincidentally, that is the
release series in which OIIO started exporting OpenImageIOConfig.cmake
and friends, making the FindOpenImageIO module obsolete. This file is
not only unnecessary, but actually because it's out of date, it won't
work properly with upcoming OpenImageIO 2.3 (due to be released later
this year, or right now for anybody working from OIIO master branch).

Simply removing FindOpenImageIO.cmake will cause OCIO's build to use
the OIIO exported config.

I did notice that your OpenImageIO requirement was for 2.1.9, and there
is an unfortunate typo in that version's exported config. So I'm also
proposing making the minimum OIIO be 2.1.10, which was a patch released
in Jan 2020 -- a full year before OCIO v2 was released. OIIO continued
to issue 2.1 patches until 2.1.20 (Sept 2020), and since then, it's
actually OIIO 2.2.x that has recieved active support. I'm not saying
that you need to use 2.2 as a minimum, but I am trying to justify that
bumping your 2.1.9 minimum to 2.1.10 is not a radical move and will
avoid possible build troubles.


lgritz:
The CI is failing, I'll need to look into that in more detail. You can hold off on reviewing while I figure this out.

EvanAW:
> The CI is failing, I'll need to look into that in more detail. 

Note that even the ones (Windows and MacOS) that are passing aren’t finding OIIO. 
Instead of failing, a warning is issued saying that the four apps that depend on OIIO will not be built, and the CI script continues.

lgritz:
I'm going to abandon this PR until I figure out what's going on.

EvanAW:
Just in case this helps @lgritz: OCIO was using the OpenImageIOConfig.cmake file, but there were issues when using OpenEXR >= 2.5 for Half dependency.  See https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1261

The decision was made to revert to using FindOpenImageIO.cmake
See https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1266

PR #1409, opened 2021-06-10
Hide dependency symbol visibility

lgritz:
When creating libOpenColorIO.so, we lacked the linker commands that
hide symbol visibility from the dependent libraries, which is
necessary to prevent OCIO from exporting the symbols from Expat and
the other dependencies that OCIO needs to use internally but is not
trying intentionally to expose via its API.

Failing to do this can result in symbol clashes and all sorts of
subtle errors if OCIO is used in an app that also uses and is linked
against a potentially different version of Expat (or any of the other
deps).

### DIFF
--- a/src/OpenColorIO/CMakeLists.txt
+++ b/src/OpenColorIO/CMakeLists.txt
@@ -273,6 +273,14 @@ set_target_properties(OpenColorIO PROPERTIES
     SOVERSION     ${SOVERSION}
 )
 
+if (UNIX AND NOT APPLE)
+    # Also hide all the symbols of dependent libraries to prevent clashes if
+    # an app using this project is linked against other versions of our
+    # dependencies.
+    set_property (TARGET OpenColorIO
+                  APPEND PROPERTY LINK_FLAGS "-Wl,--exclude-libs,ALL")
+endif ()
+
 if(MSVC AND BUILD_SHARED_LIBS)
 	# Install the pdb file if any.
 	install(FILES $<TARGET_PDB_FILE:${PROJECT_NAME}> DESTINATION bin OPTIONAL)


lgritz:
This is failing on Mac for reasons that I don't quite understand.

Should we not be doing this on Mac? Or have I failed so add something else that's needed?

I'm looking into it, but if anybody more expert than I am about this symbol visibility can spot the problem, please let me know.

lgritz:
Mystery solved. On Mac, ld does not accept that argument. PR amended.

lgritz:
Note: I have made this PR against RB-2.0, because that's where I really need it right now. But if accepted, please don't forget to cherry-pick it into master so that will also be in 2.1.

lgritz:
I'm not sure why the GPU tests failed, but looking at the logs, it doesn't appear to be related to my changes, since it never gets as far as building the software.

remia:
I was under the impression that GPU tests were not run, except on a specific branch. Maybe someone with permissions can restart the job from within GH Actions panel.

Expat symbols are there on macOS built library, not sure if it will cause the same issue you were seeing, but the linker doesn't seem to support the flag anyway as you mentioned. Looks like CMake 3.13 would allow us to use `target_link_options` instead of `set_property` but I guess we don't really care here.

michdolan:
It's odd that the GPU job ran. It should only run on push to a branch under the primary OCIO repo. I'm going to try closing and reopening this PR to restart all checks, which may help.

michdolan:
Closing and reopening didn't help, but note that the GPU check is not required.

hodoulp:
When integrating pull requests to create the release 2.0.1, the GPU CI build was also constantly failing. It took me a while at that time, to have a green build... There is definitively a weakness somewhere!

hodoulp:
Note: Here is a thread with potential solutions for macOS but that requests an extra step:
https://stackoverflow.com/questions/435352/limiting-visibility-of-symbols-when-linking-shared-libraries

From that thread, I tried the approach `-Wl,--version-script=libfoo.version` but i's also unsupported on macOS.

But `gcc` remaps to `AppleClang` on macOS so, the limitation seems to be a gcc vs. clang difference!

hodoulp (approved):

doug-walker (approved):

hodoulp:
The GPU CI build should be fixed by PR #1412

PR #1410, opened 2021-06-13
Allow choice of specific python release

lgritz:
If you're building on a system with multiple versions of Python available,
it can be very hard to make OCIO build against a specific version. In
particular, https://cmake.org/cmake/help/latest/module/FindPython.html
explains that FindPython looks preferably for version 3 of Python.

This situation doesn't really come up in the CI, because all of the containers
have only one version of python each -- the one you are trying to use.
But often people are building on systems with several versions of Python.

I also removed a comment about pybind11 2.4.3 that no longer seems
relevant since we're asking for 2.6.1 or higher.

### DIFF
--- a/share/cmake/modules/FindExtPackages.cmake
+++ b/share/cmake/modules/FindExtPackages.cmake
@@ -52,18 +52,8 @@ if(OCIO_BUILD_APPS)
     find_package(lcms2 2.2 REQUIRED)
 endif()
 
-if(OCIO_BUILD_PYTHON)
-
-    # NOTE: Depending of the compiler version pybind11 2.4.3 does not compile 
-    # with C++17 so, if you change the pybind11 version update the code to 
-    # compile pybind11 and dependencies with C++17 or higher i.e. remove the 
-    # cap of C++ version in FindPybind11.cmake and 
-    # src/bindings/python/CMakeLists.txt.
-
-    # pybind11
-    # https://github.com/pybind/pybind11
-    find_package(pybind11 2.6.1 REQUIRED)
-endif()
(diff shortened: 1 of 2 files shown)


share/cmake/modules/FindExtPackages.cmake
```diff
@@ -52,18 +52,8 @@ if(OCIO_BUILD_APPS)
...
-    # with C++17 so, if you change the pybind11 version update the code to 
-    # compile pybind11 and dependencies with C++17 or higher i.e. remove the 
-    # cap of C++ version in FindPybind11.cmake and 
-    # src/bindings/python/CMakeLists.txt.
-
-    # pybind11
-    # https://github.com/pybind/pybind11
-    find_package(pybind11 2.6.1 REQUIRED)
-endif()
+set (OCIO_PREFERRED_PYTHON_VERSION "" CACHE STRING
```
michdolan, line 55: Thinking we might want to make this an option declaration in the root `CMakeLists.txt` file for better visibility, which could then be referenced here.

lgritz: I can move it to the top level if you want. Also, do you prefer this name, or the less wordy `OCIO_PYTHON_VERSION`?

hodoulp: Less wordy one `OCIO_PYTHON_VERSION` with a warning (or error) if the option is defined but `OCIO_BUILD_PYTHON` is OFF.

lgritz: done

lgritz:
PR amended with the review suggestions.

hodoulp (approved):

doug-walker (approved):

PR #1411, opened 2021-06-13
Use CMake visibility flags

remia:
Following recent discussions around symbol visibility, we noticed that OpenColorIO only seem to use `-fvisibility-inlines-hidden` but not `-fvisibility=hidden` on the core library. This PR add it and replace the hard coded compiler flags with CMake utility variables to enable that behaviour by default overall.

Here is some tests I did on macOS to illustrate the impact:
```
cmake -DCMAKE_INSTALL_PREFIX=myinstall -DOCIO_BUILD_APPS=ON -DOCIO_INSTALL_EXT_PACKAGES=ALL .. && cmake --build . -j 12 && make install

// A few moments later..
nm myinstall/lib/libOpenColorIO.2.1.0.dylib | grep " T " | grep -i _XML_ | wc -l
nm myinstall/lib/libOpenColorIO.2.1.0.dylib | grep " T " | wc -l
ls -lh myinstall/lib/libOpenColorIO.2.1.0.dylib | awk '{print $5}'

MASTER BRANCH
66
3318
6.6M

THIS PR
66
907
6.3M
```

Notice that Expat symbols are still leaked, I don't know yet how to fix that on macOS.


share/cmake/utils/CompilerFlags.cmake
```diff
@@ -72,6 +68,14 @@ endif()
 mark_as_advanced(PLATFORM_COMPILE_FLAGS)
 
 
+###############################################################################
+# Define hidden symbol visibility by default for all targets.
+
+set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN YES)
```
lgritz, line 76: Should these be options/cache variables so that they show up in ccmake, etc. as settable, documented build options?

lgritz (approved):
This is a good improvement that (along with the related PR I made) aligns OCIO with best practices for symbol hiding. Please do also backport this to RB-2.0! Many of us would be relieved to get a 2.0.2 that does all the symbol hiding properly.

lgritz:
I don't know the right solution for macos, but I would certainly like to find out. What I tried in the other PR works for Linux, but that linker option just didn't exist on Mac.

share/cmake/utils/CompilerFlags.cmake
remia: I guess we should prefer options for modern cmake. Do you see a scenario that would justify these being exposed ?

lgritz: Not necessarily. I think you could make a good argument that the symbols should always be hidden, and so it should not be exposed as cache variables. People could still override it with `-D`, but they'd need to read the cmake to find out that it existed; by not being declared as cache variables, they are much less prominently advertised to anybody not specifically looking for them.

remia: Given that these are builtin CMake variable and not OCIO specific, I don't think there is much value aliasing them as an option or cache variable. We can probably expect developer wanting to play with these to know them, as it's probably going to be required (if ever) only in very specific / debug tests ?

One thing that I just realised though is that we should probably check that the user didn't override these through the command line or other ways, so that there is still opportunity to change the settings if need be. Maybe that was part of you concerns.

docs/quick_start/installation.rst
```diff
@@ -63,18 +63,18 @@ default, cmake will try to install all of the items labelled with * and so
...
 - \*pystring >= 1.1.3
 
 Some optional components also depend on:
 
 - \*Little CMS >= 2.2 (for ociobakelut ICC profile baking)
-- \*pybind11 >= 2.4.3 (for the Python bindings)
+- \*pybind11 >= 2.6.1 (for the Python bindings)
 - Python >= 2.7 (for the Python bindings)
 - Python 3.7 or 3.8 (for the docs, with the following PyPi packages)
-    - Sphinx
```
michdolan, line 77: FWIW: The casing of these packages is consistent with PyPI: https://pypi.org/project/Sphinx/

That being said, I don't object to consistent casing here since `pip` is case insensitive.

michdolan (approved):

share/cmake/utils/CompilerFlags.cmake
hodoulp: I think that OpenColorIO `cmake` should not have dedicated OCIO options (`ccamke` displays nothing related to symbol visibility) because most of people will/should not change the default behavior. However, expert users should have a way to change default values if needed i.e. lines 74 to 76 could be something like:

```
if(CMAKE_C_VISIBILITY_PRESET NOT DEFINED)
  set(CMAKE_C_VISIBILITY_PRESET hidden)
else()
   message(WARNING 'The .... is ${CMAKE_C_VISIBILITY_PRESET}') 
endif()
...
```

hodoulp (changes requested):
@remia I'm waiting last changes about visibility default values (if that's a valid comment / approach) before approving the pull request. Let me know.

docs/quick_start/installation.rst
remia: Good point, I reverted the change!

share/cmake/utils/CompilerFlags.cmake
remia: Implemented just what you described @hodoulp, thanks for the snippet.

hodoulp (approved):

PR #1412, opened 2021-06-14
Adsk Contrib - Fix GPU build on RB-2.0

hodoulp:

The pull request tries to improve / fix the RB-2.0 build for GPU which fails quite often...
The GPI CI build should not trigger when starting from a forked branch.

The other changes are only here to impose a compilation.

### DIFF
--- a/.github/workflows/gpu_workflow.yml
+++ b/.github/workflows/gpu_workflow.yml
@@ -25,7 +25,7 @@ jobs:
   linux_gpu:
     name: 'Linux CentOS 7 VFX CY2020 GPU <GCC 6.3.1>'
     # Don't run on OCIO forks
-    if: github.repository == 'AcademySoftwareFoundation/OpenColorIO'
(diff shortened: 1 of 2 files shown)


hodoulp:
Based on the current checks, the GPU is not started. That's a first good point. Now, let see if it succeeds.

hodoulp:
To avoid the Windows and macOs emulation mode, the GPU CI build needs a physical machine adding some costs. So, the original intent was to only run the GPU CI build on the repo itself (i.e. not running on forks) to limit fees. The ASWF offers the GPU build capability but OpenColorIO should then try to wisely use it. Starting few weeks ago, the GPU CI build was also executed on forks!

Thanks to @simran, we finally discover that the execution condition was, in fact, wrong. 
> As `github.repository` is the repo the workflow is running in, this will always be true for the official repo. Usually one would use `github.event.pull_request.head.repo.full_name != github.repository` as condition to prevent jobs from running if they originate from a fork, [...]

michdolan (approved):

doug-walker (approved):

PR #1413, opened 2021-06-15
[DO NOT MERGE] Check GPU CI for security issue

Simran-B:


Simran-B:
GPU workflow did not get triggered - which is good. Unclear why that happened in #1409

image
image

Curiously, the [workflow run](https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/926843975) links to https://github.com/AcademySoftwareFoundation/OpenColorIO/commit/70d541843ac66e117011d31c4b3c73a163cf0662

image

and GitHub even says that this commit does not belong to any branch in this repo:

image

There was a force-push, but in the branch of the fork

image

I don't see anything in the PR feed that the base branch was changed or the like, but I would not expect a workflow run in any case. Very peculiar.

Issue #1398, opened 2021-05-04
GetColorSpaceFromFilepath should handle v1 configs better


doug-walker:
OCIO v2 introduced the new FileRules section of the config to add some new features around file path mapping.  A new set of API calls related to that was also added.  The intention is that any v1 clients that were using parseColorSpaceFromString would update to use getColorSpaceFromFilepath.  The name was changed because there is a difference in behavior with the new system that developers need to think about when upgrading.  ParseColorSpaceFromString still exists and still does what it used to do, but it is listed as deprecated.

The first difference is that getColorSpaceFromFilepath uses the new FileRules, so it can do the original v1 algorithm of searching for a color space name in the path (if the config author enables it), but it can also do a lot more, such as map specific file extensions to a specified color space. 

The second difference is that getColorSpaceFromFilepath *always* returns a valid color space.  This was a request from app developers that always need a default (which is a common scenario).

The behavior of parseColorSpaceFromString (in v1 and v2) actually depends on the value of the strictparsing attribute of the config.  If strictparsing is true, parseColorSpaceFromString returns an empty string if the path did not contain a color space name.  If strictparsing is false, it will try to return the default role in that scenario, if it exists and points to a legitimate color space (otherwise it still returns an empty string).

If an application using v2 wants to do something special in the case where the path does not contain a color space name, the suggested code is as follows:
```
if (config->filepathOnlyMatchesDefaultRule(path)  &&  config->isStrictParsingEnabled())  {
     do what you want to do here 
}
```

But most apps (in my estimation) do not need to do this and should simply be able to call getColorSpaceFromFilepath to set the default color space for a file.  *This should be the case regardless of the version of the config file.*  Ideally, the application should not need to check the version of the config and sometimes call parseColorSpaceFromString and other times call getColorSpaceFromFilepath.

The reason I created this issue is that there is currently a problem with getColorSpaceFromFilepath when used with v1 configs (which of course do not have the FileRules section).  It currently is not searching the path for a color space name.  We will fix that but we also need to decide what getColorSpaceFromFilepath will return for a v1 config if the default role is not present or not valid.  (Remember, the call promises to return a valid color space and developers should be able to rely on this.)  Here is the proposal:

```
if (config->getMajorVersion() == 1) {
    if PathUtils.cpp ParseColorSpaceFromString finds a match, return it
    else if the default role is a valid color space, return it
    else if there is a color space such that lower(cs) == "raw", return the first one
[...]

PR #1417, opened 2021-06-19
Adsk Contrib - Improve File Rules support for v1 configs

hodoulp:

The changes ease the transition from the version 1 `Config::parseColorSpaceFromString()` method to the File Rules using `Config:: getColorSpaceFromFilepath()`. When reading version 1 config files, the in-memory config will now have two valid file rules. The first one is  `FilePathSearchRuleName` which evaluates like the former `Config::parseColorSpaceFromString()`, and the second one is the `DefaultRuleName`. It means that the integration of the OpenColorIO v2 library can now only use the `Config:: getColorSpaceFromFilepath()` method for all config versions.

In order to always return a valid color space (i.e. find a color space for the `DefaultRuleName` rule), the algorithm is:
1. Use the default role if it exists (i.e. that's the default implementation)
2. Use the "raw" color space (case insensitive) if it exists & is a 'data' color space
3. Use the first 'data' color space if one exists
4. Use the first active color space
5. finally, fallback to the first color space.

Two enhanced unit tests validate the v1 config read and the v1 in-memory config creation. 

Note: The `Config::parseColorSpaceFromString()` method is now flagged as deprecated.

Please refer to #1398 for the discussion details.


include/OpenColorIO/OpenColorIO.h
```diff
@@ -1009,7 +1010,7 @@ class OCIOEXPORT Config
      * * If strict parsing is disabled, return ROLE_DEFAULT (if defined).
      * * If the default role is not defined, return an empty string.
      */
-    OCIO_DEPRECATED("This is now deprecated, please use Config::getColorSpaceFromFilepath().")
+    OCIO_DEPRECATED("This is now deprecated starting at v2.0, please use Config::getColorSpaceFromFilepath().")
```
michdolan, line 1013: Just a small suggestion. Either wording works fine though.

```suggestion
    OCIO_DEPRECATED("This was marked as deprecated starting in v2.0, please use Config::getColorSpaceFromFilepath().")
```

src/OpenColorIO/FileRules.cpp
```diff
@@ -940,5 +955,96 @@ std::ostream & operator<< (std::ostream & os, const FileRules & fr)
...
+            {
+                if (config.getNumColorSpaces() > 0)
+                {
+                    // Take the first active color space.
+                    fileRules->setColorSpace(1, config.getColorSpaceNameByIndex(0));
+                }
+                else
+                {
+                    static constexpr char msg[]
+                        = "The default rule creation fallbacks to the first color space because "\
```
michdolan, line 1030:
```suggestion
                        = "The default rule creation falls back to the first color space because "\
```

michdolan (approved):
Two super minor wording suggestions, but LGTM. Thanks @hodoulp !

hodoulp:

doug-walker (approved):

PR #1418, opened 2021-06-19
Adsk Contrib - Fix FixedFunctionTransform::CreateEditableCopy()

hodoulp:

The pull request fixes a major bug with the `FixedFunctionTransform` transform. When using the `FixedFunctionTransform::CreateEditableCopy()` method for a style with mandatory params it was throwing. The problem was that it created a transform with the style only, validating the instance and only after adding the params.

The code now creates an instance with the style and params (when needed) so the validation always succeeds. I've added a unit test to validate the behavior.


michdolan (approved):

hodoulp:

doug-walker (approved):

PR #1421, opened 2021-06-22
Test CMake Config script

remia:
This PR is a follow up of #1397, adding a simple test of the generated script.
* A dummy single file project is built and run after main library has been tested,
* Currently only enabled on shared build, it seems static build would be harder to test and require the client project to import the correct dependencies,
* Build is done in source to avoid some inconsistency in GitHub Action while changing directory within a single job (I had some issues in Windows),
* Added ``cxx_std_11`` public compiler feature on the core library based on @hodoulp suggestion.

On a related note, I added an explicit warning ignore in the client project, the already known C4275. I wonder if we could do that from #pragma inside the source code (OCIO Exception class) to not impose that on the users.


.github/workflows/ci_workflow.yml
```diff
@@ -268,6 +268,14 @@ jobs:
       - name: Test
         run: ctest -V
         working-directory: _build
+      - name: Test CMake Consumer
+        if: matrix.build-shared == 'ON'
+        run: |
+          cmake -DCMAKE_PREFIX_PATH=../../../_install .
```
hodoulp, line 274: I think that `-DCMAKE_BUILD_TYPE=${{ matrix.build-type }} ` is missing; otherwise, it build a release by default.

.github/workflows/ci_workflow.yml
```diff
@@ -390,6 +398,14 @@ jobs:
...
```
hodoulp, line 404: I think that -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} is missing; otherwise, it build a release by default.

.github/workflows/ci_workflow.yml
```diff
@@ -518,3 +534,13 @@ jobs:
         run: ctest -V
         shell: bash
         working-directory: _build
+      - name: Test CMake Consumer
+        if: matrix.build-shared == 'ON'
+        run: |
+          cmake -DCMAKE_PREFIX_PATH=../../../_install .
+          cmake --build . \
```
hodoulp, line 541: I think that -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} is missing; otherwise, it build a release by default. That seems much more important for Windows (refer to the discussion in issue #11530?because of some limitations/problems.

tests/cmake-consumer/consumer.cpp
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright Contributors to the OpenColorIO Project.
+
+
+// Check that OCIO library can be consumed by a cmake build
+
+#include <cassert>
+#include <iostream>
```
hodoulp, line 8: [Minor] Are these two includes needed?

hodoulp (changes requested):

hodoulp:
> On a related note, I added an explicit warning ignore in the client project, the already known C4275. I wonder if we could do that from #pragma inside the source code (OCIO Exception class) to not impose that on the users.

That's a good point. My answer is yes in OpenColorIO.h

src/OpenColorIO/CMakeLists.txt
```diff
@@ -182,12 +182,18 @@ if(NOT WIN32)
     configure_file(res/OpenColorIO.pc.in ${CMAKE_CURRENT_BINARY_DIR}/OpenColorIO.pc @ONLY)
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/OpenColorIO.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
 endif()
+
 add_library(OpenColorIO ${SOURCES})
 
+# Require at least a C++11 compatible compiler for consumer projects.
+target_compile_features(OpenColorIO
+    PUBLIC cxx_std_11
```
michdolan, line 190: This is enforced in the root CMake project via this module:
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/share/cmake/utils/CppVersion.cmake

Do you think that's sufficient? Or that having the requirement directly on the target is beneficial too?

michdolan (commented):
Great idea @remia !

.github/workflows/ci_workflow.yml
remia: Thanks @hodoulp, good catch!

tests/cmake-consumer/consumer.cpp
remia: Will remove `<iostream>` good catch, I seem to remember `cassert` is needed at least on Windows.

src/OpenColorIO/CMakeLists.txt
remia: The main idea was to pass this requirement onto the project importing OpenColorIO. If you look at the generated `OpenColorIOTargets.cmake`, it should now include a `INTERFACE_COMPILE_FEATURES "cxx_std_11"` which was not the case before. That way client project should get a nice error if using an older standard. The `CppVersion.cmake` script only set the default CMake variables to be used to initialise subsequent targets declared within the project.

Does that that answer your question or did I misinterpret ?

.github/workflows/ci_workflow.yml
remia: Is there a typo in the issue number ?

hodoulp: Yep, fixed.

src/OpenColorIO/CMakeLists.txt
hodoulp: That's something important to enforce/impose at the API level because it's a strong requirement imposed by OpenColorIO to apps.

hodoulp (approved):

src/OpenColorIO/CMakeLists.txt
michdolan: Makes sense. Thanks.

michdolan (approved):

PR #1436, opened 2021-07-13
Update CLF test files, add CLF python scripts

doug-walker:
Made some refinements to the Common LUT Format (CLF) test set to allow it to also serve as the test kit for the ACES CLF Implementation Working Group.

Modifications to CLF test files included:
- avoid causing the max finite half float value to overflow into infinity when processing the CLF test target
- the length of some Lut1D arrays was increased to improve the match between CPU and GPU (due to the known issue of GPU quantization of fractional values in texture look-ups)
- likewise, some Lut3D interpolation values were changed to tetrahedral so the CPU and GPU would match better for small Luts (since a shader is used for the computation rather than a simple texture look-up)
- some CLFs were modified so that they produce a plausible output image (e.g. not a constant, as was happening in one case)
- a few new tests were added to improve coverage in a few areas

A new "clf" directory was added under the share directory to add files related to the usage of OCIO as an aid for validation of CLF implementations. There is a script that will process the CLF working group's test image through all the CLF examples and another script to compare reference images to images from an implementation under test using the CLF working group's test metric.  The CLF working group's small OpenEXR test image was also added so that no external files are needed to run the tests.

Also fixed a minor issue in the GPU shader used for large (typically > 4096 values), non-half-domain Lut1D arrays to correct a small error when large negative values were evaluated.

Signed-off-by: Doug Walker Doug.Walker@autodesk.com


michdolan (approved):

hodoulp:

hodoulp (approved):

hodoulp:
\easycla

PR #1446, opened 2021-08-03
Adsk Contrib - Fix a bit depth optimization bug

hodoulp:

If the color transformation ends up to do nothing it internally creates an empty list of ops. In that situation only, the bit-depth adjustment was not done at the right time corrupting the final list of CPU ops (i.e. the actual processing).

The method `FinalizeOpsForCPU()` has two main steps. The first one only computes the list of the ops, and the second computes the list of CPU ops including the potential bit-depth adjustments (which must be held by `m_inBitDepthOp` & `m_outBitDepthOp`). The bit-depth adjustments were badly done during the first step. The fix consists to remove that faulty code and let the second step correctly handing the case.

The unit test validates the fix by using two different approaches (which were ending to the same problem).


hodoulp:
This is a candidate for 2.0.2

michdolan (approved):

hodoulp:
Friendly reminder that the pull request could be merged next week so, you could use that period to review these changes knowing that the PR is also good candidate for `RB-2.0`.

doug-walker (approved):

Issue #1457, opened 2021-08-10
Error when applying a CDL as a Look


UnaiM:
In OCIO v2, if a `Look` contains a `CDLTransform` or `FileTransform` defining/pointing to a no-op CDL, picking a `View` that uses said `Look` generates an error. Here’s an example:

```python
import PyOpenColorIO as ocio
config = ocio.Config.CreateFromStream('''\
ocio_profile_version: 1
roles:
  scene_linear: raw
displays:
  default:
    - !<View>
      name: cdl
      colorspace: raw
      looks: cdl
looks:
  - !<Look>
    name: cdl
    process_space: raw
    transform: !<CDLTransform> {}
colorspaces:
  - !<ColorSpace>
    name: raw
''')
config.getProcessor('raw', 'default', 'cdl', ocio.TransformDirection.TRANSFORM_DIR_INVERSE)
```

```
Traceback (most recent call last):
  File "/path/to/the/script.py", line 9, in <module>
    config.getProcessor('raw', 'default', 'cdl', PyOpenColorIO.TransformDirection.TRANSFORM_DIR_INVERSE)
PyOpenColorIO.Exception: BuildColorSpaceOps failed, null dstColorSpace.
```

Note that this also happens if the CDL is filled up but still a no-op, i.e. the slope explicitly set to 1 in all channels etc. Also, this happens regardless of the `ocio_profile_version`, and **isn’t the expected behaviour in OCIO v1.**

hodoulp:
@UnaiM thanks for reporting the problem.
Which branch did you use for testing?

The `master` branch already contains a fix concerning an empty look but the `RB-2.0` does not (i.e. PR #1443 is waiting for approval and merge).

hodoulp:
@UnaiM The pull request is now merged in RB-2.0

UnaiM:
Thanks @hodoulp, however after trying the example above on a [fresh `RB-2.0` build](/AcademySoftwareFoundation/OpenColorIO/commit/9eb736bd84f7ac1ea1f7620dd3b585a2e9f90285), it still errors the same.

PR #1460, opened 2021-08-18
Adsk Contrib - Fix the processor creation using a DisplayViewTransform with a no-op look

hodoulp:

The pull request fixes the creation of a processor for a (display, view) color transformation when the look part exists but that's a no-op as explained in #1457.

The problem was that the original code optimized the list of ops when the look is a no-op leaving a variable (i.e `currentColorSpace`) uninitialized for the inverse computation. The forward computation was right because the variable is initialized before calling `RunLookTokens()`. The solution consists to differ the optimization of the list of ops to the optimization step because the look and its process space are both valid so, let the code adds the list of ops (even an empty one). By doing so, it correctly sets the variable. 

Note that the former code was also removing the process space ops when the look was a no-op. The new code keeps it but the optimization will latter combine the forward and reverse process space computation as highlighted by the second case of the new unit test.

### DIFF
--- a/src/OpenColorIO/transforms/LookTransform.cpp
+++ b/src/OpenColorIO/transforms/LookTransform.cpp
@@ -195,10 +195,15 @@ std::ostream& operator<< (std::ostream& os, const LookTransform& t)
 namespace
 {
 
-void RunLookTokens(OpRcPtrVec & ops,
-                   ConstColorSpaceRcPtr & currentColorSpace,
+// The method gets the list of ops from the list of looks.
+//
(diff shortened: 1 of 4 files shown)


remia:
Great catch @hodoulp, I think it also solve the earlier issue related to the LegacyViewingPipeline #1437 by addressing the root cause instead of protecting the client code.

```
// Put the new ops into a temp array, to see if it's a no-op
// If it is a no-op, dont bother doing the colorspace conversion.
```
This comment earlier in the `RunLookTokens` function is no longer accurate, correct ?

michdolan (approved):

Issue #1435, opened 2021-07-12
Add preliminary OSL support


doug-walker:
OpenShadingLanguage (OSL), which is also an ASWF project, has many similarities to the shading languages that OCIO's GPU renderer can currently generate such as GLSL and HLSL.  Therefore, for most ops, it would not be too difficult to provide the option to emit OSL as one of the supported languages.  This would be useful in the context of adding OCIO support to MaterialX, for example (another ASWF project!).

The most difficult issues are how to handle LUTs (which are represented as textures by the GPU path) and dynamic properties (which are handled as uniforms on the GPU).  We have been discussing ideas for how to handle these, but the proposal is that an initial implementation would throw if the color space conversion involves LUTs and that dynamic properties would be set to their initial values (in other words, a new Processor would need to be created to modify them).  Given that the needs for color management within the context of, for example a MaterialX shading network, tend to be less than the needs of a general purpose application, our sense is that these limitations might be acceptable for an initial implementation, but we are interested in feedback on this question.

For reference, tagging issue #261 here, since there is some overlap in some of the discussion around that one.

lgritz:
For LUTs, depending on the size, they can be just arrays in the language (in the process of optimization and JITing, an array in OSL that is only written to in the process of its declaration/initialization will be turned into a static constant, and will not be filled in anew every time the shader runs). Larger LUTs can, of course, just be textures.

I'm not sure what you mean by dynamic properties / uniforms. Can you ELI5? (I should know more internal details and nomenclature of OCIO, I know, but in practice I tend to only directly touch the API calls I need to get a color processor and use it to transform arrays of colors.)

doug-walker:
[...]

PR #1462, opened 2021-08-19
Adsk Contrib - Add preliminary OSL support

hodoulp:

The pull request provides the first step of the native implementation for the OSL translation. At the high level the translation is implemented like any other translations in OCIO in order to keep the single GPU implementation even if some GPU code changes were still needed to accommodate the OSL vs. HLSL/GLSL differences/limitations.

Refer to #1435 for some details / discussions. 

That's a major milestone for OpenColorIO as it enhances potential integrations like in MaterialX environment for example (as discussed in the issue). On the other hand, the OSL library could also benefit from this work by improving color management support in some of the `transform` methods for example.

The OSL translation works for all the existing ops without too many changes compare to the previous writing of the GPU algorithm. So, contributors should not be lost when jumping in the GPU code. The limitations are that the Lut 1D and 3D ops which are still not supported because of the Lut data exchange using OSL (i.e. no easy technical solution for now), and the dynamic properties which are only 'local' variables for short-term. But OSL supports input parameters to any shader types so it can be implemented at some point.

There is now a new unit test framework to support writing unit tests for the OSL translation with two basic unit tests. This framework only supports the OSL shader compilation and lake of 'easy' unit test writing for now. The idea is to slowly enhance the framework and the number of unit tests.

Lastly, the `findOpenShadingLanguage()` method only works with the `OpenShadingLanguage_ROOT` cmake variable (i.e. no auto discovery). Once again, that will be improve over the time to match the other find method capabilities.

This PR also does some clean-up of inconsistent usage of the GPUShaderUtils.


hodoulp:
The generated OSL shader currently processes pixel per pixel using the `color4` OSL type for input and output parameters. For now, the only validation is a successfully compilation (using `oslcomp` library).

Note: The implementation can contain global variables and methods so, I've encapsulated them in the shader method itself (hoping they will be private). 

Below is the template of the OSL shader generated by OpenColorIO.

```
/* All the includes */
  
#include "vector4.h"
#include "color4.h"

/* All the generic helper methods */

vector4 __operator__mul__(vector4 v, matrix m)
{
  return vector4(v.x * m[0][0] + v.y * m[1][0] + v.z * m[2][0] + v.w * m[3][0], v.x * m[0][1] + v.y * m[1][1] + v.z * m[2][1] + v.w * m[3][1], v.x * m[0][2] + v.y * m[1][2] + v.z * m[2][2] + v.w * m[3][2], v.x * m[0][3] + v.y * m[1][3] + v.z * m[2][3] + v.w * m[3][3]);
}

vector4 __operator__mul__(color4 c, vector4 v)
{
  return vector4(c.rgb.r, c.rgb.g, c.rgb.b, c.a) * v;
}

vector4 __operator__mul__(vector4 v, color4 c)
{
  return v * vector4(c.rgb.r, c.rgb.g, c.rgb.b, c.a);
}

vector4 __operator__sub__(color4 c, vector4 v)
{
  return vector4(c.rgb.r, c.rgb.g, c.rgb.b, c.a) - v;
}

vector4 __operator__add__(vector4 v, color4 c) {
  return v + vector4(c.rgb.r, c.rgb.g, c.rgb.b, c.a);
}

vector4 __operator__add__(color4 c, vector4 v) {
  return vector4(c.rgb.r, c.rgb.g, c.rgb.b, c.a) + v;
}

vector4 pow(color4 c, vector4 v) {
  return pow(vector4(c.rgb.r, c.rgb.g, c.rgb.b, c.a), v);
}

vector4 max(vector4 v, color4 c) {
  return max(v, vector4(c.rgb.r, c.rgb.g, c.rgb.b, c.a));
}

/* The shader implementation */

shader OSL_OCIOMain(color4 inColor = {color(0), 1}, output color4 outColor = {color(0), 1})
{
    // Declaration of all helper methods

    <..>

    // Declaration of the OCIO shader function

    color4 OCIOMain(color4 inPixel)
    {
        color4 myColor = inPixel;

        <.. some code here ..>

        return myColor;
    }

    outColor = OCIOMain(inColor);
}

**Comment by hodoulp** [comment]
Following the [2021-08-19 OSL TSC meeting](https://docs.google.com/document/d/1yf0bG6eoE2EvKZBNZX3nskdTvu99ADTDTNOknCDJd1I/edit), here are some hints from @lgritz to improve/complete the OSL unit test framework.

> Patrick has submitted a PR to OCIO where it can generate OSL source as a string to implement any color transformation supported by OCIO. 
For unit tests, LG suggests using either the shade_image() (see end of oslexec.h) or modeling a simple test harness off of the code in `testsuite/example-deformer` as examples. Of course, testshade is also a possibility, but doing in C++ may end up with faster unit tests than constantly shelling out to testshade hundreds of times.

File: share/cmake/modules/FindOpenShadingLanguage.cmake [hunk 1]
```diff
@@ -0,0 +1,80 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
lgritz, line 1: OSL 1.11+ has proper exported cmake configuration files that will set up all these targets. Nobody should need a FindOpenShadingLanguage.cmake.

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -294,6 +296,103 @@ void GpuShaderCreator::createShaderText(const char * shaderDeclarations,
...
+        kw.newLine() << "shader " << "OSL_" << getFunctionName() 
+                     << "(color4 inColor = {color(0), 1}, output color4 outColor = {color(0), 1})";
+        kw.newLine() << "{";
+
+        const std::string str = kw.string() + getImpl()->m_declarations;
+        getImpl()->m_declarations = str;
+    }
+
+    if (getLanguage() == LANGUAGE_OSL)
+    {
```
lgritz, line 386: If I'm reading this correctly, you end the "if OSL" block and then immediately re-enter a new one?

lgritz (commented):
This looks reasonable to me.

I would love to see you not use FindOpenShadingLanguage.cmake, and instead rely on our exported OSLConfig.cmake.

I'm a little concerned about what will happen when we eventually have built-in types for color4/vector4, and I wonder if all the notation will 100% work when we switch to built-in times (for example, I would expect c4.r to work, whereas now you have to awkwardly say c4.rgb.r). Maybe we shouldn't be too concerned, because it's not going to happen especially soon. Maybe I'll just have to ensure that this notation still works. But maybe it's worth building in a little bit of logic to designate this as OSL 1.x, or at least to somehow distinguish the before/after we add the new types?

share/cmake/modules/FindOpenShadingLanguage.cmake
hodoulp: The cmake method also provides a support to link against a custom OSL version using the cmake variable `OpenShadingLanguage_ROOT`. For now that's the only way for OCIO to access OSL.

hodoulp:
> I'm a little concerned about what will happen when we eventually have built-in types for color4/vector4, and I wonder if all the notation will 100% work when we switch to built-in times (for example, I would expect c4.r to work, whereas now you have to awkwardly say c4.rgb.r). Maybe we shouldn't be too concerned, because it's not going to happen especially soon. Maybe I'll just have to ensure that this notation still works. But maybe it's worth building in a little bit of logic to designate this as OSL 1.x, or at least to somehow distinguish the before/after we add the new types?

I changed the OSL enum to be `LANGUAGE_OSL_1` so any later OSL syntax changes can be handled correctly.

Concerning the OSL syntax, I suggest to first make `c4.r` working to ease the readability of any OSL code.

hodoulp:
Below is the command to 'enable' the OSL translation in OCIO:
`cmake -GNinja -DOpenShadingLanguage_ROOT=/Users/hodoulp/dev/osl_1/install_rls -DCMAKE_CXX_STANDARD=14 ../.`

michdolan:
We need to merge this today to release OCIO 2.1.0 in time for VFX Reference Platform 2022. Technically our project process rules require waiting an additional 3 days to merge without a second approval from outside the contributor's organization, but under the circumstances we need to forge ahead sooner. If there are any objections, please comment by Today at 5pm EST. Thank you.

doug-walker (approved):
Per TSC and Slack discussion, approving now.

PR #1463, opened 2021-08-22
PR: Minor Build Fixes

KelSolaar:
This PR adds the `Half` include directory to *ociodisplay* and *ociolutimage* apps `CMakeLists.txt` files.

Without them added explicitly, the apps do not build because the include directories are not added by **CMake** to the respective `<AdditionalIncludeDirectories>` elements in the `*.vcproj` files:

```
[...]\openimageio\2.2.17.0.rf1\[...]\oiio\include\OpenImageIO/Imath.h(21): fatal error C
1083: Cannot open include file: 'OpenEXR/ImathColor.h': No such file or directory [..]\opencolorio\src\apps\ociolutimage\ociolutimage.vcxproj]
```

Note that my _OpenImageIO_ and _OpenEXR/Half_ libraries are provided externally.


hodoulp:
@KelSolaar thanks for the fix.
But I think that you should check against PR #1432 from @michdolan because it completely rewrites the half dependency to use `OpenExr/half.h` or `Imath/half.h`. But that's still valuable for `RB-2.0` i.e. 2.0.2

KelSolaar:
I'm building against the VFX Platform 2021, so indeed, 2.0.x!

hodoulp:
So, you can propose a pull request against `RB-2.0` (instead of `master`) and checks if the proposed pull request from Michael is fixing the Windows / Visual Studio problem.

KelSolaar:
Changing the base will require cherry-picking the commit cleanly on the RB-2.0 branch which I will do when I hit work.

Are you suggesting that I try #1432 onto RB-2.0?

KelSolaar:
@hodoulp: I cherry-picked the commit on top of RB-2.0.

Issue #1296, opened 2021-01-31
OpenColorIO library CMakeLists.txt: Sets up GNUInstallDirs for non-win but completely ignores it in install(TARGETS...


hobbes1069:
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/bc88857f48ec8cdc2d8c669a905805261839d72d/src/OpenColorIO/CMakeLists.txt#L181-L193

and then:

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/bc88857f48ec8cdc2d8c669a905805261839d72d/src/OpenColorIO/CMakeLists.txt#L297-L301

Why go to the trouble of setting up the right directories and then just not use them?

hobbes1069:
I see this is still a problem in 2.1.0. Hardcoding install paths is *BAD*. 

Here's my fix:
```
$ cat ocio-install.patch
Index: OpenColorIO-2.1.0/src/OpenColorIO/CMakeLists.txt
===================================================================
--- OpenColorIO-2.1.0.orig/src/OpenColorIO/CMakeLists.txt
+++ OpenColorIO-2.1.0/src/OpenColorIO/CMakeLists.txt
@@ -293,7 +293,7 @@ endif()
 
 install(TARGETS OpenColorIO
 	EXPORT ${PROJECT_NAME}_EXPORTED_TARGETS
-	LIBRARY DESTINATION lib
-	RUNTIME DESTINATION bin
-	ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
 )
```

remia:
Thanks for the suggestion @hobbes1069, I implemented that, among other things, in #1471 would be great if you could have a look.

hobbes1069:
LGTM!

remia:
@hobbes1069 Would be curious if you have an example where the current hardcoded script fails or produce unexpected results, so that we are aware of what kind of changes we can expect for existing users? Thanks.

hobbes1069:
[...]

PR #1471, opened 2021-08-26
Remove OpenColorIOHeaders target and hardcoded install paths

remia:
Following @lgritz comment in #1397, I think we can remove the `OpenColorIOHeaders` to simplify the CMake scripts a bit. I had some issues with documentation build because I first generated the OpenColorIOABI.h in a new location but resolved this by using the exact same target path as before which I think is the simplest solution.

There shouldn't be a need to merge this for the upcoming release.


hodoulp (approved):

remia:
Updated install paths to use CMake automatic variables as suggested in #1296. Also updated public headers installation to use `PUBLIC_HEADER` target property. Using these new variables could update the install folder structure on Windows, while probably being more "as expected".

hodoulp:
@remia Following the last commit, do you need to update the `CMakeLists.txt` from all the `apps`, all the `libutils`, and the `Python` bindings?

hodoulp:
There is also some hard-coded `lib` to change for the `Python bindings` i.e. src/bindings/python/CMakeLists.txt  at line 229 & 231

remia:
Thanks @hodoulp good point, it seems like we should stop using `${LIB_SUFFIX}` according to https://gitlab.kitware.com/cmake/cmake/-/issues/18640#note_488642.

hodoulp (approved):

hodoulp:
@michdolan The PR is a candidate for `RB-2.1`

hodoulp:
@michdolan or @doug-walker The PR is close to the two-weeks rule, could one of you have a look? Thanks.

doug-walker (approved):

zachlewis:
Oh whoops -- yep, that's the one I was going for.

Issue #1429, opened 2021-07-02
Version handling for 2.1


doug-walker:
Some work is needed on version handling given that OCIO 2.1 will have new FixedFunction and Builtin styles to support the ACES 1.3 Gamut Compression.

The affected file formats are the config and CTF, which are both versioned and the version should be incremented.  The proposal is to increment the version of both from 2.0 to 2.1.  If these files are loaded into the released 2.0.x library, an exception will be thrown (as per existing unit tests).

When writing new files, the 2.1 code should try to write the lowest possible version of each format.  For CTF, this should be very easy since CTFTransform.cpp:GetOpMinimumVersion already implements this functionality and only a few lines of code are needed.  For the config format, this will be more work since similar machinery will need to be added.

PR #1472, opened 2021-08-30
Adsk Contrib - Update the CTF & config version support

hodoulp:

The pull request implements the issue #1429.

The OpenColorIO version 2.1 adds a new `Gamut Compression` algorithm accessible through a fixed function transform and a builtin transform (refer to the pull request #1404 and the issue #1402 for details) so, the config version must now be changed to `2.1`.

A side-effect is also to update the CTF file format version update to handle the new transforms.


michdolan (approved):

doug-walker (approved):
Per TSC and Slack discussion, approving this now because it's important it make it into the 2.1 release.  Given that 2.1 can create configs that have features that did not exist in v2.0, the config version number should be incremented to avoid confusion.

PR #1478, opened 2021-09-02
Adsk Contrib - Better manage Imath dependency

hodoulp:

Using `Imath` in the library could sometime conflict with the `Imath` from `OpenImageIO`. When you try on ubuntu 20.04 using the system `OpenImageIO` the compilation of `OpenColorIO` fails because of symbol clashes on `Imath` (i.e. 3.1.x from OpenColorIO 2.1.x vs. 3.0.x expected by OpenImageIO 2.1.x).

The pull request changes the code to use the OpenImageIO one for apps using OpenImageIO. I also removed some useless `Imath` dependencies.

### DIFF
--- a/share/cmake/modules/FindOpenImageIO.cmake
+++ b/share/cmake/modules/FindOpenImageIO.cmake
@@ -59,6 +59,10 @@ find_library ( OPENIMAGEIO_LIBRARY
                NAMES OpenImageIO${OIIO_LIBNAME_SUFFIX}
                HINTS ${OPENIMAGEIO_ROOT_DIR}
                PATH_SUFFIXES lib64 lib )
+find_library ( OPENIMAGEIO_UTIL_LIBRARY
+               NAMES OpenImageIO${OIIO_LIBNAME_SUFFIX}_Util
+               HINTS ${OPENIMAGEIO_ROOT_DIR}
+               PATH_SUFFIXES lib64 lib )
 find_path ( OPENIMAGEIO_INCLUDE_DIR
             NAMES OpenImageIO/imageio.h
             HINTS ${OPENIMAGEIO_ROOT_DIR} )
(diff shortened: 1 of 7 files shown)


src/apps/ocioconvert/main.cpp
```diff
@@ -13,6 +13,16 @@ namespace OCIO = OCIO_NAMESPACE;
 
 #include <OpenImageIO/imageio.h>
 #include <OpenImageIO/typedesc.h>
+
+// Take the half.h the same way OpenImageIO takes it i.e. do not use the Imath/OpenEXR one from
+// OpenColorIO to avoid version clashes between OpenColorIO & OpenImageIO libraries. For example,
+// OpenColorIO uses Imath 3.1.x but OpenImageIO (from the system) is using Imath 3.0.x which
+// breaks the OpenColorIO compilation.
+#if (OIIO_VERSION >= 20200)
```
hodoulp, line 21: @lgritz, Based on my limited experiments, the include file `<OpenImageIO/Imath.h>` only exists starting with oiio version 2.2. Let me know if I need to change/improve the check.

hodoulp:
I did my tests on `ubuntu 20.04` and `macOS`.

src/apps/ocioconvert/main.cpp
lgritz: That's right. I believe I never backported support for Imath 3.x to OIIO 2.1.

lgritz: I never know the right answer to this problem. In practice, I'm very fastidious about making sure that the "current" release family (2.2 until yesterday, now 2.3!) is constantly kept able to build with the latest compilers and dependencies. But once a few months have elapsed since a big release, I tend to only backport things like to the "obsolete" branches if somebody requests it. Since VFX Platform 2021 is still on OpenEXR 2.4 (!), VP2022 will be the first to dictate OpenEXR/Imath >= 3.0, the popular demand for that backport to 2.1 has not materialized.

lgritz (approved):
lgtm

hodoulp:
The PR is a candidate for RB-2.1

share/cmake/modules/FindOpenImageIO.cmake
```diff
@@ -59,6 +59,10 @@ find_library ( OPENIMAGEIO_LIBRARY
                NAMES OpenImageIO${OIIO_LIBNAME_SUFFIX}
                HINTS ${OPENIMAGEIO_ROOT_DIR}
                PATH_SUFFIXES lib64 lib )
+find_library ( OPENIMAGEIO_UTIL_LIBRARY
+               NAMES OpenImageIO${OIIO_LIBNAME_SUFFIX}_Util
```
remia, line 63: Looking at https://github.com/OpenImageIO/oiio/blob/master/src/libutil/CMakeLists.txt#L49, should this be `OpenImageIO_Util${OIIO_LIBNAME_SUFFIX}` ?

hodoulp: Good catch.

doug-walker (approved):
Approving immediately since this bug fix is needed to unbreak the CI builds.

PR #1464, opened 2021-08-23
Add virtual display python bindings

michdolan:
Adds missing Python bindings for OCIO Config virtual display interface.

Fixes #1317

PR #1481, opened 2021-09-05
ConfigTest - Virtual display python unit test

meimchu:
ConfigTest: Python unit test updated with additional virtual display unit tests. Accompanying features introduced in #1464

Follow up questions and comments:
- I would like to ask for some guidance on the cpp part of the  unit tests which deal with the different OS platforms. What would be the right way to implement that logic in the python side? Thanks!
- I have left the `test_virtual_display()` method alone for now. Let me know if you think my new ConfigVirtualDisplayTest provides sufficient coverage so that can be removed.
- I mainly followed closely to the virtual display-related unit tests I saw in the cpp unit tests per suggestion. There are some still more "traditional" testings that I can add to it later as well.

Signed-off-by: Mei Chu <meimchu@gmail.com>


tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
+        self.assertEqual('<USE_DISPLAY_NAME>',
+                         self.cfg.getVirtualDisplayViewColorSpaceName(view_name))
+        self.assertEqual('', self.cfg.getVirtualDisplayViewLooks(view_name))
+        self.assertEqual('', self.cfg.getVirtualDisplayViewRule(view_name))
+        self.assertEqual(
+            '', self.cfg.getVirtualDisplayViewDescription(view_name))
+
+    def test_get_virtual_display_views_shared(self):
+        views = self.cfg.getVirtualDisplayViews(OCIO.VIEW_SHARED)
+        self.assertEqual(views.__len__(), 1)
```
hodoulp, line 929: [minor] Why not `len(views)`?

tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
+        view_name = views[0]
```
hodoulp, line 930: You can simplify the two next lines `self.assertEqual(views[0], 'sview2')`

tests/python/ConfigTest.py
hodoulp: The comment also applies below.

tests/python/ConfigTest.py
hodoulp: The comment also applies below.

tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
+
+        # Test remove a shared view from the Virtual Display.
+
+        self.cfg.removeVirtualDisplayView('sview2')
+        views = self.cfg.getVirtualDisplayViews(OCIO.VIEW_DISPLAY_DEFINED)
+        self.assertEqual(views.__len__(), 1)
+        views = self.cfg.getVirtualDisplayViews(OCIO.VIEW_SHARED)
+        self.assertEqual(views.__len__(), 0)
+
+        # Extra serialize & deserialize validation.
```
hodoulp, line 957: Add a empty line between the comment and the code like you did at line 950

tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
+        """
+
+        self.cfg.clearVirtualDisplay()
+        views = self.cfg.getVirtualDisplayViews(OCIO.VIEW_DISPLAY_DEFINED)
+        self.assertEqual(views.__len__(), 0)
+        views = self.cfg.getVirtualDisplayViews(OCIO.VIEW_SHARED)
+        self.assertEqual(views.__len__(), 0)
+
+        # Extra serialize & deserialize validation.
+        cfg = OCIO.Config().CreateFromStream(self.cfg.serialize())
```
hodoulp, line 982: Same as line 957

tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
+        views = cfg.getVirtualDisplayViews(OCIO.VIEW_DISPLAY_DEFINED)
+        self.assertEqual(views.__len__(), 0)
+        views = cfg.getVirtualDisplayViews(OCIO.VIEW_SHARED)
+        self.assertEqual(views.__len__(), 0)
+
+    def test_virtual_display_with_active_displays(self):
```
hodoulp, line 988: The code now switches to another config file which is confusing because the previous methods were all related to another config file. To make it clear that it's now a new config to test, could you create another test case?

tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
+            cfg.validate()
+        self.assertEqual(str(cm.exception), "Config failed validation. " +
+                         "Display 'virtual_display' has a " +
+                         "view 'Raw1' that refers to a color space" +
+                         " or a named transform, 'raw1', which is not defined.")
+        
+        cfg.removeVirtualDisplayView('Raw1')
+        cfg.validate()
+
+        cfg.addVirtualDisplayView('Raw1', 'raw', 'look')
```
hodoulp, line 1055: It should be  `cfg.addVirtualDisplayView('Raw1', 'Film', 'raw1', 'look')` i.e. look is the fourth argument.

hodoulp (changes requested):
Good job.
There is few minor code improvements but the core is here.

Concerning the platform specific tests I suggest to forget them for short-term because it requires to find out the platform (possible with Python) but also having the headless mode i.e. `OCIO_HEADLESS_ENABLED`. The latter is currently not available for Python unit tests.
That could be a 'follow-up' branch.

tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
         self.assertEqual(
-            0, 
+            0,
             len(cfg.getVirtualDisplayViews(OCIO.VIEW_DISPLAY_DEFINED)))
         self.assertEqual(0, len(cfg.getVirtualDisplayViews(OCIO.VIEW_SHARED)))
+
+
+class ConfigVirtualDisplayTest(unittest.TestCase):
+    def setUp(self):
+        self.cfg = OCIO.Config().CreateFromStream(SIMPLE_CONFIG_VIRTUAL_DISPLAY)
```
michdolan, line 862:
```suggestion
        self.cfg = OCIO.Config.CreateFromStream(SIMPLE_CONFIG_VIRTUAL_DISPLAY)
```
`CreateFromStream` is a static method.

michdolan: Same comment below

tests/python/ConfigTest.py
```diff
@@ -773,83 +775,287 @@ def test_virtual_display(self):
...
+
+    def test_virtual_display_v1(self):
+        """
+        Test that the virtual display is only supported by v2 or higher.
+        """
+
+        with self.assertRaises(OCIO.Exception):
+            cfg = OCIO.Config().CreateFromStream(SIMPLE_CONFIG_VIRTUAL_DISPLAY_V1)
+
+        cfg = OCIO.Config().CreateRaw()
```
michdolan, line 1008: This is also a static method. This applies to all `OCIO.Config.Create*` methods used in this module.

michdolan:
Thanks for the contribution @meimchu !

remia:
(Re-launched the Wheel jobs, there was an issue with Chocolatey failing to install the `doxygen` package. I already saw that a few time and have no real idea on where it comes from but most likely an issue on their side.)

tests/python/ConfigTest.py
```diff
@@ -71,34 +74,34 @@
...
-#"""
-#return out_pixel;
-#}
+# """
+# return out_pixel;
+# }
 #
-#"""
+# """
 #
```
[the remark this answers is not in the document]
meimchu: I just simply ran a formatter at the end 😅 No harm done for consistency, right?

tests/python/ConfigTest.py
```diff
@@ -544,7 +551,7 @@ def test_ruled_views(self):
         self.assertEqual('View_c', next(views))
         # View_d rule is Rule_3 that lists c3.
         self.assertEqual('View_d', next(views))
-        #/ View_e rule is Rule_4 that lists c3.
+        # / View_e rule is Rule_4 that lists c3.
         self.assertEqual('View_e', next(views))
```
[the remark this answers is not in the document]
meimchu: Addressed!

tests/python/ConfigTest.py
```diff
@@ -336,7 +339,8 @@ def test_shared_views(self):
         self.assertEqual('', cfg.getDisplayViewDescription('', 'view2'))
         self.assertEqual('', cfg.getDisplayViewDescription('', 'view3'))
         self.assertEqual('', cfg.getDisplayViewDescription('', 'view4'))
-        self.assertEqual('description 5', cfg.getDisplayViewDescription('', 'view5'))
+        self.assertEqual(
+            'description 5', cfg.getDisplayViewDescription('', 'view5'))
         self.assertEqual('desc6', cfg.getDisplayViewDescription('', 'view6'))
```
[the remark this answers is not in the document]
meimchu: Reversed!

michdolan (approved):

PR #1483, opened 2021-09-05
FindExpat: allow usage of expat 2.2.9

### LINKED ISSUE #1482: OCIO requires Expat 2.2.8+, but its Find module only works for Expat 2.2.10 and higher
Hi!

I'm working on deploying OpenColorIO 2 support in Krita. I just found that OCIO's FindExpat doesn't detect correctly the version we ship as part of our 3rdparty dependency builds, 2.2.9.

This is because of two separate bugs:
- in https://github.com/libexpat/libexpat/commit/da7cda3a8579e86bb553e331cc4747075307f907 upstream renamed the target to `expat::expat`, while 2.2.9 and lower have `expat::libexpat`. OCIO requires 2.2.8 as a minimum, but doesn't account for this.
- furthermore, FindExpat mixes CMake and pkg-config style of include directories. pkg-config's use `expat_INCLUDE_DIR` via `find_path` on expat.h, but CMake's provides the same via `expat_INCLUDE_DIRS` (note the plural).

This was tested with OCIO 2.1.0, and CMake 3.21.1 on Windows 10 with MSVC 19.29.30133.

amyspark:
Fixes #1482 by:

- aliasing expat::expat to expat::libexpat if the former is missing
- patching INTERFACE_INCLUDE_DIRECTORIES if the property is missing

### DIFF
--- a/share/cmake/modules/Findexpat.cmake
+++ b/share/cmake/modules/Findexpat.cmake
@@ -34,7 +34,40 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
     endif()
 
     if(expat_FOUND)
+        if (TARGET expat::libexpat)
+            message(STATUS "Expat ${expat_VERSION} detected, aliasing targets.")
+            add_library(expat::expat ALIAS expat::libexpat)
+        endif()
+
+        get_target_property(expat_INCLUDE_DIR expat::expat INTERFACE_INCLUDE_DIRECTORIES)
+
         get_target_property(expat_LIBRARY expat::expat LOCATION)
+
+        if (NOT expat_INCLUDE_DIR)
+            # Find include directory too, as its Config module doesn't include it
+            find_path(expat_INCLUDE_DIR
+                NAMES
+                    expat.h
+                HINTS
+                    ${expat_ROOT}
+                    ${PC_expat_INCLUDE_DIRS}
+                PATH_SUFFIXES
+                    include
+                    expat/include
+            )
+            message(WARNING "Expat's include directory not specified in its Config module, patching it now to ${expat_INCLUDE_DIR}.")
+            if (TARGET expat::libexpat)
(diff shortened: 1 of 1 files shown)


hodoulp (approved):
Thanks @amyspark for the fix. We completely missed the target name change & the missing include variable.

amyspark:
Added the hotpatch of the include in all targets since it's also missing on `expat::expat` (which MinGW uses).

hodoulp (approved):

share/cmake/modules/Findexpat.cmake
```diff
@@ -34,7 +34,41 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
...
+                    expat.h
+                HINTS
+                    ${expat_ROOT}
+                    ${PC_expat_INCLUDE_DIRS}
+                PATH_SUFFIXES
+                    include
+                    expat/include
+            )
+            message(WARNING "Expat's include directory not specified in its Config module, patching it now to ${expat_INCLUDE_DIR}.")
+            if (TARGET expat::libexpat)
```
remia, line 60: Minor: Is the if / else required given that we create an alias above ? Same remark for the `get_target_property` on line 40 & 42.

hodoulp:
Friendly reminder that the pull request could now be merged next Monday.

hodoulp:
Done, with success.

hodoulp:

michdolan (approved):

amyspark:
Thanks for merging and for the reviews ! Do you think this patch should go in 2.0 and 2.1 as well?

hodoulp:
I've already made the pull request for 2.1. But I did not plan to make one for 2.0 as I only integrate critical bug fixes in that release. If you think that it is an emergency or blocker for 2.0 users, please make your point so we can evaluate it.

Thanks for your help in this issue.

PR #1485, opened 2021-09-07
Add support for MinGW

amyspark:
In these cases the build system doesn't add a suffix to yaml-cpp.

Additionally, the `L_tmpnam_s` define has been added only last year in MinGW. 7.3 (the last official release used by Qt 5) doesn't have it, so let's import the define from the official patch.

### DIFF
--- a/src/OpenColorIO/Platform.cpp
+++ b/src/OpenColorIO/Platform.cpp
@@ -174,6 +174,12 @@ std::string CreateTempFilename(const std::string & filenameExt)
 
     // Note: Because of security issue, tmpnam could not be used.
 
+    // Note 2: MinGW doesn't define L_tmpnam_s, we need to patch it in.
+    // https://www.mail-archive.com/mingw-w64-public@lists.sourceforge.net/msg17360.html
+#if !defined(L_tmpnam_s)
+#define L_tmpnam_s L_tmpnam
+#endif
+
     char tmpFilename[L_tmpnam_s];
     if(tmpnam_s(tmpFilename))
     {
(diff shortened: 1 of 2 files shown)


src/OpenColorIO/Platform.cpp
```diff
@@ -174,6 +174,12 @@ std::string CreateTempFilename(const std::string & filenameExt)
 
     // Note: Because of security issue, tmpnam could not be used.
 
+    // Note 2: MinGW doesn't define L_tmpnam_s, we need to patch it in.
+    // https://www.mail-archive.com/mingw-w64-public@lists.sourceforge.net/msg17360.html
+#if !defined(L_tmpnam_s)
+#define L_tmpnam_s L_tmpnam
```
remia, line 180: Question: Looking at the link, should we add `#define TMP_MAX_S TMP_MAX` too ?

remia:
Thanks for the fix @amyspark. Looking quickly at other Find*.cmake modules, I wonder if the same update would be required elsewhere? Maybe that could be checked by compiling with `-DOCIO_INSTALL_EXT_PACKAGES=ALL`?

amyspark:
> 
> 
> Thanks for the fix @amyspark. Looking quickly at other Find*.cmake modules, I wonder if the same update would be required elsewhere? Maybe that could be checked by compiling with `-DOCIO_INSTALL_EXT_PACKAGES=ALL`?

I compile using `MISSING` because we already ship OpenEXR and Expat; imath, pystring and yaml-cpp are OK with this patch.

amyspark:
The build error looks spurious:

```
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: CMakeFiles/ociolutimage.dir/main.cpp.o: undefined reference to symbol '_ZN16OpenImageIO_v2_310ParamValue11clear_valueEv'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: /usr/local/lib64/libOpenImageIO_Util.so.2.3: error adding symbols: DSO missing from command line
```

Can anyone retry the action?

hodoulp:
The fix is in PR #1478 that I will try to merge soon.

hodoulp (approved):

hodoulp:

PR #1487, opened 2021-09-07
A few fixes and proposed changes.

patlefort:
* Remove starting backslash if present from temporary filename. A starting backslash mean that it is safe to create it in the current working directory.
* Use mkstemps to generate a temporary file on non-Windows systems. This function will create a temporary file with a randomized name with proper permissions.
* Rename CreateTempFilename to CreateTempFile, since mkstemps will create the file.
* Remove GenerateRandomNumber which is no longer needed.
* Fix add_test in cmake files. This notation allows it to understand that the command portion is a target and translate it to the proper executable.
* Add guard to remove temporary file in create_temp_filename test.

### DIFF
--- a/src/OpenColorIO/Platform.cpp
+++ b/src/OpenColorIO/Platform.cpp
@@ -11,6 +11,9 @@
 
 #ifndef _WIN32
 #include <strings.h>
+#else
+#include <cstdlib>
+#include <cerrno>
 #endif
 
 
@@ -151,22 +154,7 @@ void AlignedFree(void* memBlock)
 #endif
 }
 
-namespace
-{
-
-int GenerateRandomNumber()
-{
-    // Note: Read https://isocpp.org/files/papers/n3551.pdf for details.
-
-    static std::mt19937 engine{};
-    static std::uniform_int_distribution<> dist{};
-
-    return dist(engine);
-}
-
-}
-
-std::string CreateTempFilename(const std::string & filenameExt)
+std::string CreateTempFile(const std::string & filenameExt)
 {
     std::string filename;
 
@@ -180,19 +168,20 @@ std::string CreateTempFilename(const std::string & filenameExt)
         throw Exception("Could not create a temporary file.");
     }
 
-    filename = tmpFilename;
+    filename = tmpFilename[0] == '\\' ? tmpFilename + 1 : tmpFilename;
+    filename += filenameExt;
 
 #else
 
     // Linux flavors must have a /tmp directory.
-    std::stringstream ss;
-    ss << "/tmp/ocio_" << GenerateRandomNumber();
-
-    filename = ss.str();
+    filename = "/tmp/ocio_XXXXXX" + filenameExt;
+    
+    if(mkstemps(filename.data(), filenameExt.size()) == -1)
+    {
+        throw std::system_error(errno, std::system_category(), "Could not create a temporary file.");
+    }
 
 #endif
-
-    filename += filenameExt;
 
     return filename;
 }
--- a/tests/cpu/Platform_tests.cpp
+++ b/tests/cpu/Platform_tests.cpp
@@ -154,7 +154,17 @@ OCIO_ADD_TEST(Platform, create_temp_filename)
     std::set<std::string> uids;
     for (size_t idx=0; idx<TestMax; ++idx)
     {
-        uids.insert(OCIO::Platform::CreateTempFile(""));
+        struct FGuard
+        {
+            ~FGuard()
+            {
+                std::remove(m_filename.c_str());
+            }
+            
+            std::string m_filename;
+        } file{ OCIO::Platform::CreateTempFile("") };
+        
+        uids.insert(file.m_filename);
     }
 
     // Check that it only generates unique random strings.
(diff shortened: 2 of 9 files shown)


patlefort:
I'm getting a 502 Internal server error when I try to get authorized with EasyCLA.

hodoulp:
Sometimes the link to the EasyCLA registration fails but retry later.

src/OpenColorIO/Platform.cpp
```diff
@@ -180,20 +168,21 @@ std::string CreateTempFilename(const std::string & filenameExt)
...
 #else
 
     // Linux flavors must have a /tmp directory.
-    std::stringstream ss;
-    ss << "/tmp/ocio_" << GenerateRandomNumber();
-
-    filename = ss.str();
+    filename = "/tmp/ocio_XXXXXX" + filenameExt;
+    
+    if(mkstemps(filename.data(), filenameExt.size()) == -1)
```
hodoulp, line 179: The change is major as the method `mkstemps` now generates a temporary file name, creates the file on disk, opens it and returns the file descriptor. I mainly have two concerns:
1. The implementation behavior is now different between `Linux` and `Windows` which is confusing.
2. The file creation & opening is clearly not always needed. You can check `OCIO_ADD_TEST(Platform, create_temp_filename)` and  `OCIO_ADD_TEST(CDLTransform, escape_xml)` as example where these steps are useless.

hodoulp (changes requested):

src/OpenColorIO/Platform.cpp
patlefort: The implementation could be made the same on both platforms. What I'd do is make it return an std::fstream. I think it would be good to also test creation and opening of these files. It could also be more secure because mkstemps will assign permission of 0600 on the file. It will also open the file with O_EXCL flag so opening will fail if it already exists, safer and one less thing to think about. The only trouble is to pass that file descriptor to a fstream, there is no standard way.

In any case, if it's too much of a change, then nevermind.

remia: Looks like we could have `CreateTempFilename` left as is and add `CreateTempFile` next to it ? Agree that the Windows path need to create a file as well.

tests/gpu/GPUHelpers.cpp
```diff
@@ -18,7 +19,7 @@
 namespace OCIO = OCIO_NAMESPACE;
 
 
-// TODO: Make OCIO::Platform::CreateTempFilename() public so it could be used here.
+// TODO: Make OCIO::Platform::CreateTempFile() public so it could be used here.
```
remia, line 22: Why CDLTransform_tests.cpp is able to use `OCIO::Platform::CreateTempFilename()` but not here ?

tests/cpu/Platform_tests.cpp
```diff
@@ -154,7 +154,17 @@ OCIO_ADD_TEST(Platform, create_temp_filename)
...
-        uids.insert(OCIO::Platform::CreateTempFilename(""));
+        struct FGuard
+        {
+            ~FGuard()
+            {
+                std::remove(m_filename.c_str());
+            }
+            
+            std::string m_filename;
+        } file{ OCIO::Platform::CreateTempFile("") };
```
remia, line 165: If we were to have two functions `CreateTempFile` and `CreateTempFileName` we could just use the latter and remove the guard I guess.

tests/gpu/GPUHelpers.cpp
hodoulp: The CPU unit tests are built using directly the cpp files from `OpenColorIO` to deeply test all classes/methods private to the the library (i.e. white box tests). That's critical as the CPU implementation is the reference to the other language translations. So, the GPU unit tests are built on top of the `OpenColorIO` public API (i.e. black box tests) and uses the CPU to validate the processing results. Note that a coming pull request implements the OSL translation validation using the same technic.

Currently the `OCIO::Platform::CreateTempFilename()` method is private. And the comment correctly explains the solution to be able using it here. I suggest to do it as the pull request revisits the implementation method.

src/OpenColorIO/Platform.cpp
hodoulp: Here is a summary of the requirements:
- The behavior must be exactly the same between the different platforms.
- The only usages is to create a temporary file with a predefined content (it implies to later automatically delete the file).
   - The only usages are currently in CPU unit tests for `CDL` tests and in the GPU unit tests for 3D LUT tests. (The test in `Platform_tests.cpp` is not a usage because it tests the method itself).
- To use the method in the GPU unit test, the method must be public but the std::stream's are strictly forbidden in the public API (in fact all std classes/methods are forbidden)

So, I can only see two approaches:
1. _Public method:_ Improve the existing method (it only creates a filename) and make it public (to refactor the GPU method `createTempFile()`).
2. _Public class:_ Create a public `TempFile` class which creates a temporary file with a predefined content and automatically deletes the temporary file in the destructor (a little bit like the `FileGuard` in CLDTransform_tests.cpp).

Note:
> Looks like we could have CreateTempFilename left as is and add CreateTempFile next to it ?

In my mind that's similar to the class proposal because the file temporary name is only needed to create a file, and the temporary file deletion must still be consider.

hodoulp (changes requested):

PR #1489, opened 2021-09-09
Adsk contrib - Various improvements 

hodoulp:

The pull request contains:

- Improvement to the `ctest` support to allow `ctest -T memcheck` on `Linux` (`valgrind` must be installed first).
- Improvement to the OpenImageIO support i.e. when version greater or equal than 2.3, OpenColorIO must compile in C++14.
- Use 'ctest -T memcheck' to reproduce Maya memory issue findings without success and it reports tons of false positives. So, I only improve the code around suspicious codes (detected by Maya) because I found no obvious errors.


CMakeLists.txt
```diff
@@ -314,3 +314,12 @@ install(
...
     DESTINATION "${OCIO_CONFIG_INSTALL_DIR}"
 )
+
+
+###############################################################################
+# ctest
+
+# To correctly generate the DartConfiguration.tcl, include the CTest CMake file.
+# By doing so 'ctest -F memcheck' works.
+include (CTest)
```
remia, line 324: [Minor] Just noting that we may move that include up and get rid of the `enable_testing()` call. Could also have this include depend on whether `OCIO_BUILD_TESTS` is `ON` or` OFF` ? I don't think there is issues here though, it just seems a bit of an unusual CMake usage.

hodoulp: Not a `cmake` expert so I follow your recommendation.

hodoulp:

doug-walker (approved):

hodoulp:
Hi @michdolan, @remia , could one of you have a look? Thanks.

CMakeLists.txt
remia: Sorry for having second thoughts on this @hodoulp, but could we add again `enable_testing()` just after the `include (CTest)` ? I think we can now have case where `OCIO_BUILD_TESTS=ON` but the built-in CMake `BUILD_TESTING=OFF` so that the end result is we don't build the tests. I feel like the OCIO variable should always take precedence here.

remia (approved):

PR #1493, opened 2021-09-15
Add Python GPU renderer implementation example

michdolan:
From discussion in Slack #dev channel about how to implement the OCIO GPU renderer in Python, this is a trimmed down version of an image viewer I wrote while stress testing the OCIO v2 Python bindings last year. It implements the OCIO GPU renderer with dynamic parameters for exposure and gamma adjustment. The goal here is to show just the pieces that need to be implemented for a functional and performant viewer GUI, which is quite a bit of code, but demonstrates the OCIO API in context, along with all the glue code needed to interface with the other Python libraries used here.

As @hodoulp mentioned in Slack, we should reimplement `oglapphelpers` in Python, which much of the rendering code in this module was adapted from. This module would be greatly reduced with that, leaving mostly Qt GUI code, but in the meantime this can serve as a working example.


KelSolaar:
Fantastic, this is extremely useful!

I hope we can also discuss some level of abstraction so that for a end user it can be as simple as calling `apply` on the GPU processor.

hodoulp:
[Question] Why not creating an apps (like `ociocheck`, `ociodisplay`, etc.) ?
So, the `Python` app could be improved when `oglapphelpers.py` library is ready and also installed like any other apps. My concern is that an 'not used' example becomes deprecated at some point like an example in the documentation for example.

michdolan:
I could rebrand this example as pyociodisplay and move it to the apps folder, refining it over time.

remia:
Thanks for sharing @michdolan ! I'm working on having it work on macOS, will report when succeeded. Would be great for it to support all 3 major platforms.

share/python/examples/gpu_viewer.py
```diff
@@ -0,0 +1,1367 @@
...
+    @QtCore.Slot(float)
+    def _on_exposure_changed(self, value):
+        self.image_plane.update_exposure(value)
+
+    @QtCore.Slot(float)
+    def _on_gamma_changed(self, value):
+        self.image_plane.update_gamma(value)
+
+
+if __name__ == "__main__":
```
remia, line 1363: Because you need to be on the Core Profile on macOS to access the programmatic pipeline, I did update the default format by using:

```
format = QtOpenGL.QGLFormat()
format.setProfile(QtOpenGL.QGLFormat.CoreProfile)
format.setSampleBuffers(True)
format.setSwapInterval(1)
format.setVersion(3, 2)
QtOpenGL.QGLFormat.setDefaultFormat(format)
```

Might need a platform conditional if that breaks things on other platform, I didn't check.

share/python/examples/gpu_viewer.py
```diff
@@ -0,0 +1,1367 @@
...
+            0,
+            GL.GL_RGBA,
+            GL.GL_FLOAT,
+            ctypes.c_void_p(0)
+        )
+
+        GL.glTexParameteri(
+            GL.GL_TEXTURE_2D,
+            GL.GL_TEXTURE_WRAP_S,
+            GL.GL_CLAMP
```
remia, line 188: Replacing by `GL.GL_CLAMP_TO_EDGE` to fix usage under the Core Profile (`GL_CLAMP` seems to be deprecated in 3.0 and removed in 3.2).

zachlewis (approved):
This is *so* *helpful*. Thanks guys. This dovetails really nicely with aspects of @hodoulp's and @michdolan's Siggraph presentations. I'd been struggling a little with figuring out what I needed to figure out vis a vis the bare minimum OpenGL + Qt to realize the python OCIO GPU image viewer of my dreams... and it looks like, in retrospect, I probably would have been better served by just biting the bullet and grokking oglapphelpers and ociodisplay. 

It's really great to see how all the pieces fit together -- I have a whole new level of appreciation for the GPU processor design and API. For example, and maybe this is intuitive to those with more experience with OpenGL, but I had no idea the GPU path represented large Lut1Ds as 2D textures. Damn clever, and probably something that would have caused me grief if left to my own devices. 

Anyway, thanks! Confirmed working quite well on macOS with Remi's tweaks (and imath bindings).

hodoulp:
@michdolan This pull request is the next one to be merged. But there is still the question about the location, and a second one (at my point of view) would be to document how to run it (or create a CMakeLists.txt with dependencies).

michdolan:
@hodoulp I will move this and add a README.

PR #1496, opened 2021-09-20
Use the fast_float library to parse floating-point numbers

### LINKED ISSUE #297: OCIO fails parsing configs with non-C locale

### LINKED ISSUE #379: Error when loading View LUTs depending on locale setting

### LINKED ISSUE #1322: Make Numeric read/write locale independent
For the last few years, the locale introduced some challenges to any OpenColorIO integration as the library does not recommend or enforce any locale requirement(s). So, the float number read could fail if the floats are saved like `10,0` instead of `10.0` (i.e. French versus English notations). For example using a French localized DCC to save a LUT file and/or a config file, and later use an English localized DCC to read one of them is the use case to correctly handle.  

Following the discussion at a TSC meeting, the proposal is then to at least enforce an English float notation independently from the host locale, to read/write all numerics.

Over the years, several pull requests were created around that problem: #1089, #297, etc.

amyspark:
This MR brings in the amalgamated header of [the fast_float library](https://github.com/fastfloat/fast-float), and applies it to our uses of sscanf with %f. This library is locale-free, and does not incur the performance penalty of using a thread locale RAII class.

I have also included the relevant entries to THIRD_PARTY, and added a succint Readme listing the library version and its source, in the style of sampleicc.

Fixes #297
Fixes #379
Fixes #1322

### DIFF
--- a/share/cmake/utils/CompilerFlags.cmake
+++ b/share/cmake/utils/CompilerFlags.cmake
@@ -25,6 +25,10 @@ if(USE_MSVC)
     # Note: Do not use /Wall (i.e. /W4) which adds 'informational' warnings.
     set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} /W3")
 
+    # Do enable C4701 (Potentially uninitialized local variable 'name' used), which is level 4.
+    # This is because strtoX-based from_chars leave the value variable unmodified.
+    set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} /we4701")
+
     if(OCIO_WARNING_AS_ERROR)
         set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} /WX")
     endif()
--- a/src/OpenColorIO/ParseUtils.cpp
+++ b/src/OpenColorIO/ParseUtils.cpp
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 // Copyright Contributors to the OpenColorIO Project.
 
+#include <cmath>
 #include <cstring>
 #include <iostream>
 #include <set>
@@ -11,6 +12,7 @@
 #include "ParseUtils.h"
 #include "Platform.h"
 #include "utils/StringUtils.h"
+#include "utils/NumberUtils.h"
 
 
 namespace OCIO_NAMESPACE
@@ -526,6 +528,7 @@ static constexpr int DOUBLE_DECIMALS = 16;
 std::string FloatToString(float value)
 {
     std::ostringstream pretty;
+    pretty.imbue(std::locale::classic());
     pretty.precision(FLOAT_DECIMALS);
     pretty << value;
     return pretty.str();
@@ -536,6 +539,7 @@ std::string FloatVecToString(const float * fval, unsigned int size)
     if(size<=0) return "";
 
     std::ostringstream pretty;
+    pretty.imbue(std::locale::classic());
     pretty.precision(FLOAT_DECIMALS);
     for(unsigned int i=0; i<size; ++i)
     {
@@ -550,9 +554,9 @@ bool StringToFloat(float * fval, const char * str)
 {
     if(!str) return false;
 
-    std::istringstream inputStringstream(str);
-    float x;
-    if(!(inputStringstream >> x))
+    float x = NAN;
+    const auto result = NumberUtils::from_chars(str, str + strlen(str), x);
+    if (result.ec != std::errc())
     {
         return false;
     }
@@ -567,6 +571,7 @@ bool StringToInt(int * ival, const char * str, bool failIfLeftoverChars)
     if(!ival) return false;
 
     std::istringstream i(str);
+    i.imbue(std::locale::classic());
     char c=0;
     if (!(i >> *ival) || (failIfLeftoverChars && i.get(c))) return false;
     return true;
@@ -575,6 +580,7 @@ bool StringToInt(int * ival, const char * str, bool failIfLeftoverChars)
 std::string DoubleToString(double value)
 {
     std::ostringstream pretty;
+    pretty.imbue(std::locale::classic());
     pretty.precision(DOUBLE_DECIMALS);
     pretty << value;
     return pretty.str();
@@ -585,6 +591,7 @@ std::string DoubleVecToString(const double * val, unsigned int size)
     if (size <= 0) return "";
 
     std::ostringstream pretty;
+    pretty.imbue(std::locale::classic());
     pretty.precision(DOUBLE_DECIMALS);
     for (unsigned int i = 0; i<size; ++i)
     {
@@ -601,9 +608,10 @@ bool StringVecToFloatVec(std::vector<float> &floatArray, const StringUtils::Stri
 
     for(unsigned int i=0; i<lineParts.size(); i++)
     {
-        std::istringstream inputStringstream(lineParts[i]);
-        float x;
-        if(!(inputStringstream >> x))
+        float x = NAN;
+        const char *str = lineParts[i].c_str();
+        const auto result = NumberUtils::from_chars(str, str + lineParts[i].size(), x);
+        if (result.ec != std::errc())
         {
             return false;
         }
--- a/src/OpenColorIO/fileformats/FileFormatHDL.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatHDL.cpp
@@ -37,6 +37,7 @@
 #include "ParseUtils.h"
 #include "transforms/FileTransform.h"
 #include "utils/StringUtils.h"
+#include "utils/NumberUtils.h"
 
 
 namespace OCIO_NAMESPACE
@@ -182,16 +183,11 @@ readLuts(std::istream& istream,
         }
         else if(inlut)
         {
-            // StringToFloat was far slower, for 787456 values:
-            // - StringToFloat took 3879 (avg nanoseconds per value)
-            // - stdtod took 169 nanoseconds
-            char* endptr = 0;
-            float v = static_cast<float>(strtod(word.c_str(), &endptr));
+            float v{};
+            const auto result = NumberUtils::from_chars(word.c_str(), word.c_str() + word.size(), v);
 
-            if(!*endptr)
+            if (result.ec == std::errc())
             {
-                // Since each word should contain a single
-                // float value, the pointer should be null
                 lutValues[lutname].push_back(v);
             }
             else
--- a/src/OpenColorIO/fileformats/FileFormatSpi1D.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatSpi1D.cpp
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: BSD-3-Clause
 // Copyright Contributors to the OpenColorIO Project.
 
+#include <cmath>
 #include <cstdio>
+#include <cstring>
 #include <sstream>
 
 #include <OpenColorIO/OpenColorIO.h>
@@ -13,7 +15,7 @@
 #include "Platform.h"
 #include "transforms/FileTransform.h"
 #include "utils/StringUtils.h"
-
+#include "utils/NumberUtils.h"
 
 /*
 Version 1
@@ -124,9 +126,25 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
             }
             else if(StringUtils::StartsWith(headerLine, "From"))
             {
-                if (sscanf(lineBuffer, "From %f %f", &from_min, &from_max) != 2)
+                char fromMinS[64] = "";
+                char fromMaxS[64] = "";
+#ifdef _WIN32
+                if (sscanf_s(lineBuffer, "From %s %s", fromMinS, 64, fromMaxS, 64) != 2)
+#else
+                if (sscanf(lineBuffer, "From %s %s", fromMinS, fromMaxS) != 2)
+#endif
                 {
                     ThrowErrorMessage("Invalid 'From' Tag", currentLine, headerLine);
+                }
+                else
+                {
+                    const auto fromMinAnswer = NumberUtils::from_chars(fromMinS, fromMinS + 64, from_min);
+                    const auto fromMaxAnswer = NumberUtils::from_chars(fromMaxS, fromMaxS + 64, from_max);
+
+                    if (fromMinAnswer.ec != std::errc() || fromMaxAnswer.ec != std::errc())
+                    {
+                        ThrowErrorMessage("Invalid 'From' Tag", currentLine, headerLine);
+                    }
                 }
             }
             else if(StringUtils::StartsWith(headerLine, "Components"))
@@ -179,7 +197,6 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
 
         int lineCount=0;
 
-        StringUtils::StringVec inputLUT;
         std::vector<float> values;
 
         while (istream.good())
@@ -192,10 +209,17 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
 
             if (line.length() != 0)
             {
-                inputLUT = StringUtils::SplitByWhiteSpaces(StringUtils::Trim(lineBuffer));
                 values.clear();
-                if (!StringVecToFloatVec(values, inputLUT)
-                    || components != (int)values.size())
+
+                char inputLUT[4][64] = {"", "", "", ""};
+#ifdef _WIN32
+                if (sscanf_s(lineBuffer, "%s %s %s %63s", inputLUT[0], 64,
+                           inputLUT[1], 64, inputLUT[2], 64, inputLUT[3],
+                           64) != components)
+#else
+                if (sscanf(lineBuffer, "%s %s %s %63s", inputLUT[0],
+                           inputLUT[1], inputLUT[2], inputLUT[3]) != components)
+#endif
                 {
                     std::ostringstream os;
                     os << "Malformed LUT line. Expecting a ";
@@ -207,6 +231,26 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
                 if (lineCount >= lut_size)
                 {
                     ThrowErrorMessage("Too many entries found", currentLine, "");
+                }
+
+                values.resize(components);
+
+                for (int i = 0; i < components; i++)
+                {
+                    float v = NAN;
+                    const auto result = NumberUtils::from_chars(inputLUT[i], inputLUT[i] + 64, v);
+
+                    if (result.ec != std::errc())
+                    {
+                        std::ostringstream os;
+                        os << "Malformed LUT line. Could not convert component";
+                        os << i << " to a floating point number.";
+
+                        ThrowErrorMessage("Malformed LUT line", currentLine,
+                                            line);
+                    }
+
+                    values[i] = v;
                 }
 
                 // If 1 component is specified, use x1 x1 x1.
--- a/src/OpenColorIO/fileformats/xmlutils/XMLReaderUtils.h
+++ b/src/OpenColorIO/fileformats/xmlutils/XMLReaderUtils.h
@@ -13,6 +13,8 @@
 #include <OpenColorIO/OpenColorIO.h>
 
 #include "MathUtils.h"
+#include "utils/StringUtils.h"
+#include "utils/NumberUtils.h"
 #include "Platform.h"
 
 
@@ -156,16 +158,17 @@ void ParseNumber(const char * str, size_t startPos, size_t endPos, T & value)
     const char * startParse = str + startPos;
 
     double val = 0.0f;
-    char * endParse = nullptr;
 
-    // The strtod expects a C string and str might not be null terminated.
-    // However since strtod will stop parsing when it encounters characters
-    // that it cannot convert to a number, in practice it does not need to
-    // be null terminated.
-    // C++11 version of strtod processes NAN & INF ASCII values.
-    val = strtod(startParse, &endParse);
+    size_t adjustedStartPos = startPos;
+    size_t adjustedEndPos = endPos;
+
+    FindSubString(startParse, endPos - startPos, adjustedStartPos, adjustedEndPos);
+
+    const auto result = NumberUtils::from_chars(startParse + adjustedStartPos, startParse + adjustedEndPos, val);
+
     value = (T)val;
-    if (endParse == startParse)
+
+    if (result.ec == std::errc::invalid_argument)
     {
         std::string fullStr(str, endPos);
         std::string parsedStr(startParse, endPos - startPos);
@@ -187,10 +190,10 @@ void ParseNumber(const char * str, size_t startPos, size_t endPos, T & value)
             << TruncateString(fullStr.c_str(), endPos, 100) << "'.";
         throw Exception(oss.str().c_str());
     }
-    else if (endParse != str + endPos)
+    else if (result.ptr != str + endPos)
     {
         // Number is followed by something.
-        std::string fullStr(str, startPos + (endParse - startParse));
+        std::string fullStr(str, endPos);
         std::string parsedStr(startParse, endPos - startPos);
         std::ostringstream oss;
         oss << "ParserNumber: '"
--- a/src/utils/NumberUtils.h
+++ b/src/utils/NumberUtils.h
@@ -1,0 +1,166 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright Contributors to the OpenColorIO Project.
+
+#ifndef INCLUDED_NUMBERUTILS_H
+#define INCLUDED_NUMBERUTILS_H
+
+#if defined(_MSC_VER)
+#define really_inline __forceinline
+#else
+#define really_inline inline __attribute__((always_inline))
+#endif
+
+#include <cstdlib>
+#include <locale>
+#include <system_error>
+#include <type_traits>
+
+namespace OCIO_NAMESPACE
+{
+namespace NumberUtils
+{
+
+struct Locale
+{
+#ifdef _WIN32
+    Locale() : local(_create_locale(LC_ALL, "C"))
+    {
+    }
+    ~Locale()
+    {
+        _free_locale(local);
+    }
+    _locale_t local;
+#else
+    Locale() : local(newlocale(LC_ALL_MASK, "C", NULL))
+    {
+    }
+    ~Locale()
+    {
+        freelocale(local);
+    }
+    locale_t local;
+#endif
+};
+
+struct from_chars_result
+{
+    const char *ptr;
+    std::errc ec;
+};
+
+static const Locale loc;
+
+really_inline from_chars_result from_chars(const char *first, const char *last, double &value) noexcept
+{
+    errno = 0;
+    if (!first || !last || first == last)
+    {
+        return {first, std::errc::invalid_argument};
+    }
+
+    char * endptr = nullptr;
+
+    double
+#ifdef _WIN32
+    tempval = _strtod_l(first, &endptr, loc.local);
+#else
+    tempval = ::strtod_l(first, &endptr, loc.local);
+#endif
+
+    if (errno != 0)
+    {
+        return {first + (endptr - first), std::errc::result_out_of_range};
+    }
+    else if (endptr == first)
+    {
+        return {first, std::errc::invalid_argument};
+    }
+    else if (endptr <= last)
+    {
+        value = tempval;
+        return {first + (endptr - first), {}};
+    }
+    else
+    {
+        return {first, std::errc::argument_out_of_domain};
+    }
+}
+
+really_inline from_chars_result from_chars(const char *first, const char *last, float &value) noexcept
+{
+    errno = 0;
+    if (!first || !last || first == last)
+    {
+        return {first, std::errc::invalid_argument};
+    }
+
+    char *endptr = nullptr;
+
+    float
+#ifdef _WIN32
+    tempval = _strtof_l(first, &endptr, loc.local);
+#elif __APPLE__
+    // On OSX, strtod_l is for some reason drastically faster than strtof_l.
+    tempval = static_cast<float>(::strtod_l(first, &endptr, loc.local));
+#else
+    tempval = ::strtof_l(first, &endptr, loc.local);
+#endif
+
+    if (errno != 0)
+    {
+        return {first + (endptr - first), std::errc::result_out_of_range};
+    }
+    else if (endptr == first)
+    {
+        return {first, std::errc::invalid_argument};
+    }
+    else if (endptr <= last)
+    {
+        value = tempval;
+        return {first + (endptr - first), {}};
+    }
+    else
+    {
+        return {first, std::errc::argument_out_of_domain};
+    }
+}
+
+really_inline from_chars_result from_chars(const char *first, const char *last, long int &value) noexcept
+{
+    errno = 0;
+    if (!first || !last || first == last)
+    {
+        return {first, std::errc::invalid_argument};
+    }
+
+    char *endptr = nullptr;
+
+    long int
+#ifdef _WIN32
+    tempval = _strtol_l(first, &endptr, 0, loc.local);
+#else
+    tempval = ::strtol_l(first, &endptr, 0, loc.local);
+#endif
+
+    if (errno != 0)
+    {
+        return {first + (endptr - first), std::errc::result_out_of_range};
+    }
+    else if (endptr == first)
+    {
+        return {first, std::errc::invalid_argument};
+    }
+    else if (endptr <= last)
+    {
+        value = tempval;
+        return {first + (endptr - first), {}};
+    }
+    else
+    {
+        return {first, std::errc::argument_out_of_domain};
+    }
+}
+} // namespace NumberUtils
+} // namespace OCIO_NAMESPACE
+#endif // INCLUDED_NUMBERUTILS_H
--- a/src/OpenColorIO/CMakeLists.txt
+++ b/src/OpenColorIO/CMakeLists.txt
@@ -227,6 +227,7 @@ target_link_libraries(OpenColorIO
         ${OCIO_HALF_LIB}
         pystring::pystring
         sampleicc::sampleicc
+        utils::from_chars
         utils::strings
         yaml-cpp
 )
--- a/src/OpenColorIO/fileformats/FileFormatIridasLook.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatIridasLook.cpp
@@ -16,6 +16,7 @@
 #include "pystring/pystring.h"
 #include "transforms/FileTransform.h"
 #include "utils/StringUtils.h"
+#include "utils/NumberUtils.h"
 
 
 /*
@@ -416,19 +417,18 @@ private:
             std::string size_raw = std::string(s, len);
             std::string size_clean = pystring::strip(size_raw, "'\" "); // strip quotes and space
 
-            char* endptr = 0;
-            int size_3d = static_cast<int>(strtol(size_clean.c_str(), &endptr, 10));
+            long int size_3d{};
+            
+            const auto result = NumberUtils::from_chars(size_clean.c_str(), size_clean.c_str() + size_clean.size(), size_3d);
 
-            if (*endptr)
+            if (result.ec != std::errc())
             {
-                // strtol didn't consume entire string, there was
-                // remaining data, thus did not contain a single integer
                 std::ostringstream os;
                 os << "Invalid LUT size value: '" << size_raw;
                 os << "'. Expected quoted integer";
                 pImpl->Throw(os.str().c_str());
             }
-            pImpl->m_lutSize = size_3d;
+            pImpl->m_lutSize = static_cast<int>(size_3d);
         }
         else if (pImpl->m_data)
         {
--- a/src/OpenColorIO/fileformats/FileFormatSpi3D.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatSpi3D.cpp
@@ -12,6 +12,7 @@
 #include "Platform.h"
 #include "transforms/FileTransform.h"
 #include "utils/StringUtils.h"
+#include "utils/NumberUtils.h"
 
 
 /*
@@ -131,7 +132,7 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
     // Parse table
     int index = 0;
     int rIndex, gIndex, bIndex;
-    float redValue, greenValue, blueValue;
+    float redValue = NAN, greenValue = NAN, blueValue = NAN;
 
     int entriesRemaining = rSize * gSize * bSize;
     Array & lutArray = lut3d->getArray();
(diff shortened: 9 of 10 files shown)


amyspark:
Force-pushed again because of a typo in one of the issue numbers.

remia:
Thanks for the contribution @amyspark, great to see some progress on this long standing issue ! I have yet to test that on actual files, though looking at the changes I already had questions (I have not followed all discussions on the topic over the years, trying to understand):
- It looks like SPI[1|3]D format are the only users of `sscanf` to parse float numbers while other formats rely on `StringToFloat` from `ParseUtils.h` which rely on `std::istringstream`. Is that affected by the system locale as well ? I would assume we should use this helper everywhere and restrict usage of raw fast_float calls only there. Do we use `sscanf` on SPI formats due to specific performance requirements ?
- Can we reproduce this locale issue in unit test and add in as part of the test suite ?

amyspark:
Hey @remia !

>     * It looks like SPI[1|3]D format are the only users of `sscanf` to parse float numbers while other formats rely on `StringToFloat` from `ParseUtils.h` which rely on `std::istringstream`. Is that affected by the system locale as well ? I would assume we should use this helper everywhere and restrict usage of raw fast_float calls only there.

*Anything* that is not `std::atof` is affected by the issue, starting with `std::strtod` up to any stringstream that uses floats. However, these could be worked around by imbuing the locale, but see below.

> Do we use `sscanf` on SPI formats due to specific performance requirements ?

I believe so; iirc @hodoulp rejected the stringstream+imbuelocale approach because it is notoriously slower compared to a plain sscanf. And using C++17 `from_chars` wasn't an option because [Clang to date still doesn't ship it](https://reviews.llvm.org/D70631).

>     * Can we reproduce this locale issue in unit test and add in as part of the test suite ?

It is tricky because you need to change the running thread's locale through your C runtime before being able to reproduce it, e.g. with `LC_ALL` or `LANG` set to any non-English locale that uses commas as thousands separator.

amyspark:
Since the parsing utils already pay the cost of `stringstream`, the last commit ensure they parse numbers according to the C locale.

michdolan:
Thanks for your contribution @amyspark ! I added this PR to the TSC meeting agenda tomorrow for more discussion.

hodoulp:
:heavy_exclamation_mark: As agreed at the last [TSC meeting](https://wiki.aswf.io/display/OCIO/2021-10-04), all the third-party library dependencies must remain external (when used as-is) and be included using a dedicated `FindXX()` cmake method. You can check [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/tree/master/share/cmake/modules) for examples.

amyspark:
I was not present at the meeting, so I would like to take the opportunity to address these points below.

> Michael: Says in PR it's amalgamated, but not certain if it's modified.

The library was fetched and committed as is from the release: https://github.com/fastfloat/fast_float/releases/tag/v3.1.0. https://github.com/fastfloat/fast_float/tree/v3.1.0#using-as-single-header lists the amalgamation process. I'm afraid the upstream doesn't list SHA256 hashes, so I cannot provide more assurances regarding that.

> Kevin: Provides amalgamated file in GH project. Apache 2 license. Is this the right code to be adding? Is this the right solution to solving locale issue? I know every month there's something better or faster, and eventually some of this will make it into standard library. Is the core idea is to solve locale parsing/serializing problem? Is this library a reasonable solution for that?

The list of issues in this PR shows that this bug has existed for almost a decade, which is a very long time in our field.
It is correct, however, that the standard library now allows users to parse floating-point numbers locale-free, with the addition of `std::from_chars`. However, when one looks more in depth, this is not entirely true.

This feature in particular not only requires a fairly new standard (for instance, we at Krita only moved to 14 last year), it also needs a very recent compiler. [According to cppreference](https://en.cppreference.com/w/cpp/compiler_support/17), to date, only GCC 11 and MSVC 2017 support `from_chars` with floating point numbers. Clang just released version 13, see [the LLVM site](https://releases.llvm.org/)), yet this particular feature is still under review. See [D70631 at the LLVM site](https://reviews.llvm.org/D70631). In all cases, we are limited to C++17 -- there is no chance to use `from_chars` with 14 or 11, which is what this library adds.

> Kevin: Actual repo is header only. If we didn't want to include it, could use submodule to pull down or something.

> Kevin: Would prefer it to be external dependency. As long as it doesn't break or regress, ok with this library. Should make sure it doesn't change a lot. Doesn't have a lot of commits.

I will port it to an external dependency in the next force push.

> Rémi: stringstream is slow

stringstream is unavoidable until OpenColorIO migrates its formatting utilities to fmt, because `from_chars` doesn't cover float-to-string.

amyspark:
I've now pushed the requested fixes. @hodoulp, Spi1D now also uses `from_chars` through the changes I made to `StringVecToFloatVec`, can you run your measurements again?

lgritz:
Indeed, to_chars is actually a C++17 feature, not even C++14.

amyspark:
@lgritz MSVC is standards compliant, i.e. it only allows `from_chars` with `/std:c++17` onwards.

hodoulp:
Here are my new numbers demonstrating a major improvement for `spi1d` 🎉
|                        | master           |  first test        | second test
| ------------- | ------------- | ------------- | ------------- |
| spi1d  [^1]            | 107 ms  | 116 ms  | 55 ms  |
| spi3d  [^2]           | 18 ms  | 16 ms  | 16 ms  |
| clf  (1D LUT) [^3]  | 44 ms  | x | 43 ms|
| clf (3D LUT) [^4]  | 27 ms | x | 26 ms|


[^1]: `./src/apps/ocioperf/ocioperf --transform ./testdata/lut1d_3.spi1d --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 1 --nocache`

[^2]: `./src/apps/ocioperf/ocioperf --transform ./testdata/spi_ocio_srgb_test.spi3d --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 1 --nocache`

[^3]: `./src/apps/ocioperf/ocioperf --transform ./testdata/clf/lut1d_long.clf --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 1 --nocache`

[^4]: `./src/apps/ocioperf/ocioperf --transform ./testdata/clf/aces_to_video_with_look.clf  --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 1 --nocache`

hodoulp:
> > Only the header only library version must be used (i.e. the current FindFastFloat() method downloads the library and not the header file).

> @hodoulp, I understood that the library was to be built from source, as is done with the rest of the dependencies?

In general, the third-party libraries are built from sources. But this one offers a header only solution which will reduce compile time. So, we should take it.

michdolan:
In terms of the `fast_float` license, both available licenses (Apache 2.0 and MIT) are OSI approved, so it should not be an issue to use with OCIO: https://spdx.org/licenses/

@jmertic and @swinslow for any additional input on use of https://github.com/fastfloat/fast_float in OCIO.

jmertic:
@michdolan Correct. It would be preferred to use under the terms of the MIT versus Apache 2.

hodoulp:
Following the discussion at the last TSC meeting, I collected numbers for different implementations even if some file format readers are currently far less optimal than others i.e. it uses the current WIP implementations from this pull request. If more performance optimizations can be done, it will be done later. The goal for now is only to compare each float conversion implementations to pick the best solution for short and long terms.

The numbers include the file read (i.e. file cache is disabled) and a processor instance creation (i.e. processor cache is disabled). As these files only contain a single big LUT, the processor creation is mainly the conversion from strings to floats.

The following numbers are from `macOS 10.15` with `Apple Clang 12`:

|                        | master (no locale fix) | fast_float           |  from_chars (C++17)        | strtof (C) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| spi1d  [^1]            | 107 ms | 32 ms  | No supported | 36 ms |
| spi3d  [^2]           | 28 ms | 16 ms  | No supported | 17 ms |
| clf  (1D LUT) [^3]  | 44 ms | 44 ms  | No supported | 45 ms |
| clf (3D LUT) [^4]  | 27 ms | 26 ms | No supported | 29 ms |

The following numbers are from `Windows 10` with `MSVC 19.27`:

|                        | master (no locale fix) | fast_float           |  from_chars (C++17)        | strtof (C) |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| spi1d  [^5]            | 185-300 ms | 71-75 ms | 88-93 ms | 97-104 ms |
| spi3d  [^6]           | 50 ms | 39-42 ms | 50 ms | 53-57 ms |
| clf  (1D LUT) [^7]  | 84 ms | 67-69 ms | 81-87 ms | 87-95 ms |
| clf (3D LUT) [^8]  | 49-53 ms | 37-40 ms | 49-52 ms | 56-59 ms |

Note: The `strtof (C)` proposal uses a static global variable to hold the C locale i.e. it could then introduce a lock. I did measurements  on Linux and Windows, the static global variable (vs. local variable) is always by far the fastest one. The local variable is in the conversion method (to keep the same method signature between the various implementations) which is not at all optimal.

Note: `OpenImageIO` also faced the same challenge and developed an home-made solution that we wanted to check. After reading the [code](https://github.com/OpenImageIO/oiio/blob/master/src/libutil/strutil.cpp#L1339), the OIIO solution relies on `strtof` and `strtod` like the `strtof (C)` proposal. But it highlights that `strtof` is way slower than `strtod` on `MacOS` so, I've updated the proposal and measure again. That's true!

Note: To make it run on Windows, I changed a little bit the proposal:

```
#if defined(_MSC_VER)
    #define really_inline __forceinline
#else
    #define really_inline inline __attribute__((always_inline))
#endif

#if defined(__has_include) && __has_include(<version>)
#include <version> // from C++20 all feature macros should be defined here
#else
#include <utility>
#endif

#ifdef USE_FAST_FLOAT
    #include <fast_float/fast_float.h>

    template <typename T>
    really_inline bool ocio_from_chars(const char * first, const char * last, T & value) noexcept
    {
        // By design that's not supported by from_chars() so handle it here.
        if (first && *first == '+') ++first;

        const auto ret = fast_float::from_chars(first, last, value);
        return ret.ec == std::errc();
    }
#elif __cpp_lib_to_chars >= 201611
    #include <charconv>

    template <typename T>
    really_inline bool ocio_from_chars(const char * first, const char * last, T & value) noexcept
    {
        // By design that's not supported by from_chars() so handle it here.
        if (first && *first == '+') ++first;

        const auto ret = std::from_chars(first, last, value);
        return ret.ec == std::errc();
    }
#else
    #include <locale>
    #include <stdlib.h>
    #include <type_traits>

    struct Locale
    {
#ifdef _WIN32
        Locale() : local(_create_locale(LC_ALL, "C")) {}
        ~Locale() { _free_locale(local); }
        _locale_t local;
#else
        Locale() : local(newlocale(LC_ALL_MASK, "C", NULL)) {}
        ~Locale() { freelocale(local); }
        locale_t local;
#endif
    };

    static const Locale loc;

    really_inline bool ocio_from_chars(const char * first, const char * last, double & value) noexcept
    {
       errno = 0;
       if (!first || !last || first == last)
        {
            return false;
        }

#ifdef _WIN32
        value = _strtod_l(first, nullptr, loc.local);
#else
        value = ::strtod_l(first, nullptr, loc.local);
#endif

        return errno == 0;
    }

    really_inline bool ocio_from_chars(const char * first, const char * last, float & value) noexcept
    {
       errno = 0;
       if (!first || !last || first == last)
        {
            return false;
        }

#ifdef _WIN32
        value = _strtof_l(first, nullptr, loc.local);
#elif __APPLE__
        // On OSX, strtod_l is for some reason drastically faster than strtof_l.
        value = static_cast<float>(::strtod_l(first, nullptr, loc.local));
#else
        value = ::strtof_l(first, nullptr, loc.local);
#endif

        return errno == 0;
    }

    really_inline bool ocio_from_chars(const char * first, const char * last, long int & value) noexcept
    {
        errno = 0;
        if (!first || !last || first == last)
        {
            return false;
        }

#ifdef _WIN32
        value = _strtol_l(first, nullptr, 0, loc.local);
#else
        value = ::strtol_l(first, nullptr, 0, loc.local);
#endif

        return errno == 0;
    }
#endif
```

[^1]: `./src/apps/ocioperf/ocioperf --transform ./testdata/lut1d_3.spi1d --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 10 --nocache`

[^2]: `./src/apps/ocioperf/ocioperf --transform ./testdata/spi_ocio_srgb_test.spi3d --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 10 --nocache`

[^3]: `./src/apps/ocioperf/ocioperf --transform ./testdata/clf/lut1d_long.clf --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 10 --nocache`

[^4]: `./src/apps/ocioperf/ocioperf --transform ./testdata/clf/aces_to_video_with_look.clf  --image ~/dev/images/aces/images/ACES/DigitalLAD.2048x1556.exr --iter 10 --nocache`

[^5]: `.\src\apps\ocioperf\ocioperf.exe --transform .\testdata\lut1d_3.spi1d --image F:\OpenSource\benchmark\images\ACES\DigitalLAD.2048x1556.exr --iter 10 --nocache`

[^6]: `.\src\apps\ocioperf\ocioperf.exe --transform .\testdata\spi_ocio_srgb_test.spi3d --image F:\OpenSource\benchmark\images\ACES\DigitalLAD.2048x1556.exr --iter 10 --nocache`

[^7]: `.\src\apps\ocioperf\ocioperf.exe --transform .\testdata\clf\lut1d_long.clf --image F:\OpenSource\benchmark\images\ACES\DigitalLAD.2048x1556.exr --iter 10 --nocache`

[^8]: `.\src\apps\ocioperf\ocioperf.exe --transform .\testdata\clf\aces_to_video_with_look.clf --image F:\OpenSource\benchmark\images\ACES\DigitalLAD.2048x1556.exr --iter 10 --nocache`

hodoulp:
Hi @amyspark,
Your work triggered some many questions that it took some time to have a clear picture on what is the right approach(es) for the library with the `locale` and the performance aspects. In conclusion, the agreement is to only implement the `fast_float` and `strtof` approaches. As the OpenColorIO C++ version is C++11 for now and the `fast_float` provides better performance, the C++17 `std::from_chars` implementation will be useless for short and mid terms. But the `fast_float` implementation must be controlled by a cmake variable (ON by default) and must only use the header only code (i.e. not the library one).
Please let me know if you have any questions.

amyspark:
>  But the fast_float implementation must be controlled by a cmake variable (ON by default) and must only use the header only code (i.e. not the library one).
Please let me know if you have any questions.

How will hexadecimal floats, and integers be handled, as they are *not* supported by fast_float?

hodoulp:
>  > But the fast_float implementation must be controlled by a cmake variable (ON by default) and must only use the header only code (i.e. not the library one).
Please let me know if you have any questions.

> How will hexadecimal floats, and integers be handled, as they are not supported by fast_float?

The hexadecimal support is dropped (as discussed at the [TSC meeting](https://wiki.aswf.io/display/OCIO/2021-10-18)), and the integer support will have to be manually coded (i.e. the template of the previous `fast_float` approach could be removed, and the integer support with `ocio_from_chars` & `strtod_l` be moved out of the ifdef).

amyspark:
I pushed the requested changes, yanking the remaining uses of strtod and strtol. Thanks for reviewing this @hodoulp!

hodoulp:
@amyspark , thanks for the push.

There are some unit test failures:

```
2: [ 431/1024] [FileFormatIridasLook / hexasciitofloat                      ] - PASSED
2: 
2: FAILED: Error parsing Iridas Look file (). Error is: Invalid LUT size value: '"8"'. Expected quoted integer. At line (22).
2: [ 432/1024] [FileFormatIridasLook / simple3d                             ] - FAILED
```
and
```
2: [ 468/1024] [XMLReaderHelper / string_to_float                           ] - PASSED
2: D:\a\OpenColorIO\OpenColorIO\tests\cpu\fileformats\xmlutils\XMLReaderUtils_tests.cpp:41:
2: FAILED: OCIO::Exception is expected to be thrown
2: [ 469/1024] [XMLReaderHelper / string_to_float_failure                   ] - FAILED
2: D:\a\OpenColorIO\OpenColorIO\tests\cpu\fileformats\xmlutils\XMLReaderUtils_tests.cpp:89:
2: FAILED: values[4] == 66.0f
2: 	values were '0' and '66'
2: D:\a\OpenColorIO\OpenColorIO\tests\cpu\fileformats\xmlutils\XMLReaderUtils_tests.cpp:119:
2: FAILED: OCIO::Exception is expected to be thrown
2: [ 470/1024] [XMLReaderHelper / get_numbers                               ] - FAILED
2: [ 471/1024] [XMLReaderHelper / trim                                      ] - PASSED
2: D:\a\OpenColorIO\OpenColorIO\tests\cpu\fileformats\xmlutils\XMLReaderUtils_tests.cpp:383:
2: FAILED: OCIO::Exception is expected to be thrown
2: [ 472/1024] [XMLReaderHelper / parse_number                              ] - FAILED
```

amyspark:
Working on them, I couldn't run the tests locally.

amyspark:
Tests are fixed, and now we have consistent handling of integers and floats.

@hodoulp I'd like to point out a few things:

- integers now pass through the locale-based `strtol`, so everything is locale agnostic
- to make builds with and without fast_float (or integer and float parsing) consistent, locale-based functions need to copy the character range to a temporary string.

Do let me know if the changes are acceptable, and whether you'd an entry in the build matrix for `OCIO_USE_FAST_FLOAT`. (Though, as OCIO specify its builds manually, I'm not sure if a full matrix duplication would be acceptable.)

share/cmake/modules/FindFastFloat.cmake
```diff
@@ -0,0 +1,128 @@
...
+
+    set(_EXT_DIST_ROOT "${CMAKE_BINARY_DIR}/ext/dist")
+    set(_EXT_BUILD_ROOT "${CMAKE_BINARY_DIR}/ext/build")
+
+    # Set find_package standard args
+    set(FastFloat_FOUND TRUE)
+    set(FastFloat_VERSION ${FastFloat_FIND_VERSION})
+    set(FastFloat_INCLUDE_DIR "${_EXT_DIST_ROOT}/${CMAKE_INSTALL_INCLUDEDIR}")
+
+    # This library is include-only. No need to add further flags.
```
hodoulp, line 75: When I look at the `fast_float` [v3.2.0 release](https://github.com/fastfloat/fast_float/releases/tag/v3.2.0), there is a single header file. But running the pull request installs several header files without static library. Bottom line it makes what was proposed (i.e. include only). However, what is the difference between the two?

src/OpenColorIO/ParseUtils.cpp
```diff
@@ -601,11 +607,12 @@ bool StringVecToFloatVec(std::vector<float> &floatArray, const StringUtils::Stri
...
     {
-        std::istringstream inputStringstream(lineParts[i]);
         float x;
-        if(!(inputStringstream >> x))
+        const char *str = lineParts[i].c_str();
+        const auto result = NumberUtils::from_chars(str, str + strlen(str), x);
+        if (result.ec != std::errc())
         {
-            return false;
+          return false;
```
hodoulp, line 615: '4 spaces' is the space to have inside parenthesis.

src/OpenColorIO/fileformats/FileFormatHDL.cpp
```diff
@@ -185,13 +186,11 @@ readLuts(std::istream& istream,
             // StringToFloat was far slower, for 787456 values:
             // - StringToFloat took 3879 (avg nanoseconds per value)
             // - stdtod took 169 nanoseconds
-            char* endptr = 0;
```
hodoulp, line 188: You can now remove the three previous lines.

src/OpenColorIO/fileformats/FileFormatSpi1D.cpp
```diff
@@ -209,6 +233,25 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
                     ThrowErrorMessage("Too many entries found", currentLine, "");
                 }
 
+                values.resize(components);
+
+                for (int i = 0; i < components; i++) {
```
hodoulp, line 238: `{` must be on the next line.

src/OpenColorIO/fileformats/xmlutils/XMLReaderUtils.h
```diff
@@ -154,18 +156,31 @@ void ParseNumber(const char * str, size_t startPos, size_t endPos, T & value)
...
+    // However, since from_chars also takes a pointer to the last character,
+    // the string does not need to be null terminated explicitly.
+    // Do note, however, that plus signs and leading whitespace
+    // need to be skipped manually -- this
+    // was left implicit in the usage of strtod.
+    // TODO: allow hex numbers -- this will probably mean we need another from_chars library
+    const std::string parsedStr = StringUtils::LeftTrim({startParse, endPos - startPos});
+    if (startPos + parsedStr.size() != endPos) {
+        adjustedParse = str + endPos - parsedStr.size();
+    }
```
hodoulp, line 173: The code should not use the StringUtils methods (which behind the scene mostly creates strings). Here is the code must only manipulate pointers.

src/OpenColorIO/fileformats/xmlutils/XMLReaderUtils.h
```diff
@@ -154,18 +156,31 @@ void ParseNumber(const char * str, size_t startPos, size_t endPos, T & value)
...
+
+    if (StringUtils::StartsWith(parsedStr, "+")) {
```
hodoulp, line 175: This code is useless (and slow) because `NumberUtils::from_chars` takes care of the `+` case.

src/utils/NumberUtils.h
```diff
@@ -0,0 +1,205 @@
...
+        return {first, std::errc::invalid_argument};
+    }
+
+    double tempval{};
+
+    char *endptr = nullptr;
+
+    // Unfortunately we need consistent behaviour
+    // between strtod and from_chars.
+    const std::string temp(first, last - first);
```
hodoulp, line 92: That kills the performance so, the code must be changed.

src/utils/NumberUtils.h
```diff
@@ -0,0 +1,205 @@
...
+        return {first, std::errc::invalid_argument};
+    }
+
+    char *endptr = nullptr;
+
+    float tempval{};
+
+    // Unfortunately we need consistent behaviour
+    // between strtod and from_chars.
+    const std::string temp{first, static_cast<size_t>(last - first)};
```
hodoulp, line 133: That kills the performance so, the code must be changed.

src/utils/NumberUtils.h
```diff
@@ -0,0 +1,205 @@
...
+        return {first, std::errc::invalid_argument};
+    }
+
+    long int tempval;
+
+    char *endptr = nullptr;
+
+    // Unfortunately we need consistent behaviour
+    // between strtod and from_chars.
+    const std::string temp(first, last - first);
```
hodoulp, line 178: That kills the performance so, the code must be changed.

src/utils/NumberUtils.h
```diff
@@ -0,0 +1,205 @@
...
+#ifdef USE_FAST_FLOAT
+#include <fast_float/fast_float.h>
+#endif
+
+#include <cstdlib>
+#include <locale>
+#include <system_error>
+#include <type_traits>
+
+namespace NumberUtils
```
hodoulp, line 22: The header code must be in the `OCIO_NAMESPACE` namespace.

hodoulp (changes requested):
Thanks for the update. 

The string creation in the `from_chars` will definitively kill the overall performance.

src/utils/NumberUtils.h
amyspark: Is this correct? 

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/64dccadba11826a2358bf192519563568462cf08/src/utils/StringUtils.h#L18

hodoulp: No. I found no historical reasons to not have the OCIO namespace so, I assume that we missed it.

hodoulp:
> (Though, as OCIO specify its builds manually, I'm not sure if a full matrix duplication would be acceptable.)

No. As we try to keep the build number under control (i.e. there is already lot of builds) It would be great to only add one or two new entries.

share/cmake/modules/FindFastFloat.cmake
amyspark: I had understood from this conversation that the team preferred to build it from scratch: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1496#issuecomment-939206100.

I'll push the changes right away.

src/utils/NumberUtils.h
amyspark: As explained in the commit message, changing this bit (and all below this review) will make locale and fast_float versions inconsistent.

tests/cpu/fileformats/xmlutils/XMLReaderUtils_tests.cpp
```diff
@@ -45,13 +45,9 @@ OCIO_ADD_TEST(XMLReaderHelper, string_to_float_failure)
...
     const size_t len2 = std::strlen(str2);
-    // All characters are parsed and this is more than the required length.
-    // The string to double function strtod does not stop at a given length,
-    // but we detect that strtod did read too many characters.
-    OCIO_CHECK_THROW_WHAT(OCIO::ParseNumber(str2, 0, len2 - 2, value),
-                          OCIO::Exception,
-                          "followed by unexpected characters");
-
+    // strtod with temporary storage will behave the same as from_chars
+    OCIO_CHECK_NO_THROW(OCIO::ParseNumber(str2, 0, len2 - 2, value));
```
amyspark, line 49: @hodoulp, this is the test that will detect the inconsistency. strtod-based from_chars will throw here without the duplication, whereas from_chars will succeed.

src/utils/NumberUtils.h
```diff
@@ -19,7 +19,7 @@
 #include <system_error>
 #include <type_traits>
 
-namespace NumberUtils
+namespace OCIO_NAMESPACE
```
hodoulp, line 22: I did not say to remove the `NumberUtils` namespace but rather to encapsulate it in the `OCIO_NAMSPACE` one i.e. two namespaces. That's to have code like `NumberUtils::from_chars()` (like `Platform::Getenv()` for example) to avoid potential confusion with `std::from_chars`.

src/OpenColorIO/fileformats/xmlutils/XMLReaderUtils.h
```diff
@@ -164,7 +164,7 @@ void ParseNumber(const char * str, size_t startPos, size_t endPos, T & value)
 
     FindSubString(startParse, endPos - startPos, adjustedStartPos, adjustedEndPos);
 
-    const auto result = NumberUtils::from_chars(startParse + adjustedStartPos, startParse + adjustedEndPos, val);
+    const auto result = OCIO_NAMESPACE::from_chars(startParse + adjustedStartPos, startParse + adjustedEndPos, val);
```
hodoulp, line 167: Based on the following discussion `OCIO_NAMESPACE::from_chars` must be `NumberUtils::from_chars`.
Note: The `OCIO_NAMESPACE` is useless because the code is already in this namespace.

share/cmake/modules/FindFastFloat.cmake
```diff
@@ -0,0 +1,129 @@
...
+
+    if(_FastFloat_TARGET_CREATE)
```
hodoulp, line 77: As the code now looks for the header only, the line 78 to line 99 seems now useless.

share/cmake/utils/CompilerFlags.cmake
```diff
@@ -12,7 +12,8 @@ if(USE_MSVC)
     set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} /DUSE_MSVC")
 
     # /we4062 Enables warning in switch when an enumeration value is not explicitly handled.
-    set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} /EHsc /DWIN32 /we4062")
+    # /we4701 Potentially uninitialized local variable 'name' used
```
hodoulp, line 15: If that's needed by `fast_float`, please add a comment like `... because of fast_float vX.X.X`

tests/cpu/fileformats/xmlutils/XMLReaderUtils_tests.cpp
hodoulp: Based on the only usage (read floats separated by a space, comma, etc. anyway the separator character is not a number) it's perhaps not a problem. Let me think about.

hodoulp (changes requested):
Want to put some time investigating the last point: parsing difference between `fast_float::from_chars` and `::strtof_l` in order to remove the `temp` string.

share/cmake/utils/CompilerFlags.cmake
amyspark: Not exactly, I added this because OCIO's current GCC/Clang setup issues `-Wall -Werror` and that detects uninitialized variables, unlike MSVC (which I work with). So I end up noticing missing bits only at CI time.

tests/cpu/fileformats/xmlutils/XMLReaderUtils_tests.cpp
amyspark: To check, you could try making the string duplication conditional on `first + strlen(first) > last`, and see what codepaths trigger it.

hodoulp: I did some checks with your pull request and the following code is working without the temporary string:
```
really_inline from_chars_result from_chars(const char * first, const char * last, double & value) noexcept
{
    errno = 0;
    if (!first || !last || first == last)
    {
        return {first, std::errc::invalid_argument};
    }

    char * endptr = nullptr;

    double
#ifdef _WIN32
        tempval = _strtod_l(first, &endptr, loc.local);
#else
        tempval = ::strtod_l(first, &endptr, loc.local);
#endif

    if (errno != 0)
    {
        return {first + (endptr - first), std::errc::result_out_of_range};
    }
    else if (endptr == first)
    {
        return {first, std::errc::invalid_argument};
    }
    else if (endptr <= last)
    {
        value = tempval;
        return {first + (endptr - first), {}};
    }
    else
    {
        return {first, std::errc::argument_out_of_domain};
    }
}
```

amyspark: The rest of the tests pass, so I'll comment the test suite (just in case it's needed in the future).

share/cmake/utils/CompilerFlags.cmake
amyspark: I will add the explanation to be further clear, as this may affect only the strtod-based from_chars (which leave the value variable explicitly untouched on failure).

tests/cpu/fileformats/xmlutils/XMLReaderUtils_tests.cpp
```diff
@@ -43,15 +43,12 @@ OCIO_ADD_TEST(XMLReaderHelper, string_to_float_failure)
...
     OCIO_CHECK_NO_THROW(OCIO::ParseNumber(str1, 0, 2, value));
 
-    const char str2[] = "12345";
-    const size_t len2 = std::strlen(str2);
-    // All characters are parsed and this is more than the required length.
-    // The string to double function strtod does not stop at a given length,
-    // but we detect that strtod did read too many characters.
-    OCIO_CHECK_THROW_WHAT(OCIO::ParseNumber(str2, 0, len2 - 2, value),
-                          OCIO::Exception,
-                          "followed by unexpected characters");
```
hodoulp, line 53: On least one macOS & AppleClang 10, the code above is now working with the last proposal. Did you face a problem?

amyspark: If enabled, this happens: https://github.com/amyspark/OpenColorIO/actions/runs/1453341456

I need to cherry-pick the last commit to fix two typos in the ci-workflow changes.

hodoulp: Based on the link, the code throws an exception as expected but the unit test code does not expect one! 
The right behavior is throw an exception, and the original unit test code handles it. 

The change is to uncomment lines 46 to 53 (in red above), and remove lines 47 to 51 (in green below). Is it what you did?

amyspark: Here's the commit: https://github.com/amyspark/OpenColorIO/commit/95e199a2f5253f5a307fc98365d3ed0d1fdec576

The real issue is that, test-wise, there is a difference between the two implementations. That's why tests without fast-float throw an exception (correctly) and tests with fast-float do not (correctly as well).

hodoulp: The changes in the commit are not the ones to do. By reverting to the original code it should work.

The original code is:
```
    const char str2[] = "12345";
    const size_t len2 = std::strlen(str2);
    // All characters are parsed and this is more than the required length.
    // The string to double function strtod does not stop at a given length,
    // but we detect that strtod did read too many characters.
    OCIO_CHECK_THROW_WHAT(OCIO::ParseNumber(str2, 0, len2 - 2, value),
                          OCIO::Exception,
                          "followed by unexpected characters");
```

amyspark: > The changes in the commit are not the ones to do. By reverting to the original code it should work.

It will work, but you'll instead break the tests when fast_float is enabled. Since those will correctly parse "123" with the specified boundaries, no exception will be thrown.

hodoulp: Yep. No good solution as `fast_float::from_chars` & `strtof` behavior differently. 
Concerning the test in discussion, the problem is the `strtof` implementation only. So, you can protect these few lines with an `#ifdef`.

src/OpenColorIO/fileformats/FileFormatSpi1D.cpp
```diff
@@ -209,6 +233,26 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
                     ThrowErrorMessage("Too many entries found", currentLine, "");
                 }
 
+                values.resize(components);
+
+                for (int i = 0; i < components; i++)
+                {
+                    float v = NAN;
+                    const auto result = NumberUtils::from_chars(inputLUT[i], inputLUT[i] + strlen(inputLUT[i]), v);
```
remia, line 241: [Question] Why using `strlen` here instead of 64 fixed size like is done elsewhere?

src/OpenColorIO/fileformats/xmlutils/XMLReaderUtils.h
```diff
@@ -156,16 +158,17 @@ void ParseNumber(const char * str, size_t startPos, size_t endPos, T & value)
...
+    size_t adjustedEndPos = endPos;
+
+    FindSubString(startParse, endPos - startPos, adjustedStartPos, adjustedEndPos);
+
+    const auto result = NumberUtils::from_chars(startParse + adjustedStartPos, startParse + adjustedEndPos, val);
+
     value = (T)val;
-    if (endParse == startParse)
+
+    if (result.ec == std::errc::invalid_argument)
```
remia, line 171: [Question] Why not using `result.ec != std::errc()` like elsewhere?

src/utils/NumberUtils.h
```diff
@@ -0,0 +1,193 @@
...
+#endif
+};
+
+struct from_chars_result
+{
+    const char *ptr;
+    std::errc ec;
+};
+
+static const Locale loc;
```
remia, line 56: [Question] Do we care about having duplication of the global variable in each translation unit here?

remia (commented):
The PR looks great @amyspark, I made a few minor points that are mostly questions and not issues. Also wondering if we can add a simple test case with, say simulating a French locale system and checking the behaviour is as expected (though I'm not sure the exact scope this test should cover). Has this been discussed and sorted already?

src/OpenColorIO/fileformats/FileFormatSpi1D.cpp
```diff
@@ -124,10 +126,26 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
             }
             else if(StringUtils::StartsWith(headerLine, "From"))
             {
-                if (sscanf(lineBuffer, "From %f %f", &from_min, &from_max) != 2)
+                char fromMinS[64] = "";
+                char fromMaxS[64] = "";
+#ifdef _WIN32
```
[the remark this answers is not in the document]
amyspark: No, that was a typo for sure!

src/utils/NumberUtils.h
amyspark: Hmm, I've seen this result in symbol clashes with very complex static library link trees. I can fix this by moving all implementation details to a cpp, and leaving a clean ABI stub here.

@hodoulp -- I think I now know why `StringUtils` is not defined inside `namespace OCIO_NAMESPACE`, that's already implied by every place where it's `#include`d. For the above to work, I'll have to define things in just `namespace NumberUtils`.

src/OpenColorIO/fileformats/xmlutils/XMLReaderUtils.h
amyspark: Because of the semantics of strtod, I signal a parsing overrun by returning `std::errc::argument_out_of_domain`. This is checked by https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/3995e744f5b6e18047129dffdbd5e11794941c5c/tests/cpu/fileformats/xmlutils/XMLReaderUtils_tests.cpp#L60-L62

src/utils/NumberUtils.h
amyspark: Update. Doing so causes a conflict when building static OCIO, as CMake wants the helper to be a public library (and doesn't embed it inside OCIO itself).

remia: Hadn't thought about that, we can live it here if there is no issues anyone can see.

remia:
Note: I believe the additional checks showing up in the list are due to the renaming of the jobs in the CI workflow file. We need to update the required list in the repository settings. Or at least that would be my guess.

hodoulp:
The pull request now seems ready but I would like to double-check the performance of this final version before approving. Great job @amyspark

remia:
As discussed [yesterday](https://wiki.aswf.io/display/OCIO/2021-11-29) here is some performance numbers (from different hardware) on the 3 major platforms using the new `ocioperf` tool as it exists on `main` (no longer need an image as input). While probably not strictly needed, I ran each commands twice and kept the second result only, trying to discard system cache.

macOS 11.6 AppleClang 13.0 CXX14

|                   | master (no locale fix) | fast_float | no fast_float |
| ----------------- | ---------------------- | ---------- | ------------- |
| spi1d  [^1]       | 126.306                | 37.7097    | 36.4144       |
| spi3d  [^2]       | 29.0348                | 18.3221    | 20.9059       |
| clf  (1D LUT) [^3]| 44.6314                | 46.5028    | 47.2037       |
| clf (3D LUT) [^4] | 27.537                 | 26.2199    | 29.2086       |

Windows 10 MSVC 19.16.27045.0 CXX14

|                   | master (no locale fix) | fast_float | no fast_float |
| ----------------- | ---------------------- | ---------- | ------------- |
| spi1d  [^1]       | 315.49                 | 65.8565    | 89.2596       |
| spi3d  [^2]       | 30.5407                | 20.152     | 30.1545       |
| clf  (1D LUT) [^3]| 71.519                 | 54.939     | 71.3029       |
| clf (3D LUT) [^4] | 33.8682                | 25.1431    | 35.5455       |

Linux CentOS 7.6 GCC 9.3.1 CXX14

|                   | master (no locale fix) | fast_float | no fast_float |
| ----------------- | ---------------------- | ---------- | ------------- |
| spi1d  [^1]       | 444.353                | 100.728    | 145.077       |
| spi3d  [^2]       | 65.8942                | 50.4903    | 74.5444       |
| clf  (1D LUT) [^3]| 126.986                | 89.4428    | 129.521       |
| clf (3D LUT) [^4] | 72.9861                | 48.4598    | 76.121        |


[^1]: `ocioperf --transform ./testdata/lut1d_3.spi1d --iter 10 --nocache`
[^2]: `ocioperf --transform ./testdata/spi_ocio_srgb_test.spi3d --iter 10 --nocache`
[^3]: `ocioperf --transform ./testdata/clf/lut1d_long.clf --iter 10 --nocache`
[^4]: `ocioperf --transform ./testdata/clf/aces_to_video_with_look.clf --iter 10 --nocache`

hodoulp:
@amyspark The add of a library dependency in the 'core' library is always a challenging discussion because of all the potential impacts on short and long term for OpenColorIO maintainers but also for package managers, external contributors, etc. In that case, the goal is to add these changes to the coming 2.1.1 release so, we cannot add a new dependency. The patch version 2.1.1 is planned for next week, will you have the time to cleanup the code?

amyspark:
@hodoulp What would be, more precisely, the changes you'd like in this MR? And what should be moved to another commit or MR?

hodoulp:
Please remove the `fast_float` dependency i.e. the code must only use the 'custom' implementation.

hodoulp (approved):
Great job

doug-walker (approved):
Thanks for this work amyspark and for your patience during the long process of finding a solution that meets the many requirements/constraints!

remia (approved):
Thanks for the work @amyspark! Just a minor point for coherence about the `sscanf` on Windows but all good to merge for me as well regardless.

src/OpenColorIO/fileformats/FileFormatSpi1D.cpp
```diff
@@ -124,10 +126,26 @@ CachedFileRcPtr LocalFileFormat::read(std::istream & istream,
...
+                if (sscanf(lineBuffer, "From %s %s", fromMinS, 64, fromMaxS, 64) != 2)
```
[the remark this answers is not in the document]
amyspark: Good catch!

PR #1497, opened 2021-09-20
Add support for OpenGL ES as a shading language

### LINKED ISSUE #1486: Support OpenGL ES
Hi,

Again here from the v2 porting effort from Krita. I noticed that OCIO v2 uses 1D textures throughout the OpenGL helpers and GPU shader utilities, which renders OCIO's new GPU pipeline incompatible with OpenGL ES.

Since our HDR support depends on ANGLE, which in turn provides OpenGL ES, is there any way to drop this optimization?

(In the meanwhile, I'll hardcode the legacy pipeline and add some #defines to patch them.)

amyspark:
This commit adds two new entries to `GpuLanguage`,
`GPU_LANGUAGE_GLSL_ES_1_0` and `GPU_LANGUAGE_GLSL_ES_3_0`.

The only meaningful differences w.r.t. stock OpenGL are:

- the 1D texture optimization isn't applied to ES, as they are not
  supported at all;

- the texture<N>D() calls are replaced in GLSL ES 3 by a call to
  texture().

Fixes #1486

### DIFF
--- a/include/OpenColorIO/OpenColorTypes.h
+++ b/include/OpenColorIO/OpenColorTypes.h
@@ -445,6 +445,8 @@ enum GpuLanguage
     GPU_LANGUAGE_GLSL_1_2,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_1_3,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_4_0,          ///< OpenGL Shading Language
+    GPU_LANGUAGE_GLSL_ES_1_0,       ///< OpenGL ES Shading Language
+    GPU_LANGUAGE_GLSL_ES_3_0,       ///< OpenGL ES Shading Language
     GPU_LANGUAGE_HLSL_DX11,         ///< DirectX Shading Language
     LANGUAGE_OSL_1                  ///< Open Shading Language                
 };
--- a/include/OpenColorIO/OpenColorTypes.h
+++ b/include/OpenColorIO/OpenColorTypes.h
@@ -445,10 +445,10 @@ enum GpuLanguage
     GPU_LANGUAGE_GLSL_1_2,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_1_3,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_4_0,          ///< OpenGL Shading Language
+    GPU_LANGUAGE_HLSL_DX11,         ///< DirectX Shading Language
+    LANGUAGE_OSL_1,                 ///< Open Shading Language
     GPU_LANGUAGE_GLSL_ES_1_0,       ///< OpenGL ES Shading Language
     GPU_LANGUAGE_GLSL_ES_3_0,       ///< OpenGL ES Shading Language
-    GPU_LANGUAGE_HLSL_DX11,         ///< DirectX Shading Language
-    LANGUAGE_OSL_1                  ///< Open Shading Language                
 };
 
 enum EnvironmentMode
--- a/src/libutils/oglapphelpers/glsl.cpp
+++ b/src/libutils/oglapphelpers/glsl.cpp
@@ -444,25 +444,27 @@ void OpenGLBuilder::useAllUniforms()
 
 std::string OpenGLBuilder::getGLSLVersionString()
 {
-    if (m_shaderDesc->getLanguage() == GPU_LANGUAGE_GLSL_1_3)
+    switch (m_shaderDesc->getLanguage())
     {
+    case GPU_LANGUAGE_GLSL_1_2:
+        // That's the minimal version supported.
+        return "#version 120";
+    case GPU_LANGUAGE_GLSL_1_3:
         return "#version 130";
-    }
-    else if (m_shaderDesc->getLanguage() == GPU_LANGUAGE_GLSL_4_0)
-    {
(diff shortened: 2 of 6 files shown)


hodoulp:
@amyspark that's great. Thanks a lot for the help.
Did you test the pull request with `Krita` ?

amyspark:
@hodoulp Yeah! The legacy GPU pipeline runs perfectly with OpenGL ES 3 under Angle, there is no need anymore for the `texture<N>D` defines. I'm still to test the new GPU pipeline, though, I got detoured by #1496.

amyspark:
@hodoulp status update, the new GPU pipeline returns completely scrambled results under OpenGL ES:

 Legacy: |  New (and ba52937fd5c39fde2f64d3e149d70413b45baf1e reverted) 
--|--
Captura de pantalla 2021-09-20 164327 | Captura de pantalla 2021-09-20 164607

With and without your 1D LUT optimization it still showed the same glitched appearance, so I'm guessing there's an encoding that's not being properly skipped? Do you know where I could have a deeper look at?

amyspark:
Upon further debugging, I believe the 1D LUT codepath may have a serious bug. I've tested with various input profiles from the ACES configuration and none of them actually trigger the tex1d codepath; however, every profile that does involve a 1D LUT results in those clamp-looking glitches. I've checked with both our OpenGL 4.x pipeline and through Angle, both against current master.

amyspark:
@remia, no, I haven't made any further progress on this. I hadn't thought of ociodisplay because it doesn't get built with our current configuration (we don't use OpenImageIO in Krita).

hodoulp:
Another option is to use `ocioconvert` (but you still need `OpenImageIO`) to compare CPU vs GPU legacy vs GPU converted images. Otherwise, you can provide us with the LUT plus the image so we can test on our side.

```
~/dev/ocio_2/build_rls (master) $ ./src/apps/ocioconvert/ocioconvert --help
ocioconvert -- apply colorspace transform to an image 

usage: ocioconvert [options]  inputimage inputcolorspace outputimage outputcolorspace
   or: ocioconvert [options] --lut lutfile inputimage outputimage
   or: ocioconvert [options] --view inputimage inputcolorspace outputimage displayname viewname


Options:
    --lut                  Convert using a LUT rather than a config file
    --view                 Convert to a (display,view) pair rather than to an output color space
    --gpu                  Use GPU color processing instead of CPU (CPU is the default)
    --gpulegacy            Use the legacy (i.e. baked) GPU color processing instead of the CPU one (--gpu is ignored)
    --gpuinfo              Output the OCIO shader program
    --help                 Print help message
    -v                     Display general information
```

amyspark:
@remia was spot on; it turns out I was using the legacy pipeline's `edgelen` value for the texture size, and not the one returned by OCIO. (For context: we cannot use the classes in `oglapphelpers` because Krita uses Qt's OpenGL facilities, so I ported the calls we needed to our shader code.)

For verification:

OpenGL | OpenGL ES (Angle)
---------|----------------
Captura de pantalla 2021-09-21 222916 | Captura de pantalla 2021-09-21 223013

include/OpenColorIO/OpenColorTypes.h
```diff
@@ -445,6 +445,8 @@ enum GpuLanguage
     GPU_LANGUAGE_GLSL_1_2,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_1_3,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_4_0,          ///< OpenGL Shading Language
+    GPU_LANGUAGE_GLSL_ES_1_0,       ///< OpenGL ES Shading Language
+    GPU_LANGUAGE_GLSL_ES_3_0,       ///< OpenGL ES Shading Language
```
hodoulp, line 449: No. Please add the new enums at the end to avoid a Public API break (which increases the chance to have the ES support in 2.1; otherwise, it will be 2.2. i.e. 2023!)

src/libutils/oglapphelpers/glsl.cpp
```diff
@@ -452,6 +452,14 @@ std::string OpenGLBuilder::getGLSLVersionString()
     {
         return "#version 400 core";
     }
+    else if (m_shaderDesc->getLanguage() == GPU_LANGUAGE_GLSL_ES_1_0)
```
hodoulp, line 455: [Not from your branch] Please replace the `ifs` by a `switch` so a new enum value generates a build break i.e. easier to see where to add something.

hodoulp (changes requested):
Thanks for the contribution.

amyspark:
@hodoulp, should I add support for using the classes in oglapphelpers with GLES?

hodoulp:
@amyspark Yes, that would be interesting to have `oglapphelpers` with GL ES support. For long-term, the most important would be to have OpenGL ES unit tests like `[]/tests/gpu`. But this pull request could be merged as-is for short-term because it's already a major improvement for `OpenColorIO`.

amyspark:
@hodoulp, it seems CMake doesn't provide a Find module for the ES equivalent of FindOpenGL, and adding support proper to the app helpers needs to re-detect from the ground up the needed dependencies.

It'll be certainly interesting! but for another MR in the future. For now, I'll wrap it up here :smiley:

hodoulp (approved):

remia (approved):

PR #1499, opened 2021-09-21
Port from master to RB-2.1 - FindExpat: allow usage of expat 2.2.9 (#1483)

### LINKED ISSUE #1482: OCIO requires Expat 2.2.8+, but its Find module only works for Expat 2.2.10 and higher

hodoulp:
- alias expat::expat to expat::libexpat if the former is missing
- patch INTERFACE_INCLUDE_DIRECTORIES if the property is missing

Fixes #1482.

Signed-off-by: L. E. Segovia <13498015+amyspark@users.noreply.github.com>

Co-authored-by: Patrick Hodoul <patrick.hodoul@autodesk.com>

### DIFF
--- a/share/cmake/modules/Findexpat.cmake
+++ b/share/cmake/modules/Findexpat.cmake
@@ -34,7 +34,40 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
     endif()
 
     if(expat_FOUND)
+        if (TARGET expat::libexpat)
+            message(STATUS "Expat ${expat_VERSION} detected, aliasing targets.")
+            add_library(expat::expat ALIAS expat::libexpat)
+        endif()
+
+        get_target_property(expat_INCLUDE_DIR expat::expat INTERFACE_INCLUDE_DIRECTORIES)
+
         get_target_property(expat_LIBRARY expat::expat LOCATION)
+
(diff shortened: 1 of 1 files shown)


hodoulp:
Same DCO issue than #1501

amyspark:
@hodoulp re DCO, I think you could amend the first commit with the correct author line (as listed in the Signed-off-by) and force push.

michdolan (approved):

Issue #1420, opened 2021-06-22
PNG Colorspace is wrong


zhoub:
Specified the target colorspace to srgb for ocioconvert command line as this
```
$ ocioconvert input.exr linear output.png srgb
```
There is sRGB LUT, image was generated properly, but by checking colorprofile by imagick as
```
magick identify -verbose output.png | grep Colorspace
```
The captured value is RGB, it should be sRGB or the colorprofile specified by ocioconvert.

lgritz:
This is probably an OIIO bug, actually.

lgritz:
On second thought, looking at OIIO code, it looks like it tries to do the right thing. I'm investigating.

lgritz:
I think it's a bug in ocioconvert.

Somewhere, there should be a

    spec.attribute("oiio:ColorSpace", new_color_space_name);

before the `f->open(outputimage, spec)` call.

What's happening is that when reading exr, the spec that's read will have the "oiio:ColorSpace" attribute set to "Linear", since exr's are presumed to be linear. Then that same spec is passed to the output -- but not modified by ocioconvert to change it to "sRGB" -- and so OIIO's png output is seeing that it's set to "linear" rather than "srgb" and therefore not putting the nugget it the PNG file that identifies it as sRGB.

zhoub:
Did a commit, verified and it generated proper .png file
```
Colorspace: sRGB
  Type: TrueColorAlpha
    oiio:ColorSpace: sRGB
```

```
From 2638bac3af5eef72a24828f86d6cf63598080511 Mon Sep 17 00:00:00 2001
From: Bo Zhou <Bo.Schwarzstein@gmail.com>
Date: Thu, 15 Jul 2021 16:21:55 +0900
Subject: [PATCH] Write converted image with output colorspace (#1420)

Signed-off-by: Bo Zhou <Bo.Schwarzstein@gmail.com>
---
 src/apps/ocioconvert/main.cpp | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/src/apps/ocioconvert/main.cpp b/src/apps/ocioconvert/main.cpp
index b56d499..9d7bc03 100644
--- a/src/apps/ocioconvert/main.cpp
+++ b/src/apps/ocioconvert/main.cpp
@@ -616,6 +616,8 @@ int main(int argc, const char **argv)
             exit(1);
         }
 
+        spec.attribute("oiio:ColorSpace", outputcolorspace);
+
[...]

PR #1500, opened 2021-09-22
Write converted image with output colorspace (#1420)

zhoub:
Signed-off-by: Bo Zhou <Bo.Schwarzstein@gmail.com>


src/apps/ocioconvert/main.cpp
```diff
@@ -625,6 +625,8 @@ int main(int argc, const char **argv)
             exit(1);
         }
 
+        spec.attribute("oiio:ColorSpace", outputcolorspace);
```
hodoulp, line 628: There are three different conversions: 1) using `inputcolorspace` and `outputcolorspace`, 2) using a LUT file, and 3) using `inputcolorspace` and a (`display`, `view`) pair. Only the first conversion can add the `oiio:ColorSpace` metadata so, the code needs to check if `outputcolorspace` is not null.

hodoulp: @lgritz is there a way with the `OpenImageIO` metadata to save the (`display`, `view`) pair?

hodoulp (changes requested):
Thanks a lot for your contribution.
There is only a small change needed to improve the proposal.

src/apps/ocioconvert/main.cpp
lgritz: You can ask it to save any name/data you want. But there is not an established convention for encoding anything other than simply the name of the color space that the pixels are believed to be in at the point they are saved. (Also note that the ability to actually honor the request to save any of this depends on the specific file format being used. OpenEXR is pretty flexible with arbitrary metadata, most other file formats are not.)

lgritz: So my understanding is that at this point, the pixels aren't in a named color space that's in the config per se.

Is there an established OCIO convention for using a string to denote a (display, view) combination? And what could we do with that if it was stored in the file? Is there any OCIO API call that can use that as a starting point? Like, can you say "these pixels are values in this view of this display, now transform it into some other space?"

hodoulp: There is no OCIO approved convention (but Maya has one). 
That's a good point to bring to the next OCIO TSC meeting (I will add it to the agenda)

hodoulp (approved):

hodoulp:
Thanks @zhoub for your contribution.

src/apps/ocioconvert/main.cpp
zachlewis: I would love if OCIO codified such a convention -- out of curiosity, what convention does Maya use? 

Related, I think it would be pretty useful if we had Display and View FormatMetadata enums, to complement the existing CLF InputDescriptor and OutputDescriptor enums. (Granted, "OutputDescriptor" can be anything; but it would be great to have a semi-formal convention for differentiating between color space conversion and display-view transforms, such that a system with a-priori knowledge of Display descriptors -- such as an OCIO config, for example -- would ostensibly be able to programmatically isolate the "Display" portions of the transform from the "View" portions, if that makes sense)

doug-walker: One of the many motivations for the changes to the Display/View system in OCIO v2 was to have a more generic display color space available for this sort of tagging.  So if it used a v2 style Display/View, you don't care about the view and would just write the display color space.  If it's a v1 style Display/View, then by definition there is already a color space in the config that implements it and the name of that could be written.  However, since that color space would incorporate the view transform, it may not be that meaningful.

src/apps/ocioconvert/main.cpp
```diff
@@ -625,6 +625,11 @@ int main(int argc, const char **argv)
             exit(1);
         }
 
+        if (outputcolorspace)
+        {
+            spec.attribute("oiio:ColorSpace", outputcolorspace);
+        }
+
```
doug-walker, line 632: You could improve this by adding an "else if (useDisplayView)" block that basically does this:
outputcolorspace = getDisplayViewColorSpaceName(display, view);

zhoub: Sorry I'm not familiar with the detailed API. The issue and PR here are just to generate a "proper enough" JPG||PNG with colorspace, so that with most of system default viewers could display the image correctly.

The display and view are 2 strings, so I'd prefer to set 2 extra metadata fields `ocio:display` and `ocio:view` to hold the vlaues.

But for `outputcolorspace`, this is single value to allow OIIO to generate actual file, what's the best value ? Display or view ? I prefer the value of display, it's more generic than view.

WDYT ?

doug-walker: I like your goal of having "most of system default viewers could display the image correctly", but unfortunately I don't follow how this would work.  The name being written here is the name from the user's OCIO config file.  These names are not standardized.  So how would the "system default viewers" act on this information?  I guess you are referring to operating system included viewers such as Preview on MacOS?  As far as I know, none of these are using OCIO, so you would need to tag the image with a string that would be meaningful to these other apps.  Is there a standard for that that you had in mind?  Do you have a prototype of this working in some app?

Regarding your suggestion of adding 2 new metadata fields for ocio:display and ocio:view, could you elaborate on how you expect these would be used by downstream software and which file formats would be able to make use of this information?

As I wrote, you should use getDisplayViewColorSpaceName to set the outputcolorspace for the oiio:ColorSpace field.  As I wrote in the issue, the display is more relevant than the view, but I would not use the raw display string since this may not actually be a color space.  Furthermore, in many configs it is not even used as a "display".  For example, in the popular 1.x ACES configs it is "ACES".

Thanks!

zhoub: `
I like your goal of having "most of system default viewers could display the image correctly", but unfortunately I don't follow how this would work. The name being written here is the name from the user's OCIO config file. These names are not standardized. So how would the "system default viewers" act on this information? I guess you are referring to operating system included viewers such as Preview on MacOS? As far as I know, none of these are using OCIO, so you would need to tag the image with a string that would be meaningful to these other apps. Is there a standard for that that you had in mind? Do you have a prototype of this working in some app?
`
I think you didn't try the issue #1420, please have a try at different platforms with and without modification, then you would see the difference. As the issue reported, specified sRGB as output colorspace, it shouldn't write Linear as metadata.

`
Regarding your suggestion of adding 2 new metadata fields for ocio:display and ocio:view, could you elaborate on how you expect these would be used by downstream software and which file formats would be able to make use of this information?
`
This is not going to be used by any software for now, just as metadata for the generated image, just as EXIF information when you take a photo. Assuming user generates several different PNGs from single EXR with difference display+view, metadata would be safer and easier for debugging purpose. But I think this should be done from another issue/PR as a feature.

`
As I wrote, you should use getDisplayViewColorSpaceName to set the outputcolorspace for the oiio:ColorSpace field. As I wrote in the issue, the display is more relevant than the view, but I would not use the raw display string since this may not actually be a color space. Furthermore, in many configs it is not even used as a "display". For example, in the popular 1.x ACES configs it is "ACES".
`
This is fine. I did the commit.

https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1500/commits/beccd66e3d1974193e18817c825eb7584f5057ee

Thanks.

zachlewis: > Regarding your suggestion of adding 2 new metadata fields for ocio:display and ocio:view, could you elaborate on how you expect these would be used by downstream software and which file formats would be able to make use of this information?

Providing both the display and view permits downstream software to defer the choice of how "far" to invert the baked-in rendering to the user; i.e., whether to decode to an output-referred image state or a scene-referred state (render-time look / data overrides notwithstanding, I suppose).  Even if the view inversion is potentially lossy, there are absolutely cases where the need to work with scene-referred relative exposure values trumps the need for 100% faithful reproduction. 

Display / View inversion is also extremely convenient for previewing multiple media sources of differing image states under a single user-selected Display / View (which, under the right conditions, might end up bypassing any lossy inversions along the way). 

Also, as Bo says, persisting both Display and View metadata is extremely useful for debugging. In a world where everyone is working remotely, with potential technical and legal constraints on one's ability to share what they're seeing and how they're viewing, when someone says "Hey, my dailies renders aren't matching the editorial reference", or "Something changed, and now everything looks wrong", being able to quickly ascertain both the Display and View used to produce the image in question (without having to dig through render logs or earlier versions of comp scripts) is worth its weight in gold. 

Lastly, always considering only the Display will likely produce unexpected and inconsistent behavior across v1- and v2-style configs, where imagery produced with v1 configs will always decode to a scene-referred state by default, while images produced with v2-style Display View transforms will always decode to a display-referred state. Personally, for this reason alone, I'd have expected downstream software's decoding behavior to default to fully inverting both the View and the Display, as default behavior discrepancies can be extremely difficult to track down and debug.

Just my two cents : )

doug-walker: Thanks for the update.  When you wrote "most system default viewers" were you referring to ImageMagick?  I'm curious if this fix also improves behavior in any other viewers you tested.

doug-walker: Zach, you raise interesting points, but it seems you are assuming the downstream software will be using OCIO, which seems to be a different use-case than what zhoub had in mind.  If the intent is to write separate display and view strings with the intent of accurately inverting this downstream, perhaps the config path and context variables also need to be saved for that to be meaningful?  In any case, I have no objections to adding display, view and other descriptive metadata like that, even with the understanding that it won't make it into most formats (OpenEXR only?).

Having faced many issues in the past with image format metadata being interpreted differently by downstream apps, I find the concept of inverting a display/view to be incredibly ambitious and still feel that it would be more realistic to simply get apps to agree on how to display the current image correctly on a monitor.  :)  But to be clear, the current PR does not even do that, all it does is prevent OIIO from incorrectly tagging it as "Linear".

doug-walker: I thought your comments added some interesting considerations that I hadn't thought of, so thank you for the posts Zach (definitely not a derailment!).  

I agree with you that the PR is ready to be merged but was waiting to hear if there are any other comments from people.  Also would be interested to hear more from Bo about which apps he tested his work in.

michdolan (approved):

src/apps/ocioconvert/main.cpp
KelSolaar: For reference purposes, and because I mentioned it on Slack here is what the VES Transfer/Interchange Spec recommends:

image

KevinJW:
As others have mentioned I'm a little cautious about doing things with displays and views...
```
displays:
  P3:
 - !<View> {name: MYSHOW, colorspace: AlexaV3LogC, looks: MYSHOW_P3}
 Rec709:
 - !<View> {name: MYSHOW, colorspace: AlexaV3LogC, looks: MYSHOW_Rec709}
```
In this made up example, I'm replicating a typical way of using client supplied 3D LUTs as display transformations. It would be common to supply a LUT that takes camera encodigns and transform them to a display, in this example the colour space is the same, but the images would be quite different.

remia:
Good point @KevinJW, I think another variant of this can be found in the [filmic blender config](https://github.com/sobotka/filmic-blender/blob/master/config.ocio).

zachlewis:
Maybe I’m misunderstanding Kevin’s caution… wouldn’t this ambiguity (different displays that share common view names) be an argument in favor of tracking both the display and the view in metadata?
________________________________

hodoulp:
Already approved, and also discussed & approved during the last TSC meeting (refer to the [meeting notes](https://wiki.aswf.io/display/OCIO/2021-10-04)).

PR #1503, opened 2021-09-26
Remove Python wheel workflow running on every pull request

remia:
This contains changes to the GH workflows, specifically related to Python wheels plus minor fixes:
* Wheel workflow set to be run nightly, on tags named `v2.*` and on pull request **only** when specifics files are updated, mainly the `wheel_workflow.yml` (this is why wheels check will run for this pull request),
* Adding early support for Python 3.10, this triggered some issues that were fixed related to `collections` Python module updates and the use of raw `ssize_t` failing on MSVC. macOS is not supported currently due to an issue building `numpy` (it could be bypassed by not using it but this will probably solve itself when 3.10 is released and numpy ships wheel for it),
* Adding wheels for Linux ARM 64bits, this takes up to 2hrs per wheel to build instead of the usual 10-20min so it could be removed if we feel it's not needed,
* Avoid duplicated CI workflow checks when a pull_request is opened from a local branch (however, these will still shows as 3 skipped checks in the pull request, one per OS matrix).

I'm planning to look at Analysis nightly build in a subsequent pull request.

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -47,6 +47,10 @@ jobs:
        sse=${{ matrix.use-sse }}, 
        cxx=${{ matrix.cxx-standard }}, 
        docs=${{ matrix.build-docs }}>'
+    # Avoid duplicated checks when a pull_request is opened from a local branch.
+    if: |
+      github.event_name == 'push' ||
+      github.event.pull_request.head.repo.full_name != github.repository
     # GH-hosted VM. The build runs in CentOS 7 'container' defined below.
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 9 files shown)


michdolan:
Why are the Linux ARM 64bit builds so much slower than others?

michdolan (approved):
Thanks @remia !

remia:
@michdolan I followed [instructions](https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation) from `cibuildwheel` to build those Linux ARM wheels, I believe the architecture emulation through QEMU is probably introducing the slowdown. I don't really know whether it would be used that much but looking at `numpy` for comparison, they [provide](https://pypi.org/project/numpy/#files) it.

One related thing I would be curious is whether we can extract some download statistics out of PyPI but it looks like my account doesn't have Manager rights so I can't check (though I didn't found anything on the dev PyPI IIRC).

hodoulp (approved):
LGTM

hodoulp:
@remia I approved and merged this pull request but I would like you to care of the back port to RB-2.1 in case there is some changes. Let me know if that's fine.

remia:
Thanks @hodoulp, I'll look at this tomorrow (I don't expect changes).

PR #1513, opened 2021-10-07
chore: update to new cibuildwheel location

henryiii:
Cibuildwheel is now part of the PyPA. The old name still works, but this is better, and if you use dependabot, it works better with the correct name.


remia:
Thanks for the patch @henryiii. It looks like your commit was not signed off which is required for this project, would you be willing to fix that (more details [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1513/checks?check_run_id=3822946292)) ?

PR #1515, opened 2021-10-09
Update analysis workflow

remia:
This contains some update to the nightly build workflow and related areas:
* Update minimum versions of yaml-cpp to 0.7.0 and expat to 2.4.1
* Add analysis jobs running under `macos-latest` and `windows-latest`
* Allow for build of external projects on the latest available version with `OCIO_INSTALL_EXT_PACKAGES_LATEST`
* Fix external project build on Windows limited to debug only
* Prevent wheel workflow nightly to run on forks

This is still a draft but pushing it if there is any objection to the way it's headed.


remia:
I wonder if we should update `yaml-cpp` and `expat` minimum version or just make sure we support newer version seamlessly. Also don't remember if any decision was made on how to handle dependency version update for new releases (major, minor or patch releases with probably different constraints).

Also noting that with this current PR, yaml-cpp 0.6.3 will still work (if we were to leave the FindExt.cmake to 0.6.3 or above) but unit test will fail because of the trailing whitespace differences.

hodoulp:
As agreed by TSC, that's a go for a port in `RB-2.0` when ready.

hodoulp:
> About the latest usage, currently it was only planned to be used for the nightly analysis build, are we saying we want to enable that for Imath? I'm not sure that would be something we want as the same library version could compile one day and the next fail due to some new Imath version introducing incompatible changes, or did I misinterpreted what was said?

Yes. Only the nightly analysis CI build should use the latest version of the third-party libraries.

My point is related to the goal of the `latest` CI build. What are libraries to validate?

As mentioned during the last TSC meeting and in previous comments from this pull request maintaining a range of versions for third-party libraries could be complex on long-term (I took the example of the yaml version conflicts I faced when updating that library). So, we agreed to stick to a specific version for the mandatory third-party libraries not part of the ASWF ecosystem (such as `expat`, `yaml-cpp`, etc.). BUT the `latest` CI build must take to the latest version for third-party libraries part of the ASWF ecosystem such as `Imath`, etc.

@michdolan @remia In conclusion I suggest that the `latest` CI build only validates OpenColorIO with the latest `Imath`, `OpenImageIO` and `OpenShadingLanguage` libraries.

.github/workflows/analysis_workflow.yml
```diff
@@ -96,6 +96,8 @@ jobs:
           share/ci/scripts/linux/install_openexr.sh latest
           share/ci/scripts/linux/install_imath.sh latest
```
michdolan, line 97: This will pre-install the latest Imath tag, which we are then doing again below (to a different location) with `-DOCIO_INSTALL_EXT_PACKAGES=ALL` and `-DOCIO_INSTALL_EXT_PACKAGES_LATEST=ON`. We will also be installing expat, lcms2, yaml-cpp, pystring, pybind11, and openfx twice with this current setup. Previously we used `-DOCIO_INSTALL_EXT_PACKAGES=NONE` for this workflow because these were all pre-installed with the scripts. If `-DOCIO_INSTALL_EXT_PACKAGES_LATEST=ON` is preferred, I would suggest only using the above install scripts for optional dependencies (and perhaps their dependencies, which may overlap with OCIO) which aren't installed by OCIO. Or, use `-DOCIO_INSTALL_EXT_PACKAGES=MISSING` to avoid duplicates where possible.

hodoulp: Perhaps the two defines approach is not the optimal one. To implement the right behavior, the code could support a fourth option dedicated to `CI latest` ?

remia: I could remove some of the dependencies install script but not much because OIIO need quite a lot of what OCIO already uses. I can remove all remaining `OCIO_INSTALL_EXT_PACKAGES_LATEST` stuff and only rely on install script, hoping that they will work on all 3 platforms (to that end we may want to move them to some non platform dependent sub folder as it seems it would only be copy and paste port). The advantage of `OCIO_INSTALL_EXT_PACKAGES_LATEST` was that it doesn't require any additional scripts to work but we already removed most of that and there might not be any use case out of this analysis workflow file.

remia: I'll try to have something ready this week to address this point.

doug-walker (approved):

hodoulp (changes requested):
Following comment from @michdolan, I'm now waiting for the new commit from @remia.

.github/workflows/analysis_workflow.yml
remia: I pushed an update implementing the requested, change. As before analysis workflow temporarily enabled on PR to validate the changes.

hodoulp (approved):
LGTM

hodoulp:
@michdolan do you want to have a look before the merge?

hodoulp:
When thinking about a patch version in a week, the pull request now needs to be merged to not delay too much the other pull requests.

Issue #1314, opened 2021-02-16
Improve ocioperf by removing the OpenImageIO (OIIO) dependency


hodoulp:
The command line tool `ocioperf` helps to investigate any performance hits independently from its integration in a host (Maya, Nuke, etc.). This first version provides the support to read an image (using OIIO). That's useful but it unfortunately increases the cyclic dependency problem between OpenColorIO and OpenImageIO.

The proposal is to replace an arbitrary external image by a in-memory generated one in float with an arbitrary value range (i.e. [0.0, 1.0], [-1.0, +2.0], etc.).

hodoulp:
Hi @kanakgautam ,
Happy to see your interest in the OpenColorIO projet.

The change consists to remove the dependency to OpenImageIO from `ocioperf`, and replace it by a generated in-memory image. If you have interest in that task, the first step is to build the OpenColorIO library on your machine, and run unit tests. 

When successful, you can then remove the dependency to OpenImageIO, and generate an in-memory image. It means a RGBA float array with values [-1.0f, 2.0f] and a size of 1024. Both are applicable to the four color components. 

As extra, you can add command line arguments to specify the range (e.g. --range -1.0 2.0) and the number of values (e.g. --values 1024). 

In the context of the `Google Summer of Code` the two first steps are good enough to comply with the OpenColorIO prerequisite.

kanakgautam:
Thanks a lot @hodoulp, I'll try to follow the steps you mentioned and send a PR asap. I'll let you know if I face any issues.

doug-walker:
Closing.  This was fixed in PR #1516.

PR #1516, opened 2021-10-14
Adsk Contrib - Default to C++14 and remove OIIO from ocioperf

hodoulp:

The pull request moves the default C++ version from 11 to 14 as discussed (but C++11 version remains supported) and removes the OpenImageIO dependency from the `ocioperf` command line tool (refer to #1314).

The `ocioperf` changes/improvements are:
- it uses a 4K image from [-1.0f, 2.0f] in float case.
- user can disable optimizations (default is false)
- user can specify the input and output bit-depths (defaults are float)


michdolan (approved):

hodoulp:
Following discussions between @doug-walker and me we would like to improve the image generation (i.e. the slow gradient proposal does not capture well typical images). Stay tuned.

hodoulp:
Sorry for the delay but we now have a better image generation.

hodoulp:
@remia @michdolan There are still four Linux CI builds pending. Any idea? Can I still merge the pull request?

src/apps/ocioperf/main.cpp
```diff
@@ -573,101 +459,197 @@ int main(int argc, const char **argv)
...
+        std::vector<float> img_f32_ref;
+        std::vector<uint16_t> img_ui16_ref;
+
+        // Generate a synthetic image by emulating a LUT3D identity algorithm that steps through
+        // many different colors.  Need to avoid a constant image, simple gradients, or anything
+        // that would result in more cache hits than a typical image.  Also, want to step through a 
+        // wide range of colors, including outside [0,1], in case some algorithms are faster or
+        // slower for certain colors.
+
+        static constexpr size_t length   = 64;
```
remia, line 487: [Question] How was this size picked? 64^3 is way lower than 3840x2160 so I think you have lot of repetition in the image but that might be by design?

remia:
> @remia @michdolan There are still four Linux CI builds pending. Any idea? Can I still merge the pull request?

I'm fine with merging on my side, it seems like a GitHub Action internal error.

src/apps/ocioperf/main.cpp
doug-walker: Good point Remi, I would suggest increasing it to 201.

doug-walker (approved):

PR #1520, opened 2021-10-21
Adds Metal Shading Language (MSL) Generation support

Morteeza:
Signed-off-by: Morteza Mostajabodaveh <smostajabodaveh@apple.com>

Co-authored-by: Ingthor Hjalmarsson <ihjalmarsson@apple.com>


src/OpenColorIO/GpuShaderUtils.h
```diff
@@ -6,16 +6,18 @@
 #define INCLUDED_OCIO_GPUSHADERUTILS_H
 
 #include <sstream>
+#include <utility>
+#include <vector>
```
hodoulp, line 10: I seems that these includes can be removed. In fact the complete file can be reverted.

src/OpenColorIO/GpuShaderUtils.cpp
```diff
@@ -2,14 +2,16 @@
 // Copyright Contributors to the OpenColorIO Project.
 
 #include <math.h>
+#include <vector>
+#include <iomanip>
+#include <string>
```
hodoulp, line 8: Are these includes still needed?

hodoulp (changes requested):
Last cleanups, but the pull request is now ready.

src/OpenColorIO/GpuShaderUtils.cpp
Morteeza: Last two can be removed. But vector is still needed.

src/OpenColorIO/GpuShaderUtils.h
Morteeza: That's right. One less file in the pull request :+1:

hodoulp:
@Morteeza If the CI builds all succeed, I approve the pull request :smile:

Morteeza:
> @Morteeza If the CI builds all succeed, I approve the pull request 😄

I hope it does! :D 
Thanks a lot for the tests and your very helpful comments.

hodoulp (approved):
Great Work!

Morteeza:
> Great Work!

Thanks @hodoulp. We still can't merge. We need @Simran-B 's approval, too.

hodoulp:
A merge can only be done if two 'official' reviewers approved the pull request. I will contact the others to have the second approval.

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -239,6 +282,110 @@ const char * GpuShaderCreator::getCacheID() const noexcept
...
+    
+    kw.indent();
+    std::string separator = "";
+    for(const auto& param : m_functionParameters)
+    {
+        kw.newLine() << separator << param.type << " " << param.name;
+        inPixelNeedsComma = true;
+        separator = ", ";
+    }
+    kw.newLine() << (inPixelNeedsComma ? "," : "") << kw.float4Keyword() << " inPixel)";
```
remia, line 361: [Minor] Can replace `","` by `", "` for easier read?

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -294,6 +441,133 @@ void GpuShaderCreator::createShaderText(const char * shaderDeclarations,
...
+}
+
+std::string MetalClassWrappingInterface::getClassWrapperFooter(const std::string& originalFooter)
+{
+    GpuShaderText st(GPU_LANGUAGE_MSL_2_0);
+    
+    st.newLine();
+    generateClassWrapperFooter(st, m_functionName);
+    
+    std::string classWrapFooter = "\n// close class wrapper\n\n";
```
remia, line 565: [Minor] Can upper case first letter of close.

remia (commented):
Thanks for the work @Morteeza, I didn't found anything wrong with the code, only minor points. The style changes can also be addressed in a follow-up PR as you already mentioned. I hope we will be able to get rid of the manual declaration parsing code in the near future. Did you test the generated code in actual GPU processing code?

src/OpenColorIO/ParseUtils.cpp
```diff
@@ -257,14 +257,15 @@ const char * GpuLanguageToString(GpuLanguage language)
...
+        case GPU_LANGUAGE_CG:           return "cg";
+        case GPU_LANGUAGE_GLSL_1_2:     return "glsl_1.2";
+        case GPU_LANGUAGE_GLSL_1_3:     return "glsl_1.3";
+        case GPU_LANGUAGE_GLSL_4_0:     return "glsl_4.0";
         case GPU_LANGUAGE_GLSL_ES_1_0:  return "glsl_es_1.0";
         case GPU_LANGUAGE_GLSL_ES_3_0:  return "glsl_es_3.0";
-        case GPU_LANGUAGE_HLSL_DX11: return "hlsl_dx11";
-        case LANGUAGE_OSL_1: return "osl_1";
+        case GPU_LANGUAGE_HLSL_DX11:    return "hlsl_dx11";
+        case GPU_LANGUAGE_MSL_2_0:      return "msl_metal";
```
remia, line 267: Please also update this to `msl_2`

src/OpenColorIO/ParseUtils.cpp
```diff
@@ -276,13 +277,14 @@ GpuLanguage GpuLanguageFromString(const char * s)
...
+    else if(str == "glsl_1.2")    return GPU_LANGUAGE_GLSL_1_2;
+    else if(str == "glsl_1.3")    return GPU_LANGUAGE_GLSL_1_3;
+    else if(str == "glsl_4.0")    return GPU_LANGUAGE_GLSL_4_0;
     else if(str == "glsl_es_1.0") return GPU_LANGUAGE_GLSL_ES_1_0;
     else if(str == "glsl_es_3.0") return GPU_LANGUAGE_GLSL_ES_3_0;
-    else if(str == "hlsl_dx11") return GPU_LANGUAGE_HLSL_DX11;
-    else if(str == "osl_1") return LANGUAGE_OSL_1;
+    else if(str == "hlsl_dx11")   return GPU_LANGUAGE_HLSL_DX11;
+    else if(str == "osl_1")       return LANGUAGE_OSL_1;
+    else if(str == "metal")       return GPU_LANGUAGE_MSL_2_0;
```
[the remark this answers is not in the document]
Morteeza: > Thanks for the work @Morteeza, I didn't found anything wrong with the code, only minor points. The style changes can also be addressed in a follow-up PR as you already mentioned. I hope we will be able to get rid of the manual declaration parsing code in the near future. Did you test the generated code in actual GPU processing code?

Hi @remia, 
Thanks a lot for your review. I applied the changes you suggested.
Previously we were not parsing the declaration part to extract variables, but since we wanted to keep this change as less intrusive as possible and not touch the interfaces defined in `OpenColorIO.h`, we decided to go this way. 

Also, yes. I have tried a few shaders generated with opencolorio changes and it seems to work fine. Just as an example, here you can see the input image is color adjusted with `spi-vfx` config from linear space to srgb8.

in_Texture
out_Texture

src/OpenColorIO/ParseUtils.cpp
Morteeza: oops... good catch... sorry I missed that. Thanks for pointing it out.

Also, we decided to go with shading language name instead of shading language and api name.

Morteeza:
> Thanks for the work @Morteeza, I didn't found anything wrong with the code, only minor points. The style changes can also be addressed in a follow-up PR as you already mentioned. I hope we will be able to get rid of the manual declaration parsing code in the near future. Did you test the generated code in actual GPU processing code?

Thanks @remia for the review.

Previously, we weren't extracting variables from declaration block, but to keep the metal code generation only in OCIO implementation and not touch `OpenColorIO.h`, we decided to do it this way.

Also, we have tried the generated code with a few configs in a standalone Metal application, and it works fine. Here you can see how an input image color is corrected using `spi_vfx` config from linear space to srgb8:
in_Texture
out_Texture

michdolan (commented):
Thanks for your contribution @Morteeza ! Some architectural questions, which as mentioned could be reworked in later PRs if needed.

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -40,6 +73,8 @@ class GpuShaderCreator::Impl
     std::string m_shaderCodeID;
 
     std::vector<DynamicPropertyRcPtr> m_dynamicProperties;
+    
+    std::unique_ptr<MetalClassWrappingInterface> m_classWrappingInterface;
```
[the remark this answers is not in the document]
Morteeza: I can rename `m_classWrappingInterface` to something more clear. At the moment all use cases of this variable in enclosed in `if (getLanguage() == GPU_LANGUAGE_MSL_2_0))`.

Morteeza:
> Thanks for your contribution @Morteeza ! Some architectural questions, which as mentioned could be reworked in later PRs if needed.

@michdolan Thanks for your review. I applied suggested changes to the code.

src/OpenColorIO/ParseUtils.cpp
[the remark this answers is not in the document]
hodoulp: > > @remia  The style changes can also be addressed in a follow-up PR as you already mentioned.

> @Morteeza Previously we were not parsing the declaration part to extract variables, but since we wanted to keep this change as less intrusive as possible and not touch the interfaces defined in OpenColorIO.h, we decided to go this way.

Yes. The idea was to try having the MSL support for the next patch release i.e. no ABI break. That's the only way I found for short-term but that's clearly not a long-term approach.

src/OpenColorIO/GpuShaderDesc.cpp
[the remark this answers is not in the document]
hodoulp: > @michdolan  Is that the intent?

Yes that's the long term approach as the OSL translation always needs a version of this class. But the short-term was to have something so @Morteeza can focus on the MSL unit test framework and in parallel, we can slowly enhance this new concept (and perhaps, change the API to make it more robust).

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -114,6 +114,10 @@ void GpuShaderCreator::setLanguage(GpuLanguage lang) noexcept
     {
         getImpl()->m_classWrappingInterface = std::unique_ptr<MetalShaderClassWrapper>(new MetalShaderClassWrapper);
     }
+    else
+    {
+        getImpl()->m_classWrappingInterface = std::unique_ptr<NullGpuShaderClassWrapper>(new NullGpuShaderClassWrapper);
+    }
```
hodoulp, line 120: The assignation operator of `GpuShaderCreator::Impl` must also be updated to create the right wrapper instance.

Morteeza:
@hodoulp Both requested changes are applied in the latest revision.

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -74,12 +74,27 @@ class GpuShaderCreator::Impl
...
+    {
+        if(m_language == GPU_LANGUAGE_MSL_2_0)
+        {
+            m_classWrappingInterface = std::unique_ptr<MetalShaderClassWrapper>(new MetalShaderClassWrapper);
+        }
+        else
+        {
+            m_classWrappingInterface = std::unique_ptr<NullGpuShaderClassWrapper>(new NullGpuShaderClassWrapper);
+        }
+    }
```
hodoulp, line 97: The usual design is to create a factory method:
```
static std::unique_ptr<GpuShaderClassWrapper> CreateClassWrapper(m_language)
{
   switch language)
   {
       case MSL: return [...]
       [..]
   }
}
```
The use of `switch` helpers to detect missing enum case(s) at compile time (thanks to compiler's flag).

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -110,14 +125,7 @@ void GpuShaderCreator::setLanguage(GpuLanguage lang) noexcept
...
     getImpl()->m_language = lang;
-    if(lang == GPU_LANGUAGE_MSL_2_0)
-    {
-        getImpl()->m_classWrappingInterface = std::unique_ptr<MetalShaderClassWrapper>(new MetalShaderClassWrapper);
-    }
-    else
-    {
-        getImpl()->m_classWrappingInterface = std::unique_ptr<NullGpuShaderClassWrapper>(new NullGpuShaderClassWrapper);
-    }
+    getImpl()->updateClassWrappingInterface();
```
hodoulp, line 128: So, the code is then
`getImpl()->m_classWrappingInterface = Impl::CreateClassWarpper(m_language);`

hodoulp (changes requested):
Thanks for the update.

src/OpenColorIO/GpuShaderDesc.cpp
Morteeza: @hodoulp Thanks for the suggestion. I applied the requested change.

src/OpenColorIO/GpuShaderClassWrapper.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+bool MetalShaderClassWrapper::operator=(const GpuShaderClassWrapper& rhs)
+{
+    if(auto* metalShader_rhs = dynamic_cast<const MetalShaderClassWrapper*>(&rhs))
+    {
+        *this = *metalShader_rhs;
+        return true;
+    }
+    return false;
+}
```
hodoulp, line 248: I'm still do not understand the need of this method. The signature itself is quite confusing (too close to the assignation operator), and the equal operator cannot return a boolean.
I think that you should remove this method (and the ones in the derived classes).

src/OpenColorIO/GpuShaderDesc.cpp
```diff
@@ -70,12 +74,35 @@ class GpuShaderCreator::Impl
             m_functionHeader = rhs.m_functionHeader;
             m_functionBody   = rhs.m_functionBody;
             m_functionFooter = rhs.m_functionFooter;
+            
+            m_classWrappingInterface = CreateClassWrapper(m_language);
+            *m_classWrappingInterface = *rhs.m_classWrappingInterface;
```
hodoulp, line 79: A self-explanatory writing is in fact `m_classWrappingInterface = rhs.m_classWrappingInterface->clone()`.

It becomes much more explicit and it avoids the overloading issue present with the assignation operator (i.e. what assignation operator is used when the variable is a `GpuShaderClassWrapper` type but the instance is a `MetalShaderClassWrapper` type).

hodoulp (changes requested):
After seeing all the code it seems that there is a better approach i.e. far less confusing.

src/OpenColorIO/GpuShaderDesc.cpp
Morteeza: @hodoulp Thanks for the comment. The change is applied. Let me know if I can make any improvement somewhere else.

src/OpenColorIO/GpuShaderClassWrapper.h
```diff
@@ -52,6 +52,7 @@ class MetalShaderClassWrapper : public GpuShaderClassWrapper
     std::string getClassWrapperFooter(const std::string& originalFooter) final;
     
     std::unique_ptr<GpuShaderClassWrapper> clone() const final;
+    const MetalShaderClassWrapper& operator=(const MetalShaderClassWrapper& rhs);
```
hodoulp, line 55: The return instance is not const for assignation operator, the method signature is then
`MetalShaderClassWrapper& operator=(const MetalShaderClassWrapper& rhs);`

hodoulp (approved):
Looks good.

michdolan (approved):

src/OpenColorIO/GpuShaderClassWrapper.h
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright Contributors to the OpenColorIO Project.
+
```
remia, line 3: [Minor] Should add include guards here.

Morteeza: @remia  good catch! include guards added.

src/OpenColorIO/GpuShaderClassWrapper.h
remia: This is the last issue I think @Morteeza

Morteeza: @hodoulp change applied.

remia (approved):

PR #1522, opened 2021-10-25
Fix add_test calls and change temporary file creation.

patlefort:
Followup from my previous PR (https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1487)

* Fix add_test in cmake files. This notation allows it to understand that the command portion is a target and to translate it to the proper executable.
* Remove starting backslash if present from temporary filename. A starting backslash mean that it is safe to create it in the current working directory.
* Implement a public TempFile class.
* Change relevant tests to use TempFile class.

I've implemented a public TempFile class. I've put it in the OCIO namespace, is it fine? Should I put it in the Platform namespace? Also, GPUHelpers.h/cpp are left empty, should they be removed?

Signed-off-by: Patrick Northon <northon_patrick3@yahoo.ca>


remia:
@patlefort Could you also add the ctest -C flag to the `analysis_workflow.yml` file (hard-coded to Release) so that this doesn't break? I believe it is then ready for merge.

remia (approved):

hodoulp (approved):

PR #1539, opened 2021-12-01
Adsk Contrib - Fix RGB spline curve inversion when the slope is zero

doug-walker:
When using the typical interface for creating a GradingRGBCurveTransform, the slopes at the ends are non-zero, but it is also possible to manually set the slopes at the control points, so a zero slope is possible.  In fact, this is the case for the ACES Output Transforms in order to exactly follow the original CTL code.

Currently, when evaluating the transform in the inverse direction, it will return an infinity, which will then often become NaN when going into following transforms that mix the channels (such as a Matrix).  This may result in NaN when inverting some of the ACES Output Transform Built-in Transforms for wide-gamut colors.

With this fix, if the slope of the curve at an end-point is below a threshold, and the requested inverse is outside the range of the forward transform, the returned result will be the left or right-most control point.  That is, it will be limited to the domain of the control points for the forward transform.  This is also similar to what happens when trying to invert a Lut1D when the slope of the forward transform becomes zero.


hodoulp:
@michdolan @remia That's a very important improvement that we need to have for a patch release. In order to have it in 2.1.1, do you think you can soon have a look? Thanks.

remia (approved):

doug-walker:
Thanks for the test Remi!

PR #1542, opened 2021-12-06
Fixes C++20 compilation errors and warnings

remia:
This was tested on macOS Apple Clang 13.0 (and brew Clang 13.0), I plan to add it on the analysis CI once #1515 is merged (if it doesn't break dependency build).

### DIFF
--- a/include/OpenColorIO/OpenColorIO.h
+++ b/include/OpenColorIO/OpenColorIO.h
@@ -3087,8 +3087,6 @@ public:
      */
     struct UniformData
     {
-        UniformData() = default;
-        UniformData(const UniformData & data) = default;
         UniformDataType m_type{ UNIFORM_UNKNOWN };
         DoubleGetter m_getDouble{};
         BoolGetter m_getBool{};
--- a/include/OpenColorIO/OpenColorTransforms.h
+++ b/include/OpenColorIO/OpenColorTransforms.h
@@ -433,7 +433,6 @@ extern OCIOEXPORT std::ostream & operator<<(std::ostream &, const DisplayViewTra
 struct OCIOEXPORT GradingRGBM
 {
     GradingRGBM() = default;
-    GradingRGBM(const GradingRGBM &) = default;
     GradingRGBM(double red, double green, double blue, double master)
         : m_red(red)
         , m_green(green)
@@ -460,7 +459,6 @@ extern OCIOEXPORT std::ostream & operator<<(std::ostream &, const GradingRGBM &)
 struct OCIOEXPORT GradingPrimary
 {
     GradingPrimary() = delete;
-    GradingPrimary(const GradingPrimary &) = default;
     explicit GradingPrimary(GradingStyle style)
         : m_pivot(style == GRADING_LOG ? -0.2 : 0.18)
         , m_clampBlack(NoClampBlack())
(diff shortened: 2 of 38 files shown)


hodoulp:
@remia The pull request changes the public API. That's not clear if it impacts the ABI and/or behavior (i.e. for copy constructor, constructor, etc. be not present vs. default could be compiler specific!). To be safe, we should not move it to RB-2.1. Is that correct?

remia:
I agree @hodoulp, I don't believe there any rush to include that in 2.1 release. It would also be interesting to have a clear answer for the ABI impact of these changes, maybe the ABI checker tool @michdolan was testing would be helpful here?

For reference here is the warnings that triggered those changes:
```
../OpenColorIO/OpenColorIO.h:3091:9: error: definition of implicit copy assignment operator for 'UniformData' is deprecated because it has a user-declared copy constructor [-Werror,-Wdeprecated-copy]
        UniformData(const UniformData & data) = default;
        ^
src/OpenColorIO/GpuShader.cpp:298:21: note: in implicit copy assignment operator for 'OpenColorIO_v2_2dev::GpuShaderDesc::UniformData' first required here
        data        = m_uniforms[index].m_data;
```
```
OpenColorTransforms.h:436:5: error: definition of implicit copy assignment operator for 'GradingRGBM' is deprecated because it has a user-declared copy constructor [-Werror,-Wdeprecated-copy]
    GradingRGBM(const GradingRGBM &) = default;
    ^
OpenColorTransforms.h:460:19: note: in implicit copy assignment operator for 'OpenColorIO_v2_2dev::GradingRGBM' first required here
struct OCIOEXPORT GradingPrimary
                  ^
src/OpenColorIO/fileformats/ctf/CTFReaderHelper.cpp:2214:34: note: in implicit copy assignment operator for 'OpenColorIO_v2_2dev::GradingPrimary' first required here
                m_gradingPrimary = values;
                                 ^
```
I guess another way of approaching this would have been to define a defaulted copy assignment operator, but overall all these are probably best removed for these POD-like structure.

hodoulp:
[Question] In C++20, what was the problem with operator == ?

remia:
@hodoulp From what I understand, C++20 introduce new rules for comparison operators overload resolution where it rewrite the method with reversed arguments (`this` and `other`) and add that to the candidates. This produces asymmetric results in our case because the parameter is always the base class type to allow virtual overrides, and the compiler do not like it. All these changes seem related to the introduction of the spaceship comparison operator (<=>). I didn't dig too deep into the details so maybe someone will expand on that.

Here I think we were using a quite unusual pattern overriding the `operator==` in a class hierarchy, it seemed more logical to use an `equals()` method with non-member comparison operator.

```
src/OpenColorIO/transforms/CDLTransform.cpp:172:19: warning: ISO C++20 considers use of overloaded operator '==' (with operand types 'const OpenColorIO_v2_2dev::CDLOpData' and 'const OpenColorIO_v2_2dev::CDLOpData') to be ambiguous despite there being a unique best viable function [-Wambiguous-reversed-operator]
    return data() == dynamic_cast<const CDLTransformImpl*>(&other)->data();
           ~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/OpenColorIO/ops/cdl/CDLOpData.h:124:10: note: ambiguity is between a regular call to this operator and a call with the argument order reversed
    bool operator==(const OpData& other) const override;
```

Hopefully that make some sense...

hodoulp:
That's strange because only clang notifies this ambiguity (using godbolt, latest msvc & gcc compilers do not).

hodoulp (approved):
LGTM.

michdolan (approved):

PR #1564, opened 2021-12-16
After Effects plug-in updates

fnordware:
Updates to After Effects and Photoshop plug-ins to keep up with the rest of OCIO.


src/OpenColorIO/GpuShaderClassWrapper.cpp
```diff
@@ -7,6 +7,8 @@
 
 #include "GpuShaderClassWrapper.h"
 
+#include <cctype>
```
doug-walker, line 10: This include is already present on line 4.

doug-walker (approved):
Minor issue needs to be fixed with GPUShaderClassWrapper.cpp, but the rest of it seems fine, so I'll approve now.

src/OpenColorIO/GpuShaderClassWrapper.cpp
fnordware: Whoops! Was not there in RB-2.1. Will get rid of the extra.

remia (approved):

PR #1565, opened 2021-12-16
RB-2.1 After Effects plug-in updates

fnordware:
Updates to After Effects and Photoshop plug-ins to keep up with the rest of OCIO.


src/OpenColorIO/GpuShaderClassWrapper.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include "GpuShaderClassWrapper.h"
 
+#include <cctype>
```
doug-walker, line 8: It's interesting that you needed to add this.  Patrick, is it expected that GPUShaderClassWrapper.cpp is different between main and RB-2.1?

doug-walker (approved):
Minor issue with GPUShaderClassWrapper.cpp, but the rest of it seems fine, so I'll approve now since we are holding the release for this.

src/OpenColorIO/GpuShaderClassWrapper.cpp
remia: From memory I added it in https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1515 to fix some errors in Windows latest analysis build. I think it's a good thing to have it here indeed, but also curious how you found this @fnordware, did you get some errors as well?

remia (approved):

src/OpenColorIO/GpuShaderClassWrapper.cpp
remia: Maybe best to pull the change directly from [main](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/GpuShaderClassWrapper.cpp) to avoid unneeded conflicts, the include order seems to be different with your change.

hodoulp: Double-checked on macOS, this include is useless!

remia: Definitively needed on Windows for some MSVC versions though.

fnordware: Yeah, I got errors when building on Windows. That was RB-2.1, didn't check main as I should have. Mac OS includes a lot of headers by default, and you get a lot more if you include something like Cocoa.h. Not the first time Windows reminded me to include a header Mac didn't need.

PR #1574, opened 2021-12-21
Fix support for ARM64 macOS python wheels

remia:
This PR address #1573 and contains the following changes:
* Update the `setup.py` build script to support cross compiling on macOS, heavily based on this PR https://github.com/pybind/cmake_example/pull/48 from @henryiii
* Build universal wheels, containing both arm64 and x86_64 architectures as it seems to be the recommended way, see https://cibuildwheel.readthedocs.io/en/stable/faq/#apple-silicon
* Build all external dependencies instead of missing ones to be sure to compile with the proper architecture flags on macOS
* Migrate to the latest cibuildwheel

For the current 2.1.1 release (and 2.1.0), the arm64 wheel is broken. Maybe one way of dealing with that without a new release would be to delete these wheels only and push the universal ones instead. PyPI doesn't allow overwriting already published wheels (even after deletion) and having both arm64 and universal wouldn't work because `pip`pick the closest match possible, here the arm64 if you are using a M1 mac.

Also not clear why I needed 13 escape characters to correctly preserve the arm64;x86_64 architecture flags in cmake ExternalProject, it interact with cmake list separator but if someone have a better way feel free to jump in!


hodoulp:
> Also not clear why I needed 13 escape characters to correctly preserve the arm64;x86_64 architecture flags in cmake ExternalProject, it interact with cmake list separator but if someone have a better way feel free to jump in!

Below are some questions / ideas:
- With this proposal, the line is now part of all the `findXX()` methods, did you manually check on your machine that `-DCMAKE_OSX_ARCHITECTURES="arm64;x86_64"` works on command line?
- As the goal is to have something like `"arm64;x86_64"`, did you try to change `setup.py` code to produce the right string (i.e. removing the 'string replace' line in all the `findXX()` methods)?
- Did you try in all the `findXX()` methods to only add `-DCMAKE_OSX_ARCHITECTURES="${CMAKE_OSX_ARCHITECTURES}"` or `-DCMAKE_OSX_ARCHITECTURES:string=${CMAKE_OSX_ARCHITECTURES}` without the 'string replace' line?

remia:
> * With this proposal, the line is now part of all the findXX() methods, did you manually check on your machine that -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" works on command line?
> * As the goal is to have something like "arm64;x86_64", did you try to change setup.py code to produce the right string (i.e. removing the 'string replace' line in all the findXX() methods)?

Yes it works and is effectively what is currently happening in setup.py, it does call CMake from a subprocess using the variable set just like you showed unless I'm missing something. This is done [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1574/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R132). Setup.py already use the correct string if I understand your meaning, just that because it contains a semicolon, it is interpreted as a list separater in the ExternalProject stuff, the main OCIO project is not affected.

> * Did you try in all the findXX() methods to only add -DCMAKE_OSX_ARCHITECTURES="${CMAKE_OSX_ARCHITECTURES}" or -DCMAKE_OSX_ARCHITECTURES:string=${CMAKE_OSX_ARCHITECTURES} without the 'string replace' line?

Not luck with that, however I found a cleaner way by doing `string(REPLACE ";" "$<SEMICOLON>" ESCAPED_CMAKE_OSX_ARCHITECTURES "${CMAKE_OSX_ARCHITECTURES}")`, I'll push that if that sounds good to you.

hodoulp:
> string(REPLACE ";" "$\<SEMICOLON\>" ESCAPED_CMAKE_OSX_ARCHITECTURES "${CMAKE_OSX_ARCHITECTURES}")

Great. That's way better to understand the change.

hodoulp (approved):

remia:
Trying to force push with the same commit amended to fix the CI builds always being cancelled after I manually re-launch the jobs.

hodoulp (approved):

doug-walker (approved):

Issue #1540, opened 2021-12-04
[Python] Syntax issue in NamedTransform setEncoding() parameter


MrLixm:
Hello, 

When using the Python API, 
the [NameTransform](https://opencolorio.readthedocs.io/en/latest/api/namedtransform.html).setEncoding() method have its parameter with a typo. The parameter name is stated as being `encodig` while it should be `encoding`

Cheers.
Liam.

remia:
Thanks for reporting @MrLixm, I included the fix the linked PR.

remia:
Closing as this was merged in `main`.

Issue #1434, opened 2021-07-12
Python bindings missing transform copy


remia:
Is there an easy way to copy Transforms from the Python binding like in v1 with `CreateEditableCopy()` ? Seems like this is missing in v2.

I tried the `copy.deepcopy()` by the way but it dont’t work because the object not are pickle-able, we would need to add `py::pickle pybind11` definitions which is probably too convoluted for what we want.

michdolan:
Originally we had implemented it, but since Python doesn't support `const` it didn't work the same way. In Python all of the transforms are always editable. I agree it would be good to support a proper Python implementation of OCIO object copying though.

remia:
Yeah that make sense, and copying might still be needed in some cases indeed.

Will have a look at it.

remia:
Closing as this was merged in `main`.

PR #1575, opened 2021-12-22
Python copy support

remia:
This PR addresses #1434 adding `copy` / `deepcopy` (same implementation) for OCIO objects having C++ `createEditableCopy()` methods. It also fix the typo reported in #1540.


zachlewis (approved):
Thanks Remi, this is super useful. Seems to work as expected.

hodoulp:
[Question] Only the `deep` copy exists in the public API of OpenColorIO i.e. `createEditableCopy()`, why also adding the Python `copy()`?

remia:
Good question.

It seems that in our context, the difference between Python `copy()` and `deepcopy()` is not relevant and both may be accurately implemented with `createEditableCopy()`. At least that's how I understand it because I don't have a clear view on the object memory layout as generated by pybind11 and it may be unnecessary complexity to strictly follow Python definition of shallow copy.

Here is the [pybind11 doc](https://pybind11.readthedocs.io/en/stable/advanced/classes.html?#deepcopy-support) and a quote from the [copy module doc](https://docs.python.org/3/library/copy.html):

> The difference between shallow and deep copying is only relevant for compound objects (objects that contain other objects, like lists or class instances):
> 
> * A shallow copy constructs a new compound object and then (to the extent possible) inserts references into it to the objects found in the original.
> * A deep copy constructs a new compound object and then, recursively, inserts copies into it of the objects found in the original.

hodoulp:
There is a major semantic difference between `shallow` and `deep` copies and, adding the two methods in Python introduces a discrepancy between the Python and C++ APIs (because the `shallow` copy is not supported in the C++ API). For example, a `deep` copy of a config instance means that the file rules are ***not*** shared between the two config instances (where a `shallow` copy would have share the file rule instance).

I think that adding the Python `copy()` (i.e. a `shallow` copy) introduces some confusion because it's in fact implemented by a `deep` copy.

remia:
I can remove it, it will go back to produce a segfault as before if we are fine with it.

zachlewis (approved):

tests/python/ConfigTest.py
```diff
@@ -863,6 +893,7 @@ def test_virtual_display(self):
 
 
 class ConfigVirtualWithActiveDisplayTest(unittest.TestCase):
+
```
hodoulp, line 896: it seems useless, you can remove it.

tests/python/ConfigTest.py
```diff
@@ -884,6 +915,7 @@ def test_virtual_display_with_active_displays(self):
 
 
 class ConfigVirtualDisplayTest(unittest.TestCase):
+
```
hodoulp, line 918: It seems useless, you can remove it.

tests/python/ContextTest.py
```diff
@@ -1,11 +1,31 @@
...
 
-import unittest, os, sys
+import copy, unittest, os, sys
 import PyOpenColorIO as OCIO
 
 class ContextTest(unittest.TestCase):
 
+    def test_copy(self):
+        """
+        Test the deepcopy() and copy() method.
```
hodoulp, line 11: The comment must be updated.

hodoulp (changes requested):
LGTM. Only made some comments for code improvements.

tests/python/ConfigTest.py
```diff
@@ -244,6 +245,35 @@
 
 
 class ConfigTest(unittest.TestCase):
+
+    def test_copy(self):
+        """
+        Test the deepcopy() and copy() method.
```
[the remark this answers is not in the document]
remia: Updated all the comments.

tests/python/ConfigTest.py
remia: I added for coherence of coding style compared to other classes but removed from the commit now to avoid polluting diff.

tests/python/ConfigTest.py
```diff
@@ -244,6 +245,35 @@
...
+
+        other = copy.deepcopy(cfg)
+        self.assertFalse(other is cfg)
+
+        self.assertEqual(other.getMajorVersion(), cfg.getMajorVersion())
+        self.assertEqual(other.getMinorVersion(), cfg.getMinorVersion())
+        self.assertEqual(other.getName(), cfg.getName())
+        self.assertEqual(other.getDescription(), cfg.getDescription())
+        self.assertEqual(list(other.getColorSpaceNames()), list(cfg.getColorSpaceNames()))
+
```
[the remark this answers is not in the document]
remia: Added the requested test for Config.

tests/python/ConfigTest.py
```diff
@@ -244,6 +245,45 @@
...
+        self.assertFalse(other is cfg)
+
+        self.assertEqual(other.getMajorVersion(), cfg.getMajorVersion())
+        self.assertEqual(other.getMinorVersion(), cfg.getMinorVersion())
+        self.assertEqual(other.getName(), cfg.getName())
+        self.assertEqual(other.getDescription(), cfg.getDescription())
+        self.assertEqual(list(other.getColorSpaceNames()), list(cfg.getColorSpaceNames()))
+        self.assertEqual(other.getFileRules().getNumEntries(), cfg.getFileRules().getNumEntries())
+
+        rules.removeRule(0)
```
hodoulp, line 283: Please add a comment to explain the goal of the following lines:
// Check that the file rules are not shared between the two config instances.

hodoulp (approved):

tests/python/ConfigTest.py
remia: Added.

doug-walker (approved):

PR #1578, opened 2022-01-03
Adsk Contrib - Improve the library install step

hodoulp:
Signed-off-by: Patrick Hodoul <Patrick.Hodoul@autodesk.com>

The pull request improves the library installation step following comments from @elsandosgrande and @Begasus  in #1415. Even if the comments focus on the 2.0.x release, I moved the fixes in `main` to later move them to `RB-2.1` (i.e. 2.1.x is now the official release).

From the root directory, the ways to test are:

- `cmake -B ./build_rls -S . -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=./install_rls -DCMAKE_INSTALL_BINDIR=binDir -DCMAKE_INSTALL_DATADIR=dataDir -DCMAKE_INSTALL_LIBDIR=libDir -DCMAKE_INSTALL_INCLUDEDIR=includeDir -DBUILD_SHARED_LIBS=ON`
- `make -C ./build_rls install -j 12`
- `ls -all ./install_rls`

and

- `cmake -B ./build_rls -S . -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=./install_rls -DBUILD_SHARED_LIBS=ON`
- `make -C ./build_rls install -j 12`
- `ls -all ./install_rls`

Tests were done on MacOS 10.15.7 with cmake 3.20 and AppleClang 12.0.

PR #1579, opened 2022-01-05
Propagate Android CMake configuration to external projects

amyspark:
:wave:

I have been testing OpenColorIO integration in Android for Krita. I found that missing requirements respect the  `CMAKE_TOOLCHAIN_FILE` of its parent project, but no further Android-specific variables are passed. This causes non-ARM builds, e.g. x86, to build OCIO for the specified architecture and its requirements for armv7a, thus failing at the link stage.

This MR fixes this issue by propagating the toolchain file, ABI and STL variables. This commit was sufficient to build a working prototype of color managed Krita (though with a non-fully-functional configuration reading, which I'll file a bug about separately).

### DIFF
--- a/share/cmake/modules/FindHalf.cmake
+++ b/share/cmake/modules/FindHalf.cmake
@@ -190,6 +190,14 @@ if(NOT Half_FOUND AND NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL NONE)
                 ${Half_CMAKE_ARGS} -DCMAKE_OSX_DEPLOYMENT_TARGET=${CMAKE_OSX_DEPLOYMENT_TARGET})
         endif()
 
+        if (ANDROID)
+            set(Half_CMAKE_ARGS
+                ${Half_CMAKE_ARGS}
+                -DANDROID_PLATFORM=${ANDROID_PLATFORM}
+                -DANDROID_ABI=${ANDROID_ABI}
+                -DANDROID_STL=${ANDROID_STL})
+        endif()
+
         # Hack to let imported target be built from ExternalProject_Add
         file(MAKE_DIRECTORY ${Half_INCLUDE_DIR})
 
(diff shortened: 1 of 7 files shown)


hodoulp:
The Windows build issue should be unlocked by a change from PR #1578.

hodoulp:
> I have been testing OpenColorIO integration in Android for Krita.

Could you give some details on how it's working with OpenColorIO / Krita / Android ?

amyspark:
>  Could you give some details on how it's working with OpenColorIO / Krita / Android ?

Mostly, it Just Works ™️, with the obvious (Qt) adjustments to the logic in [the app helpers for OpenGL](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/libutils/oglapphelpers/glsl.cpp). [This](https://invent.kde.org/graphics/krita/-/blob/7480e580c11a007452d6df0983da322ee697807c/plugins/dockers/lut/ocio_display_filter_vfx2021.cpp) is my port of the logic.

Notable caveats:

- When not in OpenGL ES 3, OCIO 2 needs the extensions `GL_OES_texture_float`, `GL_EXT_color_buffer_half_float` OR `GL_EXT_texture_storage`, and `GL_OES_texture_half_float_linear` for the texture filtering. See [this commit](https://invent.kde.org/graphics/krita/-/commit/7480e580c11a007452d6df0983da322ee697807c). 
- Due to the Scoped Storage limitations (#1580), only single-file configuration works, and only when fed via the stream config constructors; it is not possible anymore to load arbitrary files from the filesystem.
- In other words, if the file refers to a LUT (eg. the ACES 1.2 configuration), it is impossible to load it. This will cause an exception either at config construction, or at the moment of instantiating the processor (`vpt->getProcessor(config, config->getCurrentContext())`).

amyspark:
@hodoulp for the sake of completeness, these commits show how I integrated it with Krita: [1](https://invent.kde.org/graphics/krita/-/commit/06e7ca5be44374080992fd739deb44f9fc45bd9a) [2](https://invent.kde.org/graphics/krita/-/commit/7fa2ed6174ec450c21f53609ebd7d515f610813a).

It is essential to have a matching STL between the app, Qt, and OCIO, otherwise the exception handling will cause an `abort()` due to https://github.com/android/ndk/issues/1441. (Not handling the exception will at least log where it fails to `adb logcat`, but it still crashes.)

hodoulp (approved):

remia (approved):

PR #1582, opened 2022-01-07
Adsk Contrib - Remove 'expat' symbols

hodoulp:

The pull request removes the 'yaml' symbols from the library to avoid clashes with hosts (i.e. an app or another library) when it also uses the yaml with a different version. On `macOS` and `linux` (tested with ubuntu) only the OpenColorIO symbols are now present.


remia:
This worked on my macOS build. Did you saw this [-hidden-l flag](https://stackoverflow.com/a/70425525/4814046)? I wonder if there is a way to use that easily through CMake as it seems like a cleaner solution to the problem.

hodoulp:
@remia the solution is a lot less cryptic so, the code uses it now.

remia (approved):

remia:
For information about my previous comment, I was not clear how this would work with expat being already linked in the `target_link_libraries` call but this appears to take precedence and work fine which is good news.

doug-walker (approved):

Issue #1581, opened 2022-01-07
[Static] Target "ociotest" links to target "sampleicc::sampleicc" and  "utils::strings",  but the target was  not found. 


PhoebeHui:
Hi Guys,

Related to https://github.com/microsoft/vcpkg/issues/21642, the user encountered the following failure when use opencoloio with static build, we noticed that target sampleicc::sampleicc and utils::strings exported in static build, but these looks private which can't be used and exported externally. we could repro this issue on windows. Could you please help take a look?

OpenColorIOTargets.cmake:
```
set_target_properties(OpenColorIO::OpenColorIO PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_11"
  INTERFACE_LINK_LIBRARIES "OpenColorIO::OpenColorIOHeaders;\$<LINK_ONLY:expat::expat>;\$<LINK_ONLY:Imath::Imath>;\$<LINK_ONLY:pystring::pystring>;\$<LINK_ONLY:sampleicc::sampleicc>;\$<LINK_ONLY:utils::strings>;\$<LINK_ONLY:yaml-cpp>"
)
```

Failures:
```
CMake Error at /Users/matt/vcpkg/scripts/buildsystems/vcpkg.cmake:540 (_add_executable):
  Target "ociotest" links to target "sampleicc::sampleicc" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
Call Stack (most recent call first):
  CMakeLists.txt:11 (add_executable)

CMake Error at /Users/matt/vcpkg/scripts/buildsystems/vcpkg.cmake:540 (_add_executable):
  Target "ociotest" links to target "utils::strings" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
Call Stack (most recent call first):
  CMakeLists.txt:11 (add_executable)
```

Repro steps:
1. git clone https://github.com/PhoebeHui/vcpkg
2. cd vcpkg
3. git checkout dev/Phoebe/opencolorio
4. ./bootstrap-vcpkg.bat
5. ./vcpkg install opencolorio:x64-windows-static
6. ./vcpkg integrate install, copy the '-DCMAKE_TOOLCHAIN_FILE=xxxx', it will be used in test steps.
7. Create a test project that will link with OpenColorIO：
test\CMakeLists.txt
    ```
[...]

PR #1599, opened 2022-02-11
Minor build and CI fixes

### LINKED ISSUE #1593: Minor patch for gcc 12
As it typically happens, some headers are no longer "free" with major gcc upgrades. I needed to add this in order to build with gcc 12:

```
Index: OpenColorIO-2.1.1/src/OpenColorIO/FileRules.cpp
===================================================================
--- OpenColorIO-2.1.1.orig/src/OpenColorIO/FileRules.cpp
+++ OpenColorIO-2.1.1/src/OpenColorIO/FileRules.cpp
@@ -3,6 +3,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cstring>
 #include <map>
 #include <regex>
 #include <sstream>
```

Pretty minor otherwise I would have done a PR.

remia:
#### Fix the analysis workflow

The install location for built dependencies in the analysis workflow is now explicit in Linux and macOS (Windows already use a custom location). This is because OpenImageIO now started to override the install location to point in the build tree when none is specified, instead of using standard Unix paths.

Currently failing on OpenEXR Windows build, see related issue https://github.com/AcademySoftwareFoundation/openexr/issues/1236

#### Improve documentation for OFX build #1607

### DIFF
--- a/tests/cpu/Config_tests.cpp
+++ b/tests/cpu/Config_tests.cpp
@@ -8682,8 +8682,8 @@ OCIO_ADD_TEST(Config, look_fallback)
     // Test that the look fallback syntax works for look with missing file.
     // The fallback syntax allow to specify looks to try in order and use the
     // first valid one. When the syntax "my_look | " is used, the fallback
-    // is empty and there will be no look applied (no-op) if my_look in invalid.
-    // This may happen if my_look rely on a missing environment variable.
+    // is empty and there will be no look applied (no-op) if my_look is invalid.
+    // This may happen if my_look relies on a missing environment variable.
 
     {
         static const std::string CONFIG =
--- a/tests/cpu/Config_tests.cpp
+++ b/tests/cpu/Config_tests.cpp
@@ -8676,3 +8676,63 @@ colorspaces:
         OCIO_CHECK_ASSERT(proc->isNoOp());
     }
 }
+
+OCIO_ADD_TEST(Config, look_fallback)
+{
+    // Test that the look fallback syntax works for look with missing file.
+
+    {
+        static const std::string CONFIG =
+            "ocio_profile_version: 1\n"
+            "\n"
+            "search_path: " + OCIO::GetTestFilesDir() + "\n"
+            "\n"
+            "roles:\n"
+            "  scene_linear: cs\n"
+            "\n"
+            "displays:\n"
+            "  disp1:\n"
+            "    - !<View>\n"
+            "      name: view1\n"
+            "      colorspace: cs\n"
+            "      looks: missing_file_look | \n"
+            "\n"
+            "looks:\n"
+            "  - !<Look>\n"
+            "    name: missing_file_look\n"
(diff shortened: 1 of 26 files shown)


remia:
Note that this PR is not complete yet and will probably include some fix related to #1581.

tests/cpu/Config_tests.cpp
```diff
@@ -8676,3 +8676,63 @@ OCIO_ADD_TEST(Config, look_is_noop)
         OCIO_CHECK_ASSERT(proc->isNoOp());
     }
 }
+
+OCIO_ADD_TEST(Config, look_fallback)
+{
+    // Test that the look fallback syntax works for look with missing file.
```
hodoulp, line 8682: Please add a comment to explain what is the meaning of a look fallback `missing_file_look | ` in the unit test. This notation can be used for fallback but it mainly highlights that there are two choices (i.e. `missing_file_look` and `none`) and the precessing takes the first successful.

remia: Good point, comment added.

hodoulp:
@remia What is the status of this pull request?

tests/cpu/Config_tests.cpp
```diff
@@ -8680,6 +8680,10 @@ OCIO_ADD_TEST(Config, look_is_noop)
 OCIO_ADD_TEST(Config, look_fallback)
 {
     // Test that the look fallback syntax works for look with missing file.
+    // The fallback syntax allow to specify looks to try in order and use the
+    // first valid one. When the syntax "my_look | " is used, the fallback
+    // is empty and there will be no look applied (no-op) if my_look in invalid.
```
hodoulp, line 8685: typo `if my_look in invalid` to `if my_look is invalid`

tests/cpu/Config_tests.cpp
```diff
@@ -8680,6 +8680,10 @@ OCIO_ADD_TEST(Config, look_is_noop)
...
+    // This may happen if my_look rely on a missing environment variable.
```
hodoulp, line 8686: typo `rely` to `relies`

remia:
> @remia What is the status of this pull request?

This PR is ready to merge now from my point of view @hodoulp.

doug-walker (approved):
Thanks Remi!

There seems to be an analysis build failure on Windows related to OpenEXRCore, but I didn't try to figure out why it's happening on one Windows build but not another.  Any thoughts on what went wrong?

remia:
> There seems to be an analysis build failure on Windows related to OpenEXRCore, but I didn't try to figure out why it's happening on one Windows build but not another. Any thoughts on what went wrong?

Hi @doug-walker, this is the issue I mentioned in the PR description, OpenEXR doesn't compile on Windows MSVC 2022 currently, while the issue has now been fixed upstream, it's not released yet. I expect the analysis build will be fixed when that is the case. We shouldn't wait for this however.

PR #1605, opened 2022-02-23
Add python source distribution package

remia:
Adding Python source distribution package to be uploaded to PyPI, allowing source install (eg. when a binary wheel for the requested architecture is not available) as discussed in #1573.

I pushed a test PyPI archive to test, available [here](https://test-files.pythonhosted.org/packages/08/8e/3d3028dd1dd2129d2b54a70afb4ef4cfcbab1889882a84b78cbb42ccf73b/opencolorio-2.2.0.dev3.tar.gz). Download and then call pip install on it to test, there is a way to test that directly installing from pip / test.pypi.org but there are security concerns as dependency can be pulled from test PyPI which we can't trust so I won't share the command here.

### DIFF
--- a/.github/workflows/wheel_workflow.yml
+++ b/.github/workflows/wheel_workflow.yml
@@ -20,7 +20,10 @@ on:
       - v1.*
     paths:
       - .github/workflows/wheel_workflow.yml
+      - MANIFEST.in
       - pyproject.toml
+      - setup.cfg
+      - setup.py
   schedule:
     # Nightly build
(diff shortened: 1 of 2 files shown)


hodoulp (approved):
LGTM but I'm clearly not an expert. As @remia did a test I approve the PR.

hodoulp:
Friendly reminder that the pull request could now be merged next week (i.e. more than two-weeks old).

doug-walker (approved):
Thanks Remi!

PR #1610, opened 2022-03-07
Port from main to RB-2.1 - Backport improved "OpenImageIO" dependency management.

KelSolaar:
Hello,

This PR backports ec50ccbccf6d252ddcca197531d3852178f36ca3 from @hodoulp to 2.1 to make sure that the `Imath` headers are included as dependencies while building the OCIO apps.

@remia for VIS.

Cheers,

Thomas


src/apps/CMakeLists.txt
```diff
@@ -6,14 +6,16 @@ if(OCIO_BUILD_APPS)
...
 	add_subdirectory(ociochecklut)
 	add_subdirectory(ociomakeclf)
+	add_subdirectory(ocioperf)
 	add_subdirectory(ociowrite)
 
 	if(OpenImageIO_FOUND AND TARGET OpenImageIO::OpenImageIO)
 		add_subdirectory(ociolutimage)
 		add_subdirectory(ocioconvert)
 		add_subdirectory(ociodisplay)
-		add_subdirectory(ocioperf)
```
remia, line 15: I think this should be put back in because it relies on an earlier PR #1516 that is not in RB-2.0 and probably don't need to be. Looking at the commit you used mentioned in the description, it's not clear to me how it got here.

src/apps/CMakeLists.txt
```diff
@@ -6,14 +6,16 @@ if(OCIO_BUILD_APPS)
...
 	else()
-		message(WARNING "Could NOT find OpenImageIO. Skipping build for ociolutimage, ocioconvert, ociodisplay and ocioperf")
+		message(WARNING "Could NOT find OpenImageIO. Skipping build for ociolutimage, ocioconvert and ociodisplay.")
```
remia, line 17: This change should also be reverted.

src/apps/CMakeLists.txt
```diff
@@ -6,14 +6,16 @@ if(OCIO_BUILD_APPS)
 	add_subdirectory(ociocheck)
 	add_subdirectory(ociochecklut)
 	add_subdirectory(ociomakeclf)
+	add_subdirectory(ocioperf)
```
remia, line 9: This should be removed.

src/apps/CMakeLists.txt
KelSolaar: I think that I know what happened. I will rebase off this PR from RB-2.1 anyway.

remia: Ah ok, and yeah just realised I assumed it was targeting 2.0 but it's actually 2.1!

KelSolaar: Sounds good, I will rename the PR title! I have just rebased everything from RB-2.1, it should be much cleaner now.

KelSolaar: The test suite is seemingly passing now, good sign!

src/apps/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ if(OCIO_BUILD_APPS)
 		add_subdirectory(ociodisplay)
 		add_subdirectory(ocioperf)
 	else()
-		message(WARNING "Could NOT find OpenImageIO. Skipping build for ociolutimage, ocioconvert, ociodisplay and ocioperf")
+		message(WARNING "Could NOT find OpenImageIO. Skipping build for ociolutimage, ocioconvert and ociodisplay.")
```
remia, line 17: [Minor] Should be reverted

remia (commented):
Thanks for the update @KelSolaar.

I believe we should also remove the find_package(OpenImageIO ...) from [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/RB-2.1/CMakeLists.txt#L269).

share/cmake/modules/FindExtPackages.cmake
```diff
@@ -103,3 +103,39 @@ if(OCIO_BUILD_PYTHON OR OCIO_BUILD_DOCS)
...
+# The presence of OpenImageIO allows additional OCIO apps and the OSL
+# translation unit tests to be built.
+
+if(OCIO_BUILD_APPS OR OCIO_BUILD_TESTS)
+    # OpenImageIO
+    # https://github.com/OpenImageIO/oiio
+
+    set(OIIO_VERSION "2.1.9")
+
+    if(OCIO_USE_OIIO_CMAKE_CONFIG)
```
hodoulp, line 116: You should also take the `OCIO_USE_OIIO_CMAKE_CONFIG` cmake variable from main (i.e. [CMakeLists.txt line 152](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/CMakeLists.txt#L152) ).

hodoulp (changes requested):

remia (approved):

doug-walker (approved):
Thanks Thomas!

hodoulp (approved):

PR #1614, opened 2022-03-15
Adsk Contrib - Fix half-domain Lut1D issue for values above HALF_MAX

doug-walker:
This fixes an issue where float values greater than HALF_MAX but which don't round to Infinity (in other words, values between 65505 and 65519) were returning NaN. (This did not affect the GPU, only CPU.)

If this issue was encountered when inverting a normal domain Lut1D, and a work-around is needed before receiving this fix, turn off the OPTIMIZATION_LUT_INV_FAST optimization flag. For example, to use the default settings but turn off this flag, set the environment variable as follows:
export OCIO_OPTIMIZATION_FLAGS=230440899

### DIFF
--- a/src/OpenColorIO/ops/lut1d/Lut1DOpCPU.cpp
+++ b/src/OpenColorIO/ops/lut1d/Lut1DOpCPU.cpp
@@ -527,7 +527,7 @@ IndexPair IndexPair::GetEdgeFloatValues(float fIn)
     const float floatTemp = (float) halfVal;
 
     // Strict comparison required otherwise negative fractions will occur.
-    if (fabs(floatTemp)> fabs(fIn)) 
+    if (fabs(floatTemp) > fabs(fIn)) 
     {
         idxPair.valB = halfVal.bits();
         idxPair.valA = idxPair.valB;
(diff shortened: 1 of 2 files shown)


hodoulp:
Friendly reminder that the pull request can now be merged this week as it's now more than 2 weeks old.

remia (approved):
Sorry for the late review @doug-walker.

> If this issue was encountered when inverting a normal domain Lut1D

Just checking, did you mean half-domain Lut1D?

I reproduced the bug in `ocioconvert` and Nuke 13 by applying the LUT you use in your test to an EXR with a value > 65504 and can confirm the fix works by making the `fraction` a `NaN` that is then converted to 0 instead of producing an `inf` before.

doug-walker:
@remia , yes, the description is correct, I really did mean "normal domain Lut1D".  When the "fast inverse" optimization option is enabled, when a normal domain Lut1D is inverted it often becomes a forward evaluation in a half-domain Lut1D (this happens when the range of the LUT exceeds [0,1]).  That was the scenario involved in the original bug report.

hodoulp (approved):
As the PR is already more than two-weeks old, I approve it.

PR #1646, opened 2022-05-16
Adsk contrib - Scripts to facilitate building OCIO on Windows

cedrik-fuoco-adsk:
The goal of this PR is to add scripts that makes building OCIO on Windows easier.

- The script ocio_deps.bat can be used to check and install some of the difficult dependencies - such as OpenImageIO, Freeglut and Glew.

- The script ocio.bat can be used to make it easy to run cmake configure, build and install.

### DIFF
--- a/docs/quick_start/installation.rst
+++ b/docs/quick_start/installation.rst
@@ -5,31 +5,70 @@
 .. _installation:
 
 Installation
-============
-
-The easy way
 ************
 
-While prebuilt binaries are not yet available for all platforms, OCIO
-is available via several platform's package managers.
+Although OpenColorIO is available through a variety of package managers, please note that the 
+version available through a given package manager may be significantly outdated when compared to 
+the current official OCIO release. Even two years after the introduction of OpenColorIO v2, several 
+package managers continue to install OCIO 1.1.1.
 
-Please note that the package managers are still installing the previous 
-stable release, 1.1.1.  If you want OCIO v2, you currently must build from source.
-See :ref:`building-from-source`.
+.. _Python:
 
+Python
+^^^^^^
 
-Fedora and RHEL
-^^^^^^^^^^^^^^^
+If you only need the Python bindings, the simplest solution is to take advantage of the pre-built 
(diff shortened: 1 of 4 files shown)


remia:
Thanks for that @cedrik-fuoco-adsk.

I tried on my Windows box and I have issues having it work, it seems that the script doesn't handle properly whitespaces in the Python install path (`C:\Program Files\Python37`). I tried both with the `--python` flag or with the automatic detection but it fails in both case: 

> "Could not find Python. Please provide the location for python or modify PYTHON_PATH in this script.".

 Is there something I am doing wrong?

cedrik-fuoco-adsk:
The script wasn't handling spaces correctly for the python path and could potentially overwrite the path given with --python.
I've tested in my sandbox and the last commit should fix the issue.

The following example will work now : 
--python "C:\My Toolings with Spaces\Python\Python310"

cedrik-fuoco-adsk:
@remiaI need to do more investigation, but so far I found out that the parenthesis are breaking the command line before the script is actually executed. I tried to escape the parenthesis with "^" as such : 

--msvs "C:\Program Files ^(x64^)\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build"

But it seems that it break something else because Cmake can't find Visual Studio Generator anymore. But if I remove the parenthesis, it's all good.

cedrik-fuoco-adsk:
@remia I made some changes. Whenever you have time, could you try again? Just make sure to escape the parentheses in your path. e.g. --msvs "C:\with spaces and ^(p^)\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build"

This is what I tried, and it works on my Window sandbox now:

`ocio.bat --vcpkg C:\vcpkg --ocio C:\Users\ContainerUser\OpenColorIO --b C:\Users\ContainerUser\Documents\build --i C:\Users\ContainerUser\Documents\install --msvs "C:\with spaces and ^(p^)\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build" --configure`

doug-walker (approved):

Issue #1634, opened 2022-04-22
OCIO won't compile with Visual Studio 2022


doug-walker:
There may be a number of problems, but one has to do with PyBind 11:
https://github.com/pybind/pybind11/pull/3497

remia:
Hi @doug-walker,

Do you have a build log to share? I'm a bit surprised as the analysis nightly build currently pass on [Visual Studio 2022](https://github.com/AcademySoftwareFoundation/OpenColorIO/runs/6121544585?check_suite_focus=true#step:11:20) as far as I know.

doug-walker:
@remia , good point!  I did not realize that the Analysis is now using VS 2022, unlike CI which is still on VS 2019.  However, the Analysis is not building the Python binding:
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/b00877959f4497f9b8e5a08859d0ef8059fdd8e5/.github/workflows/analysis_workflow.yml#L361

So that's why it doesn't seem to be running into the PyBind 11 problem I mentioned.  

I notice the Analysis is installing PyBind 11 on Windows but then not using it (like it seems to for Linux and Mac).  Do you think Analysis should be building and testing the binding on Windows too?  @michdolan , any thoughts?

(We are working on trying to improve the Windows install instructions, so that's how we ran into the VS 2022 issue.)

remia:
Yes that's a good catch, I'm not sure why it's setup like this but that was probably me during testing that noticed some issues with Python then forgot to go back and fixing it. I'll look into it when I get some time.

The next release, assuming it's a bump of VFX platform, will probably be a good opportunity to review all dependencies and increase minimal version where appropriate.

remia:
I'm currently stuck trying to enable Python testing on the analysis build, there is a DLL load import [error](https://github.com/remia/OpenColorIO/runs/6290436198?check_suite_focus=true#step:13:3287) while running the Python unit test and I'm not able to understand where it comes from yet, it's very hard to debug using only Github Actions jobs. Note that it manages to successfully compile with pybind 2.6.2 on Visual Studio 22 so I assume it's a different issue than what you saw.

doug-walker:
Yes, definitely a different issue.  The Pybind issue we encountered was only in Debug mode and the analysis builds in Release mode.
	https://github.com/pybind/pybind11/pull/3497
	https://github.com/microsoft/onnxruntime/issues/9735

remia:
[...]

PR #1647, opened 2022-05-16
Adsk contrib - Fix issues when building OCIO with MS Visual Studio 2022

cedrik-fuoco-adsk:
All issues were discovered when trying to configure and build OCIO on Windows using Microsoft Visual Studio 2022.

Pybind11 has an issue with Microsoft Visual Studio 2022, FindGlew from CMake has a bug around setting GLEW_INCLUDE_DIRS correctly and OpenImageIO as some warnings that get flagged as error. Since those warning are in the OCIO apps, we decided that I would be ok to ignore them temporarily.

Content overview:

Updating Pybind11 to the latest version as Pybind11 <2.9.0 has issues with Microsoft Visual Studio 2022.
Add fix to make sure that GLEW_INCLUDE_DIRS is set correctly.
Ignoring warnings (C4267 and C4244) temporarily in ociolutimage, ocioconvert and oiiohelpers

Reference issue #1634.

### DIFF
--- a/share/cmake/utils/CheckSupportGL.cmake
+++ b/share/cmake/utils/CheckSupportGL.cmake
@@ -6,6 +6,7 @@
 # Define variables around the GL support.
 
 include(PackageUtils)
+include(SelectLibraryConfigurations)
 
 if((OCIO_BUILD_TESTS AND OCIO_BUILD_GPU_TESTS) OR OCIO_BUILD_APPS)
     set(OCIO_GL_ENABLED ON)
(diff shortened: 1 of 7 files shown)


share/cmake/utils/CheckSupportGL.cmake
```diff
@@ -35,6 +36,35 @@ if((OCIO_BUILD_TESTS AND OCIO_BUILD_GPU_TESTS) OR OCIO_BUILD_APPS)
                 set(GLEW_LIBRARIES GLEW::GLEW)
             endif()
         endif()
+
+        # Cmake has a bug with FindGLEW. It fails to defined GLEW_INCLUDE_DIRS in some situation.
```
remia, line 40: Question: Is the comment on line 33-34 still valid or should it be removed? It seems the CMake bug is still existing and not fixed as suggested on that comment.

share/cmake/utils/CheckSupportGL.cmake
```diff
@@ -35,6 +36,35 @@ if((OCIO_BUILD_TESTS AND OCIO_BUILD_GPU_TESTS) OR OCIO_BUILD_APPS)
...
+            if("${_GLEW_DEFS}" MATCHES "GLEW_STATIC")
+                get_target_property(GLEW_LIBRARY_DEBUG GLEW::GLEW IMPORTED_LOCATION_DEBUG)
+                get_target_property(GLEW_LIBRARY_RELEASE GLEW::GLEW IMPORTED_LOCATION_RELEASE)
+            else()
+                get_target_property(GLEW_LIBRARY_DEBUG GLEW::GLEW IMPORTED_IMPLIB_DEBUG)
+                get_target_property(GLEW_LIBRARY_RELEASE GLEW::GLEW IMPORTED_IMPLIB_RELEASE)
+            endif()
+            get_target_property(_GLEW_LINK_INTERFACE GLEW::GLEW IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE) # same for debug and release
+            list(APPEND GLEW_LIBRARIES ${_GLEW_LINK_INTERFACE})
+            list(APPEND GLEW_LIBRARY ${_GLEW_LINK_INTERFACE})
```
remia, line 58: This looks like undefined in Linux which breaks CI, should this line be removed as we don't seem to use `GLEW_LIBRARY` elsewhere? Is this variable only created on Windows? CMake [documentation](https://cmake.org/cmake/help/latest/module/FindGLEW.html) doesn't mention it.

remia:
Thanks @cedrik-fuoco-adsk, when the Linux issue has been resolved, I would suggest we merge this one quickly as the CI seem to be broken on other PR for Windows which is fixed here. Probably some update to the Windows runners (although they are not using VS 2022).

share/cmake/utils/CheckSupportGL.cmake
cedrik-fuoco-adsk: You're right. I don't see it used anywhere. I've seen some mentioned of GLEW_LIBRARY, but they are all from before 2019/2018 with an older version of cmake.

I'll remove it as it is not needed on Windows either.

share/cmake/utils/CheckSupportGL.cmake
cedrik-fuoco-adsk: I'll remove it. It is not valid anymore as that issue was closed by accident and re-opened.

doug-walker (approved):
Note:  This is a build-break fix and therefore the two-week rule does not apply.

Issue #1360, opened 2021-04-03
Add displayview inverse mode for command-line tools


doug-walker:
The command-line tools ocioconvert, ocioperf, and ociowrite provide an option to specify an input space and a display and view rather than a pair of color spaces.  When a pair of color spaces is specified, swapping the order is equivalent to inverting the transformation, but that trick doesn't work when using the display/view mode.  So a new command-line option should be added to allow this.  Under the hood, it simply sets the direction of the resulting transform to the inverse direction.  So most of the work is just handling the argument parsing.

For the command syntax, I propose adding a --inverse switch option.  (It should also be functional for the non-display-view arguments.)  (Also open to --invert, or "--dir inverse" if people prefer.)

In the case of ocioconvert, the forward syntax looks like:
ocioconvert [options] --view inputimage inputcolorspace outputimage displayname viewname
and the proposed inverse syntax should look like:
ocioconvert [options] --inverse --view inputimage displayname viewname outputimage outputcolorspace

For ocioperf and ociowrite, the forward syntax looks like:
ociowrite --displayview inputColorSpace display view --file outputfile
and the proposed inverse syntax is:
ociowrite --inverse --displayview display view outputColorSpace --file outputfile
but I would also be open to:
ociowrite --inverse --displayview inputColorSpace display view --file outputfile
if people feel that it is better to not change the order.

KelSolaar:
Like it, great idea!

hisergiorojas:
I would like to work on this issue.

doug-walker:
Thanks, that would be appreciated @hisergiorojas.  We would like to include this in OCIO 2.1, which means that we need it merged by the end of the month.

doug-walker:
This would be a great issue for someone new to the project to work on.  Pretty easy to do.

doug-walker:
Closing as fixed via PR #1648.

PR #1648, opened 2022-05-16
Adsk contrib - Implement invertview for ocioconvert, ocioperf and ociowrite

cedrik-fuoco-adsk:
Implementing invertview option for ocioconvert, ocioperf and ociowrite

Reference Issue: https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1360

### DIFF
--- a/src/apps/ocioconvert/main.cpp
+++ b/src/apps/ocioconvert/main.cpp
@@ -558,6 +558,19 @@ int main(int argc, const char **argv)
         {
             // Copy specInput into specOutput.
             specOutput = specInput;
+
+            /*
+                Set the bit-depth of the output buffer to be used by OIIO.
+
+                OIIO may change the bit-depth when writing to a file (e.g. if the output image ends in "jpg" it will be converted to 8-bit integer), 
+                and OCIO is not trying to analyze the filename to emulate OIIO's decision making process.
+                
+                Additionally, the color space conversion may require more bits than the source image.
+                For example, converting a log image to linear requires at least a half-float output format.
+                
+                For most cases, half-float strikes a good balance between precision and storage space.
+                But if the input depth would lose precision when converted to half-float, use full float for the output depth instead.
+            */
             if (specInput.format == OIIO::TypeDesc::UINT16 || specInput.format == OIIO::TypeDesc::FLOAT) 
             {
                 specOutput.set_format(OIIO::TypeDesc::FLOAT);
--- a/src/apps/ocioconvert/main.cpp
+++ b/src/apps/ocioconvert/main.cpp
@@ -79,22 +79,27 @@ int main(int argc, const char **argv)
     bool help           = false;
     bool useLut         = false;
     bool useDisplayView = false;
+    bool useInvertView  = false;
 
     ap.options("ocioconvert -- apply colorspace transform to an image \n\n"
                "usage: ocioconvert [options]  inputimage inputcolorspace outputimage outputcolorspace\n"
                "   or: ocioconvert [options] --lut lutfile inputimage outputimage\n"
-               "   or: ocioconvert [options] --view inputimage inputcolorspace outputimage displayname viewname\n\n",
+               "   or: ocioconvert [options] --view inputimage inputcolorspace outputimage displayname viewname\n"
+               "   or: ocioconvert [options] --invertview inputimage displayname viewname outputimage outputcolorspace\n\n",
                "%*", parse_end_args, "",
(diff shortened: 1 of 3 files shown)


src/apps/ocioconvert/main.cpp
```diff
@@ -513,22 +551,40 @@ int main(int argc, const char **argv)
...
-            oglApp->readImage((float *)img.getBuffer());
+            oglApp->readImage((float *)imgInput.getBuffer());
         }
         else
 #endif // OCIO_GPU_ENABLED
         {
-            const OCIO::BitDepth bitDepth = OCIO::GetBitDepth(spec);
+            // Copy specInput into specOutput.
+            specOutput = specInput;
+            if (specInput.format == OIIO::TypeDesc::UINT16 || specInput.format == OIIO::TypeDesc::FLOAT)
```
remia, line 561: [Question] Why are we promoting UINT16 to FLOAT and UINT8 to HALF? Would it be worth to add a comment in case that is indeed the right thing to do?

src/apps/ocioperf/main.cpp
```diff
@@ -179,24 +179,40 @@ int main(int argc, const char **argv)
...
+                                            "Provide the transform file to apply on the image",
+               "--colorspaces %s %s",       &inColorSpace, &outColorSpace,
+                                            "Provide the input and output color spaces to apply on the image",
+               "--view %s %s %s",           &inColorSpace, &display, &view,
+                                            "Provide the input color space and (display, view) pair to apply on the image",
+               "--displayview %s %s %s",    &inColorSpace, &display, &view,
+                                            "(Deprecated) Provide the input and (display, view) pair to apply on the image",
+               "--invertview %s %s %s",     &display, &view, &outColorSpace,
+                                            "Provide the (display, view) pair and output color space to apply on the image",
+               "--iter %d",                 &iterations, "Provide the number of iterations on the processing. Default is 10",
```
remia, line 209: [Minor] I think the default iteration count is 50 now.

src/apps/ocioperf/main.cpp
```diff
@@ -319,22 +339,34 @@ int main(int argc, const char **argv)
             }
 
             {
-                std::string msg;
-                if (!outColorSpace.empty())
+                // --colorspaces
+                useColorspaces = !inColorSpace.empty() & !outColorSpace.empty();
```
remia, line 343: [Minor] Should use `&&` instead of `&`? Same applies in other places in the PR.

src/apps/ociowrite/main.cpp
```diff
@@ -163,29 +182,80 @@ int main(int argc, const char **argv)
...
-                                                 OCIO::TRANSFORM_DIR_FORWARD);
+                // colorspaces to (display,view) pair
+
+                // validate output arguments
+                if (!display.empty() && !view.empty())
+                {
+                    processor = config->getProcessor(inputColorSpace.c_str(), 
+                                                        display.c_str(),
+                                                        view.c_str(),
+                                                        OCIO::TRANSFORM_DIR_FORWARD);
```
remia, line 231: [Minor] Indentation issue (same below).

remia (commented):
Great work @cedrik-fuoco-adsk. It's probably going to conflict with PR #1637 but I think we should merge this one first as it's ready.

src/apps/ocioconvert/main.cpp
cedrik-fuoco-adsk: It was the result of a discussion with @doug-walker. I'll add a comment based on that.

remia: Thanks, note that in the related PR #1637 all the Image IO code is gone from the apps into an ImageIO class so this code may need porting later on.

src/apps/ocioconvert/main.cpp
```diff
@@ -513,22 +551,53 @@ int main(int argc, const char **argv)
...
             oglApp->redisplay();
-            oglApp->readImage((float *)img.getBuffer());
+            oglApp->readImage((float *)imgInput.getBuffer());
         }
         else
 #endif // OCIO_GPU_ENABLED
         {
-            const OCIO::BitDepth bitDepth = OCIO::GetBitDepth(spec);
+            // Copy specInput into specOutput.
+            specOutput = specInput;
```
remia, line 560: Only seeing this now, is `specOutput` correctly initialised when taking the GPU path (`--gpu`)?

remia (approved):

doug-walker (approved):

PR #1658, opened 2022-06-07
Bump version from 2.1.2 to 2.1.3 and added dev suffix

cedrik-fuoco-adsk:
Bump version from 2.1.2 to 2.1.3 and added dev suffix

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,13 +28,13 @@ endif()
 # Project definition.
 
 project(OpenColorIO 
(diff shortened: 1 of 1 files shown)


remia (approved):

doug-walker:
Windows CI is failing due to a change on the build machine side.  An update to the Pybind version would be required to fix but we don't want to do that for the release branch at this point.

doug-walker (approved):

Issue #1621, opened 2022-03-18
Default configs


doug-walker:
Following up on recent suggestions by @lgritz and @hodoulp, I'm creating an issue to capture thoughts on the topic of whether the OCIO library should provide one or more default configs.  There are already a variety of open source configs available for use with OCIO, but it currently requires action on the part of either a developer or end-user to make use of them.  We could try to simplify this.

This raises a number of questions for the community:
1)  Should there be one or more configs that are somehow included with OCIO itself?
2)  Should OCIO try to recommend a single config as a default for everyone?

If the community decides it's something to pursue, here are some of the questions to consider:

### One or more defaults?

The Configs working group has several new ACES-related configs in progress: the Reference config, the Studio config, and the CG config.  There seems to be a sense from many people that the CG config, which attempts to be a simple, minimalist, ACES-based config might make a nice general purpose default.  But it does not include camera color spaces, so while it might be a good default for general CG artists, it might not be a good default for compositors, for example.  Likewise, should the default include any other viewing transforms in addition to ACES?  One approach would be for OCIO to provide a few defaults (with some guidelines about which to use for various user personas) but not weigh-in on a single default for everyone.

### Where are the defaults stored?

If the community decides to provide defaults, the next question is how to make them available.  One option would be to define a directory location for each platform and have the OCIO installer, when it installs the library, install one or more default configs in that location.  Another option would be for the configs to be built into the library itself, so that no external files are required.  Obviously having them be built in seems cleaner, but on the other hand, it could be nice to have a standard directory location for putting configs.  People may find this useful for adding their own defaults, to be shared among several applications on a workstation, for example.  Having them be external files also makes it simpler to handle defaults that require external LUTs.

### Backwards compatibility

We will have to discuss whether inclusion in the library itself (either directly or via the installer) is sustainable since the intent is that the CG config will evolve as ACES evolves. Applications typically need to be able to load scenes/projects/etc. that were created in previous releases and so we need to consider whether the CG config (assuming that is chosen as the single default) will always support backwards compatibility.  Because OCIO relies heavily on the color space name strings, this could potentially tie the hands of the working group as new versions of ACES are released.  If the configs working group decides it will not guarantee backwards compatibility, then we will need to decide whether it will be up to the OCIO library (e.g., CreateDefault(version)) or the client application (e.g., config.serialize(ostream), if the config is built in) to deal with it.

### API

[...]

PR #1659, opened 2022-06-10
Adsk contrib - Built-in configurations

cedrik-fuoco-adsk:
This PR covers the [Built-in configuration feature](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1621) (formerly Default configs).

I've included one CG config (cg-config-v0.1.0_aces-v1.3_ocio-v2.1.1.ocio) among the changes. In the future, more configurations and configuration types may be added.

With these modifications, the OCIO environment variable may now be set to the following values:

Using URI format:
`OCIO=ocio://default`
`OCIO=ocio://cg-config-v0.1.0_aces-v1.3_ocio-v2.1.1`

And as before the Built-in config feature: 
`OCIO=path/to/my/config/cg-config-v0.1.0_aces-v1.3_ocio-v2.1.1.ocio`

Because the Configs Working Group is still working on the CG config, we will do a follow-up PR to replace it once the final version is ready.  However, once a Built-in config is included in an official release, the intent is to preserve it for backwards compatibility.  The implementation allows for multiple versions of a given config.


include/OpenColorIO/OpenColorIO.h
```diff
@@ -266,6 +266,22 @@ class OCIOEXPORT Config
...
+     * The available configNames are:
+     * "cg-config-v0.1.0_aces-v1.3_ocio-v2.1.1" -- ACES CG config, basic color spaces for computer
+     * graphics apps. More information is available at: 
+     * %https://github.com/AcademySoftwareFoundation/OpenColorIO-Config-ACES
+     * 
+     * \throw Exception If the configName is not recognized.
+     * 
+     * \return one of the configs built into OCIO library
+     */
+    static ConstConfigRcPtr CreateFromBuiltinConfig(const char * configName);
```
remia, line 283: [Question] Should we update the `CreateFromFile()` comment to mention the `ocio://` syntax?

src/OpenColorIO/builtinconfigs/BuiltinConfigRegistry.h
```diff
@@ -0,0 +1,106 @@
...
+     * config associated with that name.
+     * 
+     * For backward compatibility, built-in configs can be set as NOT recommended. They will
+     * still be available, but not recommended for the current version of OCIO.
+     * 
+     * @param name Name for the built-in config.
+     * @param config Config as string
+     * @param isRecommended Is the built-in config recommended or not.
+     */
+    void addBuiltin(const char * name, const char * uiName, const char * const config, bool isRecommended);
```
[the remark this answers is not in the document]
cedrik-fuoco-adsk: It came from some prototyping code that I forgot to remove. I'll remove it since there is no need.

src/OpenColorIO/Config.cpp
```diff
@@ -1117,6 +1122,21 @@ ConstConfigRcPtr Config::CreateFromFile(const char * filename)
         throw ExceptionMissingFile ("The config filepath is missing.");
     }
 
+    // Check for URI Pattern: ocio://<config name>
+    static const std::regex uriPattern("ocio:\\/?\\/?([^\\s]+)");
```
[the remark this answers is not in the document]
cedrik-fuoco-adsk: As per the standard, the double forward slashes are optional. But most URI only seems to accept one of the syntaxes.
I'll change the regex to make the double forward slashes required.

tests/python/BuiltinConfigRegistryTest.py
```diff
@@ -0,0 +1,152 @@
...
+            self.assertIsInstance(iterator[i], STRING_TYPES)
+
+    def test_contains(self):
+        # Valid __contains__ for all names.
+        for name in self.REGISTRY:
+            self.assertTrue(name in self.REGISTRY)
+            self.assertFalse(name not in self.REGISTRY)
+
+        # Invalid __contains__ for non-name.
+        with self.assertRaises(OCIO.Exception) as cm:
```
remia, line 56: I think in this case we would like the `__contains__` magic method not to throw if the name is not found, so the implementation should use the for loop instead of `getBuiltinConfigByName()`.

src/bindings/python/PyBuiltinConfigRegistry.cpp
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    return false;
+                }
+                return true;
+            })
+        .def("getBuiltinConfigs", [](PyBuiltinConfigRegistry & self) 
+            {
+                return BuiltinConfigIterator(self);
+            })
+        .def("getNumBuiltinConfigs", [](PyBuiltinConfigRegistry & self)
```
remia, line 61: [Minor] Could be removed now that we have the iterator support.

tests/python/BuiltinConfigRegistryTest.py
cedrik-fuoco-adsk: Make sense, I'll change it.

src/bindings/python/PyBuiltinConfigRegistry.cpp
cedrik-fuoco-adsk: Agreed. I'll remove it.

remia (approved):
Thanks for implementing the few suggested changes @cedrik-fuoco-adsk!

doug-walker (approved):

doug-walker:
This PR is finished so I'll go ahead and merge it to allow people to experiment with the new feature.  But please take note that the current built-in is only for testing and we will remove it and replace it with the official version of the CG config (with a different name string), once it is released.

PR #1668, opened 2022-07-10
Allow loading dlls from PATH on windows and python>=3.8

anderslanglands:
python 3.8 stopped loading DLLs from PATH, which breaks PyOpenColorIO when it and its dependencies have been built as shared libraries. This PR provides an optional (gated behind an env var) fix to shim the paths in PATH back in using `os.add_dll_directory()`. This is the same approach as taken by USD.

### DIFF
--- a/src/bindings/python/CMakeLists.txt
+++ b/src/bindings/python/CMakeLists.txt
@@ -227,9 +227,9 @@ target_compile_definitions(PyOpenColorIO
 )
 
 if(WIN32)
-    set(_Python_VARIANT_PATH "${CMAKE_INSTALL_LIBDIR}/site-packages")
+    set(_Python_VARIANT_PATH "${CMAKE_INSTALL_LIBDIR}/site-packages/PyOpenColorIO")
 else()
-    set(_Python_VARIANT_PATH "${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages")
+    set(_Python_VARIANT_PATH "${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages/PyOpenColorIO")
 endif()
 
 # Create an internal global variable to access it in another scope but not publicly visible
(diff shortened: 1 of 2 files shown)


lgritz:
Is an analogous change required on all projects that have Python bindings?

KelSolaar:
> Is an analogous change required on all projects that have Python bindings?

It directly related to that: https://github.com/AcademySoftwareFoundation/Imath/issues/238

anderslanglands:
I’m working on a PR for you to review right now  Larry.

remia:
Thanks for the PR @anderslanglands! Looks like the commits are missing the sign-off to make GitHub action happy.

anderslanglands:
Hi Rémi, how do I get that?

remia:
@anderslanglands you can use the `--signoff` flag when you commit from the CLI, if you want to sign-off the last commit you can also use `git commit --amend --signoff`. GUI tools will also typically provide a way to do this, I use `git-gui` which have a button next to the commit. You will probably need to force push to remove the un-signed commit from the PR entirely.

anderslanglands:
Ahh gotcha, thanks. Sorry I’ve not done that before.

I actually was going to push another commit anyway: OIIO recently merged
the same PR, but in that case we decided to invert the logic of the feature
flag to load DLLs from PATH by default (ie match the Python <3.8
behaviour), and let the user disable that with the env var if they want the
3.8 behaviour.

This is also more similar to how USD does it (where they just do the path
loading thing without a way to disable it).

Cheers,

Anders.

remia:
Hi @anderslanglands, just checking again, do you still plan to update this PR like you mentioned?

doug-walker:
@anderslanglands , thanks for identifying this problem and proposing a solution.  We fixed this issue in PR #1759, so I'll go ahead and close this PR now.  Thanks again for your help!

anderslanglands:
Thanks Doug!

JeanChristopheMorinPerso:
As a side note, I think it's important to note that it should be opt-in, not opt-out. This can break a lot of things in an environment and should only be activated manually. Basically, it's forcing a behavior in the entire environment, which could potentially break packages which were not expecting that.

doug-walker:
@JeanChristopheMorinPerso, thank you for your comment.  We discussed this during the TSC meeting today and the group agreed that having the default be to emulate the Python 3.7 behavior is not what we want as it could be a security risk.  I've created issue #1785 for continuing this discussion, if you or Anders have any further suggestions.

JeanChristopheMorinPerso:
Great! Thanks for the update @doug-walker!

Issue #1624, opened 2022-03-26
errno checks break musl compatibility


tibequadorian:
In the 'from_chars' function, it's first checked if errno != 0 and immediately returns with std::errc::result_out_of_range aka ERANGE.
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/4fa94918c2cf572dcaf61ca07016f3b5c231c14c/src/utils/NumberUtils.h#L71-L74

According to `strtod(3p)`, if no conversion could be performed, errno may be set to EINVAL.
Unlike glibc which doesn't do this, musl does. This leads to an incorrect return value (ERANGE != EINVAL).

I discovered this through failing tests on musl.

thesamesam:
@tibequadorian seems to have written a patch for this at: https://inbox.vuxu.org/voidlinux-github/20220415165812.WIf9xdfnApCy_0So7fM1SoqKt8otlLfcokk5uD2d_E0@z/.

tibequadorian:
Yes here is the original: https://github.com/tibequadorian/void-packages/blob/openimageio/srcpkgs/opencolorio/patches/musl.patch
It's a minimal patch so I don't know if it's the best solution but it works.

PR #1678, opened 2022-08-23
Fix build and errno check on musl libc

listout:
In the 'from_chars' function, it's first checked if errno != 0 and
immediately returns with std::errc::result_out_of_range aka ERANGE.
Please refer issue [1624] (https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1624)

The function 'strtol_l' is not available on non GLIBC systems hence on
other libc use the alternative 'strtol' function.

Signed-off-by: listout <listout@protonmail.com>


remia:
Thanks for the contribution @listout. Out of curiosity, and sorry for being dumb here, why is returning the wrong error code problematic? Using `strtol()` instead of `strtol_l()` looks ok, especially if it fixes a compilation error on musl, though I am also curious what is the benefit of the locale aware variant for integers in this context of LUT file parsing.

listout:
> why is returning the wrong error code problematic? 

I'm not too sure either, in the linked issue, you see that someone mentioned that a test case was failing and some other detail

>> According to strtod(3p), if no conversion could be performed, errno may be set to EINVAL.
Unlike glibc which doesn't do this, musl does. This leads to an incorrect return value (ERANGE != EINVAL).

>> I discovered this through failing tests on musl.

As for

> Using `strtol()` instead of `strtol_l()` looks ok, especially if it fixes a compilation error on musl, though I am also curious what is the benefit of the locale aware variant for integers in this context of LUT file parsing in the first place.

It could be because of musl having only one locale

Sorry @remia for not having any concrete answers, I too don't know much about the issue itself. Picked it up from another issue and compiling opencolorio on musl

remia (approved):
Tested on Docker alpine/latest and the patch seem to fix both build and unit test so I am going to approve, thanks.

cedrik-fuoco-adsk:
Looks good to me too. I am going to approve it.

cedrik-fuoco-adsk (approved):

michdolan (approved):

PR #1689, opened 2022-09-24
Add NamedTransform Processor constructors

### LINKED ISSUE #1674: NamedTransform Processor API Proposal
Currently the only way to create a processor from a named transform is to pair the named transform name with a color space name (`config.getProcessor("cs", "nt")` or `config.getProcessor("nt", "cs")`), which results in a processor that ignores the color space and applies the named transform in isolation, or if both arguments are named transforms, joins them.

What is unclear about this interface is that the position of the named transform in the `getProcessor` arguments determines the direction that it is applied (first argument == forward, second argument == inverse). If you specify two named transforms, the first will be applied in the forward direction, and the second in inverse. This behavior doesn't appear to be documented outside of comments in the function. We can definitely document it, but I think this interface will lead to confusion in the long run since directionality is usually explicitly specified elsewhere in the API where it would otherwise be unclear. This is not an issue with color spaces since they interact with a reference space and directionality is a product of their ordering, but since named transforms have no ties to other objects in an OCIO config, it's not immediately clear what direction would be used in these interactions. 

Since named transforms have not been used much yet, I think it's a good opportunity to revise the interface (in a backwards compatible way). I propose that we add two new processor constructors to `Config` with these signatures:
```
Config::getProcessor(const ConstNamedTransformRcPtr & namedTransform, TransformDirection direction) const
Config::getProcessor(const char * & namedTransformName, TransformDirection direction) const
```
In addition, I think we could deprecate support of named transform names in the existing processor constructors. In 2.x it would continue to function as it has since 2.0, but we could consider a warning message or just clear documentation of it being deprecated. Then in 3.0 we could remove it.

To illustrate the current interface, see the following example code:

```
>>> import PyOpenColorIO as ocio
>>> config = ocio.Config.CreateRaw()
>>> ref = ocio.ColorSpace(name="ACES2065-1", isData=False)
>>> config.addColorSpace(ref)

>>> nt = ocio.NamedTransform("ST-2084 - Curve", forwardTransform=ocio.BuiltinTransform("CURVE - LINEAR_to_ST-2084"))
>>> config.addNamedTransform(nt)

# Inverse named transform
>>> proc = config.getProcessor("ACES2065-1", "ST-2084 - Curve")
>>> cpu_proc = proc.getDefaultCPUProcessor()
>>> cpu_proc.applyRGB([100, 10, 1.08])
[65504.0, 65504.0, 217.6101837158203]

# Forward named transform
>>> proc = config.getProcessor("ST-2084 - Curve", "ACES2065-1")
>>> cpu_proc = proc.getDefaultCPUProcessor()
>>> cpu_proc.applyRGB([100, 10, 1.08])
[1.0, 0.7518271207809448, 0.5158224105834961]
```

And the proposed revised interface:
```
# Inverse named transform
>>> proc = config.getProcessor(nt, direction=ocio.TRANSFORM_DIR_INVERSE)
>>> cpu_proc = proc.getDefaultCPUProcessor()
>>> cpu_proc.applyRGB([100, 10, 1.08])
[65504.0, 65504.0, 217.6101837158203]

# Forward named transform
>>> proc = config.getProcessor("ST-2084 - Curve")
>>> cpu_proc = proc.getDefaultCPUProcessor()
>>> cpu_proc.applyRGB([100, 10, 1.08])
[1.0, 0.7518271207809448, 0.5158224105834961]
```

michdolan:
This PR adds four new processor constructors to `Config`:

- `NamedTransform` object and direction
- `Context`, `NamedTransform` object and direction
- Named transform name/alias and direction
- `Context`, named transform name/alias and direction

Also included are the updated Python bindings, C++ and Python unit tests, and updated frozen Python API docs (which update more than what I changed in this PR).

Fixes #1674


remia (approved):

src/OpenColorIO/Config.cpp
```diff
@@ -3971,6 +3971,38 @@ ConstProcessorRcPtr Config::getProcessor(const ConstContextRcPtr & context,
...
+}
+
+ConstProcessorRcPtr Config::getProcessor(const ConstContextRcPtr & context,
+                                         const ConstNamedTransformRcPtr & namedTransform,
+                                         TransformDirection direction) const
+{
+    // This gets or calculates the named transform in the requested direction. The result
+    // is always applied in the forward direction.
+    auto transform = NamedTransformImpl::GetTransform(namedTransform, direction);
+    return getProcessor(context, transform, TRANSFORM_DIR_FORWARD);
```
doug-walker, line 3988: This is perfect.  

However, it makes me realize now that my reply to your issue #1674 was wrong.  What you have here is what I was thinking of but I wrote "getTransform" instead of "GetTransform".  But now that I check, it seems that GetTransform was never added to the public API.  That was an oversight on our part because that was how I was intending people would feed the getProcessor call (since it removes the need for the caller to handle which transform directions are actually in the NamedTransform).

Sorry for missing that.  I understand better now why you felt adding these methods was important.  :)

If you have time, it would be nice to bring GetTransform forward to the public API.  (But no worries, we could do it later too.)

doug-walker (approved):
Thanks for updating the frozen Python docs too!

src/OpenColorIO/Config.cpp
michdolan: Thanks Doug. I can try to bring `GetTransform` forward to the public API this evening.

Issue #1623, opened 2022-03-22
ICC parser fails when reading some common ICC profiles


est77:
When trying to use virtual views in OpenColorIO 2.1 the ICC file format parser fails for some common profiles.
One example is the Display P3 profile included with OSX (I am not sure I can redistribute this file).
The error is:

'International Color Consortium profile' failed with: Error parsing .icc file (/System/Library/ColorSync/Profiles/Display P3.icc).  Illegal curve tag in ICC profile.

remia:
Hi @est77,

I did try on my default "Color LCD" profile shipping with the MacBookPro and it worked fine. Could you check that your ICC profile contain the rTRC, gTRC and bTRC tags? OCIO only support Matrix / TRC profiles for virtual displays currently.

doug-walker:
@remia , you are right, the ICC spec defines a bunch of different Parametric Curve Types and OCIO does not support all of them currently.  I don't think they all would work as an ExponentWithLinear, but if not, one could either build a Lut1D or define a new FixedFunction for the other parametric types.

remia:
Agree @doug-walker that they won't all fit the `ExponentWithLinearTransform`, maybe it would be cleaner to have a `FixedFunctionTransform` supporting each form of parametric curve through parameters to have it all defined in one place.

est77:
Thanks for looking into it. I think that it would be great if this could be fixed. Using ICC profiles in OCIO is an important feature IMO and we have some user requests to support it in our software.

remia:
Just checking before diving into this, has anyone planed to work on this yet?

About the above proposal of adding a new `FixedFunctionTransform` for the ICC Parametric Curve's 5 different types, is there any opinions compared to simply adding a 1D LUT baking directly in `FileFormatICC` as proposed by @doug-walker ? Having this new fixed function will duplicates some of the already existing transforms (like the pure exponent and moncurve style) and using it instead may introduce some performance regression in case we don't support the same optimisations (SSE on CPU).

It seems that if the only foreseen user for this will be `FileFormatICC`, the effort of adding this new FixedFunction might be hard to justify?

doug-walker:
@remia , it would be great to have you work on this!

[...]

PR #1691, opened 2022-09-27
Support for ICC Parametric curves type 1-4

remia:
Address issue #1623 by supporting ICC Parametric Curve type 1-4 implemented as 1DLUT. I used a side-car python [script](https://gist.github.com/remia/4cab0fb18a794d699d8ae2d6b77091a4) during my tests to compare the results against pure Python or [DemoICCMax](https://github.com/InternationalColorConsortium/DemoIccMAX). Opening that now to allow time for discussion.

- Size of the 1DLUT is currently 1024, increasing that was decreasing the match against the pure Python implementation, I have a feeling this is related to inversion and flat slopes,
- Follows ICC [white paper](https://www.color.org/whitepapers/ICC_White_Paper35-Use_of_the_parametricCurveType.pdf) recommendations on parameters validation, there are no unit tests validating all possible failure case at the moment,
- Added a RangeTransform on the pure exponent path to follow ICC specs more closely. Unit tests sample updated using directly OCIO results for the case where clipping changed the expected values as I don't know exactly how the values were obtained,
- As before, all the TRC tags in the profile should be of the same type, the individual parameters can vary among red, green and blue but the function type should match.


michdolan (approved):

michdolan:
This looks good to me, but I think @doug-walker should review before merging.

src/OpenColorIO/fileformats/FileFormatICC.cpp
```diff
@@ -235,6 +244,197 @@ LocalCachedFileRcPtr LocalFileFormat::ReadInfo(std::istream & istream,
...
+        const float a = SampleICC::icFtoD(params[1]);
+        const float b = SampleICC::icFtoD(params[2]);
+        const float d = SampleICC::icFtoD(params[4]);
+
+        if ((a * d + b) < 0)
+        {
+            ThrowParaError("Expecting no negative arguments to the power law.");
+        }
+    }
+}
```
doug-walker, line 373: I notice in your Python script you implemented some additional checks (e.g. continuity at the break-point) that are warnings rather than errors.  Would it make sense to include those here and call LogWarning?

doug-walker (changes requested):
This is awesome Remi, nice work!

src/OpenColorIO/fileformats/FileFormatICC.cpp
```diff
@@ -235,6 +244,197 @@ LocalCachedFileRcPtr LocalFileFormat::ReadInfo(std::istream & istream,
...
+        const float a = SampleICC::icFtoD(params[1]);
+        const float b = SampleICC::icFtoD(params[2]);
+        const float c = SampleICC::icFtoD(params[3]);
+        const float d = SampleICC::icFtoD(params[4]);
+
+        v = v >= d ? std::pow(a * v + b, g) : c * v;
+    }
+    // Type 4:
+    // y = (ax+b)^g + e  (x >= d)
+    // y = cx+f          (x < d)
```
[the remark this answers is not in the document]
remia: Yeah I noticed that too, I think it is now fixed in the latest ICC.1:2022 (Profile version 4.4.0.0) spec.

src/OpenColorIO/fileformats/FileFormatICC.cpp
```diff
@@ -235,6 +244,197 @@ LocalCachedFileRcPtr LocalFileFormat::ReadInfo(std::istream & istream,
...
+        ThrowParaError(oss.str());
+    }
+
+    // Monotically non-decreasing (flat segments permitted)
+    const float g = SampleICC::icFtoD(params[0]);
+
+    // Forces the power law to be monotically non-decreasing.
+    if (g <= 0.0)
+    {
+        ThrowParaError("Expecting monotically non-decreasing power-law.");
```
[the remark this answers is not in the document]
remia: Will fix the typos thanks.

tests/cpu/fileformats/FileFormatICC_tests.cpp
```diff
@@ -337,6 +362,225 @@ OCIO_ADD_TEST(FileFormatICC, test_apply)
...
+OCIO_ADD_TEST(FileFormatICC, test_apply_para_t3)
+{
+    // Check processing of ParaCurve type 3.
+    {
+        static const std::string iccFileName("icc-test-pc3.icc");
+
+        float srcImage[] = {
+            -1.0f,  -1.0f,  -1.0f,  1.0f,
+             0.0f,   0.0f,   0.0f,  1.0f,
+             0.18f,  0.18f,  0.18f, 1.0f,
```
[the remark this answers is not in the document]
remia: Good catch, I'll add a smaller value for better coverage and add the parameters for the curves as comments.

src/OpenColorIO/fileformats/FileFormatICC.cpp
remia: I'll try to add them good point.

remia: I added the warning from the Python script, note that I did bump the comparison tolerance by using 8 bits quantisation for these to avoid warnings on standard ICC profiles like Display P3. What do you think @doug-walker ?

doug-walker: Thanks for adding!  Yes, I think the 8-bit quantization is appropriate to avoid unnecessary warnings.

doug-walker (approved):
All comments addressed, thank you Remi!

PR #1693, opened 2022-09-27
Baking crosstalk and NamedTransform support fix

remia:
Follow up of #1598, fix the crosstalk shaper space check making sure we use the optimised processor to allow typical collapse of matrices. There is another issue related to the use of `NamedTransform` as shaper space which appear to not work currently because we are using `LookTransform` which doesn't support these.


doug-walker (approved):

cedrik-fuoco-adsk (approved):

remia:
Tagging as draft while I investigate the issue with NamedTransform.

cedrik-fuoco-adsk (approved):

doug-walker (approved):

michdolan (approved):

PR #1694, opened 2022-09-27
Log OCIO version for debugging purposes

remia:
While debugging various DCCs OCIO support through `OCIO_LOGGING_LEVEL`, I found that this simple print of the library version could be useful.


doug-walker (approved):

KevinJW (approved):
LGTM

PR #1697, opened 2022-10-02
MinGW: Work around lack of strtof_l for locale independent parsing

amyspark:
👋 

This PR is to catch a slipup in #1496 I've only found yesterday. MinGW-based compilers lack ucrtbase's definition of `_strtof_l` completely, and `strtod_l` is also missing from GCC, so neither are viable alternatives to achieve locale-independent floating point number parsing.

In this patch, I propose to apply a workaround similar to the one existing for macOS, that covers only this flavor of compilers.

### DIFF
--- a/src/utils/NumberUtils.h
+++ b/src/utils/NumberUtils.h
@@ -99,7 +99,12 @@ really_inline from_chars_result from_chars(const char *first, const char *last, 
 
     float
 #ifdef _WIN32
+#if defined(__MINGW32__) || defined(__MINGW64__)
+    // MinGW doesn't define strtof_l (clang/gcc) nor strtod_l (gcc)...
+    tempval = static_cast<float>(_strtod_l (first, &endptr, loc.local));
+#else
     tempval = _strtof_l(first, &endptr, loc.local);
+#endif
 #elif __APPLE__
     // On OSX, strtod_l is for some reason drastically faster than strtof_l.
     tempval = static_cast<float>(::strtod_l(first, &endptr, loc.local));


src/utils/NumberUtils.h
```diff
@@ -99,7 +99,12 @@ really_inline from_chars_result from_chars(const char *first, const char *last,
 
     float
 #ifdef _WIN32
+#if defined(__MINGW32__) || defined(__MINGW64__)
```
amyspark, line 102: Alternatively, should I make this check in the CMakeLists file, so that this is safely reverted to `_strtof_l` once the mingw-w64 platform updates their SDK?

remia:
Just checking @amyspark, does that mean latest OCIO releases do not compile on MinGW?

amyspark:
@remia No, they do not. I had not noticed because our build was still running the old patch (which did not rely on the affected function).

cedrik-fuoco-adsk (approved):

src/utils/NumberUtils.h
remia: Ideally it would be a test directly in the source code if possible? Can be done later in any case I think given the current compilation issue.

remia (approved):

doug-walker (approved):

amyspark:
Thanks @doug-walke, @remia and @cedrik-fuoco-adsk ! Please don't forget to cherry-pick to 2.1 as well.

doug-walker:
@amyspark , this fix is now in OCIO 2.1.3, per your request.

Issue #1399, opened 2021-05-05
Add a robust method to query whether a color space encoding is linear


doug-walker:
It is sometimes useful for applications to know if a color space encoding is linear.  In OCIO v2 there is a new "encoding" attribute of a color space that provides this information.  However, given that the attribute is optional, and was not present at all in v1 configs, it is very possible that it will not be present and hence a more robust method with one or more fall-backs would be helpful.

There are a number of heuristics that could be used to guess if a space is linear but they all seem to involve comparison to another space in the config that is known to be linear.  The proposed assumption is that the reference space is assumed to be linear and thus may be used as the basis for the comparison.  This would be assumed for both the main (scene) reference space as well as the display reference space, if present.

To avoid the assumption of a linear reference space, the comparison could alternatively be done against the interchange roles.  But those are also new in v2 and also optional, so I'm not sure it's worth the effort.  If the config author didn't specify the encoding attribute, they probably would not have set those new roles either.

The comparison could either be via introspection of the transform to the reference space or by evaluating some values and checking the results.  Introspection of the transform seems like the more straight-forward approach.  So the proposal is to look at the color space's transform to its reference space.  The transform would be converted to ops and if it only contains Matrix or Range ops (or nothing), then it will be assumed to be linear.  

It would certainly be possible to create a linear color space that does not work with this heuristic but the easy work-around is simply to set the encoding attribute to be scene-linear or display-linear.  That will take precedence over the heuristic.

The proposed method would be called isLinear and would be on the ColorSpace class.

The suggested implementation is to also add an isLinear method to the OpData class, similar to isNoOp.  That way the ops may respond based on the current settings.  The top level method would return true if all the ops return true.

Edge case handling (I could be convinced otherwise on these, if people disagree):
 - If cs->isData() is true, the method would return false.
 - A Matrix op would return true even if it has an offset that is non-zero.
 - A Range op would return true even if it includes an offset.
 - The CDL op would return true if the power values are 1.
 - Other ops would return true if their isIdentity is true.
 - Some of the FixedFunction op styles would also be considered linear, including the ACES 1.3 gamut mapping op, once we add it.

doug-walker:
Closing as implemented via PR #1703.

PR #1703, opened 2022-10-19
Adsk contrib - Add a method to query whether a color space is linear

cedrik-fuoco-adsk:
This PR is the implementation of issue #1399. A new method, called **isColorSpaceLinear**, has been added to the Config class.

See the method header in OpenColorIO.h for detailed information.


remia:
Looks good to me, just wondering about usage of the Reference Gamut Compression transform as it is mentioned in the linked issue. Is it still going to return isLinear -> true because we don't have negative values in the linearity test on purpose?

doug-walker:
@remia , you raise a good point.  The existing test is somewhat permissive.  Certain CDL, FixedFunction, LUT, and Range ops could pass if they are linear over the tested points.  Is that desirable?  We could be more strict and just check if the optimized transform is a single Matrix, without offsets.   We're very open to feedback on this.

remia:
Thanks Doug, I think the current behaviour is good, just wanted to double check I understood correctly. Also poi ting that we are missing the python binding here.

remia (approved):

Issue #1686, opened 2022-09-06
Add new Built-in transforms to support new ACES Studio config


doug-walker:
We want to allow the upcoming Studio config to be LUT-free for version 2.2 of the library.

Here is the current list of transforms to add:

- ARRI LogC4
- CanonLog3

doug-walker:
Closing this as fixed via PR #1704.

PR #1704, opened 2022-10-21
Adsk contrib - Add Built-in Transforms for ARRI LogC4 and Canon curves

doug-walker:
Adds Built-in Transforms for:
-- ARRI LogC4
-- Canon Log2 curve
-- Canon Log3 curve

I verified that the CLF file exported from the Built-in Transform for LogC4 matches the CLF that was recently added to the OpenColorIO-Config-ACES repo for the Studio config.

The Canon curves are implemented as a normal domain rather than half-domain Lut1D based on the preference received from Canon during the development of the Studio config. (The shape of the curves currently precludes implementing them as a LogCameraTransform.)

Implements Issue #1686.

Signed-off-by: Doug Walker [doug.walker@autodesk.com](mailto:doug.walker@autodesk.com)


remia (approved):

remia:
Missed that earlier, but is there some update related to config version checks related to the use of the newly introduced builtin transforms? Similar to what is done [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/Config.cpp#L4536) or [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/fileformats/ctf/CTFTransform.cpp#L256) for the RGC.

doug-walker:
Excellent question @remia !   No updates are needed on the CTF side since Builtin Transforms are not actually saved in CTF, they are expanded first.  (And actually I should log a feature request for that since we ran into a use for that on the configs working group.)  However, I should update checkVersionConsistency in Config.cpp and will push another commit to do that.

michdolan (approved):

PR #1706, opened 2022-10-25
Add Python 3.11 wheels

remia:
Following Python 3.11 release, this PR update the wheel generator to include it.

Other updates:
* Python 3.6 is EOL so I removed it,
* All 32 bits architecture variants removed,
* cibuildwheel updated to latest version for Python 3.11 support,
* Build arm64 wheels for Apple silicon instead of universal2 following discussion in https://github.com/pypa/cibuildwheel/issues/1204 and Numpy distribution model.

If there is concerns about some of the removal, I'm not strongly opposed to adding them back, also noting that there should now be a source distribution allowing pip to build the package on the unusual version / architecture.


doug-walker:
The main item that stands out to me here is "Build arm64 wheels for Apple silicon instead of universal2".  @remia, would it be useful for you to post on the OCIO Slack to check if anyone needs the universal2?  Or is that something you're confident won't cause any issues?

remia:
@doug-walker I will ask on slack good point, from what I understand, one of the only use case is if someone wants to distribute a Python application where the end user architecture is unknown. But given that well known projects, including Numpy don't ship universal2 wheel, I think this is relatively safe choice.

JeanChristopheMorinPerso:
@remia https://cibuildwheel.readthedocs.io/en/stable/faq/#universal2 nicely describes the situation around arm64 vs universal2 wheels, in case it might be useful to anyone (in addition to the link you provided).

setup.cfg
```diff
@@ -12,6 +12,7 @@ classifiers =
     Programming Language :: Python :: 3.8
```
JeanChristopheMorinPerso, line 12: I don't seem to be able to comment on the right line, so commenting here: Should you remove 3.6?

remia: Yeah good point I did wonder about that and decided to leave it in because we still technically support it. But thinking about it again, I guess it would make sense to remove given that these metadata are probably mostly used to filter packages on PyPI and we don't ship wheels anymore for that version (although you would still be able to compile with the source distrib).

JeanChristopheMorinPerso: Actually, it's a good point. But OCIO doesn't set the https://packaging.python.org/en/latest/specifications/core-metadata/#requires-python metadata, which is kind of dangerous to not set. The classifiers are just used for actually filtering stuff. `requires-python` tells installers which Python version the package supports.

remia: I guess this was what the .toml [project] section was defining before, is that [python-requires](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires)? Looks like fairly new addition.

JeanChristopheMorinPerso: Oh yeah indeed. I don't know what I was thinking about. Anyway, it should be moved into `setup.cfg` for it to be included in the metadata, as per the warning I pointed to.

michdolan (approved):

PR #1716, opened 2022-10-31
Adsk contrib - Removing library "dev" label for official release 2.2

cedrik-fuoco-adsk:


doug-walker:
The expected CI checks passed on the three platforms, so I'll go ahead and merge this.  We will do a follow up to fix the extra CI checks.

Issue #1680, opened 2022-08-23
Install the pkg-config file when building under MSYS on windows?


agriggio:
Hi,
it would be good if the pkg-config script (`OpenColorIO.pc`) was installed when building under MSYS on windows. Here's a simple patch to do that:

```diff
--- b/src/OpenColorIO/CMakeLists.txt	2022-06-05 20:59:43.000000000 +0200
+++ a/src/OpenColorIO/CMakeLists.txt	2022-08-23 17:17:20.516786583 +0200
@@ -171,7 +171,7 @@
     SystemMonitor.cpp
 )
 
-if(NOT WIN32)
+if(NOT WIN32 OR MSYS)
 
     # Install the pkg-config file.
 
```

kmilos:
This should be `MINGW` instead of `MSYS`. (`MSYS` is a POSIX emulation environment like Cygwin where I doubt OCIO is useful; one is typically building for the WIn32 API, i.e. MINGW).

`pkg-config` runs everywhere (available via [vcpkg ](https://vcpkg.io/en/docs/maintainers/ports/vcpkg-pkgconfig-get-modules.html)for MSVC), so the `if` clause should just be simply removed and the pc filed installed always.

amyspark:
[pkgconf](https://github.com/pkgconf/pkgconf) is also a working alternative if you use Meson.

kmilos:
Indeed, vcpkg seem to use meson+pkgconf quite a bit for Win32 packaging.

doug-walker:
@agriggio , @kmilos , @amyspark  Please review PR #1720 and check that it meets your needs, thanks!

PR #1720, opened 2022-11-08
Adsk contrib - Configure the OpenColorIO.pc file on Windows

cedrik-fuoco-adsk:
This is for PR [Install the pkg-config file when building under MSYS on Windows?](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1680)

I removed the WIN32 conditions since pkg-config is available on Windows (through vcpkg for example).
In order to have the right value in OpenColorIO.pc, I've added CMAKE_INSTALL_PREFIX to the cmake build command of ocio.bat.


kmilos:
LGTM, thanks!

amyspark (approved):
LGTM too, thanks for the fix!

remia (approved):

kmilos:
See also https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1730, the relative vs absolute handling should probably be included here as well.

kmilos:
Should `src/OpenColorIO/pkgconfig/OpenColorIO.pc.in` be removed, isn't it redundant?

cedrik-fuoco-adsk:
> > No, it is needed. OpenColorIO.pc is generated from OpenColorIO.pc.in. It is a template file that changes based on variables.
> 
> I realize that. However, [the actual template used seems to be `src/OpenColorIO/res/OpenColorIO.pc.in`](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/cac1ba5feffa8a59ddd072f4fd8db62c795acf64/src/OpenColorIO/CMakeLists.txt#L185) though. So, it seems `src/OpenColorIO/pkgconfig/OpenColorIO.pc.in` is redundant, no?

I didn't realize that there were two files. It is indeed not used. I'll remove it.

src/OpenColorIO/CMakeLists.txt
```diff
@@ -178,8 +178,23 @@ set(SOURCES
...
+else()
+    set(libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
+endif()
+
+# CMAKE_INSTALL_INCLUDEDIR is not guaranteed to be relative.
+# Not using cmake_path function since it is only available from CMake ≥ 3.20.
+if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
+    set(includedir "${CMAKE_INSTALL_INCLUDEDIR}")
+else()
+    set(includedir "\${exec_prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
```
kmilos, line 195: Should be `${prefix}` for include directory.

cedrik-fuoco-adsk: I didn't want to change it since I don't know the reason why it was changed to ${exec_prefix} in the past. It was changed to ${exec_prefix} in [PR #1120](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1120/files)

Pinging @remia

kmilos: It should be restored back then.

remia: I don't remember a specific reason I made this change and the build still works fine for me when I update that. Though I'm not sure what it changes exactly given that we set on line 180 the two to be the same thing? Or did I misunderstood the logic here (I didn't write this code)?

kmilos: The GNU documentation linked above is pretty clear for `libdir` and `includedir` prescriptions, they are different, e.g.
```
libdir=${exec_prefix}/lib
includedir=${prefix}/include
```

cedrik-fuoco-adsk: Thanks for the link @kmilos. I'll make the change.

kmilos: Thanks, I think the pkg-config subject should be sorted for a while now ;)

doug-walker (approved):

PR #1723, opened 2022-11-10
replace texture2D function with texture for GLSL 1.3

bartstyczen:
texture2D is deprecated after GLSL 1.2

Signed-off-by: Bart Styczen <bart.styczen@cine.dev>


remia (approved):
Thanks for the PR @bartstyczen !

doug-walker (approved):

Issue #1718, opened 2022-11-02
Impossible to build against system/external minizip-ng


LazyDodo:
the minizip-ng build over [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/share/cmake/modules/Findminizip-ng.cmake#L136) specifies to install the headers in a subfolder called `minizip-ng` however an external build of minizip-ng is unlikely to have this option set as it is non standard. 

The problem arises [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/OCIOZArchive.cpp#L20) where the code blindly assumes this sub folder to exist, which isn't the case unless you let OCIO build minizip.

remia:
Also noticed OpenColorIO won't find minizip-ng built with `MZ_COMPAT=ON`, which is the default, because that build a library without the ng suffix. I'm not familiar enough with this library to know if this is expected or not to have that kind of incompatibility.

LazyDodo:
ran into that as well, but you can sidestep that by passing ` -Dminizip-ng_LIBRARY=/where/ever/it/lives/blah.a_lib_so`  to your cmake configure invocation for OCIO.

kmilos:
See also patches that were necessary for unbundling the MSYS2 OCIO package: https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-opencolorio

cedrik-fuoco-adsk:
Hi @kmilos,

Could you provide the steps to reproduce the issue fixed by 0005-Fix-Findminizip-ng.patch? I've tried to build it manually under MSYS2 without that patch (I removed it manually), and it seems to be working correctly.

kmilos:
@cedrik-fuoco-adsk You're probably right that `0005-Fix-Findminizip-ng.patch` is not absolutely necessary (btw, the convention is to use the plural `_LIBRARIES` and `_INCLUDE_DIRS` variables, but that's a minor detail), I think it's just the maintainer preferring to do things the "more modern" CMake way...

PR #1725, opened 2022-11-10
Adsk contrib - Fix issue with minizip build

cedrik-fuoco-adsk:
This is the PR for [Impossible to build against system/external minizip-ng](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1718).

I removed the minizip-ng subfolder part when including minizip-ng's headers. This fixes the issue with external builds that don't use that subfolder.

For internal builds, a subfolder is created like the other modules that OCIO is able to download and install manually.

I've decided that having Findminizip-ng.cmake and Findminizip.cmake is less confusing than having only a Findminizip.cmake since "minizip" is not the actual name of the library (minizip-ng without MZ_COMPAT=ON).

Therefore, the build system does the following:
1) Search for minizip-ng (libminizip-ng) (without MZ_COMPAT=ON)
2) If minizip-ng is not found,  Search minizip-ng with MZ_COMPAT=ON (which is libminizip).
3) If steps 1 and 2 are not found, download and install minizip-ng with MZ_COMPAT=OFF. 
It is set as OFF because we don't need the compatibility part. We are using the "new" API from minizip-ng.


cedrik-fuoco-adsk:
Note that the analysis workflow is failing, but it was failing before already. The PR does not introduce any new problems with the analysis workflow.

doug-walker:
@remia @michdolan , could one of you review this?  Merging this would simplify an upcoming PR for us.  TIA!

remia:
Looks good to me, thanks for updating the Analysis workflow! One thing I noticed while testing on macOS, the build with install ALL externals deps works perfectly but when using the minizip-ng from brew, it seems the `mz_strm_zlib.h` header is missing from `/usr/local/include` (all the other minizip include are present though). Does anyone else have this issue?

cedrik-fuoco-adsk:
> Looks good to me, thanks for updating the Analysis workflow! One thing I noticed while testing on macOS, the build with install ALL externals deps works perfectly but when using the minizip-ng from brew, it seems the `mz_strm_zlib.h` header is missing from `/usr/local/include` (all the other minizip include are present though). Does anyone else have this issue?

The minizip-ng packages available through brew do not have a dependency on _zlib_. That tells me that particular minizip-ng wasn't built with _zlib_ compression activated. Based on the dependencies list, it was built with _xz_ and _zstd_ compression activated - which corresponds with the missing `mz_strm_zlib.h`.

doug-walker:
Thanks for the review Remi!  Regarding the Brew issue, based on Cedrik's reply, it sounds like the Brew minizip-ng cannot be used with OCIO since it was not compiled with zlib support activated.  In any case, I think we could merge this one now to unblock Cedrik's other PR.  We could come back to the Brew issue later, if that's ok with you.

remia (approved):

remia:
Thanks @cedrik-fuoco-adsk yeah that make sense, I guess I'm surprised that zlib is not used it seemed like an obvious choice (to my naive eyes). Agree @doug-walker sorry for dragging this PR!

doug-walker:
Don't worry Remi, you did not drag the PR.  You raised an excellent point about Brew.  And in fact you sped up the PR by reviewing it, thank you!!

PR #1729, opened 2022-11-17
Adsk contrib - Hiding minizip-ng symbols on Mac

cedrik-fuoco-adsk:
This PR fixes the symbol leakage of minizip-ng on Mac.


cedrik-fuoco-adsk:
I've added a fix for a linking issue related to the code that tries to hide expat symbols (on macOS).

remia:
Can confirm this seems to fix the symbol leakage, thanks @cedrik-fuoco-adsk! Do you have more info on why the library path flags (`-Wl,L`) are now needed?

cedrik-fuoco-adsk:
> Can confirm this seems to fix the symbol leakage, thanks @cedrik-fuoco-adsk! Do you have more info on why the library search path flags (`-Wl,-L`) are now needed?

I figured it was missing since I was getting an _unknown error_ when I tried to do the same thing with minizip-ng. 
The `-l` and `-l-hidden` are the same (**they both link to the library specified**) but the second one has the hidden functionality. 
It a very different options from `--exclude-libs` from the other platform.

When we were specifying `-l-hidden-expat`, we are actually linking against expat, and since we weren't specifying any custom path, the linker tries to find it in the default paths. Since macOS comes with an installation of expat (shared version), it was always finding that shared version somewhere on the system.

Therefore, we had an odd scenario where even if you were to specify OCIO_INSTALL_EXT_PACKAGES to MISSING/ALL or trying to specify your own expat installation (shared or static), you'd get the dynamic version of expat from the existing installation on macOS.

By using <package_name>_LIBRARY, we can make sure that we are pointing to the right version (shared or static) of the library.

remia:
Thanks @cedrik-fuoco-adsk that's a great catch, I think I also saw the expat from system when using OCIO_INSTALL_EXT_PACKAGES set to ALL before so that's a nice fix. I asked the question because I assumed the library path to be already on the linker search path because the dependencies are managed by CMake but I assume from your answer they are not and CMake maybe point directly to the library using -Wl,-L (though I haven't checked).

remia (approved):

doug-walker (approved):

nieder:
Is there a minimum Xcode version needed for the -Wl,-hidden-lFOO' flag? Xcode 11.3 from macOS 10.14.5 fails on this flag.

doug-walker:
@nieder   We tried searching for when the -hidden option was added to Xcode but did not find that information.  Our CI has been running on macOS 10.15, so we know that works at least.  Please log an issue here if this is a blocker, maybe you or someone else will contribute a fix.  As a work-around, you could modify src/OpenColorIO/CMakeLists.txt to omit that option.  The symbols won't be hidden but that may be acceptable in your use case?

nieder:
I've removed the -hidden-lexpat flag and everything seems to still be working.

doug-walker:
Thanks for testing @nieder !

PR #1759, opened 2023-01-13
Adsk Contrib - Allow PyOpenColorIO module to load DLLs from Windows PATH environment variable

cedrik-fuoco-adsk:
This is a re-take of the PR (from @anderslanglands)  with modifications. 
See [Allow loading dlls from PATH on windows and python>=3.8](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1668) for context.

I made the following modifications:

- Minor change was needed to make sure that the setup_ocio.bat and setup_ocio.sh were configured correctly.
- Modified the comments in `__init__.py`.
- Changed it to an opt-out behaviour as they did in OpenImageIO. OCIO_PYTHON_LOAD_DLLS_FROM_PATH must be set to 0 to use the default behaviour of Python 3.8+ (which is to not load the DLLs from the PATH environment variable). 


remia:
I'm not really familiar with the Python 3.8+ Windows specifics but this looks good to me. Only thing is that I think the package layout has now changed to include a directory, so I would be curious to see if the wheel are still working fine with the current `setup.py` script, could you try running the wheel workflow? Maybe we should trigger a wheel workflow anytime something under src/bindings/python is changed from [there](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/.github/workflows/wheel_workflow.yml#L21)? Might be too much but could be tested.

cedrik-fuoco-adsk:
> I'm not really familiar with the Python 3.8+ Windows specifics but this looks good to me. Only thing is that I think the package layout has now changed to include a directory, so I would be curious to see if the wheel are still working fine with the current `setup.py` script, could you try running the wheel workflow? Maybe we should trigger a wheel workflow anytime something under src/bindings/python is changed from [there](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/.github/workflows/wheel_workflow.yml#L21)? Might be too much but could be tested.

I'm not familiar with the python wheel but I've tried to build it on my Macbook M1 and it seems to be built correctly. 

I have tried it on my Windows setup, but it seems to be erroring out somewhere in OCIO docs. Not really where the problem is exactly at the moment.

remia:
Thanks for testing @cedrik-fuoco-adsk, I think you did it correctly and we can monitor the wheel workflow nightly build in any case.

remia (approved):

doug-walker (approved):

Issue #1758, opened 2023-01-13
Option to force 2D texture resource for all 1D luts


num3ric:
Hi,

OCIO's GPU processor will automatically switch the texture resource type from a 2D texture to a 1D texture with sufficiently small 1D LUTs (i.e. going from 4096x2 to 2048x1). (As illustrated by the issue @remia discovered with our integration.) The OpenGL sample does implement this, but I'd unfortunately missed this behavior entirely.

Fixing this remains problematic however. Unreal does not support 1D textures, and for reference neither does Unity nor OpenGL ES / WebGL. (I believe any sampling performance difference is non-existent, making them superfluous?)

As such, having an option to disable their use and always default to 2d textures would be essential. As Remi suggested, something like a `setAllowTexture1D` exposed on the `GpuShaderCreator` or the `GpuShaderDesc` class would work.

(Happy to work on a PR if this feature request gets approved.)

doug-walker:
Thank you for exploring this problem @num3ric, we would certainly welcome your help in addressing it!

PR #1762, opened 2023-01-18
Exposed option to allow/disallow 1D texture GPU resources for 1D LUTs.

num3ric:
First pass at a solution for issue #1758.


remia:
Thanks @num3ric for the fix!

Looking at the changes, I was wondering if following the approach used for OpenGL ES wouldn't be simpler, similar to what was done [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1497/files#diff-568d96ac62004d2d38778e13d3f8223cd510d68fcf912aa9a9fb2358fbf3f63a)? Currently this change add another path for 1D texture creation which make it more complicated to parse the code, do you have a specific reason to do that, like worry about the additional computation happening in the `_computePos` helper maybe?

On the client code side, I think this could be improved but would most likely require an API breaking change. The idea would be to add an enum out parameter to `getTexture` to return the number of dimension, either `TEXTURE_1D` or `TEXTURE_2D`. I'm a bit worried about the logic as it stands: `unsigned type = (height > 1 || !allowTexture1D) ? GL_TEXTURE_2D : GL_TEXTURE_1D;`. It could be easy to miss this check and produce bug in client code, granted that if you explicitly use `setAllowTexture1D()` you should be aware of this but it seems fragile.

remia:
If we make a breaking change, the fix will have to wait for OCIO 2.3.0 probably so that's the tradeoff.

num3ric:
Could the feature be added first (having no impact unless explicitly used), but the (breaking change) API improvement pushed to 2.3.0? Just wondering :)

remia:
I think that would be ok yes!

num3ric:
If I use the same path as `OpenGL ES` (using `|| !shaderCreator->getAllowTexture1D()` in both 2d conditionals) and disallow 1d textures [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/tests/gpu/GPUUnitTest.cpp#L165), a few tests now fail, indicating that the existing 2d path would not be equivalent.

```
[169/212] [Lut1DOp / not_linear_lut1d_5_generic_shader       ] - FAILED -
Maximum error: 1 at pixel: 4 on component 0 larger than epsilon.
scr = {inf, 0, 0, 0}
cpu = {64, 0, 0, 0}
gpu = {0, 0, 0, 0}
relative tolerance=0.0005000000237
[170/212] [Lut1DOp / not_linear_lut1d_5_inverse_generic_shad ] - PASSED - (MaxDiff: 7.7486e-07 at pix[27345][1])
[171/212] [Lut1DOp / lut1d_half_domain_unequal_channels      ] - PASSED - (MaxDiff: 2.38419e-07 at pix[62508][2])
[172/212] [Lut1DOp / lut1d_file2_test                        ] - FAILED -
Maximum error: 1 at pixel: 5 on component 1 larger than epsilon.
scr = {0, inf, 0, 0}
cpu = {0, 1, 0, 0}
gpu = {0, 0, 0, 0}
absolute tolerance=9.999999747e-05
[173/212] [Lut1DOp / lut1d_file2_disallow_tex1D_test         ] - FAILED -
Maximum error: 1 at pixel: 5 on component 1 larger than epsilon.
scr = {0, inf, 0, 0}
cpu = {0, 1, 0, 0}
gpu = {0, 0, 0, 0}
absolute tolerance=9.999999747e-05
[174/212] [Lut1DOp / lut1d_hue_adjust_test                   ] - FAILED -
Maximum error: 1 at pixel: 43691 on component 0 larger than epsilon.
scr = {1.000015259, 1.000026703, 1.000038147, 1.000049591}
cpu = {1, 1, 1, 1.000049591}
gpu = {0, 0, 0, 1.000049591}
absolute tolerance=9.999999747e-06
```

Is the OpenGL ES case unit-tested? Looks like it's actually attempting to sample to the next row, but instead lands on the same row (i.e. `retVal.y` reaches 1.0)?

num3ric:
So I proceeded with the change to use the same path as OpenGL ES. However as mentioned above, this path would cause GPU test failures. Since the testing code doesn't support OpenGL Core Profile / OpenGL ES, it's not something that I could easily verify. I still went ahead and fixed the sampling code here, merging the return value construction and branching on the `height == 1` case.

I am also leaving the interface breaking change to a separate PR.

src/OpenColorIO/ops/lut1d/Lut1DOpGPU.cpp
```diff
@@ -261,29 +262,28 @@ void GetLut1DGPUShaderProgram(GpuShaderCreatorRcPtr & shaderCreator,
...
-                // (retVal.x + 0.5) / height;
-                ss.newLine() << "retVal.y = (retVal.y + 0.5) / " << float(height) << ";";
+            if (height == 1)
+            {
+                ss.newLine() << "retVal.x = (dep + 0.5) / " << float(width) << ";"; // (dep + 0.5) / width;
+                ss.newLine() << "retVal.y = 0.5;";
+            }
+            else
+            {
+                ss.newLine() << "retVal.y = floor(dep / " << float(width - 1) << ");";       // floor( dep / (width-1) ))
```
num3ric, line 282: Both path should be able to `floor`, since it's equivalent to `float(int(..))` with non-negative values.

remia:
Thanks for the update @num3ric. Assuming we aim for a patch release I still have doubts on ABI compatibility actually because we add virtual methods to `GpuShaderCreator`, I think this is not going to work, I will add that for discussion in Monday meeting.

num3ric:
Ah yes indeed, you're correct. I could take a stab at modifying the `getTexture` signatures then.

src/libutils/oglapphelpers/glsl.cpp
```diff
@@ -119,7 +121,7 @@ void AllocateTexture2D(unsigned index, unsigned & texId,
 
     glActiveTexture(GL_TEXTURE0 + index);
 
-    if (height > 1)
+    if (height > 1 || !allowTexture1D)
```
num3ric, line 124: Flipping this conditional around to `height == 1 && allowTexture1D` would make it clearer.

num3ric:
Based on the group conversation last week, I tried adding new functions that would explicitly expose texture object/resource dimensions. I looked for a way to unify the 1d/3d lut interface into a new set of methods, but couldn't find an elegant solution that would also maintain backwards compatibility. So instead I landed with two new functions in the non-3D case (which is why I chose not to add `TEXTURE_3D` to the dimensions enum). This should better preserve backwards compatibility (although the functions are now deprecated, should they be?).

Let me know what you think, thanks!

src/OpenColorIO/ops/lut1d/Lut1DOpGPU.cpp
```diff
@@ -326,7 +333,8 @@ void GetLut1DGPUShaderProgram(GpuShaderCreatorRcPtr & shaderCreator,
 
     if (height > 1 || lutData->isInputHalfDomain()
```
remia, line 334: Could that be `if (dimensions == GpuShaderDesc::TEXTURE_2D)` now like above?

src/bindings/python/PyGpuShaderCreator.cpp
```diff
@@ -67,8 +67,12 @@ void bindPyGpuShaderCreator(py::module & m)
...
         .def("setTextureMaxWidth", &GpuShaderCreator::setTextureMaxWidth, "maxWidth"_a, 
              DOC(GpuShaderCreator, setTextureMaxWidth))
-        .def("getNextResourceIndex", &GpuShaderCreator::getNextResourceIndex, 
-             DOC(GpuShaderCreator, getNextResourceIndex))
+        .def("setAllowTexture1D", &GpuShaderCreator::setAllowTexture1D, "allowed"_a,
+            DOC(GpuShaderCreator, setAllowTexture1D))
+        .def("getAllowTexture1D", &GpuShaderCreator::getAllowTexture1D,
+             DOC(GpuShaderCreator, getAllowTexture1D))
+        .def("getNextResourceIndex", &GpuShaderCreator::getNextResourceIndex,
+            DOC(GpuShaderCreator, getNextResourceIndex))
```
remia, line 76: Should add an enum type for the texture dimensions like we do for the channel type on line 113?

src/bindings/python/PyGpuShaderDesc.cpp
```diff
@@ -146,6 +147,47 @@ void bindPyGpuShaderDesc(py::module & m)
...
+                    const std::string& textureName,
+                    const std::string& samplerName,
+                    unsigned width, unsigned height,
+                    GpuShaderDesc::TextureType channel,
+                    GpuShaderDesc::TextureDimensions dimensions,
+                    Interpolation interpolation,
+                    const py::buffer& values)
+                    {
+                        py::buffer_info info = values.request();
+            py::ssize_t numChannels;
```
remia, line 160: [Minor] Seem to have an indent issue here.

src/bindings/python/PyGpuShaderDesc.cpp
```diff
@@ -285,11 +327,12 @@ void bindPyGpuShaderDesc(py::module & m)
                 const char * samplerName = nullptr;
                 unsigned width, height;
                 GpuShaderDesc::TextureType channel;
+                GpuShaderDesc::TextureDimensions dimensions;
```
remia, line 330: Should also add the read_only field to the `clsTexture` class above?

src/bindings/python/PyGpuShaderDesc.cpp
```diff
@@ -285,11 +327,12 @@ void bindPyGpuShaderDesc(py::module & m)
...
                 Interpolation interpolation;
-                it.m_obj->getTexture(i, textureName, samplerName, width, height, channel, 
+                it.m_obj->getTexture(i, textureName, samplerName, width, height, channel, dimensions,
                                      interpolation);
 
-                return { textureName, samplerName, width, height, channel, interpolation,
+                return { textureName, samplerName, width, height, channel, dimensions, interpolation,
```
remia, line 335: Assuming it's going to break existing Python code but not sure how to do it differently, same for `__next__`.

remia (commented):
Thanks for the update @num3ric, I agree with your approach. I left some comments and would also recommend we update the `pyociodisplay` Python code and the Metal code in `oglapphelpers` to use the new API.

@num3ric to be more specific here are the place using the old API:
* [pyociodisplay](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/apps/pyociodisplay/pyociodisplay.py#L809)
* [Metal oglapphelpers](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/libutils/oglapphelpers/msl.mm#L256)

src/bindings/python/PyGpuShaderDesc.cpp
num3ric: Thanks for all these review notes!

Regarding this one, in what way do you expect existing python code to break? The texture object will certainly be different but any existing code should remain compatible here since this new field is an addition, no? However admittedly... I modified the python bindings without testing it locally so I'll have to get setup for that.

num3ric: Ok, I updated the relevant python test (in the latest commit) and ran it locally. `getTextures` is in use there.

remia: Thanks for the updates @num3ric! Sorry I think my comment was wrongly assuming the Python bindings were returning a Python tuple and we were adding a field in the middle, but we are actually returning a proper aggregate object so my point is not valid!

remia (approved):

include/OpenColorIO/OpenColorIO.h
```diff
@@ -3054,6 +3054,10 @@ class OCIOEXPORT GpuShaderCreator
     virtual void setTextureMaxWidth(unsigned maxWidth) = 0;
     virtual unsigned getTextureMaxWidth() const noexcept = 0;
 
+    /// Allow 1D GPU resource type, otherwise always using 2D resources for 1D LUTs.
```
michdolan, line 3057: Do we still need these functions with the new `dimensions` parameter in the `add/getTexture` interface?

michdolan (approved):
Thanks @num3ric !

include/OpenColorIO/OpenColorIO.h
```diff
@@ -3110,19 +3114,43 @@ class OCIOEXPORT GpuShaderCreator
...
+
+    /**
+     *  Add a 1D or 2D texture
+     *
+     * \note
+     *   The 'values' parameter contains the LUT data which must be used as-is as the dimensions and
+     *   origin are hard-coded in the fragment shader program. So, it means one GPU texture per entry.
+     **/
+    virtual void addTexture(const char* textureName,
+                            const char* samplerNName,
```
num3ric, line 3148: Argument name typo, will fix.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -3110,19 +3114,43 @@ class OCIOEXPORT GpuShaderCreator
...
      **/
-    virtual void addTexture(const char * textureName,
-                            const char * samplerName,
+    OCIO_DEPRECATED("This was marked as deprecated starting in v2.3, please use addTexture() with dimensions.")
+    virtual void addTexture(const char* /*textureName*/,
+                            const char* /*samplerName*/,
+                            unsigned /*width*/, unsigned /*height*/,
+                            TextureType /*channel*/,
+                            Interpolation /*interpolation*/,
+                            const float* /*values*/) {}
```
hodoulp, line 3138: Why having an empty default implementation? 
The implementation could redirect to the new method with `dimensions` equal 2D. By doing so, the implementation remains correct.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -3366,13 +3394,22 @@ class OCIOEXPORT GpuShaderDesc : public GpuShaderCreator
...
     // 1D lut related methods
     virtual unsigned getNumTextures() const noexcept = 0;
+    OCIO_DEPRECATED("This was marked as deprecated starting in v2.3, please use getTexture() with dimensions.")
+    virtual void getTexture(unsigned /*index*/,
+                            const char *& /*textureName*/,
+                            const char *& /*samplerName*/ ,
+                            unsigned & /*width*/,
+                            unsigned & /*height*/,
+                            TextureType & /*channel*/,
+                            Interpolation & /*interpolation*/) const {}
```
hodoulp, line 3404: Why having an empty default implementation? 
The implementation could redirect to the new method and hide the `dimensions` attribute. By doing so, the implementation remains correct.

src/OpenColorIO/GpuShader.cpp
```diff
@@ -209,6 +218,8 @@ class PrivateImpl
         width         = t.m_width;
         height        = t.m_height;
         channel       = t.m_type;
+        assert(t.m_dimensions < 3);
```
hodoulp, line 221: No `assert` in the code. An error report is done using an exception.

src/libutils/oglapphelpers/glsl.cpp
```diff
@@ -119,21 +122,27 @@ void AllocateTexture2D(unsigned index, unsigned & texId,
...
-    {
-        glBindTexture(GL_TEXTURE_1D, texId);
-
-        SetTextureParameters(GL_TEXTURE_1D, interpolation);
+        glTexImage2D(GL_TEXTURE_2D, 0, internalformat, width, height, 0, format, GL_FLOAT, values);    
+        break;
 
-        glTexImage1D(GL_TEXTURE_1D, 0, internalformat, width, 0, format, GL_FLOAT, values);
+    default:
+        assert(false);
```
hodoulp, line 144: No `assert` in the code. The code must use exceptions to report errors.

hodoulp (commented):
Thanks for the pull request. Below are few high level comments.

include/OpenColorIO/OpenColorIO.h
num3ric: Thanks for the notes, will update! Although I suppose I should then remove the overrides in `GenericGpuShaderDesc`.

I would also expect `dimensions` to be `const TextureDimensions dimensions = height > 1 ? TextureDimensions::TEXTURE_2D : TextureDimensions::TEXTURE_1D;` instead of always 2D, since this would be the backward-compatible behavior.

src/OpenColorIO/GpuShader.cpp
```diff
@@ -228,20 +245,20 @@ class PrivateImpl
 
     void add3DTexture(const char * textureName,
                       const char * samplerName,
-                      unsigned dimension,
+                      unsigned edgelen,
                       Interpolation interpolation,
                       const float * values)
     {
-        if(dimension > get3dLutMaxDimension())
+        if(edgelen > get3dLutMaxLength())
```
num3ric, line 252: Renamed these to prevent confusions with the new "dimensions" definition. This is also more consistent with the exposed methods.

include/OpenColorIO/OpenColorIO.h
remia: I was wondering why you didn't use the `getAllowTexture1D()` is those implementations, is that to preserve old behaviour?

num3ric: To mimic the old behavior yes, `height > 1` was previously the implicit criteria to decide which whether the resource was 1D or 2D.

num3ric:
All notes from @hodoulp should be addressed. Let me know if anything else is needed for the merge, thanks!

doug-walker:
@num3ric , I spoke with Patrick about this some more and we realized that there is no point trying to preserve the original API.  In other words, any client app that has overridden GpuShaderCreator is going to need to be updated since OCIO is going to call the new addTexture and it won't exist (in a derived class).

Based on that, we think we should completely remove the original addTexture and getTexture methods.  It will require some client apps (such as Maya) to be updated, but we think the new API is better and there doesn't seem to be a good solution to preserve the original API.  It shouldn't be much work and it should only affect clients that don't use the built-in implementation of GpuShaderCreator.

Apologies for the late reply on this, but do you think you could make this change by Aug 29 to allow us to include it in OCIO 2.3.0 which releases on Aug 31?

num3ric:
- Removed original (previously deprecated) methods, as requested.
- Cleaned up formatting to minimize PR diff... with many force push.
- Added missing call to new texture method in `src/libutils/oglapphelpers/msl.mm`

doug-walker (approved):
Thank you @num3ric !  This is perfect, we will go ahead and merge this for 2.3.0.

PR #1772, opened 2023-02-23
Adsk contrib - Add support for minimum and recommended versions for dependencies

cedrik-fuoco-adsk:
Adding support for minimum, maximum and recommended versions for OCIO dependencies.

- Added two new macros to handle the logic: `ocio_handle_dependency` and `ocio_install_dependency`.
- The download and install part of each OCIO dependency are now split into two files. 
  - Under `share/cmake/modules/FindXYZ.cmake` and `share/cmake/modules/install/InstallXYZ.cmake`
- Added an option OCIO_VERBOSE to print more information when CMake is calling the Find modules.
- Refactored what gets printed when we build as well as some colours when modules are found.
- FindOpenImageIO.cmake was removed in favour of their CMake configuration file (as recommended by OIIO). The FindOpenImageIO.cmake that we had was coming from OIIO originally, but it is not supported anymore.
- FindOpenEXR.cmake was removed in favour of their CMake configuration files. Our FindOpenEXR.cmake was basically just doing a search for their CMake's configuration files, but we don't need a whole Find module to do that.

Here's a snippet of a build log:

image


share/cmake/modules/FindExtPackages.cmake
```diff
@@ -195,14 +190,31 @@ if((OCIO_BUILD_APPS AND OCIO_USE_OIIO_FOR_APPS) OR OCIO_BUILD_TESTS)
...
-    else()
-        find_package(OpenImageIO ${OIIO_VERSION})
-    endif()
+    set(OIIO_RECOMMENDED_VERSION "2.4")
+
+    # Supported from OIIO 2.4+. Setting this for lower versions doesn't affect anything.
+    set(OPENIMAGEIO_CONFIG_DO_NOT_FIND_IMATH 1)
+
+    ###############################################################################
+    # OpenEXR (https://github.com/AcademySoftwareFoundation/openexr)
```
lgritz, line 199: OpenImageIO?

doug-walker: Yes, sorry, copy/paste error.  We will fix.  Thanks for the review!

cedrik-fuoco-adsk:
I am working on the failure in the Linux CI. The issue comes up when OpenImageIO is found via OpenImageIOConfig.cmake.
I've identified the issue, but I'm working on a solution. The issue is present in the main branch as well.

share/cmake/modules/FindExtPackages.cmake
```diff
@@ -177,15 +165,22 @@ if(OCIO_BUILD_PYTHON OR OCIO_BUILD_DOCS)
...
 
     if(OCIO_BUILD_PYTHON)
         # pybind11
         # https://github.com/pybind/pybind11
         # pybind11 2.9 fixes issues with MS Visual Studio 2022 (Debug).
-        find_package(pybind11 2.9.2 REQUIRED)
+        ocio_handle_dependency(  pybind11 REQUIRED ALLOW_INSTALL
+                                 MIN_VERSION 2.9.2
+                                 RECOMMENDED_VERSION 2.9.2
+                                 RECOMMENDED_VERSION_REASON "Pybind 2.10.0+ does not work with Python 2.7 anymore")
```
remia, line 183: Note #1770 removes official support for Python 2.7, I'm not sure we have urgent needs to update the `pybind11` recommended version but thought I would just mention it.

CMakeLists.txt
```diff
@@ -143,6 +144,9 @@ if (WIN32)
     option(OCIO_USE_WINDOWS_UNICODE "Compile with Windows Unicode support" ON)
 endif()
 
+###############################################################################
+# Other preferences
+option(OCIO_VERBOSE "Display more information when searching or installing dependencies" OFF)
 
 ###############################################################################
 # Optimization / internal linking preferences
```
remia, line 152: Should we get read of `OCIO_USE_OIIO_CMAKE_CONFIG` now?

remia (commented):
Amazing work @cedrik-fuoco-adsk!

Would be great to finally get read of OIIO custom Find module, I think this was already tried a couple of times in the past. I would think it may be ok to bump the minimum version required if that helps in the process (for OCIO 2.3 obviously).

Thinking in terms of different compilation path now that we support more flexibility in version dependencies (tagging @Shootfast), is there something we could add to check what were the different dependencies versions from the compiled library? Not sure it's needed here for this PR but might be useful for debugging down the line. (Don't have a strong opinion about this, just raising the point. We already allowed this for Imath and OpenEXR before)

cedrik-fuoco-adsk:
Sorry, I messed up my branch while trying to rebase. I might have to do a redo.

cedrik-fuoco-adsk:
Closing in favour of https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1777 (redo)

Issue #1753, opened 2023-01-06
Add support for ARM Neon intrinsics


doug-walker:
Currently, several of the OCIO ops have a code path that uses Intel SSE intrinsics (SIMD instructions) for faster performance.  However, on Mac M1/ARM chips, these don't work and so the slower straight C++ path is used.  This task is to take advantage of the equivalent intrinsics available on ARM processors, known as Neon.

There are a number of open source projects available to help convert SSE instructions into Neon instructions.  The one we propose using is:
https://github.com/DLTcollab/sse2neon

Note that there are some precision differences and some differences in NaN handling that will need to be investigated.

doug-walker:
The work for this was done in PR #1775.

PR #1775, opened 2023-03-07
Adsk Contrib - Add support for ARM Neon intrinsics and Universal builds

cedrik-fuoco-adsk:
This PR is about [Add support for ARM Neon intrinsics #1753](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1753).

- Integrating sse2neon library to support ARM Neon intrinsic.
- Adding a new option to enable SIMD (DOCIO_USE_SIMD). OCIO_USE_SSE still works for now.
- Fixing issues that were preventing a universal build on MacOS.
- Universal build is now the default on MacOS.

Note that at least another commit will be needed once [Adsk contrib - Add support for minimum and recommended versions for dependencies](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1772) is merged into the main branch. 

Some performance results with a MacBook M1 (using a custom heavy_transform.clf):
image


share/cmake/modules/FindExtPackages.cmake
```diff
@@ -197,13 +197,10 @@ if((OCIO_BUILD_APPS AND OCIO_USE_OIIO_FOR_APPS) OR OCIO_BUILD_TESTS)
...
-    # Since OpenImageIO will try to find OpenEXR through its OpenImageIOConfig.cmake file,
-    # let's try to find OpenEXR first and if the version is too old, OCIO will not try to find
-    # OpenImageIO.
-    ocio_check_dependency_version(  OpenEXR "is_OpenEXR_VERSION_valid" 
-                                    MIN_VERSION ${OpenEXR_MININUM_VERSION} 
-                                    CONFIG)
+    set(is_OpenEXR_VERSION_valid FALSE)
+    # Check for compatibility between OpenEXR and OpenImageIO.
+    # Will set is_OpenEXR_VERSION_valid to TRUE if valid.
+    include(CheckForOpenEXRCompatibility)
```
cedrik-fuoco-adsk, line 203: Changed how we check the version of the OpenEXR used by OpenImageIO. This should be more robust and won't create any unwanted target or variables.

share/cmake/macros/ocio_check_dependency_version.cmake
```diff
@@ -1,38 +0,0 @@
-# SPDX-License-Identifier: BSD-3-Clause
```
cedrik-fuoco-adsk, line 1: removed in favour of CheckForOpenEXRCompatibility.cmake

cedrik-fuoco-adsk:
The branch is now rebased and conflicts have been resolved. 
The PR is ready to be reviewed. Thanks in advance! @remia @doug-walker

share/cmake/utils/CompilerFlags.cmake
```diff
@@ -40,7 +55,9 @@ elseif(USE_CLANG)
 
     # Use of 'register' specifier must be removed for C++17 support.
     set(PLATFORM_COMPILE_OPTIONS "${PLATFORM_COMPILE_OPTIONS};-Wno-deprecated-register")
-
+    if (HAVE_SSE2 AND HAVE_NEON)
+        set(PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS};-march=armv8-a+fp+simd+crypto+crc")
```
remia, line 59: Should that be `PLATFORM_COMPILE_OPTIONS` now?

src/OpenColorIO/SSE.h
```diff
@@ -125,19 +164,25 @@ static const __m128 PNEXP2 = _mm_set1_ps((float)2.414427569091865207710e-1);
 static const __m128 PNEXP1 = _mm_set1_ps((float)6.930038344665415134202e-1);
 static const __m128 PNEXP0 = _mm_set1_ps((float)1.000002593370603213644);
 
+// Note: The above polynomials have been chosen to acheive a precision of
```
remia, line 167: [Minor] Typo achieve.

src/OpenColorIO/SSE.h
```diff
@@ -125,19 +164,25 @@ static const __m128 PNEXP2 = _mm_set1_ps((float)2.414427569091865207710e-1);
...
 // log2 function in SSE version 2
 //
 // The function log2() is evaluated by performing argument
 // reduction and then using Chebyshev polynomials to evaluate the function 
 // over a restricted range.
 inline __m128 sseLog2(__m128 x)
 {
-    // y = log2( x ) = log2( 2^exposant * mantissa ) 
-    //               = exposant + log2( mantissa )
+    // y = log2( x ) = log2( 2^exponant * mantissa )
```
remia, line 178: [Minor] Typo exponent.

remia (commented):
Looks good to me @cedrik-fuoco-adsk, thank you for adding more comments to the SSE header. Have to say Rosetta is surprisingly good at handling the SSE x86_64 binary. I guess if we want to further optimise we would have to start more in depth profiling but the improvements we get seem already quite large as is.

share/cmake/utils/CompilerFlags.cmake
cedrik-fuoco-adsk: I read somewhere that those were the default on Apple M1, but I can't find it anymore.
I'm trying to add those back, but I ran into an issue with the universal build since the build for x86_64 doesn't accept those options.

Since there is only one "configure" for the universal build, I'm trying to find a way to set those compiler options based on the architecture at **build time**, but I haven't found a working solution so far.

Do you have any idea @remia? I thought that using generator expression would work, but it does not seems like it does.

remia: After some quick search I found [this](https://stackoverflow.com/a/67492623/4814046) but that seem tied to XCode, need to look further. Like you say, may should make sure these flags are needed, maybe bumping the compiler verbose levels or finding the command to show the default flags would help confirming. Also might be worth assessing again why we enable universal binary by default.

cedrik-fuoco-adsk: I've added a comment in my latest commit about this. These flags are needed for ARM platforms that do not enable floating point calculation and/or SIMD instructions by default. On Apple's ARM platform, these features are enabled by default. If they were not enabled by default, we would get a bunch of errors during build about SIMD instructions. (can be tested by adding "nosimd" or "nofp" compiler flags).

remia (approved):

ConnorBaker:
Hi all,

I'd love to see these changes merged -- any remaining blockers?

Thank you for your work on this, @cedrik-fuoco-adsk!

michdolan (dismissed):

michdolan:
Closing since #1828 replaces this PR

Issue #1771, opened 2023-02-20
zlib dependency requirement for 2.2.1


remia:
Starting from version 2.2.1 OpenColorIO will no longer compile with zlib version < 1.2.13. I understand this was done following a CVE report on older versions of the library. I think we should accept earlier versions as well and keep 1.2.13 as the default version only when zlib is built in source, eg. with `OCIO_INSTALL_EXT_PACKAGES=ALL`.

Any opinion or things I may have missed?

KevinJW:
This does follow what I thought we discussed in the TSC meetings:

If we are compiling without user supplied dependencies we use the recommended one, but if the user supplies there own *and* it is compatible then we should allow it to compile.

This was so that users on a particular VFX platform can update OCIO without having to update everything else.

remia:
Thanks for the feedback Doug and Kevin, glad to know it's taking care of!

doug-walker:
We've now merged #1777, which allows us to have both a min version and recommended version.  Our min zlib version is now at 1.2.10.  We could probably lower this but it would require some additional testing work.  If anyone wants us to lower the min version further for zlib, please let us know what version you'd like and we will re-open the issue.

cedrik-fuoco-adsk:
I've tested zlib 1.2.7 on multiple platforms, and there is a warning (on MacOS, appleClang) in one of zlib's files when OCIO tries to build zlib itself. Since OCIO treats warning as errors, the build fails.

I don't have those issues with zlib 1.2.8.

cedrik-fuoco-adsk:
I've tested zlib 1.2.7 on the different platforms, and there is an issue with their main CMakeLists.txt [(line 169)](https://github.com/madler/zlib/blob/30a1c7065dc1dc2c2ed68ed403792b660bfdd805/CMakeLists.txt#LL169C31-L169C41) for MacOS. The `--version-script` linker option does not exist for MacOS's version of ld. 

Zlib 1.2.7 probably works fine on MacOS, but OCIO can not built it without a patch.

It was fixed for the release of zlib 1.2.8 (see line [206](https://github.com/madler/zlib/blob/50893291621658f355bc5b4d450a8d06a563053d/CMakeLists.txt#LL206C38-L206C38)).

I would suggest to use 1.2.8 as I didn't see any errors with that version. If we really need to support CentOS 7 and the zlib version that comes with it, we will need to do a manually patch when OCIO build zlib itself.

remia:
[...]

PR #1777, opened 2023-03-09
Adsk contrib - Add support for minimum and recommended versions for dependencies (#2)

cedrik-fuoco-adsk:
Something went wrong in the previous [PR](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1772) while trying to rebase the branch.

I created a fresh branch from main and merged my changes, which went much smoother.

--------------------------------

Adding support for minimum, maximum and recommended versions for OCIO dependencies.

- Added two new macros to handle the logic: `ocio_handle_dependency` and `ocio_install_dependency`.
- The download and install part of each OCIO dependency are now split into two files. 
  - Under `share/cmake/modules/FindXYZ.cmake` and `share/cmake/modules/install/InstallXYZ.cmake`
- Added an option OCIO_VERBOSE to print more information when CMake is calling the Find modules.
- Refactored what gets printed when we build as well as some colours when modules are found.
- FindOpenImageIO.cmake was removed in favour of their CMake configuration file (as recommended by OIIO). The FindOpenImageIO.cmake that we had was coming from OIIO originally, but it is not supported anymore.
- FindOpenEXR.cmake was removed in favour of their CMake configuration files. Our FindOpenEXR.cmake was basically just doing a search for their CMake's configuration files, but we don't need a whole Find module to do that.

Here's a snippet of a build log:

image


doug-walker:
Regarding our usual 2-week clock on these Autodesk PRs, this PR is identical to PR #1772 which had already been open for two weeks.  So it is our intention to merge this one as soon as it is approved.  Other work is gated by this, thanks.

doug-walker:
@remia , we think we've addressed all of your comments from PR #1772, could you review this one again please?  Thanks in advance!

share/cmake/modules/FindExtPackages.cmake
```diff
@@ -32,130 +35,115 @@ if (APPLE)
...
-    find_package(ZLIB ${_ZLIB_FIND_VERSION} ${_ZLIB_REQUIRED})
-
-    # Restore the original find library ordering
-    if(DEFINED _ZLIB_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES)
-        set(CMAKE_FIND_LIBRARY_SUFFIXES "${_ZLIB_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES}")
-    else()
-        set(CMAKE_FIND_LIBRARY_SUFFIXES)
-    endif()
+ocio_handle_dependency(  ZLIB REQUIRED ALLOW_INSTALL
+                         MIN_VERSION 1.2.10
```
remia, line 104: [Question] Is the 1.2.10 version the minimum compatible, how did we picked it?

remia (approved):
Looks good to me, thanks for taking care of the rebase @cedrik-fuoco-adsk. Just a question about `zlib`, but that should not prevent merge and can be discussed in #1771. Great work!

share/cmake/modules/FindExtPackages.cmake
doug-walker: We basically just picked what seemed like a sufficiently old version, and did not invest the time to see what is actually the oldest that would work.  If someone has a need for an older version, please post the requirement in issue #1771.

I will go ahead and merge this now since it is blocking another task.  Thank you Remi!

doug-walker (approved):

PR #1783, opened 2023-03-31
Adsk Contrib - Dependencies latest workflow fix

cedrik-fuoco-adsk:
This PR is to address the issue with the build failure (related to documentation) in one of the CI workflows. 

In the failing environment, it is using docutils 0.18, but if you update that to 0.18.1, the build is successful (and does not seems to break the others).


remia (approved):
Great find @cedrik-fuoco-adsk !

doug-walker (approved):

PR #1788, opened 2023-04-14
Adsk Contrib - Enhanced methods to find an equivalent color space between configs

cedrik-fuoco-adsk:
Adds several new methods to be used to identify a common color space between a pair of configs that may be used as an interchange space. If the official interchange roles are not present, heuristics are used to search for a suitable color space.

`IdentifyBuiltinColorSpace` -- Searches a user config to find the name of an equivalent color space to a color space in one of the Built-in configs. For example, "What is the name in this config for the `sRGB - Texture` color space from the default Built-in config?"

`IdentifyInterchangeSpace` -- Identifies the pair of color space names to be used as the interchange space to allow conversion between color spaces in a user config and one of the built-in configs. This is the fundamental function that contains the heuristics for trying to identify known color spaces and is used by IdentifyBuiltinColorSpace and in the GetProcessorFrom/ToBuiltinColorSpace methods added in PR #1710.

This improves upon and extends upon PR #1710. One piece of feedback on that PR is that it would be helpful to be able to use the heuristics to extract the names of the interchange color spaces rather than a Processor. This gives the application more flexibility in how that information is persisted and may reduce the number of times that any heuristics (that search through a config's color spaces) need to be run.

In addition, following up on the TODO mentioned in PR #1710, we refactored most of the heuristics code to move it out of the already crowded `Config.cpp` and into the new `ConfigUtils.cpp` module. The heuristics themselves were not significantly changed.

A lot of additional unit tests were added to test more scenarios involving the heuristics.

Finally, an issue was fixed in `GetProcessorFromConfigs `so that the returned Processor is a no-op if either of the color spaces in the two configs is a data space.


src/OpenColorIO/Config.cpp
```diff
@@ -1094,414 +1100,6 @@ class Config::Impl
         // That should never happen.
         return -1;
     }
-
```
doug-walker, line 1097: Lines 1097-1504 were moved to ConfigUtils.cpp, without significant changes.  This is the heuristics code and we wanted to move this out of Config.cpp, since this module is already so long and the heuristics are essentially their own somewhat self-contained set of functions.

src/OpenColorIO/Config.cpp
```diff
@@ -5064,11 +4727,112 @@ void Config::serialize(std::ostream& os) const
...
+{
+    return getImpl()->getProcessorCacheFlags();
+}
+
+void Config::setProcessorCacheFlags(ProcessorCacheFlags flags) const noexcept
 {
     getImpl()->setProcessorCacheFlags(flags);
 }
 
+//////////////////////////////////////////////////////////////////
```
doug-walker, line 4740: Lines 4740-4835 is simply a cut/paste move from farther down in this module.  This code is from another one of our recent PRs and we decided it belongs better in this location.

remia:
Looks good to me, do we need Python bindings for the new methods?

tests/python/ColorSpaceTest.py
```diff
@@ -754,4 +790,112 @@ def check_processor_inv(self, p):
...
+
+        csname = cfg.IdentifyBuiltinColorSpace(editableCfg, builtinConfig, "lin_ap1")
+        self.assertEqual(csname, "ACES cg")
+
+        # Display-referred spaces are not supported unless the display-referred interchange
+        # role is present.
+
+        with self.assertRaises(OCIO.Exception) as cm:
+            cfg.IdentifyBuiltinColorSpace(editableCfg, builtinConfig, "sRGB - Display")
+        self.assertEqual(
```
remia, line 830: Should this be indented under the `with` (same apply below)?

remia (approved):

tests/python/ColorSpaceTest.py
cedrik-fuoco-adsk: The current code is correct. The goal of the `with` statement is to capture the exception, set the `cm ` variable to the exception message. Once the exception is thrown, then do a check on that message. If we were to put that under the `with` statement, it would not be executed because `cfg.IdentifyBuiltinColorSpace` would throw first.

cedrik-fuoco-adsk:
The failed workflow is expected as there was an issue with that workflow on the main branch. We merged the fix into the main branch yesterday.

tests/python/ColorSpaceTest.py
remia: You are perfectly right, sorry about this confusion!

doug-walker:
Taking note here that I verified that this PR fixes a bug in GetProcessorFromBuiltinColorSpace reported by a user of OCIO 2.2.1.  As a result, we will plan to include this branch in the upcoming 2.2.2 release rather than waiting for 2.3.

zachlewis (approved):

doug-walker (approved):

doug-walker:
I'm going to go ahead and merge this since we need it for the config merge feature we're working on.  The CI problem on macOS with docs=ON is unrelated to the code in this PR.

PR #1791, opened 2023-04-21
Adsk contrib - Add method to clear processor cache

cedrik-fuoco-adsk:
This PR adds to the public API a method to clear the Processor cache of a Config instance.  This is useful in situations where LUT files on disk have changed after a Processor was created.


remia (approved):
Thanks for the fix @cedrik-fuoco-adsk, I also encountered this few months back but had forgotten about it, my temporary workaround was to clone the config using `createEditableCopy()` which also clears the processor cache.

doug-walker (approved):

PR #1793, opened 2023-04-21
Adsk Contrib - Changing ZLIB min version to 1.2.8

cedrik-fuoco-adsk:
This PR changes the minimum version of ZLIB to 1.2.8.  This is a fix for issue #1771, from Remi.


remia (approved):
Thanks @cedrik-fuoco-adsk.

doug-walker (approved):

Issue #1763, opened 2023-01-21
GPU 3D lut transform causes banding


cessen:
I'm using 3d luts that span a large dynamic range and need trilinear interpolation for correct results (I'm happy to provide details of my use case if that's helpful).

Due to the nature of the luts, they should provide sufficiently accurate results even at relatively low lut resolutions.  However, much to my surprise there is *significant* banding when applied by OCIO on GPU:

![gpu posterized](https://perm.cessen.com/2023/github/ocio_3d_lut_bug/gpu_posterized.jpg)

The only way to mitigate this is to crank the lut resolution extremely high, and even then especially dark colors exhibit some subtle banding.  The results by CPU, on the other hand, are correct even with the low resolution lut:

![cpu correct](https://perm.cessen.com/2023/github/ocio_3d_lut_bug/cpu_correct.jpg)

Looking into OCIO's source, this seems to be the culprit:

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/b00877959f4497f9b8e5a08859d0ef8059fdd8e5/src/OpenColorIO/ops/lut3d/Lut3DOpGPU.cpp#L221-L236

It's even noted in the comment that this can be an issue on some GPUs.  For the record, I'm using an nVidia RTX 3060.  So not exactly niche.

I think it would make sense to replace the `sampleTex3D()` call with hand-written trilinear interpolation code, to ensure that it executes with full floating point precision and gives correct results on all platforms.  It might be a small perf hit, but for something like OCIO I suspect correctness and consistency is more important here.

Would a PR along those lines be accepted?

cessen:
Any thoughts/feedback about this?  This is a significant problem for the work that I'm doing with OCIO, and I'd like to see it fixed.

I'm happy to do the leg work, but just want to double-check that the proposed solution is acceptable first.

doug-walker:
Hi @cessen , apologies for the delayed reply.  Yes, that's a pretty shocking amount of banding in the first image.  It looks like maybe you're applying a 3d-LUT to a linear color space encoding?  If so, the banding is probably only one of several problems you will encounter.  The best practice is to convert to a log-encoded (or at least gamma-encoded) color space before applying a 3d-LUT.

In practice, OCIO users tend to be using 3d-LUTs that are a minimum of 16x16x16 (and typically larger) and being fed with log-encoded data.  So the quantization of the fractional component has not been an issue.  

[...]

PR #1794, opened 2023-04-26
WIP: add high-precision GPU trilinear interpolation for 3D LUTs

cessen:
This new code path can be enabled by disabling the new default-enabled OPTIMIZATION_NATIVE_GPU_TRILINEAR optimization flag.

The existing code path used the GPU's native trilinear texture interpolation function, which (although faster) quantized the lookup coordinates and could cause color banding.  That's still the default, but now full-precision trilinear interpolation can optionally be used instead.

## WIP

This is PR is WIP.  It functions, but it's not ready to merge yet:

- It wasn't clear how best to get the optimization flags to the shader generation code in `GetLut3DGPUShaderProgram()`.  I tried several approaches, but every approach ended up affecting *some* public API.
- The approach I finally landed on was to store the optimization flags in `GPUProcessor::Impl`, and then pass them to `Op::extractGpuShaderInfo()` via a new argument with a default value.  Of all the approaches I tried, this seemed to have the lowest impact on both the code and the API.
- Unfortunately, this still breaks ABI compatibility (I think--I'm not totally sure how default arguments work at the ABI level in C++).  Additionally, default arguments are apparently [forbidden](https://opencolorio.readthedocs.io/en/latest/guides/contributing/contributing.html#misc-rules) in the OCIO code base, so that's probably not the way forward, regardless.

I am *very much* open to suggestions for a better approach to get the optimization flags to `GetLut3DGPUShaderProgram()`.

## Performance

Some initial naive performance tests indicate that the high precision code is notably slower than than GPU-native trilinear interpolation, but about on par with OCIO's tetrahedral interpolation.  More testing is needed, however.  For example, using higher-res LUTs and testing on a variety of GPUs.  I'll update with actual data once I've had a better go at this.

### DIFF
--- a/include/OpenColorIO/OpenColorTypes.h
+++ b/include/OpenColorIO/OpenColorTypes.h
@@ -619,6 +619,13 @@ enum OptimizationFlags : unsigned long
      */
     OPTIMIZATION_NO_DYNAMIC_PROPERTIES           = 0x10000000,
 
+    /**
+     * For GPU processor, use native trilinear interpolation for 3D LUTs.  This is faster,
+     * but on many GPUs also lower precision.  With low-resolution LUTs, LUTs with large
+     * extents, or LUTs applied in a linear color space this can sometimes cause color banding.
+     */
+    OPTIMIZATION_NATIVE_GPU_TRILINEAR            = 0x20000000,
+
     /// Apply all possible optimizations.
     OPTIMIZATION_ALL                             = 0xFFFFFFFF,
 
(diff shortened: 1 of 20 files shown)


doug-walker:
We discussed this PR during the TSC meeting today. Thank you for the proposal, but the consensus is that this feature is not needed at this time. There are two workarounds to getting more precision:
- Increase the size of the LUT3D to 7x7x7 or higher.
- Set the interpolation method to tetrahedral (which is generally higher quality than trilinear).

cessen:
Thank you for discussing and considering this PR!

As I mentioned in [a later comment in the issue that prompted this PR](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1763#issuecomment-1757709348), I also agree that the *approach* in this PR is not the right one.  However, I hope it's clear that the issue itself still needs addressing, as neither of the two work-arounds you mention work for the situation that prompted filing the issue in the first place.

So I hope *a* solution is not considered off-the-table entirely.

PR #1795, opened 2023-04-28
Adsk contrib - Fix for dependencies workflow

cedrik-fuoco-adsk:
This is a fix for the failure in the dependencies workflow on macOS where OSL wasn't able to find the LLVM directory.

### DIFF
--- a/.github/workflows/dependencies_latest.yml
+++ b/.github/workflows/dependencies_latest.yml
@@ -219,6 +219,9 @@ jobs:
           share/ci/scripts/multi/install_oiio.sh latest $EXT_PATH
           share/ci/scripts/multi/install_osl.sh latest $EXT_PATH
           share/ci/scripts/multi/install_openfx.sh latest $EXT_PATH
+      - name: Print some variables
+        run: |
+          brew --prefix llvm@15
       - name: Create build directories
         run: |
           mkdir _install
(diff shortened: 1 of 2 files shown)


michdolan (approved):

remia (approved):
@cedrik-fuoco-adsk Just noting, in case that helps, you can run the workflow as part of the PR by introducing a dummy change to the GHA workflow file because we have a trigger [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/.github/workflows/dependencies_latest.yml#L19). Also, haven't tested but there is now a vscode extension for GHA [here](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions).

cedrik-fuoco-adsk:
> @cedrik-fuoco-adsk Just noting, in case that helps, you can run the workflow as part of the PR by introducing a dummy change to the GHA workflow file because we have a trigger [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/.github/workflows/dependencies_latest.yml#L19). Also, haven't tested but there is now a vscode extension for GHA [here](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions).

Thanks @remia for the information. I did the dummy change in one of the previous commits. The workflow was running fine, but now it is still not working once merged. It seems like my changes are not taken into account.

Issue #1785, opened 2023-04-03
Improve handling of Windows path change in Python 3.8+


doug-walker:
In PR #1759 we made a fix to try and solve problems caused by the change in library path behavior of Python >= 3.8 on Windows.  Following the solution used in OIIO, we basically defaulted to adopting the previous Python behavior.  However, based on feedback in PR #1668, this should be improved.

There is not agreement that the solution is simply to make the existing env var default to opt-in rather than opt-out since the people most likely to be affected by the problem are users with limited development expertise.

Based on TSC discussion, we will investigate building the Python binding statically, similar to the wheels.  At that point we would remove the OCIO_PYTHON_LOAD_DLLS_FROM_PATH env var option.  Users would need to manage the paths on their own.

Another solution would be to add only the path(s) where the OCIO required libraries are, although initial discussion suggested this might be more difficult to get right.

We're open to feedback on the best approach.  The recommended solution might be something that should be coordinated among ASWF projects, as it's likely that many of them are experiencing this issue.

JeanChristopheMorinPerso:
Should https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1759 be reverted in the meantime?

JeanChristopheMorinPerso:
It might be worth asking for feedback in ASWF's `wg-python3` Slack channel. The working group is archived, but the channel is still used from time to time to ask Python related questions.

Issue #1744, opened 2023-01-03
Building PyPI package from source requires Doxygen


remia:
When building the Python package from source, either with `pip --no-binary` or on architectures where no prebuilt wheel are available, the build process try to generate the documentation and will fail if Doxygen is not found on the system. Ideally this should be detected during the configuration of the build and documentation generated only if the required dependencies are available.

Probably have to update this hard-coded variable https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/setup.py#L80 and make sure build without the doc works.

Reported by @KelSolaar https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1573#issuecomment-1325526472

PR #1802, opened 2023-05-30
Add command line apps to Python wheel

remia:
This PR adds command line applications (C++ and Python) to the Python wheels, some points to note:
* The wheel size is growing significantly with this change, I took some actions to reduce it like striping binaries and removing the test suite but it is still around [20MB per wheel](https://test.pypi.org/project/opencolorio/2.3.0.dev4/#files). We might gain some more by using `-Os` and/or LTO to compile the apps but the most efficient way would most likely be to compile OCIO dynamically and have Python package and apps link to it. I haven't tried that and expect it will bring complexity to the build process.
* The PR also contains some C++ warnings fix,
* Documentation build fix extracted to #1809 

Note that I have no prior experience adding scripts to Python packages, I'm open to any suggestion.

Wheels are available on the test PyPI, install with this command (at your own risks):
`pip install -i https://test.pypi.org/simple/ opencolorio==2.3.0.dev4`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ class CMakeExtension(Extension):
 class CMakeBuild(build_ext):
     def build_extension(self, ext):
         extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
+        bindir = os.path.join(extdir, "bin")
 
         # required for auto-detection & inclusion of auxiliary "native" libs
         if not extdir.endswith(os.path.sep):
@@ -73,12 +74,13 @@ class CMakeBuild(build_ext):
 
         cmake_args = [
             "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={}".format(extdir),
+            "-DCMAKE_RUNTIME_OUTPUT_DIRECTORY={}".format(bindir),
             "-DPython_EXECUTABLE={}".format(sys.executable),
             # Not used on MSVC, but no harm
             "-DCMAKE_BUILD_TYPE={}".format(cfg),
             "-DBUILD_SHARED_LIBS=OFF",
             "-DOCIO_BUILD_DOCS=ON",
-            "-DOCIO_BUILD_APPS=OFF",
+            "-DOCIO_BUILD_APPS=ON",
             "-DOCIO_BUILD_TESTS=OFF",
             "-DOCIO_BUILD_GPU_TESTS=OFF",
             # Make sure we build everything for the requested architecture(s)
@@ -121,7 +123,8 @@ class CMakeBuild(build_ext):
             # Multi-config generators have a different way to specify configs
             if not single_config:
                 cmake_args += [
-                    "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{}={}".format(cfg.upper(), extdir)
+                    "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{}={}".format(cfg.upper(), extdir),
+                    "-DCMAKE_RUNTIME_OUTPUT_DIRECTORY_{}={}".format(cfg.upper(), bindir),
                 ]
                 build_args += ["--config", cfg]
 
@@ -156,11 +159,33 @@ setup(
     version=get_version(),
     package_dir={
         'PyOpenColorIO': 'src/bindings/python/package',
-        'PyOpenColorIO.tests': 'tests/python',
-        'PyOpenColorIO.data': 'tests/data',
+        'PyOpenColorIO.bin.pyocioamf': 'src/apps/pyocioamf',
+        'PyOpenColorIO.bin.pyociodisplay': 'src/apps/pyociodisplay',
     },
-    packages=['PyOpenColorIO', 'PyOpenColorIO.tests', 'PyOpenColorIO.data'],
+    packages=[
+        'PyOpenColorIO',
+        'PyOpenColorIO.bin.pyocioamf',
+        'PyOpenColorIO.bin.pyociodisplay',
+    ],
     ext_modules=[CMakeExtension("PyOpenColorIO.PyOpenColorIO")],
     cmdclass={"build_ext": CMakeBuild},
-    include_package_data=True
+    include_package_data=True,
+    entry_points={
+        'console_scripts': [
+            # Native applications
+            'ocioarchive=PyOpenColorIO.command_line:main',
+            'ociobakelut=PyOpenColorIO.command_line:main',
+            'ociocheck=PyOpenColorIO.command_line:main',
+            'ociochecklut=PyOpenColorIO.command_line:main',
+            'ocioconvert=PyOpenColorIO.command_line:main',
+            'ociodisplay=PyOpenColorIO.command_line:main',
+            'ociolutimage=PyOpenColorIO.command_line:main',
+            'ociomakeclf=PyOpenColorIO.command_line:main',
+            'ocioperf=PyOpenColorIO.command_line:main',
+            'ociowrite=PyOpenColorIO.command_line:main',
+            # Python applications
+            'pyocioamf=PyOpenColorIO.bin.pyocioamf.pyocioamf:main',
+            'pyociodisplay=PyOpenColorIO.bin.pyociodisplay.pyociodisplay:main',
+        ]
+    },
 )
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,6 @@
 
 import os
 import re
-import shutil
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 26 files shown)


include/OpenColorIO/OpenColorTypes.h
```diff
@@ -776,10 +775,7 @@ extern OCIOEXPORT const char * OCIO_OPTIMIZATION_FLAGS_ENVVAR;
  */
 extern OCIOEXPORT const char * OCIO_USER_CATEGORIES_ENVVAR;
 
-/** @}*/
-
-/** \defgroup VarsRoles
- *  @{
+/** VarsRoles
```
doug-walker, line 778: Minor:  I suspect "VarsRoles" was a name for the group, but it doesn't make sense as a description of the code.  Same for the next few groups.  These sections already have an rst header, so it might be better to just remove all three doxygen group lines.

doug-walker (approved):

remia:
Tagging @JeanChristopheMorinPerso in case you have time to check these changes, your feedback would be appreciated!

JeanChristopheMorinPerso:
Thanks for the ping @remia! I'll take a look as soon as I have time. I'm guessing you'll want to merge this before the next release? If so, when is the next release planned (so that I can prioritize on my side).

src/bindings/python/package/command_line.py
```diff
@@ -0,0 +1,18 @@
...
+import os
+import subprocess
+import sys
+
+
+BIN_DIR = os.path.join(os.path.dirname(__file__), 'bin')
+
+
+def call_program(name, args):
+    return subprocess.call([os.path.join(BIN_DIR, name)] + args, close_fds=False)
```
JeanChristopheMorinPerso, line 13: `close_fds` makes me a little bit nervous.

JeanChristopheMorinPerso (commented):
This looks great @remia ! It's exactly what I would have done.

>The wheel size is growing significantly with this change, I took some actions to reduce it like striping binaries and removing the test suite but it is still around [20MB per wheel](https://test.pypi.org/project/opencolorio/2.3.0.dev4/#files). We might gain some more by using -Os and/or LTO to compile the apps but the most efficient way would most likely be to compile OCIO dynamically and have Python package and apps link to it. I haven't tried that and expect it will bring complexity to the build process.

What I did in the past was to compile dynamically, but linking the dependencies statically. So you end up with a single library that you can now link to the executables.

```mermaid
graph TD
    A[dep1.a] -->C(libocio.so)
    B[dep2.a] -->C(libocio.so)
    C --> D(app1)
    C --> E(app2)
    C --> F(OpenColorIO.cpython310.so python lib)
```
On Linux and macOS, it will work without any additional work. On WIndows, You'll need to use `os.add_dll_directory`.

src/bindings/python/package/command_line.py
remia: If memory serve, this was directly copied from [ninja python wheel](https://github.com/scikit-build/ninja-python-distributions/blob/master/src/ninja/__init__.py#L51), I'm not sure why they used this flag exactly so I removed in for now, thanks.

remia:
Thanks for the review @JeanChristopheMorinPerso! I'll do some tests around the dynamic linking, I think I had some issues around the wheel repair steps in the past, will keep updating this PR to address the issues coming up.

JeanChristopheMorinPerso:
@remia make sure that the dependencies are beside the OCIO shared library (which is why I suggested to statically link the dependencies to the OCIO shared library, since it's easier to bundle that way).

remia:
The issue was just that on Windows the DLL for OCIO get into the bin folder on the wheel layout, so I updated the `__init__.py` to add this folder to the PATH. The OCIO build process installs all dependencies and build everything as static which is handy here.

I made a new test wheel on the [test PyPI](https://test.pypi.org/project/opencolorio/2.3.0.dev5/), I only checked it on mac OS but hopefully should work elsewhere if our tests are to be trusted.

JeanChristopheMorinPerso:
Perfect. I'll manually inspect the wheels today on all platforms.

JeanChristopheMorinPerso:
Perfect. I'll manually inspect the wheels today on all platforms.

JeanChristopheMorinPerso:
On Linux, the wheels contain:

```
├── opencolorio.libs
│   └── libOpenColorIO-94504ed6.so
└── PyOpenColorIO
    ├── bin
    │   ├── ocioarchive
    │   ├── ociobakelut
    │   ├── ociocheck
    │   ├── ociochecklut
    │   ├── ocioconvert
    │   ├── ociolutimage
    │   ├── ociomakeclf
    │   ├── ocioperf
    │   ├── ociowrite
    │   ├── pyocioamf
    │   │   ├── config-aces-reference.yaml
    │   │   ├── example.amf
    │   │   ├── example_referenced_lut.clf
    │   │   ├── pyocioamf.py
    │   │   └── README.md
    │   └── pyociodisplay
    │       ├── pyociodisplay.py
    │       ├── README.md
    │       └── requirements.txt
    ├── command_line.py
    ├── __init__.py
    ├── libOpenColorIO.so
    └── PyOpenColorIO.so
```

`libOpenColorIO.so` seems to be duplicated (`opencolorio.libs/libOpenColorIO-94504ed6.so` which was created by repairwheel and `PyOpenColorIO/libOpenColorIO.so` which is the one generated by the actual build). When I run `diff <(nm --format just-symbols -C opencolorio.libs/libOpenColorIO-94504ed6.so) <(nm -C PyOpenColorIO/libOpenColorIO.so)`, both libraries have the same symbols.

The diff of `readelf -d` is
```
[jcmorin@arch01 asd]$ diff -u <(readelf -d opencolorio.libs/libOpenColorIO-94504ed6.so) <(readelf -d PyOpenColorIO/libOpenColorIO.so)
--- /dev/fd/63	2023-07-02 14:54:36.266401410 -0400
+++ /dev/fd/62	2023-07-02 14:54:36.269734762 -0400
@@ -1,22 +1,21 @@
 
-Dynamic section at offset 0x8d1608 contains 29 entries:
+Dynamic section at offset 0x6675c0 contains 28 entries:
   Tag        Type                         Name/Value
- 0x000000000000000f (RPATH)              Library rpath: [opencolorio.libs]
- 0x000000000000000e (SONAME)             Library soname: [libOpenColorIO-94504ed6.so]
  0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
  0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
  0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
  0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
+ 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN]
  0x000000000000000c (INIT)               0x83000
  0x000000000000000d (FINI)               0x51fab0
  0x0000000000000019 (INIT_ARRAY)         0x648438
  0x000000000000001b (INIT_ARRAYSZ)       992 (bytes)
  0x000000000000001a (FINI_ARRAY)         0x648818
  0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
- 0x000000006ffffef5 (GNU_HASH)           0x670028
- 0x0000000000000005 (STRTAB)             0x68c000
+ 0x000000006ffffef5 (GNU_HASH)           0x260
+ 0x0000000000000005 (STRTAB)             0xcd70
  0x0000000000000006 (SYMTAB)             0x2c60
- 0x000000000000000a (STRSZ)              99848 (bytes)
+ 0x000000000000000a (STRSZ)              99804 (bytes)
  0x000000000000000b (SYMENT)             24 (bytes)
  0x0000000000000003 (PLTGOT)             0x669000
  0x0000000000000002 (PLTRELSZ)           23952 (bytes)
```

The executables use `opencolorio.libs/libOpenColorIO-94504ed6.so`
```
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN:$ORIGIN/../../opencolorio.libs]
 0x0000000000000001 (NEEDED)             Shared library: [libOpenColorIO-94504ed6.so]
```

And it looks like the osx wheels have the same problem.

On Windows, everything seems fine
```
C:.
└───PyOpenColorIO
    │   command_line.py
    │   PyOpenColorIO.pyd
    │   __init__.py
    │
    └───bin
        │   ocioarchive.exe
        │   ociobakelut.exe
        │   ociocheck.exe
        │   ociochecklut.exe
        │   ocioconvert.exe
        │   ociolutimage.exe
        │   ociomakeclf.exe
        │   ocioperf.exe
        │   ociowrite.exe
        │   OpenColorIO_2_3.dll
        │
        ├───pyocioamf
        │       config-aces-reference.yaml
        │       example.amf
        │       example_referenced_lut.clf
        │       pyocioamf.py
        │       README.md
        │
        └───pyociodisplay
                pyociodisplay.py
                README.md
                requirements.txt
```

I am able to import `PyOpenColorIO` without any issues.

remia:
Thanks for checking @JeanChristopheMorinPerso! I had tested the wheel locally on mac OS but only using pip wheel, I will try to see why repairwheel (or mac OS equivalent) think it has to fix the wheel which seem like it was already working. Great catch though, I was wondering why Windows wheels were half the size.

cedrik-fuoco-adsk (approved):

remia:
As a quick update, I think the reason why repairwheel / delocate is patching the generated wheel is because the disabling of the SONAME symlinks I added as a CMake build option when building the wheel is preventing CMake to generate the correct library install names (instead of @rpath/mylib.dylib I get a full path on mac OS for example). I did that because otherwise I get 3 copy of the OCIO lib in the wheel due to missing support for symlinks. On the flip side, I assume relying on a wheel repair process to get viable wheels is not really ideal, as it would presumably not work well for wheels build manually from source distribution.

I'll continue my investigations.

remia:
I uploaded a new test release, `pip install -i https://test.pypi.org/simple/ opencolorio==2.3.0.dev6`, the wheels should not contain duplicated libs anymore.

setup.py
```diff
@@ -39,6 +39,39 @@ def get_version():
         shutil.rmtree(dirpath)
 
 
+# Remove symlinks as Python wheels do not support them at the moment.
```
JeanChristopheMorinPerso, line 42: I think if you remove the `SOVERSION` it shuld fix the problem. And anyway when building the wheel, it makes little sense to have the version embedded in the dynamic library file name.

setup.py
```diff
@@ -73,12 +107,13 @@ def build_extension(self, ext):
 
         cmake_args = [
             "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={}".format(extdir),
+            "-DCMAKE_RUNTIME_OUTPUT_DIRECTORY={}".format(bindir),
             "-DPython_EXECUTABLE={}".format(sys.executable),
             # Not used on MSVC, but no harm
             "-DCMAKE_BUILD_TYPE={}".format(cfg),
-            "-DBUILD_SHARED_LIBS=OFF",
+            "-DBUILD_SHARED_LIBS=ON",
             "-DOCIO_BUILD_DOCS=ON",
```
JeanChristopheMorinPerso, line 115: Unrelated to this, but is there a reason why the docs are built here? It makes it so that to build a wheel, you need doxygen and sphinx installed.

setup.py
remia: I made a flag to disable the symlink, this is much simpler indeed, thanks.

setup.py
remia: This is used for the docstring, I made this optional (see also this issue #1744).

JeanChristopheMorinPerso: I see. I'm not too sure if I understand what you mean. Anyway, my comment is unrelated to this PR. I just think that in general, building a wheel shouldn't involve documentation tools.

src/bindings/python/package/__init__.py
```diff
@@ -16,10 +29,11 @@
 # environment variable to 0.
 #
 
-if sys.version_info >= (3, 8) and platform.system() == "Windows" and os.getenv("OCIO_PYTHON_LOAD_DLLS_FROM_PATH", "1") == "1":
-    for path in os.getenv("PATH", "").split(os.pathsep):
-        if os.path.exists(path) and path != ".":
-            os.add_dll_directory(path)
+if sys.version_info >= (3, 8) and platform.system() == "Windows":
+    if os.getenv("OCIO_PYTHON_LOAD_DLLS_FROM_PATH", "1") == "1":
```
JeanChristopheMorinPerso, line 33: Unrelated to this PR (just a general comment): it would be better if this was off by default (i.e. opt-in, not opt-out).

JeanChristopheMorinPerso (approved):
Great work @remia!

setup.py
JeanChristopheMorinPerso: Indeed, it's now much cleaner!

setup.py
remia: The documentation written in the C++ header will end up in the generated docstring for the Python wheel, developer can use help() to see the documentation, or the IDE can show it directly.

src/bindings/python/package/__init__.py
remia: I generally agree with that although I didn't investigate enough yet the pros and cons, I think we should decide the desired behaviour in the corresponding issue https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1785. When that's settle I can make the change, probably in a new PR to not block this one.

remia:
Thanks @JeanChristopheMorinPerso, the PR is looking much nicer following your reviews!

remia:
Last round of test wheels from the last update couple of days ago available: pip install -i https://test.pypi.org/simple/ opencolorio==2.3.0.dev7

doug-walker:
@remia , would you resolve the conflict here and we will go ahead and merge this?  I notice that our Wheel GitHub Action has been failing somewhat randomly for the last two weeks, not sure if this will improve that?

remia:
@doug-walker The latest fails I see on macOS are related to the recent SIMD updates:

```
    /Users/runner/work/OpenColorIO/OpenColorIO/src/OpenColorIO/CPUInfo.cpp:44:50: error: invalid output constraint '=a' in asm
        __asm__ volatile (".byte 0x0f, 0x01, 0xd0" : "=a"(eax), "=d"(edx) : "c" (index));
                                                     ^
    /Users/runner/work/OpenColorIO/OpenColorIO/src/OpenColorIO/CPUInfo.cpp:58:11: error: invalid output constraint '=a' in asm
            : "=a" (data[0]), "=S" (data[1]), "=c" (data[2]), "=d" (data[3])
```

It seems the earlier issues were solved by @cedrik-fuoco-adsk in #1826

PR #1806, opened 2023-06-17
Add support for minizip-ng 4 API

brad0:


negril:
Seeing how this breaks at the same position as https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1808 it would appear the code is broken elsewhere.

Since [PyColorSpaceSet.cpp](https://github.com/AcademySoftwareFoundation/OpenColorIO/commits/219ac86ef6b992ddf31404d6f318dbaa7540053c/src/bindings/python/PyColorSpaceSet.cpp) was changed recently this seems like the likely culprit.

num3ric:
#1762 suffers from the same CIS issue.

negril:
@brad0 you missed one call, could you add that?
```patch
diff -Naurw a/src/apps/ocioarchive/main.cpp b/src/apps/ocioarchive/main.cpp
--- a/src/apps/ocioarchive/main.cpp
+++ b/src/apps/ocioarchive/main.cpp
@@ -235,7 +235,11 @@
         }   
 
         std::string path = args[0];
+#if MZ_VERSION_BUILD >= 040000
+        reader = mz_zip_reader_create();
+#else
         mz_zip_reader_create(&reader);
+#endif
         struct tm tmu_date;
    
         err = mz_zip_reader_open_file(reader, path.c_str());
```

brad0:
> @brad0 you missed one call, could you add that?

Thanks. I was wondering why I did not notice this at first but we do not build the apps with the OpenBSD port.

brad0:
ping.

brad0:
> Looks good to me, thank you for the contribution @brad0 and sorry for the latency.

Thank you.

brad0:
ping.

cedrik-fuoco-adsk (approved):

brad0:
Thank you.

PR #1808, opened 2023-06-26
Added GetProcessorFromConfigs variants for displays and views.

num3ric:
Yield a single processor from a color space in config (a) to a display-view in config (b), through the interchange space.

Let me know if we think we need test coverage for these.


src/OpenColorIO/Config.cpp
```diff
@@ -4957,14 +4957,148 @@ ConstProcessorRcPtr Config::GetProcessorFromConfigs(const ConstContextRcPtr & sr
     }
 
     auto p2 = dstConfig->getProcessor(dstContext, dstExCs, dstColorSpace);
+    if (!p2)
```
num3ric, line 4960: Was incorrectly p1 here, so I fixed it causing the strange diff.

num3ric:
(Would be good for the library to have a `.clang-format` file.)

remia:
Thanks @num3ric, sounds like a good idea to add some unit tests inspired from the existing tests on other `GetProcessorFromConfigs()`. I think we should create an issue for the clang-format is not already existing (didn't found one), it was discussed some time ago but got lost.

remia:
There is a conflict on the branch following to recent merges in main.

num3ric:
Understood, back from summer break - will revisit either this week or the next.

num3ric:
Added unit tests. Also moved the direction argument last since in this case it reverses the entire chain of transforms.

doug-walker (approved):
Thanks Eric!  No need to make any changes (the two notes are just to myself, for later).

michdolan (approved):

PR #1823, opened 2023-08-17
CMake: Fixes for yaml-cpp and generated config

FtZPetruska:
When packaging `yaml-cpp` 0.8.0 on vcpkg, we met build issues with OpenColorIO due to how the package is linked.

With the release of `yaml-cpp` 0.8.0, the exported target changed from `yaml-cpp` to `yaml-cpp::yaml-cpp`. This only breaks when telling CMake to use the upstream package config over the `Findyaml-cpp` module.

On the other hand, the upstream CMake configs have always provided a `YAML_CPP_LIBRARIES` variable that stores the target name.

To fix this issue, I updated the `Findyaml-cpp` module to be compatible with the upstream config:
- If it does not already exist, add a `yaml-cpp::yaml-cpp` target aliasing `yaml-cpp`
- Set `YAML_CPP_LIBRARIES` to `yaml-cpp::yaml-cpp`

When `yaml-cpp` is linked, expand the `YAML_CPP_LIBRARIES` variable instead of using a target name, ensuring we always get a valid target.

Lastly, in the generated config, check that neither `yaml-cpp` and `yaml-cpp::yaml-cpp` are defined before calling `find_dependency`.

On that note, there were a few other issues with the CMake package config:
- Using `if(NOT <target name>)` always evaluates to true, `if(NOT TARGET <target name>)` should be used to check for the existence of a target.
- The loop to find the package prefix can simply be replaced by `PACKAGE_PREFIX_DIR` (provided by `@PACKAGE_INIT@`). The assumption that the config would always be in `<prefix>/lib/cmake/OpenColorIO` is pretty fragile as a user could overwrite `CMAKE_INSTALL_LIBDIR`. Similarly, vcpkg relocates CMake packages after installation to `share/<package name>` (while updating `PACKAGE_PREFIX_DIR`), also breaking that assumption.

---

An alternative I considered was to leave the Find module intact and to handle it internally as such:
```cmake
ocio_handle_dependency(yaml-cpp ...)
if(NOT TARGET yaml-cpp::yaml-cpp)
  add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
endif()
target_link_libraries(... yaml-cpp::yaml-cpp)
```

And in the exported CMake config:
```cmake
if(NOT TARGET yaml-cpp)
  find_dependency(yaml-cpp ...)
endif()

if(NOT TARGET yaml-cpp::yaml-cpp)
  add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
endif()
```

Let me know if you prefer this solution instead!


michdolan (approved):
Thanks for the fix @FtZPetruska . Looks good to me.

cedrik-fuoco-adsk (approved):

Issue #1455, opened 2021-08-10
Write OpenFX plug-in for FileTransform


doug-walker:
Issue #1117 requests OpenFX plug-ins for OCIO and Michael has PR #1371 in progress that is almost complete to implement a ColorSpaceTransform and DisplayViewTransform plug-in, which are the key ones.  It would also be great to have a few more OpenFX plug-ins, for example for FileTransform.  This would allow people to apply stand-alone LUTs in any of the formats OCIO supports, including the Academy/ASC Common LUT Format (CLF).  The implementation should include an "invert" toggle button, similar to Michael's work.

It would also be great to have OpenFX plug-ins for LookTransform and NamedTransform, although these could also be in separate PRs.

PR #1825, opened 2023-08-22
Working on OpenFX FileTransform Plugin

brkglvn01:
Started working on issue #1455, used the existing plugins as a basis for setting it up, definitely still needs to be fine tuned but would really appreciate some feedback to make sure I'm on the right track


vendor/openfx/OCIOFile.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+void OCIOFileFactory::describeInContext(OFX::ImageEffectDescriptor& desc,
+    OFX::ContextEnum /*context*/)
+{
+    baseDescribeInContext(desc);
+
+    // Define parameters
+    OFX::PageParamDescriptor* page = desc.definePageParam(PARAM_NAME_PAGE_0);
+
+    // Src color space
```
doug-walker, line 119: In the case of a FileTransform, the source argument is not actually a color space.  It's the path to the LUT file.  Since the name of the parameter is actually "src", it's ok to use source in the name, but I would recommend "src_path", "Source File Path", "Source file path name" as the strings to use here.

And please change variable names such as "srcCS" to "srcPath" or something like that.

vendor/openfx/OCIOFile.cpp
```diff
@@ -0,0 +1,142 @@
...
+    tr->setSrc(srcCsName.c_str());
+
+    // Setup and apply processor
+    OCIOProcessor proc(*this);
+
+    proc.setDstImg(dst.get());
+    proc.setSrcImg(src.get());
+    proc.setRenderWindow(args.renderWindow);
+    proc.setTransform(context, tr, (inverse ? OCIO::TRANSFORM_DIR_INVERSE
+        : OCIO::TRANSFORM_DIR_FORWARD));
```
doug-walker, line 62: Very minor nit pick ... what you've done here is not wrong, but it would be better to follow the indentation style of the other OFX plug-ins (which are more consistent with the OCIO coding style guidelines).

vendor/openfx/OCIOFile.cpp
```diff
@@ -0,0 +1,142 @@
...
+    OFX::Clip*& identityClip,
+    double& identityTime)
+{
+    std::string srcCsName = getChoiceParamOption(srcCsNameParam_);
+    OCIO::ConstColorSpaceRcPtr srcCs;
+
+    if (!srcCsName.empty())
+    {
+        OCIO::ConstConfigRcPtr config = getOCIOConfig();
+        srcCs = config->getColorSpace(srcCsName.c_str());
```
doug-walker, line 77: The source is actually the path to a LUT file, it's not a color space.

vendor/openfx/OCIODisplayView.cpp
```diff
@@ -4,6 +4,7 @@
 #include "OCIODisplayView.h"
 #include "OCIOProcessor.h"
 #include "OCIOUtils.h"
+#include "OCIOUtils.h"
```
doug-walker, line 7: Please revert this.

doug-walker (changes requested):
Thanks for this contribution Brooke, this is awesome!  

I have a few comments, but the main one is relative to the source argument being a file path rather than a color space.  You could look at the OCIO FileTransform node in Nuke for an example.  If you fix that, I think it should be pretty close to finished.

PR #1836, opened 2023-08-29
Adsk Contrib - Adding CG and Studio config v2.3

cedrik-fuoco-adsk:
The PR adds the new built-in CG and Studio config

### DIFF
--- a/src/OpenColorIO/builtinconfigs/configs/studio-config-v2.0.0_aces-v1.3_ocio-v2.3.ocio
+++ b/src/OpenColorIO/builtinconfigs/configs/studio-config-v2.0.0_aces-v1.3_ocio-v2.3.ocio
@@ -1,0 +1,1549 @@
+ocio_profile_version: 2.3
+
+environment:
+  {}
+search_path: ""
+strictparsing: true
+luma: [0.2126, 0.7152, 0.0722]
+name: studio-config-v2.0.0_aces-v1.3_ocio-v2.3
+description: |
+  Academy Color Encoding System - Studio Config [COLORSPACES v2.0.0] [ACES v1.3] [OCIO v2.3]
+  ------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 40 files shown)


michdolan (approved):

src/OpenColorIO/builtinconfigs/configs/studio-config-v2.0.0_aces-v1.3_ocio-v2.3.ocio
```diff
@@ -0,0 +1,1549 @@
...
+      ACEStransformID: urn:ampas:aces:transformId:v1.5:ACEScsc.Academy.LogC4_to_ACES.a1.1.0
+      ACEStransformID: urn:ampas:aces:transformId:v1.5:ACEScsc.Academy.ACES_to_LogC4.a1.1.0
+    isdata: false
+    categories: [file-io]
+    encoding: log
+    allocation: uniform
+    to_scene_reference: !<GroupTransform>
+      name: ARRI LogC4 to ACES2065-1
+      children:
+        - !<LogCameraTransform> {log_side_slope: 0.0647954196341293, log_side_offset: -0.295908392682586, lin_side_slope: 2231.82630906769, lin_side_offset: 64, lin_side_break: -0.0180569961199113, direction: inverse}
```
remia, line 718: I was curious why the BuiltinTransform was not used for example here, I guess it's related to https://github.com/AcademySoftwareFoundation/OpenColorIO-Config-ACES/issues/103 and might be changed in subsequent versions?

remia (approved):

src/OpenColorIO/builtinconfigs/configs/studio-config-v2.0.0_aces-v1.3_ocio-v2.3.ocio
KelSolaar: Yes, it is also made slightly more complicated because the code for the config is contributed by ARRI, so we would need to make them aware that we are using the Builtin.

Issue #1846, opened 2023-09-04
v2.3.0 Python bindings source builds failing to load libOpenColorIO


remia:
Filing this following report by @mjtitchener-fn during today's TSC meeting for tracking purpose.

On the latest release 2.3 and main branch, when building the Python package from source, there is an issue with RPATH not correctly installed preventing the import of PyOpenColorIO when using shared build.

The issue is most likely [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/bindings/python/CMakeLists.txt#L173), `CMAKE_INSTALL_RPATH` is indeed always defined at this point because it is set globally [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/share/cmake/utils/CompilerFlags.cmake#L140).

This does not affect the wheels because RPATH are defined independently [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/setup.py#L164).

KevinJW:
My build on Linux, sets an rpath based on the top level setting e.g.  cmake command

`cmake \
    -DCMAKE_INSTALL_PREFIX="$DEST" \
    -DCMAKE_INSTALL_RPATH_USE_LINK_PATH=TRUE \
    -DOCIO_INSTALL_EXT_PACKAGES=ALL \
    -DPython_EXECUTABLE=$PYTHON_EXECUTABLE \
    -DPYTHON=$PYTHON_EXECUTABLE \
    -DOCIO_USE_OPENEXR_HALF=ON \
    -Dyaml-cpp_STATIC_LIBRARY=ON \
    -DHalf_STATIC_LIBRARY=ON \
    ${BASE}/${SOURCE}
`

and then near the end of the build we see:

`-- Set runtime path of "/<my local install path>/opencolorio.v2.3.0-python2.7/bin/ociodisplay" to "$ORIGIN/../lib64:$ORIGIN"`
`-- Set runtime path of "/<my local install path>/opencolorio.v2.3.0-python2.7/lib64/python2.7/site-packages/PyOpenColorIO/PyOpenColorIO.so" to "$ORIGIN/../lib64:$ORIGIN"`

So the python bindings DSO is using the same rpath as the executables rather then the override, as @remia is suggesting.

This occurs for both Python 2.7 and 3.

mjtitchener-fn:
I was chatting to @remia about this on Slack after the TSC call yesterday.
[...]

PR #1849, opened 2023-09-06
Fix wrong RPATH being injected into v2.3.0 Python bindings DSO

KevinJW:
Suggestion for fixing #1846 

Signed-off-by: Kevin Wheatley <kevin.wheatley@framestore.com>


src/bindings/python/CMakeLists.txt
```diff
@@ -170,7 +170,7 @@ if(UNIX)
 	set_property(TARGET PyOpenColorIO PROPERTY POSITION_INDEPENDENT_CODE ON)
 endif()
 
-if (UNIX AND NOT CMAKE_SKIP_RPATH AND NOT CMAKE_INSTALL_RPATH)
+if (UNIX AND NOT CMAKE_SKIP_RPATH AND CMAKE_INSTALL_RPATH)
```
remia, line 173: Suggest to rollback to the condition we had in [RB-2.2](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/RB-2.2/src/bindings/python/CMakeLists.txt#L173) and do the same for [CompilerFlags](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/RB-2.2/share/cmake/utils/CompilerFlags.cmake#L100).
`if (UNIX AND NOT CMAKE_SKIP_RPATH)`

I originally was trying to have the [Python build](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/setup.py#L158) only ship with the required RPATH and nothing more but it turned out it created this bug and I also later realised that the wheel have a "repair" step at the end, which for macOS simply replace the RPATH entirely (as can be seen in GHA logs; I assume Linux is similar but the log is quieter) so these were probably not really needed.

remia: @KevinJW What do you think about simplifying the above as suggested in the earlier comment? Would be great to have this in the upcoming release @doug-walker is planning for this week.

doug-walker (approved):
Approving this for inclusion in 2.3.1, although would still like to pursue Remi's suggestion as a follow-up.

remia (approved):

PR #1852, opened 2023-09-08
Adsk Contrib - Changing version to 2.4.0

cedrik-fuoco-adsk:
Changing OCIO version from 2.3.0 to 2.4.0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ endif()
 # Project definition.
 
 project(OpenColorIO 
-    VERSION 2.3.0
+    VERSION 2.4.0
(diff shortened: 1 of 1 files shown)


remia (approved):

KevinJW:
A discussion point for the next TSC meeting:

Should we be bumping the version to '2.4.0' prior to there being any API/ABI changes? There are likely a number of bug fixes that will be needed for a 2.3.1 or similar patch releases.

And if not what should it be?

doug-walker (approved):

doug-walker:
The decision at the TSC was to go ahead and update the version now since we don't usually include ABI compatibility review on each new PR that goes into main.

PR #1855, opened 2023-09-13
Fix regression in cccid handling when no value is supplied

### LINKED ISSUE #1854: v2.X.X Regression/bug with cccid handling in FileTransforms
Investigating a change in behaviour seen when trying to migrate a config from v1 to v2 being unable to load a ColorCorrection file, generating the error message when running `ociocheck`:

```
The transform file: test_config/master.cdl failed while building ops with this error: You must specify a valid cccid to load from the ccc file (either by name or index). id='' is not found in the file, and is not parsable as an integer index.
```
Points of note:
- The FileTransform is decoding the file as a CDL/CCC not a CC file due to the filename
- Renaming the file to .cc  and updating the config file - works as expected
- Our documentation states that passing a cccid as NULL or empty string should use the first CC found in the file
- Adding a cccid=0 to the FileTransform YAML works

Expected:

Based on historical behaviour in v1 we should continue to try decode the file based on the contents of the file and not the file extension. we should also follow the v2 documentation accepting the file as master.cdl and default to using the first (only) colour correction found in the file, without needing users to add an explicit cccid to the config.

This should work for files named .ccc as well as .cdl

As a thought experiment I'm unclear if this should be considered as several bugs:

1. Should the FileTransform continue to try load the file as a CC file after failing as a CDL/CCC?
2. Is the empty string bug found in both FileFormatCCC.cpp and FileFormat.CDL.cpp

Reproduction files...

minimal.ocio:
```
ocio_profile_version: 1
description: Minimal
search_path: .

roles:
  default: basic
  scene_linear: basic
  data: basic
  reference: basic
  compositing_log: basic
  color_timing: basic
  color_picking: basic
  texture_paint: basic
  matte_paint: basic
  rendering: basic
  aces_interchange: basic
  cie_xyz_d65_interchange: basic

displays:
  display:
    - !<View> {name: basic, colorspace: basic }
    - !<View> {name: cdl, colorspace: basic_cdl }

colorspaces:
  - !<ColorSpace>
    name: basic

  - !<ColorSpace>
    name: basic_cdl
    from_reference: !<FileTransform> { src: master.cdl }
```

master.cdl:
```
<?xml version="1.0" encoding="UTF-8"?>
<ColorCorrection>
   <SOPNode>
      <Description>No Op CDL</Description>
      <Slope>1.00000 1.00000 1.00000</Slope>
      <Offset>0.00000 0.00000 0.00000</Offset>
      <Power>1.00000 1.00000 1.00000</Power>
   </SOPNode>
   <SatNode>
      <Saturation>1.00000</Saturation>
   </SatNode>
</ColorCorrection>
```

KevinJW:
In v1 of OCIO FileTransforms are able to load .cc files without specifying a cccid.

In v2 this broke causing an exception to be raised instead of using the first cc found in the file.

Fixes #1854

Signed-off-by: Kevin Wheatley <kevin.wheatley@framestore.com>

### DIFF
--- a/src/OpenColorIO/fileformats/FileFormatCCC.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatCCC.cpp
@@ -217,7 +217,7 @@ void LocalFileFormat::buildFileOps(OpRcPtrVec & ops,
     {
         // Use 0 for empty string.
         int cccindex=0;
-        if (StringToInt(&cccindex, cccid.c_str(), true))
+        if (cccid.empty() || StringToInt(&cccindex, cccid.c_str(), true))
         {
             int maxindex = ((int)cachedFile->m_transformVec.size())-1;
             if (cccindex<0 || cccindex>maxindex)
--- a/src/OpenColorIO/fileformats/FileFormatCDL.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatCDL.cpp
@@ -247,7 +247,7 @@ LocalFileFormat::buildFileOps(OpRcPtrVec & ops,
     {
         // Use 0 for empty string.
         int cccindex=0;
-        if (StringToInt(&cccindex, cccid.c_str(), true))
+        if (cccid.empty() || StringToInt(&cccindex, cccid.c_str(), true))
         {
             int maxindex = ((int)cachedFile->m_transformVec.size())-1;
             if (cccindex<0 || cccindex>maxindex)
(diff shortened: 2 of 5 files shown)


doug-walker (approved):
Totally unrelated to this PR, but I'm wondering about the copyright notice at the top of FileFormatCDL.cpp.  This is one of the last remaining modules that doesn't use the SPDX-License-Identifier.  Just wondering if you recently went through the process of getting your previous OIIO commits relicensed for Larry and, if so, if the same could be done here?

KevinJW:
I totally missed the fact that I might have been involved with that file from the beginning! I guess technically it would be for my previous employers to make common on it. I haven't had a requirement for migrating any other contributions to other projects, so this would be the first.

I would imagine that we might want to invite Michele @ Cinesite to a TSC or to comment.

remia (approved):
Thanks for the fix @KevinJW (we also faced this issue and had to update the YAML with a dummy cccid).

PR #1861, opened 2023-09-20
Correctly recover CXX_FLAGS in CheckSupportSSE2.cmake

licy183:
CXXFLAGS is backed up in the beginning of CheckSupportSSE2, but it is not correctly recovered.


remia (approved):
Thanks for the contribution @licy183, good catch.

PR #1862, opened 2023-09-20
Correctly reset CMAKE_CXX_FLAGS.

num3ric:
This was causing errors in my build script that used custom CMAKE_CXX_FLAGS.


num3ric:
Ah just noticed this is a duplicate of PR #1861

remia:
> Ah just noticed this is a duplicate of PR #1861

Indeed, should we close this one then?

Issue #1885, opened 2023-10-12
Default view transform missing on config copy


mdecaria:
Default view transform is dropped on config copy using the equal operator.

Python reproduction
```python
import PyOpenColorIO as ocio
config = ocio.Config.CreateFromFile("ocio://default")
config.getDefaultViewTransformName()
'Un-tone-mapped'
ocio.SetCurrentConfig(config)
ocio.GetCurrentConfig().getDefaultViewTransformName()
''
```

C++ reproduction

```C++
OCIO::ConstConfigRcPtr config = OCIO::Config::CreateFromBuiltinConfig("ocio://default");
std::cerr << "config 1 "<< config->getDefaultViewTransformName() << std::endl;
OCIO::ConstConfigRcPtr config2 = config->createEditableCopy();
std::cerr << "config 2 " << config2->getDefaultViewTransformName() << std::endl;

Output:
config 1 Un-tone-mapped
config 2
```

PR #1886, opened 2023-10-12
Fix missing Default View Transform on equal operator

mdecaria:
Fixes: https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1885


remia (approved):
Good catch @mdecaria, thanks for the PR.

cedrik-fuoco-adsk (approved):
Thanks for the PR!

doug-walker (approved):

PR #1887, opened 2023-10-12
Include headers files in cmake

num3ric:
Not having access to header files in the generated solution has been bothering me from the get-go. Though maybe not an issue if you're not working in visual studio like me - any thoughts?

(Also reordered the cpp files.)


num3ric:
Nevermind, realizing this is still missing the most important ones in `inlude/OpenColorIO`.  I'll revisit later. Should probably done cleanly with `source_group` instead.

PR #1890, opened 2023-10-19
Adsk Contrib - Fix for the doxygen installation failure in the Dependencies latest workflow (Windows)

cedrik-fuoco-adsk:
This is a tentative fix for the failure in the dependencies' latest workflow with the Doxygen installation (it seems to be failing randomly). 

We are using the package manager Chocolatey, and the Chocolatey package is getting the doxygen files from Sourceforge. It seems like there are issues with either the Chocolatey package or SourceForge and that causes the download to fail.

I've changed the script to download the doxygen release manually from GitHub and install it manually.

---

An alternative to this would be to ask the maintainer of the doxygen package on Chocolatey (https://github.com/mlt/chocolatey-packages) to download Doxygen from their GitHub release.

### DIFF
--- a/.github/workflows/dependencies_latest.yml
+++ b/.github/workflows/dependencies_latest.yml
@@ -292,7 +292,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Install docs env
-        run: share/ci/scripts/windows/install_docs_env.sh
+        run: |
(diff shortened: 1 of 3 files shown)


remia (approved):
Thanks @cedrik-fuoco-adsk, this has been a long standing issue, great to see we have a simple workaround!

doug-walker (approved):

Issue #1858, opened 2023-09-19
Broken yaml 0.8.0 detection.


SlawekNowy:
The EndeavourOS (flavour of ArchLinux) package of yaml-cpp has no `yaml-cpp` target, only `yaml-cpp::yaml-cpp`

The config file is located and read from properly.
```
[slawomir@slawomir-hppavilionlaptop15cc5xx build]$ cmake ..
-- The CXX compiler identification is GNU 13.2.1
-- The C compiler identification is GNU 13.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Setting build type to 'Release' as none was specified.
-- Setting C++ version to '14' as none was specified.
-- Performing Test COMPILER_SUPPORTS_CXX14
-- Performing Test COMPILER_SUPPORTS_CXX14 - Success
-- 
-- Checking for GPU configuration...
-- Found OpenGL (no version information)
-- Found GLEW (version "2.2.0")
-- Found GLUT (no version information)
-- GLVND supported
-- Performing Test COMPILER_SUPPORTS_ARM_NEON
-- Performing Test COMPILER_SUPPORTS_ARM_NEON - Failed
-- Performing Test COMPILER_SUPPORTS_AVX
-- Performing Test COMPILER_SUPPORTS_AVX - Success
-- Performing Test COMPILER_SUPPORTS_AVX2
-- Performing Test COMPILER_SUPPORTS_AVX2 - Success
-- Performing Test COMPILER_SUPPORTS_AVX512
-- Performing Test COMPILER_SUPPORTS_AVX512 - Success
-- Performing Test COMPILER_SUPPORTS_SSE42
-- Performing Test COMPILER_SUPPORTS_SSE42 - Success
-- Performing Test COMPILER_SUPPORTS_SSE4
-- Performing Test COMPILER_SUPPORTS_SSE4 - Success
-- Performing Test COMPILER_SUPPORTS_SSSE3
-- Performing Test COMPILER_SUPPORTS_SSSE3 - Failed
-- Performing Test COMPILER_SUPPORTS_SSE3
[...]

PR #1891, opened 2023-10-20
Simplify the Findyaml-cpp module

tobim:
This fixes compatibility with yaml-cpp 0.8, which previously failed because of a `get_property` call with the wrong target name. I took the liberty to add a few simplifications along the way.

### DIFF
--- a/share/cmake/modules/Findyaml-cpp.cmake
+++ b/share/cmake/modules/Findyaml-cpp.cmake
@@ -10,10 +10,9 @@
 #   yaml-cpp_VERSION        - Library's version
 #
 # Global targets defined by this module:
-#   yaml-cpp
+#   yaml-cpp::yaml-cpp
 #
 # For compatibility with the upstream CMake package, the following variables and targets are defined:
-#   yaml-cpp::yaml-cpp      - Alias of the yaml-cpp target
 #   YAML_CPP_LIBRARIES      - Libraries to link against yaml-cpp
 #   YAML_CPP_INCLUDE_DIR    - Include directory
 #
@@ -41,32 +40,39 @@ if(CMAKE_BUILD_TYPE MATCHES "[Dd][Ee][Bb][Uu][Gg]")
    set(BUILD_TYPE_DEBUG ON)
 endif()
 
-if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
-    set(_yaml-cpp_REQUIRED_VARS yaml-cpp_LIBRARY)
+if(yaml-cpp_FIND_QUIETLY)
+    set(quiet QUIET)
+endif()
 
+if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
+
+    # Search for yaml-cpp-config.cmake
     if(NOT DEFINED yaml-cpp_ROOT)
-        # Search for yaml-cpp-config.cmake
-        find_package(yaml-cpp ${yaml-cpp_FIND_VERSION} CONFIG QUIET)
+        find_package(yaml-cpp ${yaml-cpp_FIND_VERSION} CONFIG ${quiet})
     endif()
 
     if(yaml-cpp_FOUND)
-        get_target_property(yaml-cpp_LIBRARY yaml-cpp LOCATION)
+        # Alias target for yaml-cpp < 0.8 compatibility
+        if(TARGET yaml-cpp AND NOT TARGET yaml-cpp::yaml-cpp)
+            add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
+        endif()
+
+        set(yaml-cpp_INCLUDE_DIR ${YAML_CPP_INCLUDE_DIR})
+        get_target_property(yaml-cpp_LIBRARY yaml-cpp::yaml-cpp LOCATION)
     else()
 
         # As yaml-cpp-config.cmake search fails, search an installed library
         # using yaml-cpp.pc .
 
-        list(APPEND _yaml-cpp_REQUIRED_VARS yaml-cpp_INCLUDE_DIR yaml-cpp_VERSION)
-
         # Search for yaml-cpp.pc
-        find_package(PkgConfig QUIET)
-        pkg_check_modules(PC_yaml-cpp QUIET "yaml-cpp>=${yaml-cpp_FIND_VERSION}")
+        find_package(PkgConfig ${quiet})
+        pkg_check_modules(PC_yaml-cpp ${quiet} "yaml-cpp>=${yaml-cpp_FIND_VERSION}")
 
         # Try to detect the version installed, if any.
         if(NOT PC_yaml-cpp_FOUND)
-            pkg_search_module(PC_yaml-cpp QUIET "yaml-cpp")
+            pkg_search_module(PC_yaml-cpp ${quiet} "yaml-cpp")
         endif()
-        
+
         # Find include directory
         find_path(yaml-cpp_INCLUDE_DIR 
             NAMES
@@ -91,7 +97,7 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
             # Prefer static lib names
             set(_yaml-cpp_STATIC_LIB_NAMES 
                 "${CMAKE_STATIC_LIBRARY_PREFIX}yaml-cpp${CMAKE_STATIC_LIBRARY_SUFFIX}")
-            
+
             # Starting from 0.7.0, all platforms uses the suffix "d" for debug.
             # See https://github.com/jbeder/yaml-cpp/blob/master/CMakeLists.txt#L141
             if(BUILD_TYPE_DEBUG)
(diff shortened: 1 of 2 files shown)


doug-walker:
@tobim , thanks for the contribution.  It looks like it is failing the CI on several platforms, could you take a look?

tobim:
@doug-walker I couldn't figure out what is going wrong on the static workflow jobs locally so I took the liberty add a commit that enables a bit more output. I hope that will point me towards the problem.

tobim:
@doug-walker can we give it another shot? I hope I got it right this time.

share/cmake/modules/Findyaml-cpp.cmake
```diff
@@ -118,51 +124,35 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
         if(PC_yaml-cpp_FOUND)
             set(yaml-cpp_VERSION "${PC_yaml-cpp_VERSION}")
         endif()
+
+        # Create target
```
remia, line 128: Ideally I would stick with the way it's done in other Find scripts, that is at the end, if only to have consistency and allow for easier read and refactoring later if needed.

share/cmake/modules/Findyaml-cpp.cmake
```diff
@@ -118,51 +124,35 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
...
+        add_library(yaml-cpp::yaml-cpp UNKNOWN IMPORTED GLOBAL)
+        set_target_properties(yaml-cpp::yaml-cpp PROPERTIES
+            IMPORTED_LOCATION ${yaml-cpp_LIBRARY}
+            INTERFACE_INCLUDE_DIRECTORIES ${yaml-cpp_INCLUDE_DIR}
+        )
+        # Required because Installyaml-cpp.cmake creates `yaml-cpp::yaml-cpp`
+        # as an alias, and aliases get resolved in exported targets.
+        add_library(yaml-cpp ALIAS yaml-cpp::yaml-cpp)
```
remia, line 136: Not sure I understand exactly the issue, would it be possible to fix Installyaml-cpp instead?

share/cmake/modules/Findyaml-cpp.cmake
```diff
@@ -41,32 +40,39 @@ if(CMAKE_BUILD_TYPE MATCHES "[Dd][Ee][Bb][Uu][Gg]")
...
     endif()
 
     if(yaml-cpp_FOUND)
-        get_target_property(yaml-cpp_LIBRARY yaml-cpp LOCATION)
+        # Alias target for yaml-cpp < 0.8 compatibility
+        if(TARGET yaml-cpp AND NOT TARGET yaml-cpp::yaml-cpp)
+            add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
+        endif()
+
+        set(yaml-cpp_INCLUDE_DIR ${YAML_CPP_INCLUDE_DIR})
```
remia, line 60: [Minor] Don't think this is needed here?

share/cmake/modules/Findyaml-cpp.cmake
```diff
@@ -118,51 +124,35 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
...
+        add_library(yaml-cpp ALIAS yaml-cpp::yaml-cpp)
     endif()
 
     # Override REQUIRED if package can be installed
     if(OCIO_INSTALL_EXT_PACKAGES STREQUAL MISSING)
         set(yaml-cpp_FIND_REQUIRED FALSE)
     endif()
 
+    set(YAML_CPP_INCLUDE_DIR "${yaml-cpp_INCLUDE_DIR}")
+    set(YAML_CPP_LIBRARIES yaml-cpp::yaml-cpp)
```
remia, line 145: Would like to remove the use of `YAML_CPP_LIBRARIES` in the project and just stick to the yaml-cpp::yaml-cpp target now that it is consistently available, could you update that?

remia (commented):
Thanks @tobim, I added a few comments below.

share/cmake/modules/Findyaml-cpp.cmake
tobim: That doesn't make any sense scoping wise. You always need to create the target when you're in this branch and never when you're in the other one above.

share/cmake/modules/Findyaml-cpp.cmake
tobim: Installyaml-cpp uses `ExternalProject_Add`, so you have to deal with whatever your get from that project. The most sensible alternative to the workaround here is to upgrade the version that gets installed to yaml-cpp 0.8.

share/cmake/modules/Findyaml-cpp.cmake
tobim: The module documents that this value will be set on line 9. I would prefer to keep that consistent.

share/cmake/modules/Findyaml-cpp.cmake
tobim: Doing that would break in case [CMAKE_FIND_PACKAGE_PREFER_CONFIG](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html) is enabled and yaml-cpp is pre 0.8. Maybe an acceptable sacrifice?

share/cmake/modules/Findyaml-cpp.cmake
remia: Yes and that's handled in all other Find* scripts we have with use with conditionals at the end of the script.

share/cmake/modules/Findyaml-cpp.cmake
remia: When Installyaml-cpp is called, the Findyaml-cpp is not called after, unless I'm missing something? Is this about the static library build? Looking at the Installyaml-cpp it also always will create a yaml-cpp::yaml-cpp not matter the library version requested isn't it?

share/cmake/modules/Findyaml-cpp.cmake
remia: Afraid I'm not following what you mean here, the ALIAS target should be created when version pre 0.8 is found?

share/cmake/modules/Findyaml-cpp.cmake
remia: Fair enough, this duplicated set of exported variables (upper / lower case) is a bit confusing, might be good to remove at a later point.

share/cmake/modules/Findyaml-cpp.cmake
remia: I apologies about the last part of the answer, I was looking at my own local changes, I'd suggest to update the Installyaml-cpp script to create a yaml-cpp::yaml-cpp target for simplicity?

share/cmake/modules/Findyaml-cpp.cmake
tobim: If `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is enabled this entire module will not be used, and then there also won't be a `yaml-cpp::yaml-cpp` target for v0.7.

share/cmake/modules/Findyaml-cpp.cmake
tobim: You weren't wrong, there is an alias target in `Installyaml-cpp.cmake`. But there is one case where that doesn't help: The generated `OpenColorIOTargets.cmake` will nonetheless link the static library to `yaml-cpp` instead of `yaml-cpp::yaml-cpp`.

This was actually caught by "Test CMake Consumer with static OCIO" CI task.

share/cmake/modules/Findyaml-cpp.cmake
remia: Thanks @tobim, I was under the assumption that our `ocio_handle_dependency` macro would only use the CONFIG mode if the PREFER_CONFIG argument was specified but it appears the first `find_package` just use the CONFIG mode if `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is specified. If we wanted to force it to use the MODULE first we would have to set `ocio_dep_CONFIG_string` to MODULE by default I assume. @cedrik-fuoco-adsk is this on purpose?

share/cmake/modules/Findyaml-cpp.cmake
remia: Ok thanks for the details, I'm fine using this if needed then. I tested on my [fork](https://github.com/remia/OpenColorIO/pull/51) with similar changes but didn't get the error on the CMake Static Consumer, any idea why?

It seems like the yaml-cpp upstream have an ongoing [issue](https://github.com/jbeder/yaml-cpp/issues/1241) that might be related to this situation. Looking at the code, it looks like yaml0-cpp is still the target defined upstream and yaml-cpp::yaml-cpp is just an alias even in 0.8.0. Looking at homebrew 0.8.0 version, I have `set(YAML_CPP_LIBRARIES "yaml-cpp")`,

share/cmake/modules/Findyaml-cpp.cmake
tobim: I propose to keep `YAML_CPP_LIBRARIES` for now and clean that up at a later time. I can add a TODO comment to that end in this place if you wish?

share/cmake/modules/Findyaml-cpp.cmake
tobim: > I tested on my https://github.com/remia/OpenColorIO/pull/51 with similar changes but didn't get the error on the CMake Static Consumer, any idea why?

Sorry, no. I was able to replicate that problem by following the steps taken in the CI job.

The upstream issue was actually fixed in https://github.com/jbeder/yaml-cpp/pull/1212.

share/cmake/modules/Findyaml-cpp.cmake
remia: @tobim I agree, let's leave it here for now.

share/cmake/modules/Findyaml-cpp.cmake
tobim: Updated now.

remia:
Thanks for following up on the questions and the latest updates @tobim!

remia (approved):

opoplawski:
Seems like this should mention that it fixes #1858

doug-walker (approved):

doug-walker:
Thanks to everyone for your work on this issue!

PR #1896, opened 2023-11-01
Skip processor concatenation if the display color space is also data.

num3ric:
Our QA stumbled onto this issue with the new merged processor function. It does not act as a passthrough with `Raw` display-view(s), but it should.

### DIFF
--- a/src/OpenColorIO/Config.cpp
+++ b/src/OpenColorIO/Config.cpp
@@ -4723,16 +4723,25 @@ ConstProcessorRcPtr Config::GetProcessorFromConfigs(const ConstContextRcPtr & sr
     if (!p2)
     {
         throw Exception("Can't create the processor for the destination config "
-            "and the destination color space.");
+            "and the destination display view transform.");
     }
 
     ProcessorRcPtr processor = Processor::Create();
     processor->getImpl()->setProcessorCacheFlags(srcConfig->getImpl()->m_cacheFlags);
 
-    // If the source color spaces is a data space, its corresponding processor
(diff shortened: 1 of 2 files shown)


src/OpenColorIO/Config.cpp
```diff
@@ -4723,16 +4723,25 @@ ConstProcessorRcPtr Config::GetProcessorFromConfigs(const ConstContextRcPtr & sr
...
     ProcessorRcPtr processor = Processor::Create();
     processor->getImpl()->setProcessorCacheFlags(srcConfig->getImpl()->m_cacheFlags);
 
-    // If the source color spaces is a data space, its corresponding processor
+    const char* csName = dstConfig->getDisplayViewColorSpaceName(dstDisplay, dstView);
+    const char* displayColorSpaceName = View::UseDisplayName(csName) ? dstDisplay : csName;
+    ConstColorSpaceRcPtr displayColorSpace = dstConfig->getColorSpace(displayColorSpaceName);
+    if (!displayColorSpace)
+    {
+        throw Exception("Can't create the processor for the destination config: "
```
num3ric, line 4737: I believe this is redundant (p2 would have failed above), but left it for safety.

doug-walker: The reason for lines 4695-4709 is to give a more helpful error message if the creation of p1 is going to fail.  So please move lines 4732-4739 to above the creation of p2 in order to serve the same purpose.

doug-walker (changes requested):
One minor request, but other than that it looks good.  Thank you Eric!

doug-walker (approved):

remia (approved):

Shootfast (approved):

PR #1899, opened 2023-11-06
Restore GPU workflow and minor updates to CI

remia:
Tentatively restore runs for the GPU workflow which haven't run for the past 3 months now due to wrong filtering settings (I) introduced during a previous update. Also enabled the undefined behaviour sanitizer to the nightly build as it is easy to use alongside address, some unit tests are then removed as these were triggering it (assuming we don't need to test UB).

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -527,7 +527,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Install docs env
-        run: share/ci/scripts/windows/install_docs_env.sh
+        run: |
+          DOXYGEN_PATH=$GITHUB_WORKSPACE/doxygen
+          share/ci/scripts/windows/install_docs_env.sh "$DOXYGEN_PATH"
+          echo "$DOXYGEN_PATH" >> $GITHUB_PATH
         shell: bash
         if: matrix.build-docs == 'ON'
       - name: Install tests env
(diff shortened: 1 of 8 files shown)


remia:
Please ignore the CI failure in Platform latest, it is happening in OpenEXR (using non standard C++ type) and is not related to these changes.

`/home/runner/work/OpenColorIO/OpenColorIO/_build/ext/build/openexr/src/openexr_install/src/bin/exrcheck/main.cpp:59:13: error: unknown type name 'uintptr_t'; did you mean 'intptr_t'?`

doug-walker (approved):

remia:
New error in the platform-latest Linux / Clang jobs:

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: error: call to consteval function 'std::chrono::hh_mm_ss::_S_fractional_width' is not a constant expression
        static constexpr unsigned fractional_width = {_S_fractional_width()};
                                                      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: note: undefined function '_S_fractional_width' cannot be used in a constant expression
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2275:2: note: declared here
        _S_fractional_width()
        ^
```

[Seems like](https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1918839.html) some kind of incompatibility building Clang14 using libstdc++, hopefully this will be solved automatically when runners are updated.

doug-walker:
Per discussion, merging despite the Platform Latest CI fails.

Issue #1897, opened 2023-11-05
opencolorio 2.3.0 build failure


chenrui333:
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>error build log</summary>

```
  -- Could NOT find pystring (minimum version: "1.1.3", recommended version: "1.1.3")
  --    pystring_ROOT was: /opt/homebrew/opt/pystring
  CMake Error at share/cmake/macros/ocio_handle_dependency.cmake:220 (message):
    pystring is required, will abort at the end.
  -- Could NOT find minizip-ng (minimum version: "3.0.6", recommended version: "3.0.7")
  CMake Error at share/cmake/macros/ocio_handle_dependency.cmake:220 (message):
    minizip-ng is required, will abort at the end.
```

</details>

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/6094230289/job/16535428119
relates to Homebrew/homebrew-core#141062

remia:
Hi @chenrui333,

Thanks for the report, I noticed something similar while trying to build OCIO with `OCIO_INSTALL_EXT_PACKAGES=NONE` and homebrew packages, it seems this is the same thing going on here. I managed to have something working with the modifications in the last 2 commits of this [PR](https://github.com/remia/OpenColorIO/pull/52) though it probably need a bit more testing, note that I include the change from this other [PR](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1796) which is also needed.

PR #1900, opened 2023-11-06
Proposed fix for pystring include paths

irieger:
Make Findpystring.cmake return the full include directory to pystring so that prefixing includes with "pystring/" is no longer required.

This changes the behaviour of the included Findpystring.cmake to match the bahaviour of other regular find_package modules to report the full include directory removing the requirement to use "pystring/" when including "pystring.h" etc.
(This follows from https://github.com/conan-io/conan-center-index/pull/20703#discussion_r1383018774)

### DIFF
--- a/share/cmake/modules/Findpystring.cmake
+++ b/share/cmake/modules/Findpystring.cmake
@@ -28,12 +28,13 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
     # Find include directory
     find_path(pystring_INCLUDE_DIR
         NAMES
-            pystring/pystring.h
+            pystring.h
         HINTS
             ${pystring_ROOT}
         PATH_SUFFIXES
             include
             pystring/include
+            include/pystring
     )
 
     # Find library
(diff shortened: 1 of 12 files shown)


remia:
Thanks @irieger, FYI there was a related discussion today in #1897, here is a similar [commit](https://github.com/remia/OpenColorIO/pull/52/commits/ac64bed13bc9384155784fc191c8db367e43ef06) I did earlier.

irieger:
Good to see. Will close this and comment with some feedback on your Pull request.

irieger:
Added a comment to https://github.com/remia/OpenColorIO/pull/52 as the current state wouldn't build on Arch Linux but otherwise it contains the changes as proposed here.

remia:
Hi @irieger,

I don't see your comment, can you double check you published it? Thanks for looking into it.

irieger:
New to those github workflows. Just sent the review. Hadn't realised I need to push another button. Gitlab which is my day to day tool has a much more prominent submit button ...

PR #1901, opened 2023-11-08
Improve handling of pystring include dir

remia:
As mentioned in #1897 and #1900, there seem to be an issue with the way we look for and use pystring headers. In the case of homebrew package, the include is installed in `/usr/local/include/pystring.h` without the pystring parent folder which prevents OCIO from finding it. This PR should allow us to adapt to the filesystem layouts possible.

@irieger would be great if you have time to confirm it also work for the Conan package issue, I added your changes, thanks!


irieger:
This fixes the pystring patching we needed in the conan build up to now. (At least on my system but I assume that should be fine in general than.)

(Couldn't test outside of Conan where Findyaml-cpp is used without the yaml-cpp patch as yaml-cpp 0.8.0 will not be found. But I think #1891 should solve that, at least does for me here)

doug-walker (approved):

Shootfast (approved):
Looks good to me!

PR #1796, opened 2023-05-10
Do not assume minizip-ng uses zlib for compression

MarcusCalhoun-Lopez:
The file mz_strm_zlib.h does not exist if minizip-ng enables compression via means other than zlib.
Since none of the functions provided by mz_strm_zlib.h are used (mz_stream_zlib_*), it can be safely removed.

PR #1902, opened 2023-11-08
Improve compatibility with minizip-ng

remia:
As discussed in #1897, there is currently an issue when trying to use minizip-ng from homebrew. OCIO can't find it even if it's in theory supposed to support the package compiled with `MZ_COMPAT=ON`. This PR should fix this.


doug-walker (approved):

doug-walker:
Thanks Remi!.  Noting that this includes the fix proposed in PR #1796, so we could close that once this one is merged.

Shootfast (approved):

PR #1912, opened 2023-11-21
ocioview image and PySide6 updates

michdolan:
This PR makes the following changes to `ocioview`:

- Image viewer operates on RGB instead of RGBA, with an assumed opaque alpha
- Image viewer pan and zoom regression fixed following change from imath to NumPy for transform implementation
- Image read implementations moved to utility module
- NumPy array used for image data messages instead of oiio ImageBuf
- UI size adjustments to account for Qt6 style changes
- Pyide6 compatibility fixes


remia (approved):
Thanks for fixing the issue @michdolan, looks good to me and confirm it works fine on mac OS.

src/apps/ocioview/ocioview/message_router.py
```diff
@@ -89,32 +93,41 @@ def set_config_updates_allowed(self, allowed: bool) -> None:
...
+            message_queue.put_nowait(self._prev_image_array)
+
+    def processor_updates_allowed(self) -> bool:
+        return self._processor_updates_allowed
+
+    def set_processor_updates_allowed(self, allowed: bool) -> None:
+        self._processor_updates_allowed = allowed
+        if allowed and self._prev_config is not None:
+            # Rebroadcast last config record
+            message_queue.put_nowait(self._prev_config)
```
KelSolaar, line 121: Shall we use properties here?

```python
    @property
    def image_updates_allowed(self) -> bool:
        return self._image_updates_allowed

    @image_updates_allowed.setter
    def image_updates_allowed(self, allowed: bool) -> None:
        self._image_updates_allowed = allowed
        if allowed and self._prev_image_array is not None:
            # Rebroadcast last image record
            message_queue.put_nowait(self._prev_image_array)
```

KelSolaar (approved):
LGTM, only a minor comment about maybe using properties in the `MessageRouter` class.

doug-walker:
@michdolan , should we go ahead and merge this?

src/apps/ocioview/ocioview/message_router.py
michdolan: Good point. I will address this in a future PR

michdolan:
@remia thoughts on the CI failure I'm getting since updating the base? It's related to CMake, but this work only changed Python files internal to `ocioview`. Safe to ignore?

remia:
> @remia thoughts on the CI failure I'm getting since updating the base? It's related to CMake, but this work only changed Python files internal to `ocioview`. Safe to ignore?

I haven't time to investigate properly this week but it seems like the issue we have in https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1964 ? I didn't see the error as the job was in canceled state today, I relaunched it and it now passes so this might have been fixed with a CMake update.

PR #1918, opened 2023-12-06
Integrate main into RB-2.3 branch and update library version

doug-walker:
This PR implements the 2.3.1 release for the RB-2.3 branch.

I decided to take all of the commits since 2.3.0 since they are all bug fixes or build enhancements with nothing to break ABI compatibility.


remia:
I think we need #1849 for version 2.3.1, otherwise sounds good to me.

remia (approved):

doug-walker:
There are some understandable Actions fails. The Windows Wheel failures seem to be caused by Chocalatey download problems, similar to previous fails, though they worked on the previous commit.  The Platform Latest Linux compilation error only affects one of the command-line tools and is not a blocker since the Linux CI passes.

Going ahead to merge this and get the release out.

doug-walker:
The Windows Wheels now pass, so that was a download issue, as suspected.

PR #1926, opened 2024-01-05
Issue 1922: Add a GitHub Action to use the Mac ARM runners

chandyn:


.github/workflows/ci_workflow.yml
```diff
@@ -347,7 +347,7 @@ jobs:
             python-version: '3.11'
           # Keeping one universal build
           - build: 4
-            arch-type: "x86_64;arm64"
+            arch-type: "x86_64"
```
doug-walker, line 350: Please revert this line, we want to keep at least one universal build.

.github/workflows/ci-macarm.yml
```diff
@@ -0,0 +1,156 @@
...
+       shared=${{ matrix.build-shared }}, 
+       simd=${{ matrix.use-simd }}, 
+       cxx=${{ matrix.cxx-standard }}, 
+       python=${{ matrix.python-version }}, 
+       docs=${{ matrix.build-docs }}, 
+       oiio=${{ matrix.use-oiio}}>'
+    # Avoid duplicated checks when a pull_request is opened from a local branch.
+    if: |
+      github.event_name == 'push' ||
+      github.event.pull_request.head.repo.full_name != github.repository
```
doug-walker, line 59: Larry has some additional logic here to allow a PR containing 'macarm' to run, but it looks like you moved that up to the pull_request branches section.  Seems fine.

.github/workflows/ci-macarm.yml
```diff
@@ -0,0 +1,156 @@
...
+  # us to test specific PRs that we think need ARM verification).
+  pull_request:
+    tags:
+      - v**
+    branches:
+      - '*macarm*'
+      - '*alpha*'
+      - '*beta*'
+      - 'release'
+
```
doug-walker, line 37: Maybe we should schedule a job here, as Larry did, to run once a month regardless?

doug-walker (commented):
Thanks Chandy!  

Should you change the branch name so this PR triggers the new workflow?

.github/workflows/ci-macarm.yml
```diff
@@ -0,0 +1,156 @@
...
+       simd=${{ matrix.use-simd }}, 
+       cxx=${{ matrix.cxx-standard }}, 
+       python=${{ matrix.python-version }}, 
+       docs=${{ matrix.build-docs }}, 
+       oiio=${{ matrix.use-oiio}}>'
+    # Avoid duplicated checks when a pull_request is opened from a local branch.
+    if: |
+      github.event_name == 'push' ||
+      github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: macos-11
```
doug-walker, line 60: What part of this actually causes this to run on an ARM runner rather than our usual x86 runner?

.github/workflows/ci-macarm.yml
lgritz: You should double check that it really is/isn't running where you expect. I got caught by this -- there are some places where `if` isn't respected and I had to experiment a bit.

.github/workflows/ci-macarm.yml
lgritz: You would need that line to read `macos-13-xlarge` to get an ARM based machine.

chandyn:
> Thanks Chandy!
> 
> Should you change the branch name so this PR triggers the new workflow?

That is correct. The triggers in the action are setup to execute on a push or pull provided the branch name contains alpha, beta, release or macarm. I will rename my branch to show the execution of this action.

.github/workflows/ci_workflow.yml
chandyn: > Please revert this line, we want to keep at least one universal build.

Will do

.github/workflows/ci-macarm.yml
chandyn: OCIO did not have that schedule so I decided to remove it from the OIIO action that I copied over. It makes sense to run it at least once a month so we don't discover issues very late into the release cycle.

.github/workflows/ci-macarm.yml
chandyn: Will do. Still learning the ropes with OCIO.

.github/workflows/ci-macarm.yml
chandyn: OIIO runs every month on the 27th a few days before their patch releases. Does OCIO have a regular patch release schedule? Wondering when we should run for OCIO.

PR #1921, opened 2023-12-19
fix: use system `include pystring.h` for `ConfigUtils.cpp`

chenrui333:
closes #1920

cc @doug-walker @wahn

PR #1930, opened 2024-01-15
Fix some compatibility issues for building on some platforms

irieger:
There is a file that was added/changed to include pystring which is a regression to the changes in #1901 which fixed problems with the include directory of pystring by removing the `pystring/` directory prefix from the include path.

I also suggest adding std for strlen which also otherwise fails to build in some compiler/OS constellations, at least when building for conan.

### DIFF
--- a/src/OpenColorIO/ConfigUtils.cpp
+++ b/src/OpenColorIO/ConfigUtils.cpp
@@ -3,7 +3,7 @@
 
 #include "ConfigUtils.h"
 #include "MathUtils.h"
-#include "pystring/pystring.h"
+#include "pystring.h"
 #include "utils/StringUtils.h"
 
 namespace OCIO_NAMESPACE
--- a/src/OpenColorIO/FileRules.cpp
+++ b/src/OpenColorIO/FileRules.cpp
@@ -62,7 +62,7 @@ std::string ConvertToRegularExpression(const char * globPattern, bool ignoreCase
 
     if (ignoreCase)
     {
-        const size_t length = strlen(globPattern);
+        const size_t length = std::strlen(globPattern);
         bool respectCase = false;
         for (size_t i = 0; i < length; ++i)
         {


irieger:
Please feel free to use my suggestions. To my understanding I'd need to publish my postal address (that is what I assumed is meant by "Mailing Address" in the CLA form) to sign the CLA which I currently don't see doing as this seems shady to me.

remia:
Thanks @irieger, the pystring issue should be fixed by #1921 but the namespacing of strlen looks good.

Not sure why the CLA wants the postal address, any idea @carolalynn22 ?

carolalynn22:
Hi! All Linux Foundation CLAs require some sort of physical address, it's pretty standard practice in open source. When you contribute under a company CLA, you get to use your company's address instead, so it seems less "personal". I understand being uncomfortable with it - if you would like to read more about the CLA & LF's data protection / security standards (which are pretty good) here's the link: https://www.linuxfoundation.org/legal/privacy-policy

remia:
Regarding std::strlen, I see a number of other places using just strlen, you didn't encounter issues with those?

irieger:
> Regarding std::strlen, I see a number of other places using just strlen, you didn't encounter issues with those?

That was the only issue when packaging for conan at least.

irieger:
> Hi! All Linux Foundation CLAs require some sort of physical address, it's pretty standard practice in open source. When you contribute under a company CLA, you get to use your company's address instead, so it seems less "personal". I understand being uncomfortable with it - if you would like to read more about the CLA & LF's data protection / security standards (which are pretty good) here's the link: https://www.linuxfoundation.org/legal/privacy-policy

Thanks for the hint. However just having a quick scan through this already makes my feeling about the linux foundation worse than it was before. Openly using tracking is just shit, especially for an organization that I thought was proposing Open Source philosophies. There is no technical need for tracking users. If this was a company contribution I wouldn't care much about the address topic but as this is free time work I'm happy for the team to copy my suggestions but I'm not willing to dive into such a lengthy legal document in a non-native language that doesn't provide me anything but making me more angry about an organization that I thought works for open things. (Part of the open source philosophy for me is to give the users the freedom of their data!)

carolalynn22:
@irieger totally respect your position. Thank you for the pointers regardless!

doug-walker:
Closing inactive PRs.

PR #1932, opened 2024-01-17
Add AVX512 accelerated 1D/3D LUTS

markreidvfx:
`ocioperf.exe --transform tests/data/files/clf/lut1d_32f_example.clf`

Line by Line Average, lut dim 65, 3840x2160 image, Intel(R) Xeon(R) Gold 6226R CPU @ 2.90GHz 
image

`ocioperf.exe --transform tests/data/files/clf/lut3d_preview_tier_test.clf`

Line by Line Average, lut dim 33x33x33, 3840x2160 image, Intel(R) Xeon(R) Gold 6226R CPU @ 2.90GHz 
image

I've only been able to test on one machine with AVX512. Not exactly the performance gains I was hoping for. I'm still new to the instructions set, maybe there are some more optimizations we could do. There are quite a few AVX512 extensions. I've limited this implementation to just the AVX512F (foundation) instructions. That basically means any AVX512 capable CPU should be able run it.

Github actions use to have more intel CPU's with AVX512 available. Lately I've been getting only AMD EPYC CPU's without AVX512 for CI. I don't think there is anyway to request a specific cpu. This is very frustrating and will make this more difficult to maintain and test.

### DIFF
--- a/src/OpenColorIO/AVX512.h
+++ b/src/OpenColorIO/AVX512.h
@@ -431,10 +431,10 @@ struct AVX512RGBAPack<BIT_DEPTH_F32>
 
         avx512RGBATranspose_4x4_4x4_4x4_4x4(r, g, b, a, rgba0, rgba1, rgba2, rgba3);
 
-        _mm512_storeu_ps((__m256*)(out+0),  rgba0);
-        _mm512_storeu_ps((__m256*)(out+16), rgba1);
-        _mm512_storeu_ps((__m256*)(out+32), rgba2);
-        _mm512_storeu_ps((__m256*)(out+48), rgba3);
+        _mm512_storeu_ps((__m512*)(out+0),  rgba0);
+        _mm512_storeu_ps((__m512*)(out+16), rgba1);
+        _mm512_storeu_ps((__m512*)(out+32), rgba2);
+        _mm512_storeu_ps((__m512*)(out+48), rgba3);
     }
 
     static inline void StoreMasked(float *out, __m512 r, __m512 g, __m512 b, __m512 a, uint32_t pixel_count)
@@ -450,13 +450,13 @@ struct AVX512RGBAPack<BIT_DEPTH_F32>
         avx512RGBATranspose_4x4_4x4_4x4_4x4(r, g, b, a, rgba0, rgba1, rgba2, rgba3);
 
         k = _mm512_int2mask((mask >> 0) & 0xFFFF);
-        _mm512_mask_storeu_ps((__m256*)(out+0), k, rgba0);
+        _mm512_mask_storeu_ps((__m512*)(out+0), k, rgba0);
         k = _mm512_int2mask((mask >> 16) & 0xFFFF);
-        _mm512_mask_storeu_ps((__m256*)(out+16), k, rgba1);
+        _mm512_mask_storeu_ps((__m512*)(out+16), k, rgba1);
         k = _mm512_int2mask((mask >> 32) & 0xFFFF);
-        _mm512_mask_storeu_ps((__m256*)(out+32), k, rgba2);
+        _mm512_mask_storeu_ps((__m512*)(out+32), k, rgba2);
         k = _mm512_int2mask((mask >> 48) & 0xFFFF);
-        _mm512_mask_storeu_ps((__m256*)(out+48), k, rgba3);
+        _mm512_mask_storeu_ps((__m512*)(out+48), k, rgba3);
     }
 };
 
--- a/src/OpenColorIO/AVX512.h
+++ b/src/OpenColorIO/AVX512.h
@@ -1,0 +1,466 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright Contributors to the OpenColorIO Project.
+
+
+#ifndef INCLUDED_OCIO_AVX512_H
+#define INCLUDED_OCIO_AVX512_H
+
+#include "CPUInfo.h"
+#if OCIO_USE_AVX512
+
+#include <immintrin.h>
+
+#include <OpenColorIO/OpenColorIO.h>
+#include "BitDepthUtils.h"
+
+// Macros for alignment declarations
+#define AVX512_SIMD_BYTES 64
+#define AVX512_ALIGN(decl) alignas(AVX512_SIMD_BYTES) decl
+
+namespace OCIO_NAMESPACE
+{
+
+inline __m512 av512_clamp(__m512 value, const __m512& maxValue)
+{
+    value = _mm512_max_ps(value, _mm512_setzero_ps());
+    return _mm512_min_ps(value, maxValue);
+}
+
+inline __m512 avx512_movelh_ps(__m512 a, __m512 b)
+{
+    return _mm512_castpd_ps(_mm512_unpacklo_pd(_mm512_castps_pd(a), _mm512_castps_pd(b)));
+}
+
+inline __m512 avx512_movehl_ps(__m512 a, __m512 b)
+{
+    // NOTE: this is a and b are reversed to match sse2 movhlps which is different than unpckhpd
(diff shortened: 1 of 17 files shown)


src/OpenColorIO/ops/lut1d/Lut1DOpCPU_AVX512.cpp
```diff
@@ -0,0 +1,154 @@
...
+        case BIT_DEPTH_UINT8:
+            return linear1D<inBD, BIT_DEPTH_UINT8>;
+        case BIT_DEPTH_UINT10:
+            return linear1D<inBD, BIT_DEPTH_UINT10>;
+        case BIT_DEPTH_UINT12:
+            return linear1D<inBD, BIT_DEPTH_UINT12>;
+        case BIT_DEPTH_UINT16:
+            return linear1D<inBD, BIT_DEPTH_UINT16>;
+        case BIT_DEPTH_F16:
+#if OCIO_USE_F16C
```
doug-walker, line 109: You have the F16C check here, but other parts of the code seem to assume it's always on?

tests/cpu/UnitTestMain.cpp
```diff
@@ -78,6 +79,7 @@ int main(int argc, const char ** argv)
                "--sse2",          &sse2,             "Enable SSE2 Accelerated features",
                "--avx",           &avx,              "Enable AVX Accelerated features",
                "--avx2",          &avx2,             "Enable AVX2 Accelerated features",
+               "--avx512",        &avx512,           "Enable AVX512 Accelerated features",
                "--f16c",          &f16c,             "Enable F16C Accelerated features",
```
doug-walker, line 83: Perhaps indicate which of the other options supports the f16c option?

tests/cpu/AVX512_tests.cpp
```diff
@@ -0,0 +1,514 @@
...
+            break;
+    }
+}
+
+}
+
+DEFINE_SIMD_TEST(packed_uint8_to_float_test)
+{
+    std::vector<uint8_t> inImage(256);
+    std::vector<float> outImage(256);
```
doug-walker, line 190: 256 --> 512?

tests/cpu/AVX512_tests.cpp
```diff
@@ -0,0 +1,514 @@
...
+                                        0,    0,    0,    0,
+                                        0,  270,  500,    2,
+                                        0,    0,    0,    0,
+                                     1023, 1023,    0,    0,
+                                     1023, 1023,    0,    0};
+
+    OCIO::AVX512RGBAPack<OCIO::BIT_DEPTH_F32>::Load(&pixels[0], r, g, b, a);
+    OCIO::AVX512RGBAPack<OCIO::BIT_DEPTH_UINT10>::Store(&outImageU16[0], r, g, b, a);
+
+    for (unsigned i = 0; i < outImageU8.size(); i++)
```
doug-walker, line 404: Minor:  outImageU8 --> outImageU16 (same below).

src/OpenColorIO/AVX512.h
```diff
@@ -0,0 +1,466 @@
...
+    static inline void Store(float *out, __m512 r, __m512 g, __m512 b, __m512 a)
+    {
+        __m512 rgba0, rgba1,rgba2, rgba3;
+
+        avx512RGBATranspose_4x4_4x4_4x4_4x4(r, g, b, a, rgba0, rgba1, rgba2, rgba3);
+
+        _mm512_storeu_ps((__m256*)(out+0),  rgba0);
+        _mm512_storeu_ps((__m256*)(out+16), rgba1);
+        _mm512_storeu_ps((__m256*)(out+32), rgba2);
+        _mm512_storeu_ps((__m256*)(out+48), rgba3);
```
doug-walker, line 437: __m256 --> __m512 ??

src/OpenColorIO/AVX512.h
```diff
@@ -0,0 +1,466 @@
...
+        avx512RGBATranspose_4x4_4x4_4x4_4x4(r, g, b, a, rgba0, rgba1, rgba2, rgba3);
+
+        k = _mm512_int2mask((mask >> 0) & 0xFFFF);
+        _mm512_mask_storeu_ps((__m256*)(out+0), k, rgba0);
+        k = _mm512_int2mask((mask >> 16) & 0xFFFF);
+        _mm512_mask_storeu_ps((__m256*)(out+16), k, rgba1);
+        k = _mm512_int2mask((mask >> 32) & 0xFFFF);
+        _mm512_mask_storeu_ps((__m256*)(out+32), k, rgba2);
+        k = _mm512_int2mask((mask >> 48) & 0xFFFF);
+        _mm512_mask_storeu_ps((__m256*)(out+48), k, rgba3);
```
doug-walker, line 459: __m256 --> __m512 ??

doug-walker (commented):
Thanks Mark!  

I'd like to clarify the F16C option in relation to this.  I guess if AVX512 is supported then we should assume F16C is always supported too, right?  I have a few comments below related to this.

Yes, it's surprising the timings are not faster. Maybe it's bound by memory accesses? I've seen cases where rearranging how the LUTs are stored in memory (at the cost of taking more space) resulted in a speed-up, though not sure if that would help here.

markreidvfx:
> I'd like to clarify the F16C option in relation to this. I guess if AVX512 is supported then we should assume F16C is always supported too, right? I have a few comments below related to this.
> 
Yes, the half float conversion instructions are all part of the AVX512F (Foundation) [extension](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_cvtph_ps&ig_expand=2110,2058,2058&avx512techs=AVX512F).  

The exact overlap between AVX and AVX2 and F16c support has never been exactly clear to me. I think AVX2 pretty much guarantees F16c but I think its best to check for those extensions.

tests/cpu/UnitTestMain.cpp
markreidvfx: Good idea. I can add some clarification on which options support f16c. Only AVX and AVX2 are valid with the f16c. I was thinking of adding f16c emulation like sse2's to AVX. I've been toying around some alternative algorithms trying them make it faster.

src/OpenColorIO/ops/lut1d/Lut1DOpCPU_AVX512.cpp
markreidvfx: Will remove this #if its should always be on

tests/cpu/AVX512_tests.cpp
markreidvfx: This test is correct. The test is to check all 8 bit values, translate to float as expected.

markreidvfx:
I did a bit more perf testing of this with my old [lut3d_perf](https://github.com/markreidvfx/lut3d_perf) tool

image

It also turns out that github runners on a private repos are different then the public repo ones.  The private ones can have [avx512](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#cloud-hosts-used-by-github-hosted-runners). 

I was able to test this pull request on windows with avx512 by setting up a private fork. I kinda used up all my free minutes for the month doing it but all the tests pass 😆

doug-walker (approved):
LGTM!

tests/cpu/AVX512_tests.cpp
remia: Is that a case where using similar code compared to subsequent tests would help make it clearer?
```
size_t maxValue = OCIO::BitDepthInfo<OCIO::BIT_DEPTH_UINT8>::maxValue + 1;
std::vector<uint8_t> inImage(maxValue);
std::vector<float> outImage(maxValue);
```

markreidvfx: @remia I think that is clearer. I can add that to the PR.

markreidvfx:
@remia I added your suggestion to all the SIMD tests. I also rebased on top of the current main.

Issue #1923, opened 2023-12-21
Linux PyPI wheels are incompatible with VFX platform due to ABI issue


doug-walker:
The current Linux wheels for PyPI are based on manylinux2014 and are built on CentOS 7, [which means they use the old libstdc++ ABI](https://bugzilla.redhat.com/show_bug.cgi?id=1546704). Therefore they don't conform to the [VFX Reference Platform for CY2023 and CY2024](https://vfxplatform.com).

It turns out that even if a library does not use C++ std types in its API, it still may cause crashes. In the case of OCIO, this is because of the way [std::regex is implemented in the standard library](https://stackoverflow.com/questions/51382355/stdregex-and-dual-abi).  (OCIO uses std::regex in several internal functions.)

This may cause crashes when the OCIO Python bindings are used in applications built with the new libstdc++ ABI to comply with the current VFX Platform, or in Python on recent versions of Linux, where the new ABI is now the default.

We should update the wheels to include one of the more recent manylinux options.

remia:
Thanks for the report @doug-walker, do you have an example of Linux distribution / version we could use to reproduce the issue (in Docker) or any recent version of the majors distro will do?

remia:
I went ahead and opened a PR #1933 to add generation of `manylinux_2_28` Python wheels (alongside `manylinux2014`), this can be tested from test PyPI with version 2.4.0dev0 https://test.pypi.org/project/opencolorio/. I'm not exactly sure how to confirm it is indeed using the new API, but it seems that this new wheel doesn't contain symbol related to the old string implementation (std::__cow_string::__cow_string*) which is probably a good sign.

I tried to reproduce the issue using some of the latest Linux distro available as Docker images but couldn't find any problem yet, I assume pip wouldn't allow the download of incompatible binary wheels anyway? I guess the main problem occurs when trying to use the dynamic library inside the wheel to link against code using a different ABI. Or when trying to import multiple packages interacting into a Python script with incompatible ABIs? I'll try to look further into this.

Tagging @JeanChristopheMorinPerso in case you have more clues, is this something that has been discussed in the ci wg?

JeanChristopheMorinPerso:
[...]

PR #1933, opened 2024-01-21
Add manylinux_2_28 Python wheels

remia:
Add manylinux_2_28 Python wheels, following #1923.

From https://github.com/pypa/manylinux, the two currently active and non-EOL recommended environments (for which Docker images are provided) are manylinux2014 and manylinux_2_28. Hence why I used the latter here instead of manylinux_2_24.


doug-walker (approved):
Awesome, thank you Remi!

michdolan (approved):

PR #1935, opened 2024-01-24
Fix AVX instructions used on CPUs that don't support them

brechtvl:
It's supposed to check all bits are enabled, not just one of them. This causes a crash using OpenColorIO on older CPUs.

Thanks to Ray Molenkamp for help tracking this down.

### DIFF
--- a/src/OpenColorIO/CPUInfo.cpp
+++ b/src/OpenColorIO/CPUInfo.cpp
@@ -107,7 +107,7 @@ CPUInfo::CPUInfo()
             flags |= X86_CPU_FLAG_SSE42;
 
         /* Check OSXSAVE and AVX bits */
-        if (info.reg.ecx & 0x18000000)
+        if ((info.reg.ecx & 0x18000000) == 0x18000000)
         {
             xcr = xgetbv();
             if(xcr & 0x6) {
@@ -129,7 +129,7 @@ CPUInfo::CPUInfo()
 
         /* OPMASK/ZMM state */
         if ((xcr & 0xe0) == 0xe0) {
-            if ((flags & X86_CPU_FLAG_AVX2) && (info.reg.ebx & 0xd0030000))
+            if ((flags & X86_CPU_FLAG_AVX2) && ((info.reg.ebx & 0xd0030000) == 0xd0030000))
                 flags |= X86_CPU_FLAG_AVX512;
         }
     }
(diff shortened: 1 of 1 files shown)


brechtvl:
CC @markreidvfx

doug-walker:
Thank you for the fix @brechtvl !  Do you have any more information about specifically which CPUs are affected?

brechtvl:
This was based on a bug report from a user, we haven't heard back from them yet to know the exact CPU model. But it was a relatively old one that support instructions only up to SSE 4.1.

Still it's the AVX check that fails. So it wouldn't be surprising if CPUs that support up to SSE 4.2 are affected too, and that's the current minimum requirement for Blender, Maya and Houdini.

But I don't know the exact conditions for when that XSAVE bit is enabled while the AVX bit is not.

src/OpenColorIO/CPUInfo.cpp
```diff
@@ -107,7 +107,7 @@ CPUInfo::CPUInfo()
             flags |= X86_CPU_FLAG_SSE42;
 
         /* Check OSXSAVE and AVX bits */
-        if (info.reg.ecx & 0x18000000)
+        if ((info.reg.ecx & 0x18000000) == 0x18000000)
         {
             xcr = xgetbv();
             if(xcr & 0x6) {
```
markreidvfx, line 113: hmm, I think this should also be.

```
if((xcr & 0x6) == 0x6) 
```

https://en.wikipedia.org/wiki/Control_register#XCR0_and_XSS

bit 1 SSE (XSAVE feature set enable for MXCSR and XMM regs)
bit 2 AVX (AVX enable, and XSAVE feature set can be used to manage YMM regs)

doug-walker:
Linking this Blender issue, [which seems to be the original report](https://projects.blender.org/blender/blender/issues/117435), for reference.

markreidvfx (commented):
Good catch! That must have been tricky to track down. I think we also need to check both bits of the Extended Control Register register too.

markreidvfx (approved):
LGTM. We will probably want to backport this to 2.3.

doug-walker (approved):
Thank you @brechtvl for this fix!  My goal is to accelerate a 2.3.2 release due to this.

KelSolaar (approved):
LGTM! (Nasty one!)

remia (approved):

doug-walker:
@brechtvl , we noticed that Ray Molenkamp posted an OCIO_CPUINFO.exe in the Blender forum as a way of troubleshooting this. If you or he would like to contribute that to OCIO (e.g., under src/apps or share/troubleshooting), we would welcome that.

PR #1954, opened 2024-03-04
minizip seems to be changing to a hex number for MZ_VERSION_BUILD

opoplawski:
This hasn't been released yet but:

https://github.com/zlib-ng/minizip-ng/commit/a7d612adf210353ba0391e20d1d1abbb72d1ad83
https://github.com/zlib-ng/minizip-ng/issues/758

### DIFF
--- a/src/OpenColorIO/OCIOZArchive.cpp
+++ b/src/OpenColorIO/OCIOZArchive.cpp
@@ -226,7 +226,7 @@ void archiveConfig(std::ostream & ostream, const Config & config, const char * c
     std::string configStr = ss.str();
 
     // Write zip to memory stream.
-#if MZ_VERSION_BUILD >= 040000
+#if MZ_VERSION_BUILD >= 0x40000
     write_mem_stream = mz_stream_mem_create();
 #else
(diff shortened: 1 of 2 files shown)


doug-walker (commented):
Thanks for proposing a fix, but does it actually work with all releases of minizip-ng? Maybe we need to switch to parsing MZ_VERSION rather than trying to deal with their switching MZ_VERSION_BUILD from decimal to octal to hex?

markreidvfx:
I think we should parse `MZ_VERSION` instead too. `MZ_VERSION_BUILD` will be too unreliable going forward.

markreidvfx:
The cmake module that detects minizip does parse the [MZ_VERSION](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/91e8826a8bd78f5c6df5e5294e70671383186244/share/cmake/modules/Findminizip-ng.cmake#L161) from `mz.h` as `minizip-ng_VERSION`. So it should be just a matter of extracting the major version and creating a macro definition for it.

doug-walker:
Closing as this is no longer relevant after PR #2192.

PR #1962, opened 2024-03-30
Add a helper function to check if an extension is supported by file transform

### LINKED ISSUE #1938: Add a helper function to check if an extension is supported by FileTransform
The FileTransform class has several static methods such as GetFormatExtensionByIndex that will allow client programs to iterate over all of the supported formats to check if a given extension is supported by OCIO. However, it would be nice to have a helper function to just take the extension and return true/false.  Such as:

```
class FileTransform
    /// Returns true if the extension corresponds to a format supported by FileTransform.
    /// The argument is case-insensitive, and a leading dot, if present, is ignored.
    static bool IsFormatExtensionSupported(const char * ext);
```

meimchu:
resolves #1938 issue. Please let me know if this is what Doug is looking for!
My plan:
- [x] Write the function
- [x] Write C++ tests
- [x] Write python binding
- [x] Write python tests

(Hope my auto whitespace cleanup is ok. I recognize it can make the git history slightly more dirty.)


meimchu:
For the python side, I have a question about the python binding side as I am not very familiar with pybind11. I have the following in `PyFileTransform.cpp`.
```cpp
.def("IsFormatExtensionSupported", &FileTransform::IsFormatExtensionSupported, "extension"
    DOC(FileTransform, IsFormatExtensionSupported));
```
However, I get a static analysis error when trying to build with that. Any help and pointer about this is greatly appreciated.
```sh
error: static assertion failed due to requirement 'expected_num_args<pybind11::name, pybind11::is_method, pybind11::sibling, pybind11::arg, const char *>(sizeof...(Args), argument_loader<const char *>::args_pos >= 0, argument_loader<const char *>::has_kwargs)': The number of argument annotations does not match the number of function arguments
```

include/OpenColorIO/OpenColorTransforms.h
```diff
@@ -1070,6 +1070,9 @@ class OCIOEXPORT FileTransform : public Transform
     static const char * GetFormatNameByIndex(int index);
     /// Get the LUT reader extension at index, return empty string if an invalid index is specified.
     static const char * GetFormatExtensionByIndex(int index);
+    /// Returns true if the extension corresponds to a format supported by FileTransform.
+    /// The argument is case-insensitive, and a leading dot, if present, is ignored.
```
doug-walker, line 1074: It might be helpful to add something like the following:

/// Note that FileTransform will attempt all format readers on a given file until it is successful, even files that contain an unsupported extension or no extension. However, this function is useful for applications that want to know which files are likely to be LUT files, based on their extension.

doug-walker (changes requested):
Thanks Mei, this looks great!

For the Python binding, please try this:
```
.def_static("IsFormatExtensionSupported", &FileTransform::IsFormatExtensionSupported, "extension"_a
    DOC(FileTransform, IsFormatExtensionSupported))

```
Regarding the white space clean-up, let's bring this up at the TSC meeting to see if people object to that.

tests/cpu/transforms/FileTransform_tests.cpp
```diff
@@ -290,6 +290,17 @@ OCIO_ADD_TEST(FileTransform, format_by_index)
...
 
+OCIO_ADD_TEST(FileTransform, is_format_extension_supported)
+{
+    OCIO::FormatRegistry & formatRegistry = OCIO::FormatRegistry::GetInstance();
+    OCIO_CHECK_EQUAL(formatRegistry.isFormatExtensionSupported("foo"), false);
+    OCIO_CHECK_EQUAL(formatRegistry.isFormatExtensionSupported("bar"), false);
+    OCIO_CHECK_ASSERT(formatRegistry.isFormatExtensionSupported("cdl"));
+    OCIO_CHECK_ASSERT(formatRegistry.isFormatExtensionSupported(".cdl"));
+    OCIO_CHECK_ASSERT(formatRegistry.isFormatExtensionSupported("3dl"));
+    OCIO_CHECK_ASSERT(formatRegistry.isFormatExtensionSupported(".3dl"));
```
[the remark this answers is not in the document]
meimchu: I have added two additional tests for upper case letters as the initial character in the latest commit.

include/OpenColorIO/OpenColorTransforms.h
meimchu: I have appended those additional comments to those existing comments in the latest commit.

tests/cpu/transforms/FileTransform_tests.cpp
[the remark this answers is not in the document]
meimchu: I have also added in the python binding following your wonderful suggestions and put in the same tests for python side. Thank you.

(I shall look into the difference between .def() and .def_static() for pybind11 on my own time.)

src/OpenColorIO/transforms/FileTransform.cpp
```diff
@@ -522,6 +527,32 @@ const char * FormatRegistry::getFormatExtensionByIndex(int capability, int index
     return "";
 }
 
+bool FormatRegistry::isFormatExtensionSupported(const char * extension) const
+{
+    // Early return false with an input of just the dot.
```
remia, line 532: Suggest to also protect against invalid pointer:

```
if (!extension || !*extension)
{
    return false;
}
```

remia (commented):
Thank you @meimchu, for the white space changes, it needs to be in separated commit(s) and those listed in a .git-blame-ignore-revs at the root of OpenColorIO folder like described [here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view). I started looking at clang-format last week, so I might make a PR for that later as well, that would only cover the C++ files though.

src/OpenColorIO/transforms/FileTransform.cpp
meimchu: I have added those conditions for early return.

meimchu: I have also reverted all the whitespaces removed as part of my initial commits.

remia (approved):

doug-walker (approved):
LGTM!

PR #1963, opened 2024-04-09
Code formatting with clang-format

remia:
This adds a clang-format target, based on OpenImageIO implementation, to automatically format the source code with the .clang-format profile. Hoping to have feedback on the coding style to nail down any particular aspects that would currently be suboptimal. Here are a few notes:
* In some places, clang-format makes things worse I added some clang-format off / on around places I found relevant to preserve. Includes the GPU shader text based generation, SSE related code (including SSE.h as a whole), multiple places where matrices or image array data readability were getting negatively impacted.
* Unit tests are currently off the auto formatting as there would be too many places to manually protect for the time being.
* Variable initialization using initializers list alignment tends to get lost, for exemple [this](https://github.com/AcademySoftwareFoundation/OpenColorIO/compare/main...remia:OpenColorIO:clang-format#diff-1d6399bbe3c9289bfe361a75f78f001a5bdf363eca9c5bf11d8c0b997c0381fcL471) (OpenColorTransforms.h line 461), I don't think clang-format can do any better at the moment.
* Include order sorting is currently enabled and it broke compilation which I had to patch (for exemple when Platform.h gets included before OpenColorIO.h or gl.h before glew.h), can be disabled if we think it's safer.
* clang-format version matters and different version produce different results, we should probably pin the version in the CI. For developer, this means using the same version which is made easier with the Python wheel provided for clang-format.


remia:
Just rebased the PR and did some minor adjustments, this still very much follows OpenImageIO setup. I'm still ignoring the test suite for the time being, I believe this could be addressed in a follow up once we let some time pass for things to settle. PR is split into individual commits to easily separate out the auto formatting pass.

Include are automatically re-ordered, this will conflict with Kevin's work.

remia:
Added the output diff showing the formatting issue in the CI job, similar to how OIIO is doing it, with a test commit to trigger the issue on purpose.

PR #1965, opened 2024-04-12
build(windows): Do not try to build minizip before zlib.

SlawekNowy:


doug-walker:
@SlawekNowy , thanks for the PR. Would you please provide more detail about what scenarios require this fix? In other words, most Windows builds succeed already without this, so it must be something more specific about the way you are building.

SlawekNowy:
This is the local windows build. Fetched the repository and built it using Visual Studio generator. Both minizip and zlib are managed by opencolorio. Visual studio 2022 Community x64.

SlawekNowy:
Oh. I forgot to mention I run `--parallel` at cmake build step.

Also it seems that did not fix my problems?

SlawekNowy:
The build is done using `cmake ..` from `build` directory (from ocio dir), and then building generated project.

SlawekNowy:
I'll admit this is not standard setup, but this repo's CMakeLists does fetch deps...
And besides that is standard setup on Linux...

SlawekNowy:
From parent project's discord:
>ok cmake. You get a race condition without the deps' root set, but you don't if I do set the root directories of deps? what?

share/cmake/modules/FindExtPackages.cmake
```diff
@@ -114,6 +114,10 @@ ocio_handle_dependency(  minizip-ng REQUIRED ALLOW_INSTALL
                          MIN_VERSION 3.0.6
                          RECOMMENDED_VERSION 3.0.7
                          RECOMMENDED_VERSION_REASON "Latest version tested with OCIO")
+#minizip-ng requires zlib headers, so we force the dep.
+if(TARGET ZLIB::ZLIB AND TARGET MINIZIP::minizip-ng)
+add_dependencies(MINIZIP::minizip-ng ZLIB::ZLIB)
+endif()
```
doug-walker, line 120: @SlawekNowy , Stephen Yee's suggestion was to revert the changes here in FindExtPackages.cmake since it should be unnecessary after the change to Installminizip-ng.cmake.

doug-walker (changes requested):

doug-walker:
@SlawekNowy , thanks for the update, the code looks perfect now, but unfortunately it looks like you did not sign your latest commit?  The DCO sign-off check needs to pass before we're able to merge it.    Usually "git commit --amend -s" and then re-pushing will fix it.

SlawekNowy:
Closing in favor of #1986.

PR #1966, opened 2024-04-17
ocioview updates & fixes

michdolan:
This PR makes a number of updates and improvements to `ocioview`:

- When closing the application or creating a new config, users will now only be prompted to save their config if changes have been made. Previously the initial application state would be detected as a modified config. Note: config modification detection is limited to the data included in the config's cache ID, which does not currently include the config's name or description.
- Per suggestion from @KelSolaar in #1912 , changed the message router interface to use properties instead of getters and setters.
- Added a warning when loading a config that contains a view definition that references a view transform and non-display color space, which results in that view being interpreted as a OCIO v1 style view, dropping the view transform. Technically OCIO supports a view with a view transform and a scene reference color space, but in ocioview we intentionally limit color space selection to display color spaces when using a view transform.
- Improved consistency of icon sizing throughout application, and improved pixmap transformation quality for consistent icon filtering across tabs, buttons, and items.
- Added `ProcessorContext` interface to reference input and output processor parameters so application components that receive a viewer's processor can inspect the config items that created it. This dataclass includes the processor's: input color space name, output transform item type, output transform item name, and transform direction. This should resolve questions raised in #1914 .

### DIFF
--- a/src/apps/ocioview/ocioview/message_router.py
+++ b/src/apps/ocioview/ocioview/message_router.py
@@ -329,6 +329,54 @@ class MessageRouter(QtCore.QObject):
         """Forward unknown attribute requests to internal runner."""
         return getattr(self._runner, item)
 
+    @property
+    def gpu_language(self) -> ocio.GpuLanguage:
+        return self._runner.gpu_language
+
+    @gpu_language.setter
+    def gpu_language(self, gpu_language: ocio.GpuLanguage) -> None:
+        self._runner.gpu_language = gpu_language
+
+    @property
+    def config_updates_allowed(self) -> bool:
+        return self._runner.config_updates_allowed
+
+    @config_updates_allowed.setter
+    def config_updates_allowed(self, allowed: bool) -> None:
+        self._runner.config_updates_allowed = allowed
+
+    @property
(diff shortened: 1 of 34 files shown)


src/apps/ocioview/ocioview/message_router.py
```diff
@@ -156,14 +169,19 @@ def start_routing(self) -> None:
                     self._handle_config_message(msg_raw)
 
             # OCIO processor
-            elif isinstance(msg_raw, ocio.Processor):
-                self._prev_cpu_proc = msg_raw
+            elif (
```
KelSolaar, line 172: Feels a bit "magical" :)

KelSolaar (approved):
Thanks @michdolan, LGTM!

doug-walker (approved):

src/apps/ocioview/ocioview/message_router.py
KelSolaar: I was talking about this block specifically:

```python
                isinstance(msg_raw, tuple)
                and len(msg_raw) == 2
                and isinstance(msg_raw[0], ProcessorContext)
                and isinstance(msg_raw[1], ocio.Processor)
```

PR #1974, opened 2024-05-07
build(windows):Add msvs parameter to ocio_deps.bat for ocio.bat argument parity

SlawekNowy:


share/dev/windows/ocio_deps.bat
```diff
@@ -103,20 +109,24 @@ if ErrorLevel 1 (
...
 )
 
-if !MSVS!==0 (
-    echo No Microsoft Visual Studio installation was found in !programfiles!.
-    echo For non-standard installation path, please use the following option:
-    rem The double dash are in quote here because otherwise the echo command thow an error. 
-    echo "ocio_deps --vs <path>"
+rem ****************************************************************************************************************
+rem Setting up the environment using MS Visual Studio batch script
+set VCVARS64_PATH="!MSVS_PATH!\vcvars64.bat"
```
remia, line 122: Looking at ocio.bat, I see that the vcvars64 script is called [later](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/share/dev/windows/ocio.bat#L153) to initialize the environment. Is this needed here as well?

SlawekNowy: Does `ocio.bat` call `ocio_deps.bat`? if not, then this is needed.

SlawekNowy:
Might be, thing is nowadays we would be using vswhere either way.
Closing in favor to #1980.

PR #1976, opened 2024-05-11
Adsk Contrib - Allow read and write of file rules without color space validation

doug-walker:
It is generally possible to read and write configs that are syntactically valid but which are incomplete and hence would not pass validation. (For example, it is currently possible to serialize or load a config without any displays and views and would hence fail validation.) This is useful in scenarios where one wants to serialize configs that are only partially complete.

Currently, OCIOYaml.cpp is doing some inappropriate validation during config serialization and parsing regarding the default role and the FileRules Default rule. This PR removes those checks from OCIOYaml.cpp but adds unit tests to confirm that config->validate would still catch these problems. 

This gives the calling program the flexibility to serialize a partial config, for example that has a role that refers to a color space that may not have been added yet. Of course, before any config is used the caller should call validate, which is normally where problems of this type are supposed to be surfaced.

### DIFF
--- a/src/OpenColorIO/OCIOYaml.cpp
+++ b/src/OpenColorIO/OCIOYaml.cpp
@@ -4034,7 +4034,7 @@ inline void load(const YAML::Node & node, ViewingRulesRcPtr & vr)
     catch (Exception & ex)
     {
         std::ostringstream os;
-        os << "File rules: " << ex.what();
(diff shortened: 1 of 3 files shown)


remia (approved):

cozdas (approved):
Looks good

PR #1980, opened 2024-05-29
Adsk Contrib - Improve Windows batch files to find installed Visual Studio location

cozdas:
- Instead of assuming a hard-coded visual studio installation path, we're now using the vswhere.exe tool that the visual studio installs in every system. This removes the necessity to specify the msvc path in the command line if visual studio was not installed into the default path. Option to override is still possible as before.


doug-walker (approved):

KelSolaar (approved):
LGTM!

PR #1981, opened 2024-05-30
Adsk Contrib - Modify half-domain LUT1D GPU shader to improve zero handling

doug-walker:
Made a small improvement in the GPU shader for half-domain LUT1D ops.

The half-domain LUT1D is structured like a half-float and so it has two entries for zero (positive and negative). The shader was looking up positive 0 at the negative 0 entry. Normally, a half-domain LUT1D should set +0 and -0 to the same value, so in practice this should not be an issue, but it is now more technically correct. I added a unit test to verify this (it failed prior to the shader adjustment, the GPU matched the CPU at -0 rather than +0).

In addition, I clarified the code that scales denorms in that shader. This is numerically equivalent but might be somewhat clearer to read.

Finally, I added an unrelated but potentially useful unit test to verify that the cacheID of a Processor containing a LUT3D changes if one of the LUT entries changes.


cozdas (approved):
looks good.

PR #1982, opened 2024-05-30
Adsk Contrib - Add cmake option to turn off half-float lookup table in Imath

cozdas:
Adding OCIO_USE_HALF_LOOKUP_TABLE cmake option. If set to OFF, it adds IMATH_HALF_NO_LOOKUP_TABLE compiler definition which in return turns off the look up table that's used for half-float to float conversion, shaving off 256kB of binary size and run time footprint.

Please see half.h for the details of the half->single conversion table: https://github.com/AcademySoftwareFoundation/Imath/blob/main/src/Imath/half.h#L116-L152


doug-walker:
@KevinJW , thanks for the feedback. I had thought that this option may have some general utility, but after further discussion, Cuneyt and I decided we will keep this change confined to the nanoColor branch for now.  So I'm closing this PR.

Issue #1956, opened 2024-03-14
Add support for RGBA format in GpuShaderCreator for better Metal support?


wRosie:
Hi everyone, 

Only single channel and RGB texture formats are currently supported in `GpuShaderCreator::TextureType.` However, it appears that Apple does not support simple RGB format textures in Metal. Please refer to https://developer.apple.com/documentation/metal/mtlpixelformat. 

Working with OCIO shaders in Metal becomes inconvenient. When I query the GpuShaderDesc for a Metal shader and its textures, I sometimes get RGB textures that are not accepted by Metal. I have to insert a placeholder alpha channel to convert them to RGBA format. 

Is it an intentional decision that OCIO does not support RGBA texture? If not, is it a plan for the future?

Thanks a lot!

doug-walker:
@Morteeza, since you are our Metal expert, could you please offer some help with this question?

Morteeza:
As you mentioned, we don't have RGB texture support in Metal. You may have noticed we similarly insert placeholder alpha channel in OCIO [metal app.mm](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/libutils/oglapphelpers/metalapp.mm#L69), too.

I believe main reason we didn't add RGBA texture type was that, it will touch OCIO interface. So, maybe in future we can add that? @hodoulp do you think we can make this change in a major release?

wRosie:
I am happy to help fix it. This will save us all from copying textures in memory. 
I agree it will touch the OCIO interface. But.. When I query for a metal shader, it makes more sense to get textures that work with it, right? Please consider doing it in a major release.

doug-walker:
@wRosie , thank you for your offer to help, that would be great! This is something we could include in the OCIO 2.4 release this fall.  I'd suggest starting by just proposing the API change, to get feedback, before doing the full implementation.

wRosie:
Thanks @doug-walker. Here are some thoughts on interface change.

- Add TEXTURE_RGBA_CHANNEL value to `enum TextureType`

- Add a TextureType field to `add3DTexture()`. We could make the field optional, with a default value of TEXTURE_RGB_CHANNEL (the current behavior). This would avoid breaking existing function calls. However, I prefer adding the field to keep the signature consistent with `addTexture()`. 

- Add a `TextureType&` field to get3DTexture(). 

[...]

PR #1984, opened 2024-06-06
Add RGBA format in GpuShaderCreator for better Metal support (Issue #1956)

wRosie:
Hi OCIO team, 

As per the discussion in #1956, I added RGBA texture format in GpuShaderCreator and made GpuShaderCreator store LUTs in RGBA when the shader language is in metal.

Here is the interface and behavior change involved in this PR:

1. AddTEXTURE_RGBA_CHANNEL value to enum TextureType

2. Add a TextureType parameter to add3DTexture(). The signature is consistent with addTexture().

3. Add a TextureType& field to get3DTexture().

The addTexture()/getTexture() functions already take TextureType as a parameter, so I will not change their function signature. However, I updated their behavior -- I changed GPUShaderCreator to use RGBA formats when GPULanguage is Metal. That involves updating functions such as GetLut1DGPUShaderProgram() and GetLut3DGPUShaderProgram(). Also, getTexture() and get3DTexture() will return textures in RGBA type in Metal mode.

Additionally, I updated the tests and function calls to work with the new interface.

I am aware that the behavior is not tested on a per-platform basis, and I would love suggestions on that. If possible, I'd like to add tests to ensure that RGBA is used only in metal mode. 

Thanks!


doug-walker:
Thank you for the PR @wRosie !   Have you started the process of getting the corporate CLA agreement signed? Please don't hesitate to reach out on Slack if you have any questions about that.

src/OpenColorIO/GpuShaderUtils.cpp
```diff
@@ -222,6 +222,28 @@ std::string getMatrixValues(const T * mtx, GpuLanguage lang, bool transpose)
     return vals;
 }
 
+
+void RGBtoRGBATexture(const float* lutValues, int valueCount, std::vector<float>& float4AdaptedLutValues){
```
remia, line 226: Minor, please add the opening { on a newline. Until we get clang-format!

src/OpenColorIO/ops/lut1d/Lut1DOpGPU.cpp
```diff
@@ -155,27 +155,51 @@ void GetLut1DGPUShaderProgram(GpuShaderCreatorRcPtr & shaderCreator,
...
+        channel = GpuShaderCreator::TEXTURE_RED_CHANNEL;
+        break;
+    case 3:
         CreatePaddedLutChannels(width, height, lutData->getArray().getValues(), values);
+        channel = GpuShaderCreator::TEXTURE_RGB_CHANNEL;
+        break;
+    case 4: {
+        std::vector<float> paddedChannels;
+        values.reserve(width * height * 3);
+        CreatePaddedLutChannels(width, height, lutData->getArray().getValues(), paddedChannels);
```
remia, line 192: Maybe we can avoid the extra copy here by adding a parameter in `CreatePaddedLutChannels()` to directly create a 4 channels array? But as this is only 1D LUTs and only done at initialization, can probably stay as is.

src/OpenColorIO/ops/lut1d/Lut1DOpGPU.cpp
```diff
@@ -155,27 +155,51 @@ void GetLut1DGPUShaderProgram(GpuShaderCreatorRcPtr & shaderCreator,
...
-    {
+        channel = GpuShaderCreator::TEXTURE_RED_CHANNEL;
+        break;
+    case 3:
         CreatePaddedLutChannels(width, height, lutData->getArray().getValues(), values);
+        channel = GpuShaderCreator::TEXTURE_RGB_CHANNEL;
+        break;
+    case 4: {
+        std::vector<float> paddedChannels;
+        values.reserve(width * height * 3);
```
remia, line 191: Should call `paddedChannels.reserve()` instead?

remia (commented):
Thanks for the contribution @wRosie and sorry for the delay in reviewing.

I triggered the CI workflows, please ignore the macOS stalled jobs, we use a macOS version no longer available in Github and will need to update the workflows shortly. Unfortunately, I don't think we have the Metal tests running in Github currently, you can try running them on your side and using ociodisplay in metal mode as well (use `-metal` flag).

I believe you will need to update the msl.mm file and remove the `RGB_to_RGBA` conversions after this change to have it work?

I think we are missing a Python binding updates, in PyGPUShaderCreator.cpp, you need to add the new textureType in enumTextureType.

I don't think we have GPU unit tests dedicated to specific APIs for now, might be a good addition, making sure Metal always returns Red or RGBA LUTs only.

src/OpenColorIO/GpuShaderUtils.cpp
wRosie: Sure. Fixed this one and other '{'s in a later commit

src/OpenColorIO/ops/lut1d/Lut1DOpGPU.cpp
wRosie: Oops. My bad. Thanks for catching it!

src/OpenColorIO/ops/lut1d/Lut1DOpGPU.cpp
wRosie: Thanks for the suggestion! I once considered it when I wrote up the code. Here's my consideration.
`CreatePaddedLutChannels()` takes advantage of std::transform() to do the padding. To my understanding, this lambda style runs fast for one-to-one output. However, if I want to create the 4-channel array at initialization, I can't think of a way to take advantage of std::transform(). It is hard to say which will be faster -- "run a fast padding function and do a copy" vs "run a slightly slower padding function and insert the alpha channel at the same time". 
Anyway, the choice may not make much difference. As you said, It is a 1D lut and is only done at initialization.

wRosie (commented):
Thank you for the review, and I apologize for the delayed response. 
Please let me know if you have additional comments. Thanks!

@doug-walker I am aware of the CLA approval issue. Thank you for the patience here, I am actively tracking it down -- it is just a lengthy process :(

src/libutils/oglapphelpers/msl.mm
```diff
@@ -210,8 +210,9 @@ void RGB_to_RGBA(const float* lutValues, int valueCount, std::vector<float>& flo
         const char * textureName = nullptr;
         const char * samplerName = nullptr;
         unsigned edgelen = 0;
+        GpuShaderDesc::TextureType channel = GpuShaderDesc::TEXTURE_RGBA_CHANNEL;
         Interpolation interpolation = INTERP_LINEAR;
-        m_shaderDesc->get3DTexture(idx, textureName, samplerName, edgelen, interpolation);
+        m_shaderDesc->get3DTexture(idx, textureName, samplerName, edgelen, channel, interpolation);
```
doug-walker, line 216: GitHub won't let me leave a comment at that line, but on line 233, AllocateTexture3D seems to assume it is getting an RGB array. It calls RGB_to_RGBA on it. Isn't the idea that this should not be necessary anymore?

Are the Metal GPU unit tests passing on your local build?

doug-walker:
@wRosie , just wanted to check in with you about this PR.  Given that this would be an API-breaking change, it would need to be included in OCIO 2.4.0 which will be released on Sept. 30.  Do you think you will be able to make the requested changes in time?  Otherwise it will likely have to wait for OCIO 2.5.0 next fall.

src/OpenColorIO/ops/lut3d/Lut3DOpGPU.cpp
```diff
@@ -38,12 +37,30 @@ void GetLut3DGPUShaderProgram(GpuShaderCreatorRcPtr & shaderCreator, ConstLut3DO
     {
         samplerInterpolation = INTERP_NEAREST;
     }
-    // (Using CacheID here to potentially allow reuse of existing textures.)
-    shaderCreator->add3DTexture(name.c_str(),
+
+    if(shaderCreator->getLanguage() == GPU_LANGUAGE_MSL_2_0){
```
remia, line 41: [Minor] Brace opening should be on the next line.

remia (commented):
Thanks for the follow up @wRosie!

There are still a couple of things to address (see the code comments as well):
- Can you please add the new Texture type to the [Python bindings](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/bindings/python/PyGpuShaderCreator.cpp#L120)?
- The unit tests for Metal currently Segfault, having a deeper look, this is due to the internal copy of the LUT array within the [GPUShader](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/GpuShader.cpp#L34) not handling RGBA yet.

I have another issue on Metal test locally but it is only happening for the first test being run and is due to an OpenGL setup error where a [non existing texture is attached to a framebuffer](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/libutils/oglapphelpers/oglapp.cpp#L183), simply initialising m_imageTexID to 0 and checking before calling glFramebufferTexture2D seems to fix it. But we can address this in a later PR and review if the Metal / OpenGL interactions are correctly setup.

Issue #1640, opened 2022-05-08
[documentation][python] GradingPrimary: missing expected value type doc


MrLixm:
Hello,
When looking at the GradingPrimary class documentation (https://opencolorio.readthedocs.io/en/latest/api/grading_transforms.html),
we can see a bunch of property is available but none of them specifies which type of value is expected when **set**.

This would have not been an issue if they all except the same type but seems some except a `GradingRGBM` instance while others just expect a `float`.

Would be cool to have in the documentation what type is expected.

Here are the tests I ran to check who accepts what

```python
import unittest

import PyOpenColorIO as ocio


class OcioGradingPrimary(unittest.TestCase):
    @staticmethod
    def to_rgbm(v):
        rgbm = ocio.GradingRGBM()
        rgbm.master = v
        return rgbm

    def test_properties(self):

        gp = ocio.GradingPrimary(ocio.GRADING_LIN)

        gp.brightness = self.to_rgbm(0.18)
        gp.clampBlack = 0.18
        gp.clampWhite = 0.18
        gp.contrast = self.to_rgbm(0.18)
        gp.exposure = self.to_rgbm(0.18)
        gp.gain = self.to_rgbm(0.18)
        gp.gamma = self.to_rgbm(0.18)
        gp.lift = self.to_rgbm(0.18)
        gp.offset = self.to_rgbm(0.18)
        gp.pivot = 0.18
        gp.saturation = 0.18

        gp.brightness = self.to_rgbm(0.18)
        with self.assertRaises(TypeError): gp.clampBlack = self.to_rgbm(0.18)
        with self.assertRaises(TypeError): gp.clampWhite = self.to_rgbm(0.18)
        gp.contrast = self.to_rgbm(0.18)
        gp.exposure = self.to_rgbm(0.18)
        gp.gain = self.to_rgbm(0.18)
        gp.gamma = self.to_rgbm(0.18)
        gp.lift = self.to_rgbm(0.18)
        gp.offset = self.to_rgbm(0.18)
        with self.assertRaises(TypeError): gp.pivot = self.to_rgbm(0.18)
        with self.assertRaises(TypeError): gp.saturation = self.to_rgbm(0.18)


if __name__ == "__main__":
    unittest.main()
```

Cheers.
Liam.

sumana-2705:
@MrLixm  can you help me where can I get to know more about this documentation to gain knowledge in type of expected value? I want to work on this

[...]

PR #1989, opened 2024-07-01
Updated documentation Specifying types for GradingPrimary attributes #1640

sumana-2705:
# Updated Documentation: Specify Types for GradingPrimary Attributes

## Description:
This pull request updates the documentation for the GradingPrimary class in OpenColorIO to specify the expected types for its attributes, enhancing clarity for developers mentioned in the issue #1640 

Changes Made:
Added type specifications for the following attributes:
- brightness: `GradingRGBM`
- clampBlack: `float`
- clampWhite: `float`
- contrast: `GradingRGBM`
- exposure: `GradingRGBM`
- gain: `GradingRGBM`
- gamma: `GradingRGBM`
- lift: `GradingRGBM`
- offset: `GradingRGBM`
- pivot: `float`
- saturation: `float`

### DIFF
--- a/docs/api/python/src/pyopencolorio_gradingprimary.rst
+++ b/docs/api/python/src/pyopencolorio_gradingprimary.rst
@@ -37,5 +37,11 @@
    .. attribute:: pivot
       :type: float
 
+   .. attribute:: pivotBlack
+      :type: float
+
+   .. attribute:: pivotWhite
+      :type: float
+
    .. attribute:: saturation
       :type: float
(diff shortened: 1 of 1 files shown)


doug-walker (approved):

cozdas (commented):
Do we want to include pivotWhite and pivotBlack in the list as well?

sumana-2705:
While I was running initial tests mentioned in the issue, it did not mention pivotWhite and pivotBlack, which is why they were not included in my first response. However, considering your feedback, I think it is reasonable to assume that these properties also follow the pattern observed for similar properties like `pivot`. If that's not the case, please let me know.

cozdas (approved):
looks good

PR #1990, opened 2024-07-06
Fix for GitHub Actions node.js glibc issue

doug-walker:
The Actions runners have stopped working for the Linux Docker containers for years 2022 and earlier due to a change on the GitHub side related to Node.js updates.

The proposed work-around is to continue to use the older actions/checkout@v3 and set the env vars in this PR to use the older node version.

For reference:
https://github.com/actions/runner/issues/2906#issuecomment-2209653912
https://academysoftwarefdn.slack.com/archives/C0169RX7MMK/p1720211076837729

In addition, there is another CentOS-related problem (a yum mirror was shut down) which is preventing the Linux builds from working if the documentation support is on, so I modified the installer script to work around that.

Finally, the macos-11 runners no longer work, so I updated to the macos-12 runners. (The ARM macos-14 runners were not modified.)

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -59,7 +59,7 @@ jobs:
       # Source: https://github.com/AcademySoftwareFoundation/aswf-docker
       image: aswf/ci-ocio:${{ matrix.vfx-cy }}
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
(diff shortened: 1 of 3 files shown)


remia (approved):

cozdas (approved):

doug-walker:
The macOS 11 checks are now obsolete, so merging regardless of those checks.

PR #1991, opened 2024-07-09
Correction for build with minizip-ng-3.0.10

1div0:
This PR fixes the build with minizip-ng-3.0.10-7.fc40 package.


doug-walker:
@1div0 , thanks for the contribution, but this fix seems like it would not work with the previous releases of minizip that OCIO still supports.  The PR would need to be updated to handle both scenarios before we could merge it.

cozdas:
Now that we've dropped the support for v3.x of minizip-ng, I think we can close this PR.

PR #2006, opened 2024-07-22
ocioview: app mode, tests, color space menus, bug fixes

michdolan:
This PR includes the following list of `ocioview` updates:

- Add "Application Mode" with two options: `Edit` and `Preview`. `Edit` is the default mode, which provides the existing config authoring interface. `Preview` is a new mode for previewing the config in a UX reference integration. This PR includes a new preview viewer with a standard display/view selection interface, which is currently the only change when switching to `Preview` mode. Future work will add new interfaces for demonstrating OCIO UX best practices and for validating config behavior in a typical DCC. The `Preview` mode is being designed by the UX Working Group.
image
image

- Add unit testing framework that leverages `pytest` and `pytest-qt` for automated testing of the `ocioview` application. This PR includes three initial tests for the color space edit interface to prove its functionality, but there remains much work to do in building out a test suite for broader coverage, which I think is what this application will need to be out of beta.
image

- Add new color space combo box which leverages `ColorSpaceMenuHelper` to provide color spaces organized by family, along with color space roles. This new widget replaces all color space selection throughout the application, and will be used in the new `Preview` mode to illustrate different color space selection use cases.
image

- Add a `SignalRouter` interface for fast application-wide thread-safe communication of specific config changes, like when color spaces or roles change, or when the config is reloaded. This router doesn't route any message content like the `MessgeRouter`, making it simpler and more performant. Both have useful application, but the new router will likely grow in scope while the message router focuses on cases where data needs to be modified in a thread before being delivered to subscribers.

- Several bug fixes and stability improvements.

There is much more work to do on some of these new features, but this update aims to complete some foundational work for these changes.


michdolan:
Note: I ran Black reformatting on the full application, and since our `pyproject.toml` configuration overrides the default 88 character line length down to 79, many source files have updates with line wrap reformatting.

src/apps/ocioview/ocioview/setup.py
```diff
@@ -0,0 +1,81 @@
...
+    font = app.font()
+    font.setPointSize(8)
+    app.setFont(font)
+
+
+def setup_app(
+    app: Optional[QtWidgets.QApplication] = None,
+) -> QtWidgets.QApplication:
+    """Create and configure QApplication."""
+    # Setup environment
```
KelSolaar, line 71: As discussed on Slack, we should probably disable the native macOs menu here otherwise the new Preview mode button does not show:

```python
    # Disabling macOs native menus
    QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_DontUseNativeMenuBar)
```

KelSolaar: (We will have to fiddle with menu text alignment a bit but it does work this way at least.)

michdolan: I moved the mode menu to the config dock so it will no longer impact the main menu:

image

michdolan:
I recombined the viewer types which I had previously split out per application mode, so now there is one viewer widget which adapts to the current mode. This should resolve issues with loaded images behaving inconsistently when switching mode.

doug-walker (approved):

KelSolaar:
Tested, it works for me!

KelSolaar (approved):

Issue #2010, opened 2024-07-28
OCIOView: The Display / View Transform processor should consider optionally clamp negative values.


KelSolaar:
This one took me a few hours to understand what is going on: The ACES Builtin transforms do not clamp which is fine. However, Nuke `OCIODisplay` node does but OCIOView does not which can generate unpredictable results, especially in the Chromaticity Inspector.

![image](https://github.com/user-attachments/assets/859e22a5-1ccd-450d-9aa6-8576da62cd6f)

We should probably add some options to clamp values between 0 and 1.

PR #2023, opened 2024-08-16
PR: Implement clamping support to chromaticity inspector.

KelSolaar:
I added some new buttons to clamp the chromaticity inspector chromaticities by default in range [0, 1]. This helps giving some control over the transformation chain and alleviate some problems pointed out in https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/2010.

**No Clamp**

image

Note how the negative values exceed the boundaries of sRGB.

**Clamped**

image

### DIFF
--- a/src/apps/ocioview/ocioview/inspect/chromaticities_inspector.py
+++ b/src/apps/ocioview/ocioview/inspect/chromaticities_inspector.py
@@ -100,13 +100,31 @@ class ChromaticitiesInspector(QtWidgets.QWidget):
             ".QLabel { font-size: 10pt;qproperty-alignment: AlignCenter;}"
         )
 
-        self._chromaticities_color_spaces_label = QtWidgets.QLabel(
+        self._chromaticities_color_space_label = QtWidgets.QLabel(
             "Chromaticities Color Space"
         )
-        self._chromaticities_color_spaces_combobox = QtWidgets.QComboBox()
-        self._chromaticities_color_spaces_combobox.setToolTip(
-            "Chromaticities Color Space"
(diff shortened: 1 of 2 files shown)


doug-walker:
@KelSolaar , I'm looking at the second screenshot you posted but I don't see the new button. This has the potential for confusion if ocioview is doing something different than applications that just apply the config "as is". My advice would be:
1. Make the UI clearly indicate when a clamp is being applied.
2. Have the default be "no clamp".

KelSolaar:
Hello @doug-walker,

The buttons are here, clamping negative and positive values: 
image

This only affects the chromaticity inspector and is enabled by default, like with DCC typically, if we want something application wide for the Display Transforms, we will need to discuss about it.

Note that DCC do not tell you when they clamp, they do it, here we can toggle the behaviour for inspection.

Cheers,

Thomas

doug-walker:
Ah, thanks for pointing out the buttons @KelSolaar !  It was not obvious to me what those did. Would it be clearer to change them to "< 0" and "> 1" rather than just "<" and ">"?

Regarding the default, yes, if the DCC needs to send the values to a display eventually they will become integers and lose the extended range values. However, not all DCCs will insert an explicit clamp. Furthermore, not all DisplayViewTransform output will go to a display, for example, it could go into a file format which may or may not clamp. Even the integer video file formats often allow some amount of extended range. That was my thinking for making the default "no clamp".

That's my opinion, but I will defer to you and Michael on all of this.

michdolan (approved):
Awesome! I think since there are tooltips the buttons are ok as-is for now. We can revisit later with dedicated icons if needed.

PR #2024, opened 2024-08-16
Fix typo where apostrophe was used in forming possessive of 'it'

JGoldstone:
The actual purpose of this is not to fix the documentation typo (though of course it should be remedied), but rather to trigger a CLA warning so I can EasyCLA my way onto the project as an individual contributor; I haven't worked for ARRI since 15 July.


doug-walker (approved):
Thank you Joseph!

I know we may count on you to always be on the lookout for [Word Crimes](https://youtu.be/8Gv0H-vPoDc?feature=shared&t=80).   ;)

remia (approved):

Issue #1992, opened 2024-07-18
Make LUT-free implementations of certain built-in transforms


doug-walker:
The nanoColor project will not be using LUTs, but does want to support all built-in transforms. However, some of the built-in transforms are implemented using a Lut1D. It would be helpful to replace these with FixedFunctions. Some judgement will be necessary to decide what aspects of these to expose as editable parameters that may be useful in the future. In some cases, they may be more appropriate as LogCameraTransform enhancements rather than fixed functions. 

Here is the list of affected built-in transforms:

• ACES
                ○ ACEScc_to_ACES2065-1
                ○ ADX10_to_ACES2065-1
                ○ ADX16_to_ACES2065-1
                ○ ACES-OUTPUT - ACES2065-1_to_CIE-XYZ-D65 - SDR-CINEMA-D60sim-DCI_1.0
                ○ ACES-OUTPUT - ACES2065-1_to_CIE-XYZ-D65 - SDR-CINEMA-D65sim-DCI_1.1
• Apple Camera
                ○ APPLE_LOG_to_ACES2065-1
                ○ CURVE - APPLE_LOG_to_LINEAR
• Canon
                ○ CANON_CLOG2-CGAMUT_to_ACES2065-1
                ○ CURVE - CANON_CLOG2_to_LINEAR
                ○ CANON_CLOG3-CGAMUT_to_ACES2065-1
                ○ CURVE - CANON_CLOG3_to_LINEAR
• Displays
                ○ CURVE - ST-2084_to_LINEAR
                ○ CURVE - LINEAR_to_ST-2084
                ○ DISPLAY - CIE-XYZ-D65_to_REC.2100-PQ
                ○ DISPLAY - CIE-XYZ-D65_to_ST2084-P3-D65
                ○ DISPLAY - CIE-XYZ-D65_to_REC.2100-HLG-1000nit

doug-walker:
The items in the above list under Apple, Canon, and Displays are now done as of PR #1992.

Issue #1968, opened 2024-04-26
Make display EOTFs in built-in transforms mirrored and unclamped to preserve sub-black and super-white


nick-shaw:
In broadcast scenarios it is necessary for all processing to pass both sub-black and super-white values. BT.2111 test patterns, for example, include PLUGE, and the ramp goes from -7 to 109%. Broadcast engineers expect conversion to pass these values, so a config for broadcast use cannot currently be created using the built-in XYZ-D65 to display transforms.

doug-walker:
@nick-shaw could you please review PR#2029 to make sure the math is what you expect for the PQ and HLG functions?

nick-shaw:
The PQ and HLG curves appear to be correctly mirrored about zero to the best of my understanding (see my comment in #1969 regarding the OOTF). However since `LINEAR_TO_PQ(0.0) != 0.0` I am not sure if there is a standard way to remove the discontinuity resulting from a simple mirroring. I have in the past used an approach similar to the ACES "stretch black" to join a mirrored PQ curve at zero. But there may be other approaches that I am not aware of.

doug-walker:
@nick-shaw , thanks for having a look.  Regarding your comment about 0, LINEAR_TO_PQ(0.0) does equal zero, so I'm not following you about a discontinuity at 0.

nick-shaw:
By my calculations is close to zero but not zero. 
![image](https://github.com/user-attachments/assets/4890f027-4f59-4d0d-8de5-d8a248aa1aeb)
For input of zero the PQ value is `c_1^m_2`.

doug-walker:
@nick-shaw , what you wrote is true, but that is a PQ value of 0.836^78.84 = 7.36e-07.  A single 16-bit code value in PQ would be 1/65536 = 1.53e-05.   So we're talking about a value in PQ that is over 20x smaller than a single 16-bit code value.  To me, that is 0, for all practical purposes.  Are there any real-world situations you're aware of where that causes a problem?

nick-shaw:
No. I am not aware of any actual problem. I am just a pedant, for which I apologise!

doug-walker:
Ha ha, don't apologize Nick!  You raised a valid point.  In fact, you made me realize that the CPU and GPU paths are giving slightly different results at 0 due to how the mirroring is done. It is way below visual threshold, but it still warrants a comment in the code to make people aware of the issue.  So thank you again for raising the question.

[...]

Issue #1705, opened 2022-10-24
Allow Builtin Transforms to be saved (without expansion) in CTF


doug-walker:
Currently, Builtin Transforms are expanded into the underlying op pieces before being saved to CTF.  Although this is exactly equivalent in terms of the numerical results, it could obfuscate the original contents of a GroupTransform.  

For example, in the ACES Config generator, it would be convenient to be able to write a Builtin to the CTF, so that the generator could more simply create a Builtin entry in the config's Yaml.  (A specific case is with the Canon Log, which becomes a Lut1D when the Builtin is expanded.)

However, in other cases, people may prefer the current behavior.  For example, when debugging processing issues, one wants to see the actual ops that a Builtin Transform would be expanded into.  So ideally there should be a way to obtain either behavior.

zachlewis:
This is a great idea. Maybe we could use an optimization flag for governing expansion behavior? 

If one wants to see the actual ops, it shouldn't be too difficult to derive the expanded ops from the collapsed form with the API.

PR #2029, opened 2024-08-27
Adsk Contrib - Issues #1968 (mirrored) and #1992 (LUT-free builtins)

cozdas:
### For issue #1992 - "Make LUT-free implementations of certain built-in transforms"
- Adding fixed-function FIXED_FUNCTION_LIN_TO_PQ. This will be used to handle following built-in transforms if the LUT support is turned off: 
-- CURVE - ST-2084_to_LINEAR
-- CURVE - LINEAR_to_ST-2084
-- DISPLAY - CIE-XYZ-D65_to_REC.2100-PQ
-- DISPLAY - CIE-XYZ-D65_to_ST2084-P3-D65

- Adding fixed function FIXED_FUNCTION_LIN_TO_GAMMA_LOG. This is a parametrized gamma + log segment curve with adjustable mirroring point. This will be used to handle following built-in transform if the LUT support is turned off: 
--CURVE - HLG-OETF
--CURVE - HLG-OETF-INVERSE
-- DISPLAY - CIE-XYZ-D65_to_REC.2100-HLG-1000nit
-- CURVE - APPLE_LOG_to_LINEAR
-- APPLE_LOG_to_ACES2065-1

- Adding fixed function FIXED_FUNCTION_LIN_TO_DOUBLE_LOG which has two log-affine segments with an optional linear segment in-between. This will be used to handle following built-in transform if the LUT support is turned off: 
-- CANON_CLOG2-CGAMUT_to_ACES2065-1
-- CURVE - CANON_CLOG2_to_LINEAR
-- CANON_CLOG3-CGAMUT_to_ACES2065-1
-- CURVE - CANON_CLOG3_to_LINEAR

- Implemented CPU renderers for scalar, SSE2 (with fastPower) and SSE2 (with Intel SVML) intrinsics targets for the new PQ fixed-function. The other two fixed-functions are implemented in the basic form and will need more optimization in the future when we introduce the option to turn the LUT support off.

- Implemented GPU shader generator for all the new fixed-functions,

### For issue #1968 - "Make display EOTFs in built-in transforms mirrored and unclamped to preserve sub-black and super-white"
- Both the existing LUT-based and the new LUT-free implementations using PQ or HLG curves are now unclamped at both ends and mirrored around the origin.
- Added new built-in transforms:
--CURVE - HLG-OETF
--CURVE - HLG-OETF-INVERSE

### Aux Changes
- GetFixedFunctionCPURenderer() now takes a new bool parameter fastLogExpPow.
- Added a stand-in OCIO_LUT_SUPPORT preprocessor macro in preparation for ocio-lite where the lut support can be turned off.
- Added util functions to Config_tests.cpp to help creating configs with arbitrary version.
- Added ability to test fixed-function cpu renderers with and without fastLogExpPow.
- Added capability to specify custom extended ranges in the GPU unit tests. Default value is [-1.0,2.0] same as the previously hard-coded range.
- Fixed a bug in the GPU unit tests where the computed domain values would overshoot.
- In BuiltinTransform_tests it's now possible (and necessary) to specify the error threshold for each built-in transform. This was needed as the hard-coded 1e-6 threshold was too tight for the PQ curve which uses nested power functions and thus numerical stability was depending on the implementation.
- Added GpuShaderText::floatXGreaterThanEqual() functions


src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -1177,15 +1219,217 @@ void Renderer_LUV_TO_XYZ::apply(const void * inImg, void * outImg, long numPixel
...
+}
+
+// all platforms support ssePower()
+template<>
+__m128 Renderer_PQ_TO_LINEAR_SSE<true>::myPower(__m128 x, __m128 exp)
+{
+    return ssePower(x, exp);
+}
+
+#ifdef _WIN32
```
cozdas, line 1322: On a second thought, it might be safer if this is based on the compiler, not the target platform. I'll do some research on this.

tests/cpu/transforms/BuiltinTransform_tests.cpp
```diff
@@ -365,128 +363,190 @@ AllValues UnitTestValues
 {
     // Contains the name, the input values and the expected output values.
 
-    { "IDENTITY",
-        { { 0.5f, 0.4f, 0.3f }, { 0.5f,            0.4f,            0.3f } } },
+    { "IDENTITY",
```
cozdas, line 366: adding ability to specify different thresholds for different built-ins as some are more sensitive than others to the underlying implementation change (i.e. lut-based vs scalar vs SSE implementation). Increased the threshold for PQ curve which has nested power() invocation thus numerically less stable. Also adding extended range values to make sure that all implementations can handle wider than [0,1] domain. Target values are computed with double-precision math.

src/OpenColorIO/transforms/builtins/Displays.cpp
```diff
@@ -31,40 +35,80 @@ static constexpr double c2 = 32. * 2413. / 4096.;
...
 
+namespace HLG
+{
+static constexpr double Lw = 1000.;
+static constexpr double E_MAX = 3.;
+
+static constexpr double a = 0.17883277;
+static constexpr double b = (1. - 4. * a) * E_MAX / 12.;
+static const     double c0 = 0.5 - a * std::log(4. * a);
+static const     double c = std::log(12. / E_MAX) * 0.17883277 + c0;
```
cozdas, line 88: should I substitute these 0.17883277's with "a"?

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp
```diff
@@ -1177,15 +1282,412 @@ void Renderer_LUV_TO_XYZ::apply(const void * inImg, void * outImg, long numPixel
...
+        // RGB
+        for (int ch = 0; ch < 3; ++ch)
+        {
+            float v = *(in++);
+            // Input is in nits/100, convert to [0,1], where 1 is 10000 nits. 
+            const T L = std::abs(v * T(0.01));    
+            const T y = std::pow(L, T(m1));
+            const T ratpoly = (T(c1) + T(c2) * y) / (T(1.) + T(c3) * y);
+            const T N = std::pow(ratpoly, T(m2));
+            *(out++) = std::copysign(float(N), v);
```
doug-walker, line 1363: Please add this comment to capture Nick's comment in Issue #1968.
// Note: the PQ value for zero is 0.836^78.84 = 7.36e-07 so there is a very small jump in the mirroring at zero. However, this is 20x smaller than a single 16-bit code value, so it is not visually significant.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -523,6 +523,113 @@ void Add_LUV_TO_XYZ(GpuShaderCreatorRcPtr & shaderCreator, GpuShaderText & ss)
...
+void Add_LINEAR_TO_PQ(GpuShaderCreatorRcPtr& shaderCreator, GpuShaderText& ss)
+{
+    using namespace ST_2084;
+    const std::string pxl(shaderCreator->getPixelName());
+
+    ss.newLine() << ss.float3Decl("sign3") << " = sign(" << pxl << ".rgb);";
+    ss.newLine() << ss.float3Decl("L") << " = abs(0.01 * " << pxl << ".rgb);";
+    ss.newLine() << ss.float3Decl("y") << " = pow(L, " << ss.float3Const(m1) << ");";
+    ss.newLine() << ss.float3Decl("ratpoly") << " = (" << ss.float3Const(c1) << " + " << c2 << " * y) / ("
+        << ss.float3Const(1.0) << " + " << c3 << " * y);";
```
doug-walker, line 562: Please add a comment:
// The sign transfer here is very slightly different than in the CPU path, resulting in a PQ value of 0 at 0 rather than the true value of 0.836^78.84 = 7.36e-07, however, this is well below visual threshold.

remia (commented):
Great work @cozdas and @doug-walker!

Do we need to update the [CTF minimum version](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/fileformats/ctf/CTFTransform.cpp#L224) as well?

doug-walker:
@remia , yes absolutely right about the CTF version.  I saw you bumped the version in your ACES PR, so my plan is to merge that first to update the CTF version to 2.4 and then we will add the corresponding tests to this PR.  There will be some integration work as well in FixedFunction.

remia (approved):

tests/cpu/Config_tests.cpp
```diff
@@ -5011,7 +5143,7 @@ R"([OpenColorIO Warning]: FixedFunction style is experimental and may be removed
 
         OCIO::ConstConfigRcPtr config;
         OCIO_CHECK_THROW_WHAT(config = OCIO::Config::CreateFromStream(is), OCIO::Exception,
-            "Only config version 2.4 (or higher) can have FixedFunctionTransform style 'ACES_OUTPUT_TRANSFORM_20'.");
+            "Only config version 2.4 (or higher) can have FixedFunctionTransform style 'ACES2_OutputTransform'.");
```
doug-walker, line 5146: @remia , I'm not sure why this was passing in the unit tests in your ACES PR.

doug-walker:
@remia , the merge of the ACES2 PR into this branch was fairly complicated, so I would appreciate it if you could make another pass over this one.  And the CTF version check was added.

tests/cpu/Config_tests.cpp
remia: I was not using `FixedFunctionStyleToString(ffstyle)` to print the style in the original PR. But it's better the way you have done it!

remia:
> @remia , the merge of the ACES2 PR into this branch was fairly complicated, so I would appreciate it if you could make another pass over this one. And the CTF version check was added.

Looks ok to me!

zachlewis (approved):
Man, this is _a lot_ of work... Cuneyt, you're a maniac. This is *awesome*. 

Looks like you've taken care of #1705 here too, that's cool to see. 

I think these "Double Log" and "Gamma Log" FixedFunctions and Builtins are excellent additions, and are quite reasonably named. Having LUT-free alternatives for some of these builtins makes it a lot easier to sling representations of processors around, both in memory and on printed paper. 

For the future, I think we need to figure out a better way of documenting FixedFunction parameters. 

Fantastic work.

Issue #1781, opened 2023-03-22
Alpha channel should be 1 by default, rather than 0


doug-walker:
Currently if you make an apply call with a source ImageDesc that is RGB and a destination that is RGBA, the A value of the destination is 0.  This is inconvenient since most clients will need to then set the alpha channel to 1 to make the image visible.

The proposal is that the alpha channel should start out as 1 in this scenario.

royashams:
I would be interested in taking a look at this issue for aswf devdays!

pennelee:
Hi @doug-walker , Is there a sample test case that exists already that has destination RGBA and shows the issue where alpha is 0?  Thank you!

doug-walker:
Any CPUProcessor that starts with RGB input and produces RGBA output will have the A set to 0. You could start with a ColorSpaceTransform between any two color spaces to generate the Processor.

doug-walker:
This was fixed via PR #2197.

PR #2033, opened 2024-09-02
Alpha channel should be 1 by default, rather than 0

pennelee:
Related to issue #1781 .
https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1781

For CPU processor, when source has no alpha (RGB) but dest does (RGBA) make it default to 1.
GPU Processors (new and legacy) did not seem to have this issue, alpha was already set to 1.  Let me know if this is not the case.

Tested changes with ocioconvert and changed the output to force RGBA, since the  default is to use inplace conversion (did not check in those changes).
Tested with all the ocioconvert test cases (basic, --lut, --view, --invertview --namedtransform, --invnamed transform), and various EXR file sizes.  

Before the changes alpha would be 0, so image was not viewable.  After the changes then images can be viewed and from inspection the alpha channel is now 1.0.

### DIFF
--- a/src/OpenColorIO/ImagePacking.cpp
+++ b/src/OpenColorIO/ImagePacking.cpp
@@ -70,7 +70,7 @@ void Generic<Type>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
         inBitDepthBuffer[4*pixelsCopied+0] = *rPtr;
         inBitDepthBuffer[4*pixelsCopied+1] = *gPtr;
         inBitDepthBuffer[4*pixelsCopied+2] = *bPtr;
-        inBitDepthBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (Type)0.0f;
+        inBitDepthBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (Type)1.0f;
 
         pixelsCopied++;
         xIndex++;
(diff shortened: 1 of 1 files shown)


pennelee:
Hello @doug-walker , 
It looks like there were some CI failures and timeouts, but I'm not clear if they are related to this change I made.  Do you have any suggestion on how to resolve the CI issues?  Thank you!

KelSolaar:
@pennelee : The tests are failing here: https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/10661264522/job/29546503311?pr=2033#step:9:8497, the workflow file does not set `continue-on-error` so it will kill all the jobs.

doug-walker:
@pennelee , yes I would definitely expect that you will need to update the unit tests based on your changes. You may even want to add a new unit test, if the existing ones don't cover this behavior sufficiently well.

Were you able to run the tests locally?  The [Contributor Guide](https://opencolorio.readthedocs.io/en/latest/guides/contributing/contributing.html) has info about running the tests (ctest -V).

pennelee:
Thanks @KelSolaar , @doug-walker , I will run the test manually and see what needs to be changed and/or added.

pennelee:
Stale, creating a new one.

PR #2038, opened 2024-09-16
Add bitdepth flag to ocioconvert

remia:
This simply adds a `--bitdepth` flag to ocioconvert to explicitly control the CPU processing bitdepth and CPU / GPU write bitdepth.


doug-walker (approved):

src/apps/ocioconvert/main.cpp
```diff
@@ -611,7 +646,7 @@ int main(int argc, const char **argv)
             imgOutput->attribute("oiio:ColorSpace", outputcolorspace);
         }
 
-        imgOutput->write(outputimage);
+        imgOutput->write(outputimage, userOutputBitDepth);
```
cozdas, line 649: Feels like this should be outputBitDepth instead of userOutputBitDepth...

remia: Hi @cozdas, not sure I follow exactly, can you expand on that? Thanks

cozdas: When the user provides a bit depth for the buffer, that bit depth is used for the write call too. When the user doesn't force a bit depth, then we don't pass the buffer depth to the write call, I'm guessing that it figures out the depth by itself. If this is intended behavior, then code is correct, otherwise passing the outputBitDepth to the write call would provide more consistent behavior, which may or may not be the behavior we want.

So I just wanted to make sure that it's what you intended, not a typo.

cozdas (approved):
Look good.

src/apps/ocioconvert/main.cpp
remia: My understanding is:

By default, the output buffer uses the [same](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/apps/ocioconvert/main.cpp#L286) bitdepth as the input.
-  If the GPU path is used, the input image is read as F32, processed as F32, then written to disk in F32 (unless the new --bitdepth flag is used).
- If the CPU path is used, the input image is read at its original bitdepth, processed in a bitdepth [determined](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/apps/ocioconvert/main.cpp#L462) by ocioconvert (unless the --bitdepth flag is used to override it), then written to disk at that same bitdepth (note for this case, it is not needed to specify explicitly the bitdepth to the write call).

So I believe this is working as expected.

Issue #1999, opened 2024-07-20
Update built-in configs for OCIO 2.4 library


doug-walker:
There are a number of updates happening in the OCIO Configs for ACES project, including:
- Support for ACES 2.0
- Addition of AppleLog
- Addition of AdobeRGB
- Addition of ASWF Color Interop Forum color spaces and alias names
- Making display color spaces active
- Addition of new aliases
- Miscellaneous bug fixes

Support for ACES 2.0 will require new built-in transforms which is tracked via #1934.

Once the configs are available from the configs project, these will be integrated as built-in configs.

The ASWF Color Interop Forum draft recommendation is [available here](https://docs.google.com/document/d/1IV3e_9gpTOS_EFYRv2YGDuhExa4wTaPYHW1HyV36qUU/edit?usp=sharing).

PR #2039, opened 2024-09-17
Adsk Contrib - Adding the new cg and studio configs (issue #1999)

cozdas:
For issue #1999 adding the new cg and studio configs.

- Added new configs cg-config-v2.2.0_aces-v1.3_ocio-v2.4.ocio and studio-config-v2.2.0_aces-v1.3_ocio-v2.4.ocio
- Made default config cg-config-v2.2.0_aces-v1.3_ocio-v2.4
- Updated the tests accordingly.

TODO: Some documents such as aces_cg.rst and aces_studio.srt need to be updated.


KelSolaar (approved):
LGTM!

KelSolaar:
Related question: The list of builtin configs is growing, is there a deprecation/removal plan?

doug-walker:
@KelSolaar , I'm guessing you're asking that question due to a concern over the increase in size it adds to the library?

The goal is to improve the system so that it only implements the deltas between configs rather than generating each from scratch. In the nanoColor branch we have a way of generating the built-in configs via code (starting from the YAML files you provided). Currently this simply regenerates each config from the YAML, but this approach could be extended to avoid a significant increase in size by only implementing the changes from one version to the next.

zachlewis (approved):
Dunno if I love the idea of a color space called "CIE XYZ-D65 - Scene-referred", but it does make more sense in the context of the Display-referred equivalent also existing.

This has no doubt come up in the config meetings, but are we absolutely certain we want to expose the Scene-referred XYZ D65, and not the Display-referred? Maybe it's just me, but I've been using XYZ D65 to encode display-referred EXRs, to differentiate them from scene-referred RGB EXRs. (e.g., when one department is working on supers and graphic elements "on top of the LUT", while others are working scene-referred, "beneath the LUT").

Either way, these builtin configs look great to me!

doug-walker:
@zachlewis , thanks for the feedback. Let's discuss removing "CIE XYZ-D65 - Display-referred" from the inactive list at the TSC meeting on Monday.

doug-walker:
Following the discussion at the TSC meeting today, we decided to add the color space "CIE XYZ-D65 - Scene-referred" to the inactive list. I logged this separately [here](https://github.com/AcademySoftwareFoundation/OpenColorIO-Config-ACES/issues/136). To recap, the group agreed that this is more for use via the API and is not something that should be exposed in artist menus.

@zachlewis , thanks for starting the discussion. I pushed a commit to implement the change.

KelSolaar:
> @zachlewis , thanks for starting the discussion. I pushed a commit to implement the change.

I did put it first on my side, with the idea that we go from Scene to Display.

PR #2051, opened 2024-09-21
Adsk Contrib - Add DCDM displays and improve HLG OOTF implementation

doug-walker:
Add SDR and HDR DCDM built-in displays, which are in the set of Output Transforms recommended by ACES 2.

Improve the Rec.2100 surround algorithm used for the Rec.2100-HLG OOTF and EOTF:
- Fix an issue with the inverse
- Mirror the result around the origin for negative values


src/OpenColorIO/transforms/builtins/Displays.cpp
```diff
@@ -201,6 +201,50 @@ void RegisterAll(BuiltinTransformRegistryImpl & registry) noexcept
...
+            CreateScaleOp(ops, scale4, TRANSFORM_DIR_FORWARD);
+
+            const GammaOpData::Params rgbParams   = { 2.6 };
+            const GammaOpData::Params alphaParams = { 1.0 };
+            auto gammaData = std::make_shared<GammaOpData>(GammaOpData::BASIC_REV,
+                                                           rgbParams, rgbParams, rgbParams, alphaParams);
+            CreateGammaOp(ops, gammaData, TRANSFORM_DIR_FORWARD);
+        };
+
+        registry.addBuiltin("DISPLAY - CIE-XYZ-D65_to_DCDM-D60-BFD",
```
remia, line 243: Is this display intended to be used with the ACES2 Builtins (same for the HDR variant)? Note that there is no [CAT](https://github.com/remia/OpenColorIO/blob/aces2-output-transforms/src/OpenColorIO/transforms/builtins/ACES.cpp#L562) to D65 performed at the moment in the ACES2 builtins, which I believe was handled similarly in ACES 1.x D60 transforms. Please let me know if I need to update something on the ACES2 PR side.

src/OpenColorIO/transforms/builtins/Displays.cpp
```diff
@@ -285,6 +329,46 @@ void RegisterAll(BuiltinTransformRegistryImpl & registry) noexcept
                             CIE_XYZ_D65_to_ST2084_P3_D65_Functor);
     }
 
+    {
+        auto CIE_XYZ_D65_to_ST2084_DCDM_D65_Functor = [](OpRcPtrVec & ops)
+        {
+            // Scale 1.0 in the reference space to 0.48. The PQ function will interpret
```
remia, line 335: IS the scaling related to ACES2 `linear_scale_factor`? Because it is already applied in the [ACES2 builtins](https://github.com/remia/OpenColorIO/blob/aces2-output-transforms/src/OpenColorIO/transforms/builtins/ACES.cpp#L554). Sorry if I'm getting confused here.

doug-walker: You are totally right, the scaling actually makes no sense here for several reasons.  Removed!

src/OpenColorIO/transforms/builtins/Displays.cpp
doug-walker: Yes, it's intended to be used for the ACES 2 config. But I think what you're proposing is that you would use the "DISPLAY - CIE-XYZ-D65_to_DCDM-D65" for both the D65 and D60 view transforms. The D60 version will essentially have a "D60 sim" type of skew to it. I will remove the D60 versions.

zachlewis (approved):
El gee tee em!
Nice work, Doug!

remia (approved):
The only other display I see missing is the Display P3 / Gamma 2.2 variant, but maybe we already decided not to include. LGTM in any case!

Issue #2040, opened 2024-09-18
Platform Latest CI -- Python tests seg-fault on Windows


doug-walker:
This issue is to solve the issue that is causing the GitHub Actions Platform Latest CI runs to fail due to a seg-fault in the Python 3.11 unit tests on Windows.

remia:
I believe this is the same issue that was affecting the Wheel workflow, the issue disappear when one set `OCIO_PYTHON_LOAD_DLLS_FROM_PATH` to 0 which seem to indicate one the path added to the Python DLL directories cause the crash. On platform latest, the crash only occurs on Release build it seems.

Through some iterative tests, I have determined that one need to explicitly exclude these _two_ paths from being added to the DLL directories for the workflow succeed:
```
'C:\\Program Files\\Microsoft Service Fabric\\bin\\Fabric\\Fabric.Code'
'C:\\Program Files\\LLVM\\bin'
```
[Dumpbin](https://learn.microsoft.com/en-us/cpp/build/reference/imports-dumpbin?view=msvc-170) reports the following dependencies on PyOpenColorIO:
```
python310.dll
OpenColorIO_2_4.dll
MSVCP140.dll
VCRUNTIME140.dll
VCRUNTIME140_1.dll
api-ms-win-crt-runtime-l1-1-0.dll
api-ms-win-crt-heap-l1-1-0.dll
api-ms-win-crt-math-l1-1-0.dll
api-ms-win-crt-stdio-l1-1-0.dll
api-ms-win-crt-filesystem-l1-1-0.dll
KERNEL32.dll
```
When PyOpenColorIO is loaded, the following DLL are loaded (as seen from [dlltracer](https://pypi.org/project/dlltracer/)), obviously this is on a build that's working, otherwise the segfault prevent from having any log from dlltracer (even with python -u for un-buffered outputs, which seem to indicate the crash is quite early in the import process):
```
LoadLibrary \Device\HarddiskVolume2\Windows\System32\kernel.appcore.dll
LoadLibrary \Device\HarddiskVolume2\hostedtoolcache\windows\Python\3.11.9\x64\DLLs\_socket.pyd
LoadLibrary \Device\HarddiskVolume2\hostedtoolcache\windows\Python\3.11.9\x64\DLLs\select.pyd
[...]

PR #2052, opened 2024-09-21
Python 3.13 wheels

remia:
Here is a list of changes included:
- Bumping cibuildwheel, we should have working macOS ARM testing
- Bumping the default pybind11 version used in the build (no strong reason but seems like a good time)
- Adding Python 3.13 wheels and removing 3.7 wheels
- Fixing macOS build issues related to the deployment target and recent updates in cibuildwheel
- FIxing Windows test issues related to DLL search directory, the exact cause has not been identified but the workflow should now complete successfully. Debugging was made harder as I haven't been able to get any print out of the OpenColorTestSuite.py, even after increasing build-verbosity to 3 in cibuildwheel (warnings worked though). Use of dlltracer (stdout or stderr or audit events) didn't yield any results, but might if the debug print issue is fixed (though it probably depends on when the segfault is triggered exactly).

Tagging @KelSolaar


setup.cfg
```diff
@@ -13,6 +13,7 @@ classifiers =
     Programming Language :: Python :: 3.10
     Programming Language :: Python :: 3.11
     Programming Language :: Python :: 3.12
+    Programming Language :: Python :: 3.13
```
KelSolaar, line 16: Should we pop the `Programming Language :: Python :: 3.7` classifier above?

KelSolaar:
LGTM besides the classifier.

setup.cfg
remia: Good point and I was also wondering about it, here we simply remove wheels for 2.7 but users could still build manually for Python 3.7 and use it. We are still testing Python 3.7 with the VFX 2020 platform docker images, though it's about to be removed in the VFX 2024 PR. So all that to say that I have no objections to remove :)

KelSolaar: Looking at the Scientific Python SPEC0, support for Python 3.10 is about to be dropped: https://scientific-python.org/specs/spec-0000/, CY2023 is Python 3.10, and Python 3.7 is CY2021.

I think that as a whole, there should be an ASWF consensus on which minimum Python version to support, especially when some of the core packages we are relying on, e.g., Numpy have seen their support dropped. In Colour, we split the apple in two and support the previous VFX Reference Platform year as minimum. Let's discuss at TSC.

doug-walker (approved):
Thanks for fixing this Remi!

zachlewis (approved):
Nice work! 
Really strange, that the OCIO_PYTHON_LOAD_DLLS_FROM_PATH thing is causing problems. That's pretty baffling. 

Do we want to support free-threaded mode too for Python 3.13?

remia:
> Nice work! Really strange, that the OCIO_PYTHON_LOAD_DLLS_FROM_PATH thing is causing problems. That's pretty baffling.
> 
> Do we want to support free-threaded mode too for Python 3.13?

For the segfault, let's continue the discussion in https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/2040!

I don't think we have the resources to add support for GIL-free Python in the coming release and not exactly sure yet what work will be required to be honest. From a quick look, I assume we have to upgrade to pybind 2.13 and start using `py::mod_gil_not_used()` to check whether we need to handle unlocking / locking the GIL or not?

setup.cfg
remia: I forgot that VFX 2021 is still using Python 3.7, for now let's keep it in there as there was no strong feeling against at yesterday TSC. We should definitively make sure we have a defined plan for older platform support so worth discussing at the next TSC when you are there :)

PR #2056, opened 2024-09-23
Fix iridas cube parser to accept and convert double precision.

Shootfast:
The old Iridas cube parser would silently convert double precision values to floats and continue working, wheras the current code will raise an "Invalid color triplets" error.

This patch restores the older behaviour.

### DIFF
--- a/src/OpenColorIO/fileformats/FileFormatIridasCube.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatIridasCube.cpp
@@ -336,9 +336,12 @@ LocalFileFormat::read(std::istream & istream,
             }
             else
             {
-                float r = NAN;
-                float g = NAN;
-                float b = NAN;
+		// Use doubles here as some iridas cube files may be written with double precision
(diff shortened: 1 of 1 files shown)


src/OpenColorIO/fileformats/FileFormatIridasCube.cpp
```diff
@@ -336,9 +336,12 @@ LocalFileFormat::read(std::istream & istream,
             }
             else
             {
-                float r = NAN;
-                float g = NAN;
-                float b = NAN;
+		// Use doubles here as some iridas cube files may be written with double precision
+		// string values, and these will raise result_out_of_range errors in NumberUtils::from_chars
+		// instead of silently converting to float like they used to
+                double r = NAN;
```
remia, line 342: It probably needs an explicit cast to double to make the Windows checks happy?

Shootfast:
Actually I've just noticed this problem extends to other lut file formats as well - at least spi1d/spi3d

PR #2058, opened 2024-09-24
Fix platform latest

remia:
This should fix the Platform Latest workflow:
* Disabling OpenFX as not compatible with C++23 (I tried the latest 1.5 release to no avail)
* OCIO_PYTHON_LOAD_DLLS_FROM_PATH disabled for the test run and OCIO build directory manually added to the DLL lookup directory in OpenColorIOTestSuite.py


doug-walker (approved):
Awesome!

zachlewis (approved):

PR #2061, opened 2024-09-27
Update getting_started.rst

mbriamon:
The directory should be: OpenColorIO/share/docker. Instead of OpenColorIO/shared/docker. I wasn't able to find /shared/docker.


doug-walker (changes requested):
Hi Mary, thanks for the contribution!

You change is completely correct about the location of that file. Unfortunately, no one has been maintaining that docker file for years and so it is not something we want people to be using in its current state.  

It would be great if you could update that section to point people at this docker image, if they want a pre-built OCIO environment:
https://github.com/AcademySoftwareFoundation/aswf-docker/tree/main/ci-vfxall

Additionally, I would remove the share/docker directory.

PR #2063, opened 2024-09-27
Docs issue 2009

czerouni:
This attempts to address issue 2009, however it's not really possible to do the things requested there because of limitations of the Sphinx theme used. 

So instead this cleans up and extends the documentation for contributing to the docs. It also fixes a few Sphinx warnings. And it recreates a new set of frozen .rst files from the code. That modified several files and deleted many more, which reduced the number of Sphinx warning dramatically.


carolalynn:
Hey @czerouni - thanks for this! I have some questions on the content, but also, we required signed commits (commit -s) to acknowledge the DCO. So you'll need to:

```
git commit --amend -s
git push <your_remote_fork_name> <your_branch_name> --force
```

docs/api/python/frozen/pyopencolorio_colorspacehelpers.rst
```diff
@@ -3,7 +3,7 @@
   Copyright Contributors to the OpenColorIO Project.
   Do not edit! This file was automatically generated by share/docs/frozendoc.py.
 
-.. py:function:: AddColorSpace(config: PyOpenColorIO.Config, name: str, transformFilePath: str, categories: str = '', connectionColorSpaceName: str) -> None
+.. py:function:: AddColorSpace(config: PyOpenColorIO.PyOpenColorIO.Config, name: str, transformFilePath: str, categories: str = '', connectionColorSpaceName: str) -> None
```
carolalynn, line 6: Why is this now adding a second PyOpenColorIO for all of these?? Seems fishy to me....?

czerouni: I thought I did the signed commit thing, but I'll redo it.

And that's a good question - I don't know why that happened. I can try rerunning with the `BUILD_FROZEN_DOCS=ON` flag and see what happens.

czerouni: I did it again and nothing changed. Unfortunately the 'frozen' files are only used with BuildTheDocs, and I couldn't get them to work locally without that, so I couldn't QA them.

remia: I have built the docs often, but the PyOpenColorIO.PyOpenColorIO I think is due to a change made last year (https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1778) whereby the PyOpenColorIO is now a package (folder) which internally ship the PyOpenColorIO compiled extensions module (as well as other things like command line apps). So this can lead to duplication like here. Is this the first time we re-build python doc since March 2023?

michdolan: It may have been quite a while since we built the docs, yes. We really need to remove the frozen Python docs implementation, which makes it pretty painful to keep everything up to date. For context, the frozen docs generation was implemented to work around the issue of RTD not having an environment for compiling OCIO in order to generate the Python docs. Now that we have OCIO wheels in PyPI, we could instead pip install the latest binary from an RTD build machine and use that for generating docs. I would suggest for now reverting changes made to the frozen docs in this PR, and just contribute updated RST content.

czerouni: I've restored the original frozen rst files

doug-walker (changes requested):
Thank you for this contribution Craig, this is awesome!

Unfortunately there are some conflicts that you will need to resolve before this can be merged. And as Carol wrote, you will need to make sure all of your commits are signed off so that the DCO check passes.

Apologies for the delayed review, and sorry for the issue with the frozen documentation.

czerouni:
I don't know git well enough to go back and sign those old commits. So I am going to close this PR, create a new fork, and just do a single commit with the changes (and deal with the conflicts at the same time). Then open a new PR for that.

Issue #2066, opened 2024-09-30
Floating Point Domain Error triggered by IdentifyBuiltinColorSpace()


stolk:
`IdentifyBuiltinColorSpace()` applies transformations on pixel values that can leave the pixel value containing negative values.

Subsequently, when a gamma curve is applied in `OpenColorIO_v2_4dev::GammaMoncurveOpCPURev::apply()`  the `powf()` function is called with a negative base and a fractional power, which causes a domain error.

This leads to a **SIGFPE** and will abort applications that have floating point exceptions enabled.

```
#0  0x00007ffff6fb9c4e in __math_invalidf (x=-0.0485877097) at ../sysdeps/ieee754/flt-32/math_errf.c:80
#1  0x00007ffff50f6271 in std::pow (__x=-0.0485877097, __y=0.416666657) at /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/cmath:373
#2  0x00007ffff535681f in OpenColorIO_v2_4dev::GammaMoncurveOpCPURev::apply (this=0x555556e01fa0, inImg=0x555556e2e190, outImg=0x555556e2e190, numPixels=5)
    at /home/stolk/src/OpenColorIO/src/OpenColorIO/ops/gamma/GammaOpCPU.cpp:634
#3  0x00007ffff5583874 in OpenColorIO_v2_4dev::GenericScanlineHelper<float, float>::finishRGBAScanline (this=0x555556ebe780)
    at /home/stolk/src/OpenColorIO/src/OpenColorIO/ScanlineHelper.cpp:164
#4  0x00007ffff517e898 in OpenColorIO_v2_4dev::CPUProcessor::Impl::apply (this=0x555556e0be20, srcImgDesc=..., dstImgDesc=...)
    at /home/stolk/src/OpenColorIO/src/OpenColorIO/CPUProcessor.cpp:431
#5  0x00007ffff517edbd in OpenColorIO_v2_4dev::CPUProcessor::apply (this=0x555556e7b3b0, srcImgDesc=..., dstImgDesc=...)
    at /home/stolk/src/OpenColorIO/src/OpenColorIO/CPUProcessor.cpp:543
#6  0x00007ffff516dec3 in OpenColorIO_v2_4dev::ConfigUtils::isIdentityTransform (
[...]

PR #2067, opened 2024-09-30
[Math] Fix NaN from powf() domain error.

stolk:
This change prevents calling powf(x,y) with negative x.

The SIMD versions using ssePower() already seem to be resistent to negative pixel values, but the scalar version was not.

This would cause a SIGFPE on apps that have floating point exceptions enabled.

FIXES: #2066


doug-walker (commented):
Thanks Bram, good catch.  What do you think about moving the std::pow call down into the ternary operator, to avoid adding the extra std::max call?

In addition, I think this would fix the unit test errors.  (You're clamping the value before it's used in the test.)

stolk:
> Thanks Bram, good catch. What do you think about moving the std::pow call down into the ternary operator, to avoid adding the extra std::max call?

Hi Doug,

I advise against the ternary operator...

I checked, and the compilers (gcc, clang) actually produce more efficient code with `std::max` than with `?:`

See [this godbolt output](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIACcpAAOqAqETgwe3r7%2ByanpAiFhkSwxcYl2mA4ZQgRMxARZPn4BldUCtfUERRHRsQm2dQ1NOa1D3aG9pf3xAJS2qF7EyOwc5gDMocjeWADUJutuYsAkhAgsB9gmGgCC1zdUtKhMBLuiSUmY6FzUTy%2B7ADcxLN9gB2Ky3XZQ3bETAEJYMQFifaHXYaAB0GioKIAYmjMdiQEjaAcIXdQQARe73R7PV7vT7oMy/OnEkEmcH3aEwuEI3YKAjoEAgFhMVQQDFY0hs0nUykcea0TgAVl4fm4vFQnDc1ms/MWy0w%2BzM6x4pAImgV8wA1v5leiAByg0FmSRmDSSABs8Uk8U96wCSo4kjVltIWo4vAUIA05st8zgsCQaBYSTosXIlBTafocWAUgCNFoBFi0YgUTDUVC9QAnpwzVXmMQawB5KLaKoWjWkFNsQQthi0OscLSkLBRLzAI60WjR7tYUVGcQj3j4WHVAGYOejzCqKpeEv13ihEtB0e0PBRYi1jxYMMEYh4FhH0ib4hRVKYCmYRfAC9GeM%2BAMYAFAANTwTAAHcW0%2BdUzX4QQRDEdguHWGRBEUFR1BXUhdC4fRDGMXVLH0S9o0geZUCSRwBDnABaFt1k1N9HywciIHmNoaL8CBXBGPwuACYJJhKMoJGVPI0m4/iJACFIpIyHpRP6LgJK4mpxhkwTbEwexuM6BolL6OJVMGLotLGLojOmEzlU4g0VgkRUVVDHCI12VQHU9OjPUkXZgGQZBdikdEzF2CBcEIEhjVNWZeC7LRZnmBBMCYLA4g40hbTMUF0WdDR3TdH03U9DRPUDTgQ1IdVRwjKMYzjFcExgRAUFQVN0zICgIGzTqUAMIwuAdLhYyLEtiDLCscMbWsXxm5s2w7BwX17RgCAHIcw3HSdp1nF8F0I5dRzXTs8E3bdeF3fdD27E9dLDC8rxvDBVlHB8nxfN8PyUb9f3/UAmqApgQPAqCYMYF8EOEURxCkdD5CUNQw10QNCJQYibEe9jKOojJ6MY5jYlYrd4E43TTucXiGHcTxmj0YTimMvR5IKTJaZyfCWe46yxPw9SOk09mBJ0vSNKskSmb5wXsmFgVxcZmynIWJZHK4ZyOFVaqw3czzvN8t4BuAYKHXRLhMXCyKiGIGK1fi%2BMbX8M31mVMxlWVdZPWVLzXQDfRKtc2rOHq2MEoVUhE1axYCCSA9Mx69qc1icJWFWXWfL8gKgogB8vAYa04sCfArbwIV8KhpDYekKHMKRnC8NISDrySI91c1mrNU4FsDxj15UGxNP9cz4LJFC8KPA63MbYL0Okqyu0/eDAOO8jWwGpn9WzCX8Mg8axL5jfNJnEkIA%3D).

stolk:
> In addition, I think this would fix the unit test errors. (You're clamping the value but the ternary expression needs the unclamped value.)

Oops, did I break the unit test? I'll revisit.

doug-walker:
I was not proposing to replace the max with a new ternary. I was suggesting moving the pow call down into the existing ternary. As I mentioned, your unit tests are failing, so some action is needed.

The CLAs for all ASWF projects are managed through the same account but need to be signed separately for each project.

stolk:
> I was not proposing to replace the max with a new ternary. I was suggesting moving the pow call down into the existing ternary. As I mentioned, your unit tests are failing, so some action is needed.

Thanks Doug, I will redo this PR.

Strangely, the unmodified main branch fails the unit test as well. I have to find out what is going on here.

```
$ make test
Running tests...
Test project /home/stolk/src/OpenColorIO/build
    Start 1: test_utils_exec
1/6 Test #1: test_utils_exec ..................   Passed    0.01 sec
    Start 2: test_cpu
2/6 Test #2: test_cpu .........................***Failed    5.78 sec
    Start 3: test_cpu_no_accel
3/6 Test #3: test_cpu_no_accel ................***Failed    5.41 sec
    Start 4: test_cpu_avx512
4/6 Test #4: test_cpu_avx512 ..................***Failed    5.99 sec
    Start 5: test_gpu
5/6 Test #5: test_gpu .........................***Failed    8.83 sec
    Start 6: test_python
6/6 Test #6: test_python ......................   Passed    1.38 sec

33% tests passed, 4 tests failed out of 6
```

Does the main branch HEAD pass the unit tests for you? Thanks.

doug-walker:
All the tests should pass, except for the GPU tests on some mac platforms currently.

stolk:
I have changed this PR to be less invasive, and only correct the value passed to `powf()` and not the entire pixel value as used in the rest of the function.

Note: I cannot get the unit test to pass on my machine, even when using HEAD of main. Pushed the PR to see if CI will pass.

KevinJW:
just to note, micro optimisation difference comes from the testing vs return order, if you flip the ternary around you can get the same code ... https://godbolt.org/z/Wzafa3nsb

stolk:
Can a maintainer please approve the CI run, and then merge? Thanks.

doug-walker (commented):
Hi Bram, what do you think about changing this:

out[0] = pixel[0]<=red[3] ? pixel[0] * red[4] : data[0];

to this:

out[0] = pixel[0]<=red[3] ? pixel[0] * red[4] : 
     std::pow(pixel[0] * red[0] + red[1], red[2]);

That way it avoids the extra branch.

stolk:
> Hi Bram, what do you think about changing this:
> 
> out[0] = pixel[0]<=red[3] ? pixel[0] * red[4] : data[0];
> 
> to this:
> 
> out[0] = pixel[0]<=red[3] ? pixel[0] * red[4] : std::pow(pixel[0] * red[0] + red[1], red[2]);
> 
> That way it avoids the extra branch.

No, that does not address the issue.

We should not be feeding negative numbers to the base of a powf().

My change makes sure of that, your version does not.

doug-walker:
Hi Bram,

Is the reason that you don't like my proposal because you are worried that `pixel[0] * red[0] + red[1]` will be negative? In practice, that is never the case. In theory, someone may be able to set parameter values that would place the breakpoint below zero, but if that is the edge case you are worried about, it would be better dealt with in the GammaOpData.cpp:GammaOpData::validateParameters function rather than inside the performance critical pixel loop.

PR #2072, opened 2024-10-01
Build: enable parallel source file compilation on MSVC

aras-p:
By default Visual Studio projects do not compile source files in parallel. Add "/MP" flag to enable that.

From scratch OCIO build on Ryzen 5950X, VS2022: 580sec -> 208sec. Most of remaining single-threaded time is cloning external libraries or building external libraries (most of them don't do parallel compilation either). But within building OCIO itself, CPU usage goes close to 100% instead of 5%.


remia:
We should probably have added this long ago... thanks @aras-p!

remia (approved):

cozdas (approved):

PR #2078, opened 2024-10-10
Replaced DX11 HLSL shading language with the more accurate SM_5_0

num3ric:
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/overviews-direct3d-11-hlsl


include/OpenColorIO/OpenColorTypes.h
```diff
@@ -449,11 +449,14 @@ enum GpuLanguage
...
-    GPU_LANGUAGE_HLSL_DX11,         ///< DirectX Shading Language
+    GPU_LANGUAGE_HLSL_SM_5_0,       ///< HLSL Shading Language
     LANGUAGE_OSL_1,                 ///< Open Shading Language
     GPU_LANGUAGE_GLSL_ES_1_0,       ///< OpenGL ES Shading Language
     GPU_LANGUAGE_GLSL_ES_3_0,       ///< OpenGL ES Shading Language
-    GPU_LANGUAGE_MSL_2_0            ///< Metal Shading Language
+    GPU_LANGUAGE_MSL_2_0,           ///< Metal Shading Language
+
+    // Deprecated enum(s)
+    GPU_LANGUAGE_HLSL_DX11 = GPU_LANGUAGE_HLSL_SM_5_0
```
num3ric, line 459: This could actually be marked as `OCIO_DEPRECATED("This was marked as deprecated starting in v2.5, please use GPU_LANGUAGE_HLSL_SM_5_0.")`, but the macro would have to be changed as follows:
```
#if __cplusplus >= 201402L || _MSVC_LANG >= 201402L
    #define OCIO_DEPRECATED(msg) [[deprecated(msg)]]
#elif defined(__GNUC__) || defined(__clang__)
    #define OCIO_DEPRECATED(msg) __attribute__((deprecated(msg)))
#else
    #define OCIO_DEPRECATED(msg) /* unsupported on this platform */
#endif
```

src/OpenColorIO/ParseUtils.cpp
```diff
@@ -265,7 +265,7 @@ const char * GpuLanguageToString(GpuLanguage language)
         case GPU_LANGUAGE_GLSL_4_0:     return "glsl_4.0";
         case GPU_LANGUAGE_GLSL_ES_1_0:  return "glsl_es_1.0";
         case GPU_LANGUAGE_GLSL_ES_3_0:  return "glsl_es_3.0";
-        case GPU_LANGUAGE_HLSL_DX11:    return "hlsl_dx11";
+        case GPU_LANGUAGE_HLSL_SM_5_0:  return "hlsl_sm_5.0";
```
num3ric, line 268: Don't remember... so let me know if any precautions are needed here as this change would cause cache id invalidation.

include/OpenColorIO/OpenColorTypes.h
```diff
@@ -449,11 +449,14 @@ enum GpuLanguage
...
-    GPU_LANGUAGE_HLSL_DX11,         ///< DirectX Shading Language
+    GPU_LANGUAGE_HLSL_SM_5_0,       ///< DirectX High Level Shading Language
     LANGUAGE_OSL_1,                 ///< Open Shading Language
     GPU_LANGUAGE_GLSL_ES_1_0,       ///< OpenGL ES Shading Language
     GPU_LANGUAGE_GLSL_ES_3_0,       ///< OpenGL ES Shading Language
-    GPU_LANGUAGE_MSL_2_0            ///< Metal Shading Language
+    GPU_LANGUAGE_MSL_2_0,           ///< Metal Shading Language
+
+    // Deprecated enum(s)
+    GPU_LANGUAGE_HLSL_DX11 = GPU_LANGUAGE_HLSL_SM_5_0
```
remia, line 459: Don't remember exactly but didn't you mention the use of OCIO_DEPRECATED macro? Would it work here? Looking at the attribute [[depracated]] I think it should but not sure about the other fallback methods.

num3ric: Removed my previous comment about this because I thought it might be needlessly disruptive.

If we did want it however, it would require modifications to the macro for MSVC: the fix for `__cplusplus` in Aras P's other PR, and the deletion of `__declspec(deprecated(msg))` which doesn't work with enums.

doug-walker (approved):

cozdas (approved):
Looks good.

Issue #2082, opened 2024-10-17
getVirtualDisplayViewColorSpaceName is broken for shared views


doug-walker:
This method returns an empty string instead of the color space name. The same issue probably affects the other virtual display related methods when using shared views.

doug-walker:
Closed as completed via PR #2083.

Issue #2049, opened 2024-09-20
Add some view helper functions


doug-walker:
Here are some utility functions related to views that would be helpful to add to OCIO. The API of the functions below should be altered, as appropriate, when adding them to the Config class. As always, unit tests must be added.

```
bool displayHasView(const ConstConfigRcPtr & cfg, const char * dispName, const char * viewName)
{
    // This returns null if either the display or view doesn't exist.
    // It works regardless of whether the display or view are active,
    // and it works regardless of whether the view is display-defined
    // or if the display has this as a shared view.
    //
    // It will only check config level shared views if dispName is null.
    // It will not check config level shared views if dispName is not null.
    const char * cs = cfg->getDisplayViewColorSpaceName(dispName, viewName);

    // All views must have a color space, so if it's not empty, the view exists.
    return (cs && *cs);
}

```
```
bool hasVirtualView(const ConstConfigRcPtr & cfg, const char * viewName)
{
    const char * cs = cfg->getVirtualDisplayViewColorSpaceName(viewName);

    // All views must have a color space, so if it's not empty, the view exists.
    return (cs && *cs);
}
```
```
void clearSharedViews(ConfigRcPtr & cfg)
{
    int numViews = cfg->getNumViews(VIEW_SHARED, nullptr);
    for (int v = numViews - 1; v >= 0; v--)
    {
        const char * sharedViewName = cfg->getView(VIEW_SHARED, nullptr, v);
        if (sharedViewName && *sharedViewName)
        {
            cfg->removeSharedView(sharedViewName);
        }
    }
}
```
```
bool viewIsShared(const ConstConfigRcPtr & cfg,
                  const char * dispName,
                  const char * viewName)
{
    // Check if a view within a given display is a display-defined view or is referencing
    // one of the config's shared views.

    for (int v = 0; v < cfg->getNumViews(VIEW_SHARED, dispName); v++)
    {
        const char * sharedViewName = cfg->getView(VIEW_SHARED, dispName, v);
        if (sharedViewName && *sharedViewName && Platform::Strcasecmp(sharedViewName, viewName) == 0)
        {
            return true;
        }
    }

    return false;
}
```
```
bool virtualViewIsShared(const ConstConfigRcPtr & cfg,
                         const char * viewName)
{
    for (int v = 0; v < cfg->getVirtualDisplayNumViews(VIEW_SHARED); v++)
    {
[...]

PR #2083, opened 2024-10-17
Add view helper functions to Config class

annieln:
#2049
#2082


include/OpenColorIO/OpenColorIO.h
```diff
@@ -852,13 +852,18 @@ class OCIOEXPORT Config
...
      */
+
+    bool viewIsShared(const char * dispName, const char * viewName) const;
+
     /// Will throw if view or colorSpaceName are null or empty.
     void addSharedView(const char * view, const char * viewTransformName,
                        const char * colorSpaceName, const char * looks,
                        const char * ruleName, const char * description);
     /// Remove a shared view.  Will throw if the view does not exist.
     void removeSharedView(const char * view);
```
doug-walker, line 864: /// Clear all shared views. This will throw if any displays are still using the shared views.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -883,6 +888,11 @@ class OCIOEXPORT Config
     int getNumViews(const char * display, const char * colorspaceName) const;
     const char * getView(const char * display, const char * colorspaceName, int index) const;
```
doug-walker, line 890:
```
/**
 * \brief Compare views in a pair of configs.
 *
 * Will return false if either of the views does not exist. This will return true even
 * if the view is display-defined in one config and a reference to a shared view in the
 * other config (both within the same display), as long as the contents match. The
 * description text (if any) is ignored, since it is not a functional difference.
 */
```

include/OpenColorIO/OpenColorIO.h
```diff
@@ -900,6 +910,8 @@ class OCIOEXPORT Config
     /// Returns the description attribute of a (display, view) pair.
     const char * getDisplayViewDescription(const char * display, const char * view) const noexcept;
```
doug-walker, line 912:
```
/**
 * \brief Determine if a display and view exist.
 *
 * This returns false if either the display or view doesn't exist. It works regardless
 * of whether the display or view are active, and it works regardless of whether the
 * view is display-defined or if the display has this as a shared view. It will only
 * check config-level shared views if dispName is null. It will not check config level
 * shared views if dispName is not null.
 */

```

src/OpenColorIO/Config.cpp
```diff
@@ -3812,6 +3944,13 @@ const char * Config::getVirtualDisplayViewColorSpaceName(const char * view) cons
 {
     if (!view) return "";
 
+    // TODO: Remove the following work-around once bug for shared views is fixed.
```
doug-walker, line 3947: Actually, despite what I wrote, you could remove the TODO, I think your fix is sufficient.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -963,6 +975,9 @@ class OCIOEXPORT Config
      *
      */
```
doug-walker, line 977: Please add comments for the virtual case, similar to what I wrote above.

tests/cpu/Config_tests.cpp
```diff
@@ -7338,7 +7349,10 @@ OCIO_ADD_TEST(Config, add_remove_display)
 
     // Add a (display, view) pair.
 
+    // todo (annie): just a note here that this config will have two (display, view) pairs): sRGB and disp1
```
doug-walker, line 7352: You could remove the todo and make this a comment if you feel it clarifies the test.

tests/python/ConfigTest.py
```diff
@@ -413,11 +413,12 @@ def test_shared_views(self):
...
-        cfg.removeSharedView('view5')
-        cfg.removeSharedView('view6')
-        cfg.removeSharedView('view1')
-        cfg.removeSharedView('view2')
+        cfg.clearSharedViews()
+        # cfg.removeSharedView('view4')
+        # cfg.removeSharedView('view5')
+        # cfg.removeSharedView('view6')
+        # cfg.removeSharedView('view1')
+        # cfg.removeSharedView('view2')
```
doug-walker, line 421: Remove commented code.

doug-walker (changes requested):
Thanks Annie, this is excellent work!

I made a few suggestions below. If you could look at the comments I suggested for OpenColorIO.h and check that there is a test for each aspect of the documented functionality for each new function, that would be appreciated.

I notice that one of the Linux CI runs is failing due to a frozen Python docs problem. Currently the frozen docs system is broken and so we will handle that separately. As long as the Windows and Mac tests are passing, that's all you need to be concerned with for now.

include/OpenColorIO/OpenColorIO.h
annieln: Sorry, I'm just checking - is this meant for clearSharedViews instead of viewIsShared?

doug-walker: Yes, sorry, I put the comment on the wrong line.

annieln:
Hi, Doug! I haven't forgotten about this issue, but it has been a while, so I may have forgotten some items or fixes. Let me know if that's the case.

While writing tests, I found that the `getVirtualDisplayViewTransformName` method returns `''` for shared virtual views, even when a view transform is defined. I modified the method similar to my solution for `getVirtualDisplayViewColorSpaceName`, but wanted to flag and relate to #2082.

doug-walker:
Thank you for this update @annieln !  We have a few large high-priority projects happening at the moment, so it may take a few weeks for us to review your latest changes.  I notice that the Linux CI builds failed but that looks to be unrelated to your work. Once we update the build scripts, I will re-run the Linux checks on this PR.  Thanks again!

carolalynn:
Hi Annie! I believe if you update this branch against main, it will re-run the CI! :)

doug-walker (commented):
@annieln, apologies for the delay in getting this reviewed!

I still need to review your unit tests, but there is one addition below I'd like you to make. When you push the commit, we will make sure the CI runs.

src/OpenColorIO/Config.cpp
```diff
@@ -3812,6 +3948,12 @@ const char * Config::getVirtualDisplayViewColorSpaceName(const char * view) cons
 {
     if (!view) return "";
 
+    // Get the colorspace name for the case where a virtual view is shared.
+    if (virtualViewIsShared(view))
+    {
+        return getDisplayViewColorSpaceName(nullptr, view);
+    }
+
```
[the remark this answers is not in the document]
doug-walker: Actually, removeVirtualDisplayView already handles the shared view case, so that one is ok as is.

tests/cpu/Config_tests.cpp
```diff
@@ -8933,6 +9243,238 @@ OCIO_ADD_TEST(Config, virtual_display_exceptions)
...
+        OCIO_REQUIRE_EQUAL(2, config1->getVirtualDisplayNumViews(OCIO::VIEW_SHARED));
+
+        const char * viewName1 = config1->getVirtualDisplayView(OCIO::VIEW_SHARED, 0);
+
+        OCIO_CHECK_EQUAL(std::string("Film"), viewName1);
+        OCIO_CHECK_EQUAL(std::string("display_vt"), config1->getVirtualDisplayViewTransformName(viewName1));
+        OCIO_CHECK_EQUAL(std::string("<USE_DISPLAY_NAME>"), config1->getVirtualDisplayViewColorSpaceName(viewName1));
+        OCIO_CHECK_EQUAL(std::string(""), config1->getVirtualDisplayViewLooks(viewName1));
+        OCIO_CHECK_EQUAL(std::string(""), config1->getVirtualDisplayViewRule(viewName1));
+        OCIO_CHECK_EQUAL(std::string(""), config1->getVirtualDisplayViewDescription(viewName1));
```
doug-walker, line 9354: Looks, rule, and description are actually broken, but the test is not picking it up. Please change the config to use this:
`  - !<View> {name: Film, view_transform: display_vt, display_colorspace: <USE_DISPLAY_NAME>, looks: look1, rule: Linear, description: Test view}`

You may need to add something like this for the config to load:
```
looks:
  - !<Look>
    name: look1
```

tests/cpu/Config_tests.cpp
```diff
@@ -3749,6 +3749,271 @@ OCIO_ADD_TEST(Config, display_view_order)
...
+display_colorspaces:
+  - !<ColorSpace>
+    name: display_cs
+    to_display_reference: !<CDLTransform> {sat: 1.5}
+
+colorspaces:
+  - !<ColorSpace>
+    name: raw
+)" };
+
```
doug-walker, line 3834: Please add a few usages of viewIsShared somewhere in this test, returning both true and false.

tests/cpu/Config_tests.cpp
```diff
@@ -8933,6 +9243,238 @@ OCIO_ADD_TEST(Config, virtual_display_exceptions)
                           " which is not defined.");
 }
```
doug-walker, line 9245: Please move this test to tests/cpu/Display_tests.cpp.

tests/cpu/Config_tests.cpp
```diff
@@ -3749,6 +3749,271 @@ OCIO_ADD_TEST(Config, display_view_order)
     OCIO_CHECK_EQUAL(std::string(config->getView("sRGB_B", 1)), "View_1");
 }
```
doug-walker, line 3751: Please move this test to tests/cpu/Display_tests.cpp.

And please add a test of clearSharedViews to the "basic" test in that module.

tests/python/ConfigTest.py
```diff
@@ -842,12 +838,22 @@ def test_virtual_display(self):
                                              "sRGB")))
         self.assertEqual(1, len(cfg.getViews(OCIO.VIEW_SHARED, "sRGB")))
 
+        self.assertTrue(cfg.displayHasView("sRGB", "sview1"))
+        self.assertTrue(cfg.viewIsShared("sRGB", "sview1"))
```
doug-walker, line 842: For all the functions that return booleans, please have a test where it returns true, another where it returns false, and a third where the arguments are bogus, e.g. a name that's either empty or not in the config.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -883,6 +890,19 @@ class OCIOEXPORT Config
...
     const char * getView(const char * display, const char * colorspaceName, int index) const;
 
+    /**
+     * \brief Compare views in a pair of configs.
+     *
+     * Will return false if either of the views does not exist. This will return true even
+     * if the view is display-defined in one config and a reference to a shared view in the
+     * other config (both within the same display), as long as the contents match. The
+     * description text (if any) is ignored, since it is not a functional difference.
+     */
```
doug-walker, line 900: Please add the following to this and to virtualViewsAreEqual:

Note that the comparison is only on the strings contained in the view definition, the function does not attempt to compare that the color spaces or view transforms being referenced are identical (only that they have the same name).

doug-walker (changes requested):
@annieln, here is the rest of my review. I have a few more requests, but you've done really great work on this!  Big thanks for all the work on the unit tests, we love to see this!

annieln:
I believe I've addressed all the suggested changes - let me know if I missed anything or if there's anything else!

tests/cpu/Config_tests.cpp
doug-walker: @annieln , sorry, my request was not clear.  What I meant was that I wanted you to move your new tests to Display_tests.cpp.  This is because Config_tests.cpp is already too big and so we try to avoid adding new tests there if they could easily go somewhere else.

So please move the new test "compare_displays" and restore the other test "display_view_order".

tests/cpu/Config_tests.cpp
doug-walker: Sorry, my instructions were not clear.  Please move the new test "compare_virtual_displays" to Display_tests.cpp and restore the old test "virtual_display_exceptions" to Config_tests.cpp.

doug-walker:
Your tests look great!  Please ignore the failed Linux CI test, that is due to an unrelated problem in our build setup.

tests/cpu/Config_tests.cpp
annieln: Updated, sorry about that!

doug-walker (approved):
Great work @annieln!!

Apologies for the delay getting it reviewed and thank you for being so diligent in addressing the comments. This is one of the biggest contributions we've received in the history of our Dev Days! These new capabilities will be very useful to the community!

include/OpenColorIO/OpenColorIO.h
```diff
@@ -852,13 +852,20 @@ class OCIOEXPORT Config
      * \ref Config::validate will throw if the config does not contain
      * the matching display color space.
      */
+
+    /// Check if a view within a given display is referencing one of the config's shared views.
+    bool viewIsShared(const char * dispName, const char * viewName) const;
```
cozdas, line 857: This is a comment more on the original issue rather than the PR:

The wording order `{object(s)}{Is/Are}{State}` for the function name is different than the more typical convention of `{is/are}{Object(s)}{State}`. Config class already has function in the second form, like Config::isColorSpaceUsed(), Config::isColorSpaceLinear(). We also have functions named in yet another form like Config::isInactiveColorSpace().

This is a question for @doug-walker I guess: Having the other forms already in the API, do we really want to use this form? Personally, I believe that `isViewShared()` would be a better name choice.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -883,6 +890,23 @@ class OCIOEXPORT Config
...
+     * Will return false if either of the views does not exist. This will return true even
+     * if the view is display-defined in one config and a reference to a shared view in the
+     * other config (both within the same display), as long as the contents match. The
+     * description text (if any) is ignored, since it is not a functional difference.
+     * 
+     * Note that the comparison is only on the strings contained in the view definition,
+     * the function does not attempt to compare that the color spaces or view transforms
+     * being referenced are identical (only that they have the same name).
+     */
+    static bool ViewsAreEqual(const ConstConfigRcPtr & first,
```
cozdas, line 905: `areViewsEqual()` ?

include/OpenColorIO/OpenColorIO.h
```diff
@@ -900,6 +924,17 @@ class OCIOEXPORT Config
...
+    /**
+     * \brief Determine if a display and view exist.
+     *
+     * This returns false if either the display or view doesn't exist. It works regardless
+     * of whether the display or view are active, and it works regardless of whether the
+     * view is display-defined or if the display has this as a shared view. It will only
+     * check config-level shared views if dispName is null. It will not check config level
+     * shared views if dispName is not null.
+     */
+    bool displayHasView(const char * dispName, const char * viewName) const;
```
cozdas, line 936: Since this function is valid for the dispName being nullptr, I think we can drop the display in the name of the function making it more general. previous arguments I'd suggest `hasView()`

include/OpenColorIO/OpenColorIO.h
```diff
@@ -963,6 +998,18 @@ class OCIOEXPORT Config
...
+     * \brief Determine if a virtual view exists.
+     *
+     * This returns false if the virtual view doesn't exist. It works regardless of
+     * whether the virtual view is active, and it works regardless of whether the virtual
+     * view is display-defined or if the display has this as a shared virtual view.
+     */
+    bool hasVirtualView(const char * viewName) const;
+
+    /// Check if a given virtual view is referencing one of the config's shared views.
+    bool virtualViewIsShared(const char * viewName) const;
```
cozdas, line 1011: `isVirtualViewShared()` ?

include/OpenColorIO/OpenColorIO.h
```diff
@@ -977,6 +1024,23 @@ class OCIOEXPORT Config
...
+     * even if the virtual view is display-defined in one config and a reference to a shared
+     * virtual view in the other config, as long as the contents match.
+     * 
+     * The description text (if any) is ignored, since it is not a functional difference.
+     * 
+     * Note that the comparison is only on the strings contained in the view definition,
+     * the function does not attempt to compare that the color spaces or view transforms
+     * being referenced are identical (only that they have the same name).
+     */
+    static bool VirtualViewsAreEqual(const ConstConfigRcPtr & first,
```
cozdas, line 1040: `areVirtualViewsEqual()` ?

src/OpenColorIO/Config.cpp
```diff
@@ -3305,6 +3305,21 @@ void Config::setViewingRules(ConstViewingRulesRcPtr viewingRules)
     getImpl()->resetCacheIDs();
 }
 
+bool Config::viewIsShared(const char * dispName, const char * viewName) const
+{
+
```
cozdas, line 3310: we're checking the `getView()` result against null or empty string, so the passed parameter `viewName` should be checked here as well.

src/OpenColorIO/Config.cpp
```diff
@@ -3358,6 +3373,19 @@ void Config::removeSharedView(const char * view)
...
 
+void Config::clearSharedViews()
+{
+    int numViews = getNumViews(VIEW_SHARED, nullptr);
+    for (int v = numViews - 1; v >= 0; v--)
+    {
+        const char * sharedViewName = getView(VIEW_SHARED, nullptr, v);
+        if (sharedViewName && *sharedViewName)
+        {
+            removeSharedView(sharedViewName);
```
cozdas, line 3384: A little bit inefficient implementation as removeSharedView() will call FindView() to loop again to find it but it's fine as run-time cost will be very small.

src/OpenColorIO/Config.cpp
```diff
@@ -3695,6 +3775,28 @@ void Config::clearDisplays()
...
+{
+    const char * cs = getVirtualDisplayViewColorSpaceName(viewName);
+
+    // All views must have a color space, so if it's not empty, the view exists.
+    return (cs && *cs);
+}
+
+bool Config::virtualViewIsShared(const char * viewName) const
+{
+    for (int v = 0; v < getVirtualDisplayNumViews(VIEW_SHARED); v++)
```
cozdas, line 3788: `viewName` should be checked against being null or empty too.

cozdas:
Great work!
I added few comments, please check.

include/OpenColorIO/OpenColorIO.h
cozdas: Fair point. If we decide to, we can rename the functions before releasing a new version.

annieln:
I've added the null/empty view name check to `viewIsShared` and `virtualViewIsShared`. I've also included a couple of lines for these checks in the `virtual_display` test in Config_tests.cpp. Let me know if there's anything else I should address!

cozdas (approved):
Approved.
Thanks for the great work.

doug-walker:
Merging this as the issues with the failing documentation-related CI tests are unrelated to this PR.

Thanks again, @annieln !!

PR #2090, opened 2024-11-11
Adsk Contrib - Fix bit-depth attributes on inverse LUT writing to CTF format

doug-walker:
When writing CTF files with inverse LUT1D or LUT3D, the in & out bit-depth attributes were swapped.

The scaling of the Array values for a forward LUT is communicated via the outBitDepth. An inverse LUT writes the same Array as the forward LUT, but the in and out depths need to be swapped. So for inverse LUTs, it's the inBitDepth that signals the Array scaling.

This PR fixes the issue, existing tests were reviewed, and new tests were added to ensure that inverse Lut1D and Lut3D are both read and written correctly with integer bit-depth settings.


zachlewis (approved):

zachlewis (approved):

cozdas (approved):
Look good

PR #2094, opened 2024-11-15
Update for Adobe Photoshop and AfterEffects

kingsawyer:
Both have moved from plugin to native support


kingsawyer:
Looks like the Windows version of Hugo puts out different line endings. Try diff ignoring whitespace and line endings. Notice the generator changed from "Hugo -- gohugo.io" to "Hugo"

zachlewis:
Hmm. I think instead of altering the rendered html on the `gh-page` branch, you should adjust [docs/site/homepage/data/en/supported_apps.yml](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/docs/site/homepage/data/en/supported_apps.yml) on the `main` branch. 

I _believe_ if we merge a change to that file, the relevant site-building gh workflow will trigger.

carolalynn:
We don't have auto-generation set up for the website. You do have to commit to the gh-pages branch separately, for now.

Time permitting, I'd love to get an action set up that does this auto-magically, for sure.

Super weird about the white space - I work on a Mac. If I get some time (likely won't be till this weekend) I can give it a spin and confirm suspicions.

zachlewis:
Ah, my mistake!

Still, we need to make sure the change is reflected in supported_apps.yml in the main branch, or these changes will be lost the next time we rebuild the site...

kingsawyer:
@zachlewis we thought alike. Turns out there's one more step. And it was documented. https://github.com/AcademySoftwareFoundation/OpenColorIO/tree/main/docs/site#building-the-static-website

zachlewis:
lol dammit zach!

zachlewis:
> Time permitting, I'd love to get an action set up that does this auto-magically, for sure.

Alright, I've submitted PR #2098, which should hopefully do exactly this. 

Merging that PR should itself invoke a rebuild and republish of the site, which should effectively reproduce the exact changes made here.

(FWIW, rebuilding the site with the latest version of Hugo seems to elicit the same differences we're seeing in here, whether invoked on a Mac locally or via an Ubuntu GH runner. I think the line-ending changes etc. are explained by the fact that the site was last built with Hugo 1.11 vs the current 1.39)

kingsawyer:
@zachlewis I guess we still need this pushed or do you want to try pushing something trivial and see if the supported apps section changes?

zachlewis (approved):

zachlewis:
Ah, thanks for reminding me -- immediately after merging #2098, the new site-build-and-deploy workflow properly triggered and built; but the workflow itself [didn't have the necessary permissions](https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/12015624905/job/33494082567#step:6:317) to do the actual publishing. We noted this briefly in our last TSC, but I forgot to follow up with Doug.

Basically, we need to either relax permissions for pushing to the gh-pages branch, mess around with creating a workflow-specific Personal Access Token, or by adjusting our settings to use a GH workflow directly as the means for deploying the site (instead of updating a gh-pages branch). In any case, something I need to look more into elsewhere.


In the mean time, though, I think we agreed we'd be totally fine with merging your PR just to get the page updated already! 

(We appreciate your patience)

@doug-walker / @carolalynn what do you guys think? Merge this sucker?

carolalynn (approved):
yeah agreed fine with me!

PR #2080, opened 2024-10-15
Combine LUTs use pairIdentityReplacement when possible

remia:
In some cases a pair of inverse / forward LUTs don't get optimized away by OCIO but instead combined in a single half-domain LUT. This seems to happen when there are a number of other operators in between, like a pair of MatrixOp to handle domain scales.

For context, the example I saw this in our config was in the case of an ACEScg to ARRI LogC4 -> CDL (identity) -> ARRI LogC4 to ACEScg conversion, the LogC4 LUT was designed to preserve -1 to 500 range. Here is the Group transform from the Processor for reference. Note that the ExponentOp clamp is ignored when optimizing.
```
<MatrixTransform direction=forward, fileindepth=unknown, fileoutdepth=unknown, matrix=0.9183876682136227 0.02254918208978807 0.05906314969658851 0 0.04367542261395546 0.8533787609135172 0.1029458164725261 0 -0.005022624839390726 0.004758976255720934 1.00026364858367 0 0 0 0 1, offset=0 0 0 0>
<Lut1DTransform direction=inverse, fileoutdepth=32f, interpolation=best, inputhalf=0, outputrawhalf=0, hueadjust=0, length=8192, minrgb=[-1 -1 -1], maxrgb=[500 500 500]>
<MatrixTransform direction=forward, fileindepth=unknown, fileoutdepth=unknown, matrix=1.430002152919769 0 0 0 0 1.430002152919769 0 0 0 0 1.430002152919769 0 0 0 0 1, offset=-0.4241785407066345 -0.4241785407066345 -0.4241785407066345 -0>
<MatrixTransform direction=forward, fileindepth=unknown, fileoutdepth=unknown, matrix=1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1, offset=0 0 0 0>
<ExponentTransform direction=forward, value=1 1 1 1, style=clamp>
<MatrixTransform direction=forward, fileindepth=unknown, fileoutdepth=unknown, matrix=1 0 0 0 0 1 0 0 0 0 1 0 0 0 0 1, offset=0 0 0 0>
<MatrixTransform direction=forward, fileindepth=unknown, fileoutdepth=unknown, matrix=0.6992996464782982 0 0 0 0 0.6992996464782982 0 0 0 0 0.6992996464782982 0 0 0 0 1, offset=0.2966279035598299 0.2966279035598299 0.2966279035598299 0>
[...]

PR #2104, opened 2024-12-08
Adsk Contrib - Optimizer should detect pair inverses before combining multiple op pairs

doug-walker:
This is an alternate fix to [an optimizer issue identified by Remi](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/2080).

Currently, if a pair of fwd/inverse Lut1D ops are separated by ops that combine into an identity, it results in the Lut1D ops being combined as well. The result is a combined Lut1D rather than a Range.

The proposed fix here is to combine at most one pair of ops per pass of the optimizer. This allows the op chain to get another pass through more preferred techniques, such as removing no-ops and pair identities, before trying to combine more op pairs.

Because this could require more passes, MAX_OPTIMIZATION_PASSES was raised from 8 to 80. This should have no performance impact since the optimizer stops iterating as soon as no further improvements are found, so in practice it never gets anywhere near the max passes limit.


src/OpenColorIO/Config.cpp
```diff
@@ -5268,7 +5268,9 @@ void Config::Impl::checkVersionConsistency(ConstTransformRcPtr & transform) cons
                         || 0 == Platform::Strcasecmp(blt->getStyle(), "ACES-OUTPUT - ACES2065-1_to_CIE-XYZ-D65 - HDR-500nit-REC2020-D60-in-REC2020-D65_2.0")
                         || 0 == Platform::Strcasecmp(blt->getStyle(), "ACES-OUTPUT - ACES2065-1_to_CIE-XYZ-D65 - HDR-1000nit-REC2020-D60-in-REC2020-D65_2.0")
                         || 0 == Platform::Strcasecmp(blt->getStyle(), "ACES-OUTPUT - ACES2065-1_to_CIE-XYZ-D65 - HDR-2000nit-REC2020-D60-in-REC2020-D65_2.0")
-                        || 0 == Platform::Strcasecmp(blt->getStyle(), "ACES-OUTPUT - ACES2065-1_to_CIE-XYZ-D65 - HDR-4000nit-REC2020-D60-in-REC2020-D65_2.0") )
+                        || 0 == Platform::Strcasecmp(blt->getStyle(), "ACES-OUTPUT - ACES2065-1_to_CIE-XYZ-D65 - HDR-4000nit-REC2020-D60-in-REC2020-D65_2.0")
+                        // NB: This one was added in OCIO 2.4.1.
+                        || 0 == Platform::Strcasecmp(blt->getStyle(), "DISPLAY - CIE-XYZ-D65_to_DisplayP3-HDR") )
```
doug-walker, line 5273: This one-liner is unrelated to the optimization issue. It's just a validation check I noticed was missing related to the new built-in that Thomas added recently.

zachlewis (approved):

src/OpenColorIO/Config.cpp
doug-walker: Yes, agreed. There will be configs that pass this validation but would not work in a 2.4.0 library, which is not ideal, but probably acceptable in this case.

PR #2109, opened 2024-12-17
Fix issue with ocio_depts handling spaces in file paths

DevTGHa:
`share/dev/windows/ocio_deps.bat` was not able to find the installed visual studio location due to the space in the paths. This is because `for /f` splits on spaces and tabs by default.

For example, `C:\Program Files (x86)` becomes `C:\Program`.

Adding `delims=` to avoid splitting the output of the command and keeps the entire path in `%%i`.


cozdas (approved):
Good catch! Thanks for fixing this.

doug-walker (approved):

Issue #2111, opened 2025-01-03
FileTransforms with absolute paths should work with configs from OCIOZ archive files


doug-walker:
When an OCIOZ archive file is used as the config, FileTransforms are loaded from the .ocioz file. Currently, if an absolute path is used with a FileTransform, it won't work since it won't exist in the archive (OCIO restricts what it archives to only support paths that are relative to the working directory). 

However, an application sometimes needs to load LUT files that are not part of the config (e.g., the user simply wants to apply a stand-alone LUT file). Currently the only way to do this is to create the Processor using a config that is different from the .ocioz file. This complicates the situation for applications since they would need to have different code paths based on whether the config is an archive or not. However, the intent for archives is that they be usable the same as a normal config. 

The proposal is that FileTransforms that use absolute paths should be loaded directly from the file system and not via the .ocioz file. This should apply to all ConfigIOProxy workflows, not only OCIOZ archives.

Here is a test that may be added at line 438 in tests/cpu/OCIOZArchive_tests.cpp that demonstrates the problem:
```
        {
            const std::string filePath = OCIO::GetTestFilesDir() + "/matrix_example4x4.ctf";
            OCIO::FileTransformRcPtr transform = OCIO::FileTransform::Create();
            transform->setSrc(filePath.c_str());
            OCIO::ConstProcessorRcPtr processor = cfg->getProcessor(transform);
            OCIO::ConstTransformRcPtr tr = processor->createGroupTransform()->getTransform(0);
            auto mtx = OCIO::DynamicPtrCast<const OCIO::MatrixTransform>(tr);
            OCIO_REQUIRE_ASSERT(mtx);
            mtx->getMatrix(mat);
            OCIO_CHECK_EQUAL(mat[0], 3.24);
        }
```

doug-walker:
This was fixed in PR #2112 and will be in OCIO 2.4.2.

PR #2112, opened 2025-01-17
Adsk Contrib - Issue #2111 Absolute paths not working through proxy

cozdas:
Fixing issue #2111 
- Added fallback mechanism for files with absolute paths so that if the proxy doesn't provide hash or the content, the file system will be used. We still expect the proxy to provide files with relative paths.

- Removed closeLutStream() function as both the ifstream and the stringstream objects that getLutData may return use RAII idiom. Since we already use unique pointers to hold the objects, we don't need to worry about closing the file or freeing memory.

- Added OCIOZ unit tests for existing and non-existing files with absolute paths.

### DIFF
--- a/src/OpenColorIO/transforms/FileTransform.cpp
+++ b/src/OpenColorIO/transforms/FileTransform.cpp
@@ -190,7 +190,7 @@ std::unique_ptr<std::istream> getLutData(
     const std::string & filepath, 
     std::ios_base::openmode mode)
 {
-    if (config.getConfigIOProxy())
+    if (!pystring::os::path::isabs(filepath) && config.getConfigIOProxy())
     {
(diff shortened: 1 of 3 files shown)


src/OpenColorIO/transforms/FileTransform.cpp
```diff
@@ -635,9 +633,8 @@ void LoadFileUncached(FileFormat * & returnFormat,
                 filepath, 
                 tryFormat->isBinary() ? std::ios_base::binary : std::ios_base::in 
             );
-            auto & filestream = *pStream;
             
-            if (!filestream.good())
```
cozdas, line 640: Not good if pStream becomes null somehow. Sticking with the pointer and adding a null check instead of using a reference.

remia (approved):

doug-walker (approved):

PR #2113, opened 2025-01-20
Fix deploy site workflow

zachlewis:
New and improved github workflow for building and deploying our Hugo-based site. 

Instead of maintaining a "gh-pages" branch, this workflow uses the preferred "deploy-site" GH action. This requires updating the repo settings (Code and Automation > Pages > Build and Deploy) to use the workflow-based deployment settings, as opposed to the classic branch-based deployment settings. 

(NB: I've already updated the repo settings in anticipation of merging this PR).

Also, I haven't thoroughly tested this. The workflow is based on github's template for building and deploying Hugo sites, but if something goes wrong... well... I'll try again with another PR!


doug-walker:
@zachlewis , currently the original OCIO v1 website is being maintained as part of the site [at this link.](https://opencolorio.org/old/index.html) We reference it on the landing page of the documentation [here](https://opencolorio.readthedocs.io/en/latest/). Since that is in the gh_pages branch, and is not generated from Hugo, I'm guessing that would be dropped. Am I correct?

At this point, I'm not sure we still need to preserve the old website. Any opinions on that?

zachlewis:
Ah, yes, I forgot to mention that. You are correct, we would no longer be preserving the old site with this workflow. If that's a concern, we can work something else out that still uses the "classic" gh-pages branch method.

carolalynn:
I think we can say goodbye to our old site and remove the link :)

carolalynn (approved):
Let's gooooo!

remia (approved):

doug-walker:
This seems to be working beautifully, thank you @zachlewis !

carolalynn:
Amazing, thanks both for getting this working!

PR #2115, opened 2025-02-03
ci: Use ARM-native runners for linux ARM Python wheels

zachlewis:
Apparently, as of almost two weeks ago, our Linux ARM Python Wheels nightlies have gone haywire and are no longer completing successfully. 

Previously, we were using QEMU to emulate the ARM builds on x86 runners. This PR adds a "linux-arm" job for building the ARM-flavored linux wheels, using the new `ubuntu-24.04-arm` now-freely-available runner, which seems to keep things copacetic. 

I've also removed the bits where we install QEMU, since it's no longer needed.


remia (approved):
Thanks a lot for the fix @zachlewis, it's great to have the wheel workflow run in reasonable time as well without QEMU!

zachlewis:
Thanks for reviewing, Remi!


As discussed in the TSC meeting, I should add a linux-ARM CI check as well.

remia:
Note that the QEMU wheels are working again, but I think it would still be good to have this merged for the next release. I don't necessarily see updating the CI as a requirement for this PR as we have been shipping Linux ARM wheel for a while now and they are tested as part of the cibuildwheel process.

doug-walker (approved):
LGTM!

doug-walker:
@zachlewis , I merged main into your branch in preparation for merging this but the CI didn't pass. Please take a look when you have time.

zachlewis:
I know exactly what's going on... Apparently the GH runners were all updated to use CMake 4.0, which causes problems for zlib. (See https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4688)

One moment...

doug-walker:
Awesome, thanks for solving that Zach!

Issue #1764, opened 2023-02-09
Negative input values resulting in unexpected output values on some CPUs


mjtitchener-fn:
I've been investigating a report of negative input values resulting in unexpected output values on some CPUs.
We have been able to reproduce the issue on one of our Windows machines and it appears to be happening in the OCIO library.

### Reproduction steps:
1. Using input values R: `0.00532` G: `0.00075` B: `-0.00293`
2. Transform from `ACEScg` to `ACEScc`.

### Actual result:
R: `0.11312` G: `-0.07889` B: `1.50000`

### Expected result:
R: `0.11312` G: `-0.07889` B: `-0.31065`

Here's a visual example of the differences seen. Top is actual result, bottom is expected result.

<img width="408" alt="cpu_differences" src="https://user-images.githubusercontent.com/72390925/217841723-7a700d69-7b32-40ca-ac31-83366231bf46.png">

The user experiences this issue on machines with `Dual Intel Xeon E5-2680 v2 @ 2.80GHz` but doesn’t on machines with `Dual Intel Xeon E5-2640 v3 @ 2.60GHz`

The machine we can reproduce the issue on has `Dual Intel Xeon E5-2640 0 @ 2.50GHz`

One difference between the CPUs that stood out to me is Intel AVX (reproducible) vs. Intel AVX2 (not reproducible). Not sure if this is relevant though.

So far we’ve only seen this on Windows. We can only reproduce this in our latest Nuke release (14.0) but the report says they’ve encountered it since Nuke was upgraded to use OCIO v2.
It’s reproducible with the ACES 1.2 and ACES Studio configs.

### Workaround:
I've found you can workaround the issue by setting the environment variable `OCIO_OPTIMIZATION_FLAGS=0` to turn optimization off, or `OCIO_OPTIMIZATION_FLAGS=20479` to set optimization to lossless.
In the case where the ACES Studio config is being used switching from `ACEScc` to `ACEScct` also fixes the issue.

doug-walker:
@mjtitchener-fn , I'm not able to reproduce the bug.

I'm uploading a short Python script.  Could you please run it on the machine that shows the problem and post the result here?
[bug_1764.py.zip](https://github.com/AcademySoftwareFoundation/OpenColorIO/files/10773531/bug_1764.py.zip)

mjtitchener-fn:
Sure, no problem @doug-walker. 
[...]

PR #2120, opened 2025-02-21
Change recommended Imath version to 3.1.12.

mjtitchener-fn:
Chage the recommended Imath version to 3.1.12 (latest available).

Imath 3.1.10 included a fix that should resolve Issue #1764.

Signed-off-by: Mark Titchener <mark.titchener@foundry.com>

### DIFF
--- a/share/cmake/modules/FindExtPackages.cmake
+++ b/share/cmake/modules/FindExtPackages.cmake
@@ -77,7 +77,7 @@ ocio_handle_dependency(  pystring REQUIRED ALLOW_INSTALL
 # https://github.com/AcademySoftwareFoundation/Imath
 ocio_handle_dependency(  Imath REQUIRED ALLOW_INSTALL
                          MIN_VERSION 3.1.1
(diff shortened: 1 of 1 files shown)


remia:
There is an important bug fix in [3.1.13](https://github.com/AcademySoftwareFoundation/openexr/releases/tag/v3.1.13) related to decoding of 32 bits float DWA compressed EXRs. I would vote to jump directly to this instead!

mjtitchener-fn:
Thanks @remia. So you mean bump the OpenEXR version as part of this too? There isn't a 3.1.3 version of Imath.

remia:
Sorry @mjtitchener-fn indeed I misread you are correct, I thought the change was on OpenEXR.

No need to change anything then, unless we want to do an upgrade pass on every deps default version.

carolalynn:
We should do a pass through our deps versions for 2.5 maybe, and keep it to
imath for 2.4.2?

remia (approved):

doug-walker (approved):

PR #2121, opened 2025-02-21
Integrating matrix multiplication fix from OSL

JGamache-autodesk:
See https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1513 for more details.

As requested in review of MaterialX https://github.com/AcademySoftwareFoundation/MaterialX/pull/1917

See similar fix in MaterialX here: https://github.com/AcademySoftwareFoundation/MaterialX/pull/972


doug-walker (approved):

cozdas (approved):
Verified that the transform() function in OSL multiplies by the transposed matrix.
Approved.

PR #2123, opened 2025-02-26
Address GPU unit test failures

cozdas:
- Weights for cos(3h) and sin(h) in chroma_compress_norm() looks wrong. Fixing the weights makes the forward GPU tests pass now (except for the inverse output transform which had a separate issue).

- Updated the CPU test target values accordingly.

- The remaining GPU test failures were caused by a simple typo where we were passing h instead of J to ocio_tonescale_inv() function. With the fix all the unit tests are happy now.

- Since we decided not to include any SIMD implementation in this version, I removed the conditional code paths and left the current SSE & AVX implementations as commented out for future guidance.


src/OpenColorIO/ops/fixedfunction/ACES2/Transform.cpp
```diff
@@ -302,8 +302,8 @@ float chroma_compress_norm(float cos_hr1, float sin_hr1, float chroma_compress_s
         sin_hr1, sin_hr2, sin_hr3, 1.0f
     };
     alignas(AVX_ALIGNMENT) static constexpr float weights[8] = { // TODO: investigate reordering of the entries so we are summing equal magnitude values first?
-        11.34072f, 16.46899f, 14.66441f,  0.0f,
```
cozdas, line 305: Based on https://draftdocs.acescentral.com/output-transforms/technical-details/chroma-compression/#normalization the GPU side seems to have the correct weights thus updated the CPU-side. Please verify.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -682,14 +683,15 @@ void _Add_ChromaCompressionNorm_Shader(
     ss.newLine() << "{";
     ss.indent();
 
-    ss.newLine() << ss.floatDecl("cos_hr2") << " = cos_hr * cos_hr - sin_hr * sin_hr;";
+    // TODO: optimization: can bake weights into terms and convert dotprods to addition. /coz
+    ss.newLine() << ss.floatDecl("cos_hr2") << " = 2.0 * cos_hr * cos_hr - 1.0;";
```
cozdas, line 687: Will slightly be better in terms of robustness and speed, I guess.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -663,6 +663,7 @@ std::string _Add_Tonescale_func(
     }
 
     ss.newLine() << ss.floatDecl("J_ts") << " = " << ACES2::J_scale << " * pow((F_L_Y / ( " << ACES2::cam_nl_offset << " + F_L_Y)) * " << p.inv_A_w_J << ", " << p.cz << ");";
+    // TODO: copysign is missing here. /coz
```
cozdas, line 666: see https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/89ed0fb201a7a1d40a1f5cef725c293c2863b88b/src/OpenColorIO/ops/fixedfunction/ACES2/Transform.cpp#L393

src/OpenColorIO/ops/fixedfunction/ACES2/Transform.cpp
KevinJW: yes, this looks like an editing mistake on my part.

doug-walker (approved):

doug-walker:
Deciding to go ahead and merge this given:
- Kevin's positive comment above
- The changes here are basically fixing typos on both the CPU & GPU side
- The changes now get all the existing unit tests to pass
- We want to get Alex a fixed CPU version to continue his testing

PR #2124, opened 2025-03-05
Remove unused code for old gamut table calculations

KevinJW:
This code is no longer  used since we have changed the hue table hue distribution


doug-walker (approved):

cozdas (approved):
Yay! spring cleaning.

PR #2125, opened 2025-03-06

cozdas:
This is a slightly polished version of @KevinJW's optimization work. Attaching below Kevin's original PR comment

----------------------

1. Precomputing a number of the scaling factors, etc during the init stage

General observations

GPU shader generation notes:

- log10()
- atan2()


cozdas:
This turned out to have some sign-off confusion for some reason. I'll re-create the PR.

PR #2127, opened 2025-03-07

cozdas:
This is a slightly polished version of @KevinJW's optimization work. Attaching below Kevin's original PR comment

----------------------

1. Precomputing a number of the scaling factors, etc during the init stage

General observations

GPU shader generation notes:

- log10()
- atan2()


src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -532,14 +614,99 @@ std::string _Add_Toe_func(
...
+    else
+    {
+        // Tonescale applied in Y (convert to and from J)
+        ss.newLine() << ss.floatDecl("f") << " = " << t.m_2  << " * pow(Y / (Y + " << double(t.s_2) * p.F_L_n << "), " << t.g << ");";
+        ss.newLine() << ss.floatDecl("Y_ts") << " = max(0.0, f * f / (f + " << t.t_1 << "));";
+        ss.newLine() << ss.floatDecl("F_L_Y") << " = pow(" << double(p.F_L_n) * ACES2::reference_luminance << " * Y_ts, 0.42);";
+    }
+
+    ss.newLine() << ss.floatDecl("J_ts") << " = " << ACES2::J_scale << " * pow((F_L_Y / ( " << ACES2::cam_nl_offset << " + F_L_Y)) * " << p.inv_A_w_J << ", " << p.cz << ");";
+    ss.newLine() << "return J_ts;";
```
cozdas, line 666: abs(J) in line 643 above suggests we're expecting J to be negative, but here we're not copying the sign of it at the end to take a symmetry as the CPU implementation does.

see https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/89ed0fb201a7a1d40a1f5cef725c293c2863b88b/src/OpenColorIO/ops/fixedfunction/ACES2/Transform.cpp#L393

src/OpenColorIO/ops/fixedfunction/ACES2/Transform.cpp
```diff
@@ -13,214 +17,321 @@ namespace ACES2
...
+    return Aab;
+}
 
-    const float PI = 3.14159265358979f;
-    const float h_rad = std::atan2(b, a);
-    float h = std::fmod(h_rad * 180.f / PI, 360.f);
-    if (h < 0.f)
+f3 Aab_to_JMh(const f3 &Aab, const JMhParams &p)
+{
+    if (Aab[0] <= 0.f)
```
remia, line 190: The test `style_aces2_output_transform_invfwd` is failing on my AMD macOS GPU OpenGL tests (but strangely it passes on Metal). If I add this early exit condition in the GPU path it seem to fix the issue.

```
10: [ 42/228] [FixedFunction / style_aces2_output_transform_invf ] - FAILED - 
10: Maximum error: 9.620073318 at pixel: 0 on component 2 larger than epsilon.
10: scr = {-1, -0.9999885559, -0.9999771118, -0.9999656677}
10: cpu = {0, 0, 0, -0.9999656677}
10: gpu = {9.618265152, 9.619031906, 9.620073318, -0.9999656677}
10: absolute tolerance=0.000699999975
```

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -349,57 +349,143 @@ void Add_GamutComp_13_Inv_Shader(GpuShaderText & ss,
...
+void _Add_Aab_to_JMh_Shader(
+    GpuShaderCreatorRcPtr & shaderCreator,
+    GpuShaderText & ss,
+    const ACES2::JMhParams & p)
+{
+    const std::string pxl(shaderCreator->getPixelName());
+
+    ss.newLine() << "{";
+    ss.indent();
+
```
doug-walker, line 404: @cozdas , please add something here per Remi's suggestion, so that it returns {0,0,0} if A is <= 0, so that it more closely follows the CPU path and will not be subject to various GPU negative handling in pow().

cozdas: Sure!
Will changing that make the "J==0.0" check below on line 407 for M computation redundant as well?

src/OpenColorIO/ops/fixedfunction/ACES2/Transform.cpp
cozdas: added the same trap in GPU code, please verify that it fixed the issue on your end.

remia: Thank you @cozdas, it does fix the failure.

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
cozdas: Added the copysign and a TODO comment per our discussion in the meeting.

remia:
While I didn't review in details the line by line changes, I tested that this was indeed improving speed and had a similar or better roundtrip than 2.4.1. The initialisation time (LUTs creations) also seems similar compared to 2.4.1. I also ran the GPU tests on a few different platforms (macOS GLSL and MSL, Linux GLSL and Windows GLSL) successfully.

In previous work in progress commits I saw issues with C++11 support but this is now fixed.

Also checked for NaN in the D65 roundtrips in the CMS and Cube faces images.

Great work @KevinJW !

zachlewis (approved):
The improvements seem well reasoned... although I can't claim to understand everything that's happening, nor can I really speak to the relative efficiency of the code... the code looks good and smart and not malicious and well-tested to me, and doesn't cause anything to break when I try to build it!

doug-walker:
Thank you @KevinJW , @cozdas , @remia , and @zachlewis for your work on this!  I will merge this for inclusion in OCIO 2.4.2.

remia:
@doug-walker it looks like the builtin transform round trip tests are triggering a warning that makes the platform latest workflow [fail](https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/13867396289/job/38809091160#step:6:2680). Will need to initialise all three variables to 0 to make it go away.

PR #2132, opened 2025-03-17
Propose NaN fix for the ACES2 inverse output transforms

doug-walker:
The Aab_to_RGB function on both CPU and GPU does a division by a number which may be zero. Indeed, it is 0. at least for the D60 transforms with equal RGB values that are below but approaching the target nit level. This PR proposes the same fix for CPU and GPU that limits a value to prevent the divide by zero.

In addition, quite a few additional unit tests were added.

I processed the cms and cube_faces test images through all inverse transforms (D65 & D60) and verified that my proposed fix does not cause other significant changes (other than removing the NaNs) to the resulting ACES values.

I checked to see if the OCIO 2.4.1 code produced NaNs for the D60 inverses and it does, although in a different place along that equal code value ramp.

While working on this, I noticed that there was a bug in the way the GPU unit tests handle NaN values. This was masking several failures in unrelated tests, so I commented out several failing tests until those are fixed. For some others I turned off the Infinity or NaN part of the test. These should be revisited and addressed.

### DIFF
--- a/tests/gpu/FixedFunctionOp_test.cpp
+++ b/tests/gpu/FixedFunctionOp_test.cpp
@@ -1173,8 +1173,8 @@ OCIO_ADD_GPU_TEST(FixedFunction, style_RGB_TO_HSV_fwd)
 
     test.setErrorThreshold(1e-6f);
 
-#ifdef __APPLE__
     test.setTestNaN(false);
+#ifdef __APPLE__
     test.setTestInfinity(false);
 #endif
 }
--- a/tests/gpu/FixedFunctionOp_test.cpp
+++ b/tests/gpu/FixedFunctionOp_test.cpp
@@ -474,7 +474,9 @@ OCIO_ADD_GPU_TEST(FixedFunction, style_aces2_output_transform_invfwd)
 namespace
 {
 
-OCIO::GroupTransformRcPtr BuildRoundTripTransform(const char * display_style, const char * view_style)
+OCIO::GroupTransformRcPtr BuildDisplayViewTransform(const char * display_style, 
+                                                    const char * view_style, 
+                                                    bool doRoundTrip)
 {
     // Built-in transform for the display.
     OCIO::BuiltinTransformRcPtr display_builtin = OCIO::BuiltinTransform::Create();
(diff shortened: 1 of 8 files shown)


tests/gpu/FixedFunctionOp_test.cpp
```diff
@@ -491,18 +493,21 @@ OCIO::GroupTransformRcPtr BuildRoundTripTransform(const char * display_style, co
...
     // Assemble inverse and forward transform into a group transform that goes from
-    // display code values to ACES and back to code values.
+    // display code values to ACES2065-1 and (optionally) back to display code values.
     OCIO::GroupTransformRcPtr group = OCIO::GroupTransform::Create();
     group->appendTransform(display_builtin_inv);
     group->appendTransform(view_builtin_inv);
-    group->appendTransform(view_builtin);
-    group->appendTransform(display_builtin);
-
+    if (doRoundTrip)
```
cozdas, line 500: If I'm not mistaken, GPU unit tests compare optimized CPU and GPU processors with default optimization.
If that's the case, I wonder if fwd and inv transforms will be optimized out.

cozdas: Ah! line 537 below the my answer to my question. 
It may be a good future addition to have an optimization parameter to make it more explicit and robust though.

tests/gpu/FixedFunctionOp_test.cpp
```diff
@@ -1206,15 +1370,18 @@ namespace HLG
...
-OCIO_ADD_GPU_TEST(FixedFunction, style_LIN_TO_GAMMA_LOG_fwd)
-{
-    auto func = OCIO::FixedFunctionTransform::Create(OCIO::FIXED_FUNCTION_LIN_TO_GAMMA_LOG, HLG::params, 10);
-    func->setDirection(OCIO::TRANSFORM_DIR_FORWARD);
-
-    test.setWideRangeInterval(-0.1f, 3.35f); // Output ~[-0.3, 1.02]
-    test.setProcessor(func);
-    test.setErrorThreshold(1e-6f);
-}
+// FIXME: This no longer passes after fixing a bug in GPUUnitTest.cpp.
```
cozdas, line 1373: I need to look closer, but my guess is that for both fixed function GPU implementations to process vec3 values in a single, branch-free pass, we compute all the branches and the combine them with an arithmetic operation. Something like:

` = isSegment1 * logSeg1 + isSegment2 * linSeg + isSegment3 * logSeg2` 

where isSegmentN can be either 1.0 or 0.0. This will convert any INF values that may result in a "non-taken" branch to NAN as `0 * INF == NAN`.

cozdas (approved):
Possible effects of clamping the Ra value is above my understanding of the ACES2 output transform math, but I know Nick and Doug agreed on that change. Rest of the code changes look good to me.

doug-walker:
Note: Although the GPU CI run failed on this last commit, it was due to an unrelated issue. Merging this per the discussion at the TSC meeting yesterday.

PR #2138, opened 2025-03-21
Add AJA ColorBox as a supported application

seanblit:
- Add AJA ColorBox as supported hardware to the home page
- Add AJA OG-ColorBox as supported hardware to the home page
- Add logos for ColorBox and OG-ColorBox
- Add a ColorBox entry to the Compatible Software page in the documentation


doug-walker (changes requested):
Thanks for this PR Sean!  We would definitely love to add ColorBox to the web page!

However, for other products we generally only include one logo image to represent all members of a family.  So for example, there is only one logo for all of the Nuke family products (Nuke, Nuke Studio, etc.), and only one logo for all the Maya family products (Maya, Maya Creative, Maya Indie, etc.).  The ColorBox and OG-ColorBox seem like they fall into the same situation.  Would you be willing to trim this down to one image? You could potentially use the name field to mention both in the tooltip. And of course both will be listed in the Compatible products list.

seanblit:
Doug thanks for the feedback, I have made the changes as requested and rebased onto the latest main.

doug-walker (approved):
Thanks Sean!

carolalynn (approved):
Exciting! Thanks, Sean!

Issue #2137, opened 2025-03-21
FTBFS with Clang + Arm64 on Windows


lazka:
```
In file included from C:/M/B/src/OpenColorIO-2.4.2/src/OpenColorIO/ops/lut1d/Lut1DOpCPU_SSE2.cpp:10:
C:/M/B/src/OpenColorIO-2.4.2/src/OpenColorIO/SSE2.h:354:19: error: incompatible pointer types assigning to '__fp16 *' from 'int16_t *' (aka 'short *')
  354 |         vst1q_f16((float16_t *)(out+0), rgba);
      |         ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/clangarm64/lib/clang/20/include/arm_neon.h:39424:26: note: expanded from macro 'vst1q_f16'
 39424 |   __builtin_neon_vst1q_v(__p0, (int8x16_t)__s1, 40); \
       |                          ^~~~
In file included from C:/M/B/src/OpenColorIO-2.4.2/src/OpenColorIO/ops/lut1d/Lut1DOpCPU_SSE2.cpp:10:
C:/M/B/src/OpenColorIO-2.4.2/src/OpenColorIO/SSE2.h:357:19: error: incompatible pointer types assigning to '__fp16 *' from 'int16_t *' (aka 'short *')
  357 |         vst1q_f16((float16_t *)(out+8), rgba);
      |         ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/msys64/clangarm64/lib/clang/20/include/arm_neon.h:39424:26: note: expanded from macro 'vst1q_f16'
 39424 |   __builtin_neon_vst1q_v(__p0, (int8x16_t)__s1, 40); \
       |                          ^~~~

2 errors generated.
```

This makes it build: https://github.com/msys2/MINGW-packages/blob/3027f4fc33ee83d31bdcefb2cc07442e02ff8e50/mingw-w64-opencolorio/0006-fix-neon-build.patch

I don't know how that plays with MSVC though, and if this needs to be changed instead: https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/05466120f3f9c23b23778f3ccbafc4dde93abe4b/src/OpenColorIO/SSE2.h#L30-L34

PR #2140, opened 2025-03-27
Fix build issue when using clang (#2137)

anthony-linaro:
Fixes the issue described in #2137 by defining the correct type when using clang-cl (which masquerades as MSVC).

Tested using LLVM 20.1.0's `clang-cl`, with the following command line:
```
mkdir build
cd build
cmake -G"Ninja" .. -DCMAKE_C_COMPILER="clang-cl" -DCMAKE_CXX_COMPILER="clang-cl"
cmake --build .
ctest
```


anthony-linaro:
FYI @lazka and @kmilos

anthony-linaro:
@kmilos I've just tried it with this line instead, which uses the GNU command line:
```
cmake -G"Ninja" .. -DCMAKE_C_COMPILER="clang" -DCMAKE_CXX_COMPILER="clang++"
```

And it still appears to work. Are you able to test to see if it fixes your issue?

lazka:
> Thanks, but I'm not 100% sure this covers #2137 - we're not using `clang-cl` in MSYS2 CLANGARM64, so `_M_ARM64` is not defined AFAIK. llvm-mingw based `clang` is used instead.

https://github.com/mingw-w64/mingw-w64/blob/90da6c6535c503159e952dc8b44f8778bed6f622/mingw-w64-headers/crt/_mingw_mac.h#L87

> ACK, I'm not able to use that triplet without an msys2 env, so I'll await the response from lazka.

I can't easily test right now, but this looks functionally equivalent to the patch I was using, so lgtm!

anthony-linaro:
@kmilos I can't easily test the new version of GCC, but I think an inversion may well work here (assuming it does `__fp16` as per the docs), so I'll update it to that, and add an explicit check for `__aarch64__` too (just in case headers end up differing for some reason in the future and not defining `_M_ARM64` anymore)

anthony-linaro:
Given that, looking at the code, it's already entirely in an `#elif defined(__aarch64__) || defined(_M_ARM64)` block, so I'll just clean it up and check for `#if !defined(float16_t)`, and drop the `_M_ARM64` bit entirely - the lack of definition for `float16_t` should suffice

doug-walker (approved):

cozdas (approved):

Issue #2153, opened 2025-04-29
Add color space attribute for an "interop" ID


doug-walker:
The TSC has discussed many scenarios where applications need to create a correspondence between color spaces in a config and "known" external color spaces. Examples of this are:
- Color space enums in external graphics APIs (macOS, iOS, Windows, Wayland, Android, Vulkan, etc.)
- Color space IDs in asset file formats (MaterialX, OpenUSD, etc.)
- Color spaces for common CICP values

It would be cumbersome for OCIO to try and maintain all of these mappings, there are dozens of them. Furthermore, they evolve over time and this would create challenges if we tried to store these values in the configs themselves.

Instead, OCIO should focus on a method for applications to identify the correspondence between a color space in a config and an unambiguous external definition.

The Color Interop Forum has already worked on producing an example of an unambiguous external definition: the [Color Space Recommendations for Texture Assets and CG Rendering](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/01_TextureAssetColorSpaces/TextureAssetColorSpaces.md). The intent is for the Color Interop Forum to next take on the task of producing a similar recommendation for display color spaces that would include CICP values and enums for external graphics APIs.

The proposal is that OCIO add a new attribute for color spaces named "interop_id" that would hold a string that is used to establish this connection to an external reference. (This would be the "compact name" defined in the texture asset recommendation mentioned above, for example, "lin_rec709_scene".)

The intention is that this interop ID would be more stable over time than the user-facing color space names (which we've seen evolve, for example, in the ACES configs).

The interop ID should be a good choice for embedding in external asset formats such as OpenEXR, OpenUSD, and MaterialX.

It is intended that if a config author includes an interop ID from a Color Interop Forum recommendation, that they are using the exact transform specified in that recommendation. At some future point, it might be desirable to provide a function that would check that is true. In the meantime, it would be helpful to provide built-in transforms for all of the Color Interop Forum color spaces.

For color spaces that do not exist in Color Interop Forum recommendations, it is proposed that config authors should be allowed to populate their own interop ID string. Some sort of namespace mechanism could be used, if desired.

[...]

Issue #2152, opened 2025-04-28
Add color space attribute to hold AMF transform IDs


doug-walker:
Support for ACES Metadata File (AMF) requires that ACES transform IDs be stored with color spaces.

In the AMF prototype (available as a Python script in the shared directory), it looked for these IDs in the description field of a color space. Our current Studio configs automatically set the IDs into the description field to support this.

Unfortunately, more than one ID is needed per color space. It needs the ID for both directions (to and from ACES). It needs the ID for each Output Transform that makes use so a given display color space. And currently we are storing both the ACES and CLF IDs.  

So some of the descriptions are now quite large, for example 38 lines of text and more than 4k characters. This interferes with the more common use of the description field, which is as a brief description of the color space that is suitable for use, for example, in tool-tips.

This was discussed at the TSC today and the decision was to create a new "amf_transform_ids" field for color spaces to hold this info. It is not anticipated that the formatting of the text would need to change. It would just move from the description to the new attribute.

doug-walker:
@KelSolaar , you posted a comment in PR #2165 about AMF transform IDs. Sorry we didn't get to discuss it during the TSC meeting today.  I thought I would redirect the conversation here since it is specific to AMF.

As you know, currently the AMF IDs are being put in the description field, which really interferes with applications that want just a simple description (e.g., for use in a tooltip next to color space names). My question for you is: are you objecting to putting the AMF transform IDs into the config at all, or are you asking that they be stored differently somehow?  How would you prefer to match up color spaces in a config with the IDs in an AMF file?

One option would be to have a sidecar file, but that could easily get lost. Another option would be to build them into the library itself, but that doesn't allow users to add, e.g. vendor specific IDs that I know people are using with AMF.  To me, putting them in the config with a dedicated field seems like the best solution, but I would like to get your ideas.

thomasmansencal:
I think (and prefer) that they should be in the file, but to preserve agnosticism, I only see two ways:

- We serialise them as we currently do in the description.
- We allow arbitrary metadata to be added.

doug-walker:
For the record, the compromise that was agreed with @KelSolaar at the 2025-08-25 working group meeting is to add an amf_transform_id attribute but put it under a sub-group. Furthermore, there will be no bespoke getter/setter for it, it will use a generic attribute getter/setter that will be used for all attributes in the sub-group.

doug-walker:
[...]

Issue #1975, opened 2024-05-10
Embeddable colorspace metadata for media exports.


codycuellar:
One big gap in OCIO that is not easily solvable without external tooling and secondary processes (prone to error) is a way for config authors to specify embeddable metadata that describes a colorspace, most importantly display colorspaces, for output media destined for non-OCIO apps and workflows.

For example, when exporting a still image from a Nuke write node I can chose to use a display/view transform. The config author should be able to provide an ICC profile that describes that display characteristics, and if OCIO provided an API, Nuke could ask OCIO for the ICC profile associated with the user selected display colorspace.

While the list of possible metadata types might be long and difficult to implement in all cases, ICC, and perhaps CICP seems like a good place to start and should be relatively easy to implement. I propose something along the lines of this:

```yaml
 - !<ColorSpace>
    name: Rec.2100-PQ - Display
    aliases: [rec2100_pq_display]
    family: Display
    equalitygroup: ""
    bitdepth: 32f
    description: Convert CIE XYZ (D65 white) to Rec.2100-PQ
    isdata: false
    categories: [file-io]
    encoding: hdr-video
    allocation: uniform
    embeddableColorMetadata:
      icc: Rec.2100-PQ.icc
      cicp:
        transferFunction: 16
        colorPrimaries: 9
        matrixCoefficients: 14
    from_display_reference: !<BuiltinTransform> {style: DISPLAY - CIE-XYZ-D65_to_REC.2100-PQ}
```

The config author would then put the icc profiles in the search path, just like any LUT reference, and OCIO would provide APIs to extract the desired metadata type from a colorspace depending on the file type it's trying to write out.

```cpp
struct CICPMetadata {
  uint8 transferFunction;
  uint8 colorPrimaries;
  uint8 matrixCoefficients;
}

// fully resolved path to ICC for app to read/embed for ICC compatible file types.
const char* Colorspace::getColorMetadataICC(); 

// the standardized CICP tag data for CICP compatible file types.
CICPMetadata Colorspace::getColorMetadataCICP();
```

With an API like this - applications can rely on config authors to provide the correct metadata for file types destined for non-OCIO applications and workflows. Without something like this, an application is forced to either: 
- Not embed any metadata at all
- Provide an arbitrary list of possible metadata for the user to associate with their chosen colorspace
- Jump through hoops trying to guess the appropriate metadata

[...]

PR #2165, opened 2025-06-08
Adding color space attribs: interop_id, amf_transform & icc_profile_name

cozdas:
- Addressing the issues #1975, #2152 and #2153
- Bumped the config version to 2.5
- newly added attributes require v2.5 config both for serialization and de-serialization and will throw if they appear in older configs.
- Hardened multiple existing functions against null parameters. Previously those functions were crashing if null is passed, as assigning null to std::string is undefined behavior.
- Expanded tests to some affected functions' unit tests to include null passing. Setters will take null as empty string. compare functions will throw.

### DIFF
--- a/src/OpenColorIO/OCIOYaml.cpp
+++ b/src/OpenColorIO/OCIOYaml.cpp
@@ -3202,10 +3202,25 @@ inline void load(const YAML::Node& node, ColorSpaceRcPtr& cs, unsigned int major
                 cs->addAlias(alias.c_str());
             }
         }
+        else if (key == "interop_id") 
+        {
+            load(iter->second, stringval);
+            cs->setInteropID(stringval.c_str());
+        }
         else if(key == "description")
         {
             loadDescription(iter->second, stringval);
             cs->setDescription(stringval.c_str());
+        }
+        else if (key == "amf_transform_ids") 
+        {
+            load(iter->second, stringval);
+            cs->setAMFTransformIDs(stringval.c_str());
+        }
+        else if (key == "icc_profile_name") 
+        {
+            load(iter->second, stringval);
+            cs->setICCProfileName(stringval.c_str());
         }
         else if(key == "family")
         {
@@ -3323,11 +3338,37 @@ inline void save(YAML::Emitter& out, ConstColorSpaceRcPtr cs, unsigned int major
         }
         out << YAML::Flow << YAML::Value << aliases;
     }
+    
+    const std::string interopID{ cs->getInteropID() };
+    if (!interopID.empty())
+    {
+        out << YAML::Key << "interop_id";
+        out << YAML::Value << interopID;
+    }
+
     out << YAML::Key << "family" << YAML::Value << cs->getFamily();
+
     out << YAML::Key << "equalitygroup" << YAML::Value << cs->getEqualityGroup();
+
     out << YAML::Key << "bitdepth" << YAML::Value;
     save(out, cs->getBitDepth());
+    
     saveDescription(out, cs->getDescription());
+
+    const std::string amfTransformIDs{cs->getAMFTransformIDs()};
+    if (!amfTransformIDs.empty()) 
+    {
+        out << YAML::Key << "amf_transform_ids";
+        out << YAML::Value << amfTransformIDs;
+    }
+
+    const std::string iccProfileName{cs->getICCProfileName()};
+    if (!iccProfileName.empty()) 
+    {
+        out << YAML::Key << "icc_profile_name";
+        out << YAML::Value << iccProfileName;
+    }
+
     out << YAML::Key << "isdata" << YAML::Value << cs->isData();
 
     if(cs->getNumCategories() > 0)
@@ -4754,10 +4795,12 @@ inline void save(YAML::Emitter & out, const Config & config)
 {
     std::stringstream ss;
     const unsigned configMajorVersion = config.getMajorVersion();
+    const unsigned configMinorVersion = config.getMinorVersion();
+
     ss << configMajorVersion;
-    if(config.getMinorVersion()!=0)
+    if(configMinorVersion != 0)
     {
-        ss << "." << config.getMinorVersion();
+        ss << "." << configMinorVersion;
     }
 
     out << YAML::Block;
--- a/src/bindings/python/PyColorSpace.cpp
+++ b/src/bindings/python/PyColorSpace.cpp
@@ -90,7 +90,10 @@ void bindPyColorSpace(py::module & m)
                          const std::vector<float> & allocationVars,
                          const TransformRcPtr & toReference,
                          const TransformRcPtr & fromReference,
-                         const std::vector<std::string> & categories) 
+                         const std::vector<std::string> & categories,
+                         const std::string & interopID,
+                         const std::string& AMFTransformID,
+                         const std::string& ICCProfileName) 
             {
                 ColorSpaceRcPtr p = ColorSpace::Create(referenceSpace);
                 if (!aliases.empty())
@@ -102,11 +105,14 @@ void bindPyColorSpace(py::module & m)
                     }
                 }
                 // Setting the name will remove alias named the same, so set name after.
-                if (!name.empty())          { p->setName(name.c_str()); }
-                if (!family.empty())        { p->setFamily(family.c_str()); }
-                if (!encoding.empty())      { p->setEncoding(encoding.c_str()); }
-                if (!equalityGroup.empty()) { p->setEqualityGroup(equalityGroup.c_str()); }
-                if (!description.empty())   { p->setDescription(description.c_str()); }
+                if (!name.empty())           { p->setName(name.c_str()); }
+                if (!family.empty())         { p->setFamily(family.c_str()); }
+                if (!encoding.empty())       { p->setEncoding(encoding.c_str()); }
+                if (!equalityGroup.empty())  { p->setEqualityGroup(equalityGroup.c_str()); }
+                if (!description.empty())    { p->setDescription(description.c_str()); }
+                if (!interopID.empty())      { p->setInteropID(interopID.c_str()); }
+                if (!AMFTransformID.empty()) { p->setAMFTransformIDs(AMFTransformID.c_str()); }
+                if (!ICCProfileName.empty()) { p->setICCProfileName(ICCProfileName.c_str()); }
                 p->setBitDepth(bitDepth);
                 p->setIsData(isData);
                 p->setAllocation(allocation);
@@ -150,6 +156,9 @@ void bindPyColorSpace(py::module & m)
              "toReference"_a = DEFAULT->getTransform(COLORSPACE_DIR_TO_REFERENCE),
              "fromReference"_a = DEFAULT->getTransform(COLORSPACE_DIR_FROM_REFERENCE),
              "categories"_a = getCategoriesStdVec(DEFAULT),
+             "interopID"_a = DEFAULT->getInteropID(),
+             "amfTransformIDs"_a = DEFAULT->getAMFTransformIDs(),
+             "iccProfileName"_a = DEFAULT->getICCProfileName(),
              DOC(ColorSpace, Create, 2))
 
         .def("__deepcopy__", [](const ConstColorSpaceRcPtr & self, py::dict)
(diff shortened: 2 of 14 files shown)


lgritz:
I'm extremely curious to hear if @cozdas feels like the copilot review comments are helpful and correct.

thomasmansencal:
Not a fan of seeing all those interchange attributes stuffed at the base level of the `ColorSpace` transform. Feels like a lot of pollution induced by secondary non-critical information. Could we consider them to be nested in a new bag-like field, e.g.:

```yaml
  - !<ColorSpace>
    name: raw
    family: raw
    equalitygroup: ""
    bitdepth: 32f
    description: Some text.
    isdata: true
    allocation: uniform
    interchange:
        interop_id: foo
        amf_transform_id:
            bar
            baz
        icc_profile_name: bam
```

remia:
> I'm extremely curious to hear if @cozdas feels like the copilot review comments are helpful and correct.

Do we have an idea on the associated cost?

cozdas:
> I'm extremely curious to hear if @cozdas feels like the copilot review comments are helpful and correct.

Not too bad, it caught a valid issue and a typo. For this one the signal to noise ratio was not too bad but I'm curious to see it in a more involved PR. Sometimes this "vibe coding" is eating more time that it saves.

lgritz:
> Do we have an idea on the associated cost?

Nothing, this is just a feature of GitHub. 

Or did you mean some metaphorical cost to the world or our souls or something?

remia:
> Nothing, this is just a feature of GitHub.
> 
> Or did you mean some metaphorical cost to the world or our souls or something?

Good to know, just wanted to check whether or not it was on a shared ASWF budget.

thomasmansencal:
> Not a fan of seeing all those interchange attributes stuffed at the base level of the `ColorSpace` transform. Feels like a lot of pollution induced by secondary non-critical information. Could we consider them to be nested in a new bag-like field, e.g.:
> 
> ```yaml
>   - !<ColorSpace>
>     name: raw
>     family: raw
>     equalitygroup: ""
>     bitdepth: 32f
>     description: Some text.
>     isdata: true
>     allocation: uniform
>     interchange:
>         interop_id: foo
>         amf_transform_id:
>             bar
>             baz
>         icc_profile_name: bam
> ```

When looking at the ACES configs,  I found what was bothering me but could not put my finger on: *AMF* (`amf_transform_id`) is something very specific to *ACES*, as much as I like *ACES* and the work we do there, I feel like this is probably too much to have an attribute added to *OCIO* just to support *ACES*. I already felt that the hardcoded ACES interchange role was too much and this is continuing the trend. We should strive for agnosticism as much as possible.

carolalynn:
Hey Thomas - it's a good point you make around being agnostic, and one we definitely should keep in mind. 

We definitely put our eggs in the ACES basket for configs, but the library is less so (other than the fact that all the ACES transforms are built-in). To me, I was looking at this as a way to help us clean up the configs a bit, make it easier for folks using AMF & OCIO to operate. It's not a required attribute, so it's not locking anyone into an ACES workflow that doesn't want to be. 

However, your point is fair and we should discuss at the TSC today.

doug-walker:
I propose we move the sub-thread about AMF IDs [to this issue](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/2152#issuecomment-3153170025).

doug-walker:
Per the discussion at the 2025-08-25 working group meeting, the interop_id will stay as is but the amf_transform_id and icc_profile_name will move under a sub-group. The items in the sub-group will all use a generic getter/setter that takes an attribute name.

In addition, we are looking at allowing the interop_id to be written in profile versions >= 2.0.

Cuneyt will be making another commit in the coming days with these changes.

tests/cpu/CPUProcessor_tests.cpp
```diff
@@ -57,8 +57,9 @@ OCIO_ADD_TEST(CPUProcessor, flag_composition)
 //       a missing/partial optimization.
 
 
-template<OCIO::BitDepth inBD, OCIO::BitDepth outBD, unsigned line>
-OCIO::ConstCPUProcessorRcPtr ComputeValues(OCIO::ConstProcessorRcPtr processor,
+template<OCIO::BitDepth inBD, OCIO::BitDepth outBD>
```
cozdas, line 60: No no no! We don't want an individual copy of this function for every line number it's called from.

src/libutils/imageioapphelpers/imageio.cpp
```diff
@@ -192,6 +192,9 @@ ptrdiff_t ImageIO::getImageBytes() const
 void ImageIO::init(const ImageIO & img, BitDepth bitDepth)
 {
     m_impl->init(*img.m_impl, bitDepth);
+
+    // Do not propagate colorInteropID.
+    attribute("colorInteropID", "unknown");
```
zachlewis, line 197: Should this be the default behavior? 
I feel the "colorInteropID" should be left unset by default (indicating to downstream color-space-identifying efforts that they should move on to the config file rules heuristics; whereas if `colorInteropID` is explicitly set to "unknown", the config file rule heuristics never have a chance to take over, if that makes sense.

doug-walker: @zachlewis , I'm not opposed to leaving it unset but am curious why you think a value of "Unknown" should prevent the file rules from being used?  In the OpenEXR white paper, "Unknown" and unset are described as being essentially equivalent, so perhaps we need to make some edits there. But I'm not sure I understand what the difference should be. It seems like the decision whether to use file rules is a decision for the application and may depend on many other factors.

cozdas: I was under the impression that the setting to "unknown" would almost mean the same as not setting for a receiving application, if the application wants to assign an id, it can still use other heuristics including file rules.

I preferred setting to unknown instead of clearing the field as that signals that it's a deliberate action, not an omission: "this is set by an application that's aware of interop_id attribute but could not determine the id when the image was created". I'm not sure how that information might be used but still didn't want to lose that information.

zachlewis: Hahaha, unsurprisingly, you're absolutely right... I apologize. I can't remember what I've discussed, and with whom...! 
At some point, my conception of the heuristics for the value "unknown" diverged from what you'd actually proposed! I'd been conceiving "unknown" as an "error state" color space (e.g., with an "inverse magenta" matrix to indicate that something has gone very wrong); meaning, there would be no need to fail over to further heuristics. Kind of like a passive-aggressive exception manifested in the image state. 

Personally, I feel if "unknown" is going to be a meaningful value for _interop_, it _must_ communicate an intent _other than_ "do what you normally do when `colorInteropID` is unset (or empty) (or invalid?)". It should, IMO, represent an "end-state", an indication that all heuristics for inferring the color space from hints have been exhausted, and we don't want to fail over to a perhaps-incorrect default file rule (e.g., perhaps if strictparsing=true...) -- instead, a value of "unknown" can be used as a "sentinel" value for communicating that manual intervention is _required_, and that further color-space-identifying heuristics should _not be_ used. (As opposed to "data", which is passively treated as a NoOp.) 

And, likewise, the only reason for persisting a `colorInteropID` value of "unknown" is to communicate that downstream consumers _must_ take care to manually identify the encoding color space if they're going to do ocio stuff. 

In any case, probably worth discussing elsewhere. But in the mean time, if it's all the same, I'd prefer to err on the side of not setting the attribute at all by default.

zachlewis: You know what, @cozdas, I do see your point... 

Hmm. Maybe what I'm after is a third thing, an "unmanaged" value or something... 

In any case, as long as we're consistent, I'm okay with whatever you guys think makes the most sense!

src/bindings/python/PyColorSpace.cpp
```diff
@@ -193,6 +197,14 @@ void bindPyColorSpace(py::module & m)
              DOC(ColorSpace, getDescription))
         .def("setDescription", &ColorSpace::setDescription, "description"_a, 
              DOC(ColorSpace, setDescription))
+        .def("getInteropID", &ColorSpace::getInteropID,
+             DOC(ColorSpace, getInteropID))
+        .def("setInteropID", &ColorSpace::setInteropID, "interopID"_a,
+             DOC(ColorSpace, setInteropID))
+        .def("getInterchangeAttribute", &ColorSpace::getInterchangeAttribute, "attrName"_a,
```
remia, line 204: Are we missing the `getInterchangeAttributes()` function or Pythonic equivalent?

src/apps/ociocheck/main.cpp
```diff
@@ -26,6 +27,83 @@ const char * DESC_STRING = "\n\n"
...
+        "g22_adobergb_display",
+        "g26_p3d65_display",
+        "g26_xyzd65_display",
+        "pq_xyzd65_display",
+    };
+
+    if (id.empty()) 
+        return true;
+
+    // Check if has a namespace.
```
remia, line 73: Just pointing that the checks here seem a bit less strict than the ones from `ColorSpace::setInteropID()`, allowed characters checks, empty namespace before the `:`, etc. Is that on purpose? I guess it could happen if someone edit a config manually?

src/OpenColorIO/OCIOYaml.cpp
```diff
@@ -3348,6 +3417,37 @@ inline void save(YAML::Emitter& out, ConstColorSpaceRcPtr cs, unsigned int major
...
+        out << YAML::Key << "interchange";
+        out << YAML::Value;
+        out << YAML::BeginMap;
+        for (const auto& keyval : interchangemap)
+        {
+            // Remove trailing newlines so that only one is saved because they won't
+            // be read back.
+            std::string valStr{keyval.second};
+            {
+                auto last = valStr.back();
```
remia: Same as `SanitizeLoadedNewlines()`?

src/apps/ociocheck/main.cpp
remia: Or maybe this will simply be caught during loading the YAML and setting the interopID there, please ignore this comment then.

cozdas: Exactly! 
Initially this function had checks on syntax too, but once I added those checks to the set function, which is called during yaml load too, then I removed those from there. I believe not duplicating and scattering the logic is a better approach for the maintainability of the codebase

src/OpenColorIO/OCIOYaml.cpp
cozdas: Good catch. saveDescription() can also call the sanitize function as it's doing the same thing.

src/bindings/python/PyColorSpace.cpp
doug-walker: Yes, Cuneyt was going to loop back to that. Sorry, we should have put a comment to mention that! We should be able to do that tomorrow.

src/libutils/imageioapphelpers/imageio.cpp
doug-walker: @zachlewis , the main use for this code is in ocioconvert and whenever that is doing conversions based on a config it will be replacing Unknown with something specific. The only case where that would not happen is if someone is applying a LUT file.

Let's go with Cuneyt's preference for this PR. I will rely on you to continue the discussion of the semantics of "Unknown" with the appropriate people and then let's validate with the TSC, at which point we could loop back to change this, if necessary.

doug-walker: Zach, I view the work that you and Larry are doing on OIIO as a kind of reference implementation, so once that gets further along we may want to update the OCIO tools to follow suit.

src/apps/ociocheck/main.cpp
remia: Agreed, best to avoid duplication, thanks for confirming.

src/OpenColorIO/OCIOYaml.cpp
cozdas: combined them all. please re-check

src/bindings/python/PyColorSpace.cpp
cozdas: After some struggle with iterators, I realized that it's possible to expose this as a read-only dictionary property. With that it became a single-liner implementation with very python style usage. Please check.

src/bindings/python/PyColorSpace.cpp
```diff
@@ -205,6 +205,8 @@ void bindPyColorSpace(py::module & m)
              DOC(ColorSpace, getInterchangeAttribute))
         .def("setInterchangeAttribute", &ColorSpace::setInterchangeAttribute, "attrName"_a, "attrValue"_a,
             DOC(ColorSpace, setInterchangeAttribute))
+        .def_property_readonly("interchangeAttributes", &ColorSpace::getInterchangeAttributes,
```
remia: `.def("getInterchangeAttributes", &ColorSpace:: getInterchangeAttributes,`

Minor point but just curious if something like this would work (adjusting the use in unit tests)? Just trying to improve consistency with the rest of the API.

remia (approved):
Just a last minor comment but otherwise looks good to me, thanks @cozdas!

src/bindings/python/PyColorSpace.cpp
cozdas: replaced with a function as suggested.

doug-walker (approved):

PR #2167, opened 2025-06-10
Adsk Contrib - Renaming the view helper functions to follow the OCIO conventions.

cozdas:
- The wording order `{object(s)}{Is|Are|Has|Have}{State}` for the newly added function names are different than the more typical convention of `{is|are|have|has}{Object(s)}{State}`.

- Config class already has functions in the second form, such as `Config::isColorSpaceUsed()` or `Config::isColorSpaceLinear()`. We also have functions named in yet another form like `Config::isInactiveColorSpace()`. So not to introduce a 3rd variant, I'm renaming these newly added ones.


zachlewis:
I think maybe we should continue using TitleCaseCapitalization for the two Config "static methods" (`AreViewsEqual` and `AreVirtualViewsEqual`)


*edit: sorry, I didn't see you'd already addressed this! Carry on, LGTM!

remia (approved):

PR #2168, opened 2025-06-14
Update Python frozen docs

remia:
This is an attempt to fix the frozen doc errors in CI. I tried to fix a few warnings in the doc generation (there are still a lot, eg. undefined labels and cross references), and added a simple hack to remove the `PyOpenColorIO.PyOpenColorIO` duplication in the documentation generated type hints. I tried to use some [autodoc settings](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints) to change these hints but no success, and it is not probably an ideal solution either. The same duplication might still come up again if we get rid of the frozen doc workflow, maybe we will need to rename the pybind11 module to something like `Core` (`PyOpenColorIO.Core`) or something else if there are no other ways?

There is an issue in the updated frozen docs where some of Config:: prefixes in the ref links disappears.

I'm not sure if someone else is looking at this issue and have come up with a better solution, in which case this can be ignored.


doug-walker:
This is awesome Remi!  Cuneyt was looking at it a little but didn't come up with anything and he just left for vacation for the next six weeks, so your fix is good timing.

doug-walker (approved):
LGTM!  Looks like this fixed the DOCS=ON CI runs.

PR #2171, opened 2025-07-02
Refactor RTD Python API doc generation

michdolan:
When we first set up API documentation for OCIO v2, one limitation was that we needed to build OCIO in order to generate docstrings for the Python bindings, through a tool stack involving Doxygen, Breathe, Sphinx, and some bespoke Python extensions. This is fine for a local build, but presented challenges on the RTD (Read The Docs) platform since RTD had access to Sphinx and Doxygen, but not CMake or a C++ compiler. This meant that we would have to pre-build python documentation and allow Sphinx to pull that data without needing to build the full project. The initial approach we took to this was to "freeze" Sphinx reST output into dedicated modules, committed alongside their source in PRs that altered the API, and employ checks on CI to diff frozen documentation with generated documentation to ensure we kept the Python docs up to date. This worked, but has been clunky and error prone, and this PR aims to rework the RTD build process to leverage PyOpenColorIO wheels from PyPi for autodoc inspection, instead of the frozen output. This update is long overdue, so thank you for your patience.

The changes in this PR include:

- Removal of frozen Python .rst modules.
- Folding of Python API .rst module directives into core set of API .rst modules, alongside sibling C++ directives.
- Both of the above changes significantly reduce file count in OCIO's docs directory.
- Removal of CMake option, machinery, and related documentation for handling frozen doc creation and validation.
- Update conf.py to handle PyOpenColorIO install from PyPi when needed.

This PR will involve some iteration to make sure build results on RTD are as-expected.


doug-walker (approved):
Very excited to see this Michael!!!

What's the best way to test what happens on RTD, we should just merge it here and then make follow-on PRs with any adjustments?

It looks like the Windows CI failed for reasons unrelated to your PR. Looks like the Windows 2019 runner is no longer available.

michdolan:
Thanks Doug! RTD can create builds from a PR prior to it being merged, so now that this pull request is open, we should be able to inspect the results, and see any errors in the logs. I'll look into this tomorrow and post a link here once the output looks good.

Yes, it looks like we'll need to update the Windows runners and change the required checks. One of these days I hope we can support dynamic required checks, based on a regex pattern or something. I know that's another place that needs continual maintenance, but I suppose in this case our CI matrix needs an update too.

michdolan:
I'm going to close and reopen this PR to try and trigger RTD to build the docs for it.

michdolan:
Trying to cycle the PR again, now that our RTD integration is watching for PRs.

michdolan:
I've been iterating on some different sphinx/RTD configuration options to debug the docs being generated from the PyPi-sourced Python wheel, but oddly the Linux build (.so) appears to be missing most of its docstrings, except for `CPUProcessor`. Testing the same wheel, but the Windows build (.pyd), all the docstrings are present. So there appears to be an issue with our PyPi wheel submissions on some platforms, which in this case is preventing full docs generation.

michdolan:
The docs present in the Linux build are hardcoded into the bindings, which explains why `CPUProcessor` is documented...

michdolan:
Looks like docs are only built for wheels in our workflow when CMake can find Doxygen

michdolan:
I'm done making changes. In summary, we had started bypassing Linux doc generation on Python wheel creation when yum installs were failing on CentOS-7 images due to their package repositories being migrated to a new vault location. This resulted in our Linux Python wheels lacking docstrings from that point until now. I have updated the Doxygen install scripts to use the vault repository on CentOS images, and to use dnf for newer RockyLinux images. I have confirmed in this PR's CI that docs are being built on all platforms, including in our wheel workflows.

Following merge of this PR, we will need to trigger new Python wheels to fix the missing documentation, and then trigger a RTD rebuild to update the docs from those wheels. I'm fairly confident that will all work, but follow-up fixes may be needed if there are unexpected issues.

I also updated our Windows CI workflows to use the 2022 runners, replacing the previous Windows checks that were failing. Following merge of this PR, we will need to update required checks for `main` the `RB-2.*` branches.

The checks that are failing in this PR were either already failing (latest dependency), or lack credentials from my fork (SonarCloud), so can be ignored for now. That being said, we do need an overall CI matrix update (to include VFX Ref 2025 and 2026) and a resolution to the workflows that have been failing for a while. I can work on this in a follow-up PR. Thanks!

share/ci/scripts/linux/dnf/install_doxygen.sh
```diff
@@ -0,0 +1,55 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright Contributors to the OpenColorIO Project.
+
+set -ex
+
+DOXYGEN_VERSION="$1"
+
+if ! command -v doxygen >/dev/null; then
+    if command -v dnf >/dev/null; then
+        if [ "$DOXYGEN_VERSION" == "latest" ]; then
```
cozdas, line 11: Following the fix to the apt/install_doxygen.sh file, I think this should be a singe "=" too.
```suggestion
        if [ "$DOXYGEN_VERSION" = "latest" ]; then
```

share/ci/scripts/linux/dnf/install_doxygen.sh
```diff
@@ -0,0 +1,55 @@
...
+name=CentOS-7 - Extras
+baseurl=http://vault.centos.org/$BASE_DIR/extras/\$basearch/
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
+EOF
+            yum clean all
+            yum makecache
+        fi
+
+        if [ "$DOXYGEN_VERSION" == "latest" ]; then
```
cozdas, line 49: Following the fix to the apt/install_doxygen.sh file, I think this should be a singe "=" too.
```suggestion
        if [ "$DOXYGEN_VERSION" = "latest" ]; then
```

doug-walker:
I'm going to go ahead and merge this super helpful PR. 

Cuneyt or I will address the remaining "=" vs. "==" issue separately.

PR #2176, opened 2025-07-25
Adsk Contrib - Vulkan support

michaelHADSK:
This PR adds Vulkan support to OCIO.
- Added GPU_LANGUAGE_GLSL_VK_4_6 and the corresponding functionality
- Split up declare shader code into textures and parameters
- Added code to add vulkan uniform buffer code around variables
- Functions to add textures now return the index of the texture which is used as the binding index in the shader code
- UniformData now has a bufferOffset to be able to copy the uniform data to the right place in the uniform buffer
- There's a new function to return the size of the generated uniform buffer
- There are new functions to to set and get the Vulkan descriptor set index and texture binding start index
- Added a castToBool method to the GPUShaderText. Support for bools in uniform buffers is unfortunately not well defined/supported in Vulkan, that's why they are declared as int variables. Since there is no implicit cast to bool from int in GLSL, I had to add the explicit cast at some places. This is not so nice because other coders need to be aware that this is required for Vulkan, other ideas on how to deal with that are appreciated.


mrvux:
As requested, here is my feedback on the PR (forked the branch to check the code).
It's very similar to what I did in my personal fork for dx11, with a few key differences (given the fact that vulkan GPU_LANGUAGE_GLSL_VK_4_6 got added it could go away with it, in my case it was not possible):

1. In case of dx11/dx12, sampler states live in a separate register space than resources, so another index was required 

`void setDescriptorSetIndex(unsigned index, unsigned textureBindingStart = 1);`

so for direct3d an additional samplerBindingStart is required.

Also note that registers in Direct3d are zero indexed

`Texture2D myLut : register(t0); // (up to 127)
SamplerState myLutSampler : register(s2); //up to 15`

2. Given that in my case another enum was not required and I wanted not to break existing implementations (to have the implementation use GPU_LANGUAGE_HLSL_SM_5_0).

There is another function in GpuShaderCreator called :

`void setUseExplicitBindings(bool useExplicitBindings);`

This is off by default and keeps explicit implementations to use the default "automatic" bindings" (eg : current implementation) , and use explicit as an Opt-In.

3. The change, given we add samplerBindingStart  in the gpu creator, would fix Direct3d12 in a nearly identical way as per Vulkan:

`kw << "layout(set=" << descriptorSetIndex << ", binding = " << textureIndex << ") ";`

becomes :

`t << "Texture" << N << "D " << textureName : register(t" << textureIndex << ", space" << descriptorSetIndex  << ");"`

It would require a new enum in that case GPU_LANGUAGE_HLSL_SM_5_1

But given that it would not break the GpuCreator interface, this could be implemented at a later stage with no risk of breaking existing implementations.

src/OpenColorIO/GpuShaderUtils.cpp
```diff
@@ -895,27 +921,58 @@ std::string GpuShaderText::sampleTex3D(const std::string& textureName,
 
 void GpuShaderText::declareUniformFloat(const std::string & uniformName)
 {
-    newLine() << (m_lang == GPU_LANGUAGE_MSL_2_0 ? "" : "uniform ") << floatKeyword() << " " << uniformName << ";";
+    std::string uniformDeclString("uniform");
```
remia: Very minor but should we add the space here and other declaration methods to avoid having declarations beginning with a white space like we see in the tests?

michaelHADSK: That should work. I agree that the white space at the beginning is not so nice, I'll try to fix it before the 2.5 release.

tests/cpu/GpuShader_tests.cpp
```diff
@@ -195,6 +195,51 @@ OCIO_ADD_TEST(GpuShader, generic_shader)
...
+        OCIO_CHECK_NE(shaderDesc->getLanguage(), OCIO::GPU_LANGUAGE_GLSL_VK_4_6);
+        OCIO_CHECK_NO_THROW(shaderDesc->setLanguage(OCIO::GPU_LANGUAGE_GLSL_VK_4_6));
+        OCIO_CHECK_EQUAL(shaderDesc->getLanguage(), OCIO::GPU_LANGUAGE_GLSL_VK_4_6);
+
+        OCIO_CHECK_THROW_WHAT(shaderDesc->setDescriptorSetIndex(123, 0),OCIO::Exception, "Texture binding start index must be greater than 0.");
+        OCIO_CHECK_NO_THROW(shaderDesc->setDescriptorSetIndex(123, 456));
+        OCIO_CHECK_EQUAL(shaderDesc->getDescriptorSetIndex(), 123);
+        OCIO_CHECK_EQUAL(shaderDesc->getTextureBindingStart(), 456);
+
+        auto getSize = []() -> float { return 2; }; //simulate only 2 elements in the array
```
cozdas: This should fix the broken Windows build.

```suggestion
        auto getSize = []() -> int { return 2; }; //simulate only 2 elements in the array
```

cozdas (approved):
Just needs the broken Windows unit tests fixed. Otherwise LGTM!

tests/cpu/GpuShader_tests.cpp
michaelHADSK: Fixed as suggested.

PR #2184, opened 2025-09-03
CI fixes

doug-walker:
The latest versions of the 2023 and 2024 Linux Docker containers caused some CI failures due to several issues. One is that "pip" is no longer available. JF indicated that "pip3" is more preferred now, so I changed all "pip install" instances to "pip3 install".

Additionally, they updated from minizip-ng 3.x to 4.x which breaks a number of unit tests. This does not have a quick fix, so I temporarily rolled back the container version. This will be addressed as part of the third-party version upgrade task.


cozdas (approved):
Looks good.

carolalynn (approved):
Thanks Doug!

PR #2186, opened 2025-09-05
Adsk Contrib - Enhance active display/view list API

doug-walker:
The API for the active display and view lists was quite primitive. It only supported setting all active displays or views at once as a long string using commas or colons as separators. Given that it's legal to use commas and colons in display and view names, this places a burden on client apps. For example, it was very easy for a client app to overlook the fact that names may contain separators and thereby create invalid configs.

This PR adds a set of new methods on the Config class to make it easier for client apps to manage the active display and view lists without worrying about details such as whether display or view names contain separator characters relevant to serialization as YAML.

A slight change was made to the Python interface. Previously, getActiveDisplays and getActiveViews returned one long string. Now they return an iterator with the individual strings. This is much more in line with the other Python functions, but will be a breaking change for client Python code. I think it's worthwhile to make this change though since if the code was simply splitting on commas it would not be very robust and should be adjusted anyway.

Lots of unit tests were added.


src/bindings/python/PyConfig.cpp
```diff
@@ -564,12 +576,33 @@ void bindPyConfig(py::module & m)
...
+
         .def("setActiveViews", &Config::setActiveViews, "views"_a, 
              DOC(Config, setActiveViews))
-        .def("getActiveViews", &Config::getActiveViews, 
-             DOC(Config, getActiveViews))
+        .def("getActiveViews", [](ConfigRcPtr & self) 
+            { 
+                return ActiveViewsListIterator(self); 
+            })
+        .def("addActiveView", &Config::addActiveView, "display"_a,
```
remia: Argument name should be "view".

doug-walker: Fixed, thank you Remi!

PR #2190, opened 2025-09-16
Fixes for the Wheel workflow

doug-walker:
1. Huge thanks to Zach for figuring out the Mac problem. For now, I'm using his brew-uninstall technique.
2. The Linux x86 problem seems to be that it was trying to compile OpenEXR 3.1.5 and the runner now seems to be using a version of GCC that needs an extra include file. This was not fixed on the OpenEXR side until 3.1.6.

### DIFF
--- a/.github/workflows/wheel_workflow.yml
+++ b/.github/workflows/wheel_workflow.yml
@@ -386,8 +386,7 @@ jobs:
 
 
   upload_pypi:
-#    needs: [sdist, linux, macos, macos-arm, windows]
-    needs: [sdist, linux-arm, macos, macos-arm, windows]
+    needs: [sdist, linux, linux-arm, macos, macos-arm, windows]
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
     steps:
--- a/share/cmake/modules/FindExtPackages.cmake
+++ b/share/cmake/modules/FindExtPackages.cmake
@@ -184,7 +184,6 @@ if(OCIO_BUILD_PYTHON OR OCIO_BUILD_DOCS)
 endif()
 
 # Set OpenEXR Minimum version as a variable since it it used at multiple places.
-# set(OpenEXR_MININUM_VERSION "3.0.5")
 set(OpenEXR_MININUM_VERSION "3.1.6")
 if((OCIO_BUILD_APPS AND OCIO_USE_OIIO_FOR_APPS) OR OCIO_BUILD_TESTS)
     # OpenImageIO is required for OSL unit test and optional for apps.


zachlewis:
The OIIO MacOS-Intel wheels were having very similar troubles with OpenEXR as we seem to have here. 
It seems that the macos-13 runner image somehow has OpenEXR-3.4.0 already pre-installed via homebrew; and I had to add a step in OIIO's wheels-building workflow's "macos" job to brew-uninstall OpenEXR and Imath before building the wheels. Maybe clearing out the brew-installed OpenEXR / Imath on the macos-13 runner will fix stuff here, too?

e.g., in the 'macos' step in wheel_workflow.yml, between the "Install Python" and Build wheels" steps"
```yaml
  ...
  steps:
      - uses: actions/checkout@v4

      - uses: actions/setup-python@v5
        name: Install Python
        with:
          python-version: '3.9'
          
      - name: Remove brew OpenEXR/Imath
        run: |
          brew uninstall --ignore-dependencies openexr imath || true
            
      - name: Build wheels
        uses: pypa/cibuildwheel@v2.22.0
        env:
          CIBW_BUILD: ${{ matrix.python }}
          CIBW_ARCHS: ${{ matrix.arch }}

      - uses: actions/upload-artifact@v4
        with:
          name: cibw-wheels-${{ matrix.python }}
          path: ./wheelhouse/*.whl

...
```

lgritz:
I think the issue is that we are bitten once again between subtle binary incompatibilities between OpenEXR built against Imath 3.1 versus OpenEXR built against Imath 3.2.

One strategy is to remove the Homebrew OpenEXR 3.4/Imath 3.2, but possibly another solution is to bump the versions you build for your wheels to the latest so they don't conflict with what's on the runners.

zachlewis:
> I think the issue is that we are bitten once again between subtle binary incompatibilities between OpenEXR built against Imath 3.1 versus OpenEXR built against Imath 3.2.
> 
> One strategy is to remove the Homebrew OpenEXR 3.4/Imath 3.2, but possibly another solution is to bump the versions you build for your wheels to the latest so they don't conflict with what's on the runners.

Hmmm... You know, I'd tested with bumping OIIO's OpenEXR recipe default to 3.4.0, but I hadn't thought to try bumping Imath to 3.2 as well...! 
(In any case, we (OIIO) will have to revisit our "-DIGNORE_HOMEBREWED_DEPS" option that we use in CI to make it do a better job of ignoring homebrewed dependencies!)

doug-walker:
Thank you @zachlewis , you saved me so much time!  We will use the uninstall for now to unblock Thomas.

@lgritz , we will be working on supporting OpenEXR 3.4 over the next week or so, which should allow us to switch over to that.

The solution to the Linux problem was to bump the OpenEXR version to at least 3.1.6 to be compatible with the more recent GCC version now being used on the runner.

This PR is now ready for review.

thomasmansencal:
I was able to build macOs wheels yesterday but the upload was skipped... Will do surgery in the Action tonight.

remia:
Great catch @zachlewis !

Issue #2157, opened 2025-05-05
OCIOArchive unit test failing on minizip-bng 4.0.10


remia:
While building OCIO with the dependencies from up to date brew, I'm noticing new unit test failures related to OCIOArchive, specifically when extracting the Windows test archive.

`FAILED: exception thrown from procConfigFromExtractedArchive = configFromExtractedArchive->getProcessor( "plain_lut1_cs", "shot1_lut1_cs" ): "The specified file reference 'shot1/lut1.clf' could not be located. The following attempts were made: '/tmp/OCIOTestTemp_context_test1/shot4/shot1/lut1.clf' : '/tmp/OCIOTestTemp_context_test1/shot1/shot1/lut1.clf' : '/tmp/OCIOTestTemp_context_test1/shot2/shot1/lut1.clf' : '/tmp/OCIOTestTemp_context_test1/shot3/shot1/lut1.clf' : '/tmp/OCIOTestTemp_context_test1/shot3/subdir/shot1/lut1.clf' : '/tmp/OCIOTestTemp_context_test1/shot1/lut1.clf'."
`

My understanding is that with this new minizip-ng version, backward compatibility with archives made from Windows with minzip version less than 3.0.7 is no longer maintained. Starting from 3.0.7, Windows backslash path separators are converted to forward (https://github.com/zlib-ng/minizip-ng/pull/644) and since version 4.0.10, the extracting code no longer consider backslash as a folder separator by default on Unix (https://github.com/zlib-ng/minizip-ng/pull/834). In the unit test failure, I can indeed see that we have files containing backslashes in their names being extracted, instead of producing folders, like macOS unzipper do from the Finder.app for example.

Simple fix would be to update our test Windows archive I guess, I'm not sure if people have been making .ocioz archive from Windows but these might eventually break in the current state of things.

ncopa:
We bumped into the same issue when building the package for Alpine Linux:

```
======================================================================
ERROR: test_extract_config_and_compare_to_original (OCIOZArchiveTest.ArchiveAndExtractComparison.test_extract_config_and_compare_to_original)
Extract an OCIOZ archive that came from config X and create a new Config object Y with it,
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ncopa/aports/community/opencolorio/src/OpenColorIO-2.4.2/tests/python/OCIOZArchiveTest.py", line 395, in test_extract_config_and_compare_to_original
[...]

Issue #2150, opened 2025-04-26
ci-ocio 2025 build container is now available


jfpanisset:
The 2025 version of the ci-ocio build container is now available and could be added to the CI build matrix.

https://hub.docker.com/r/aswf/ci-ocio/tags

You can specify aswf/ci-ocio:2025.0 to get this specific version, or aswf/ci-ocio:2025 to automatically upgrade to any future releases (keeping VFX Platform 2025 compatibility).

This container includes all dependencies required to build OpenColorIO, but not OpenColorIO itself.

doug-walker:
Thank you JF!!

doug-walker:
This container is now used in our CI workflow action.

Issue #2135, opened 2025-03-18
Raise various required minimum versions for OCIO 2.5


doug-walker:
As discussed at the 2025-03-17 TSC meeting, we want to raise several minimum supported or recommended versions for the OCIO 2.5 release, which will become the VFX Platform release for CY2026. This issue is to pull together all of the planned work for better visibility to stakeholders and for comment from the community.

-- Raise the minimum supported language version to C++17, issue #2134 
-- Update OpenImageIO recommended version to 3.x, issue #2129 
-- Raise the minimum supported expat version to solve issue #2093 
-- Raise the minimum supported minizip-ng version above 3.0.8 to solve issue #1818

Please let us know in the comments if you have concerns or would like to request other components be updated.

For discussion purposes, this issue is relevant: #2070.

jfpanisset:
LCMS dependency is calling for 2.2 which is 14 years old, development has resumed and there is a 2.17 as of February 2025:

https://github.com/mm2/Little-CMS/releases/tag/lcms2.17

That would be a good candidate to update.

doug-walker:
Adding for reference this interesting thread about Imath [in #vfx_reference_platform](https://academysoftwarefdn.slack.com/archives/C017TA11X2S/p1755887714904789).

scoopxyz:
With @remia's work on this I would remove the https://github.com/AcademySoftwareFoundation/OpenColorIO/labels/help%20wanted label, otherwise this will be the first item seen for DevDays.

doug-walker:
This was implemented for OCIO 2.5. There is a table in the description of PR #2192 that shows how the minimum and recommended versions have been updated.

Issue #2134, opened 2025-03-18
Raise the minimum supported language version to C++ 17


doug-walker:
Based on discussion at the 2025-03-17 TSC meeting, the decision has been made to raise the minimum supported C++ language version from C++11 to C++17.  Note that the VFX Platform raised to C++17 for CY2021, so OCIO supported the older versions for an extra four years, which seems like more than enough.

The work involved is in the CMake scripts and in the GitHub Actions.

doug-walker:
This was done for OCIO 2.5.

PR #2192, opened 2025-09-19
Bump C++, dependencies and update CI workflows

remia:
This PR update various C++/Python and dependencies versions and GitHub workflows:
- Raise various minimum / recommended versions of dependencies (minimum mostly raised to simplify the build, but the minimum versions picked should be at least roughly 2 years old, obviously up for discussion)
- Raise minimum C++ version to 17
- Add VFX 2025 CI and remove VFX 2021
- Add Python 3.14 wheel, remove Python 3.8 (no longer advertise support for Python 3.7 / 3.8 in Python metadatas as no longer tested)
- Remove OIIO / OSL from dependencies latest workflow for simplicity (OIIO now requires OCIO)

Work on this PR done with @cozdas.

### Issues addressed in this PR
- #2135 
- #2129 
- #2134 
- #2093 
- #1818 
- #2157
- #2150 

### Dependency version update chart

```
               OCIO v2.4.2          OCIO v2.5
             ----------------    ---------------- 
Expat      : [2.4.1,  2.5.0 ] -> [2.6.0,  2.7.2 ]
Yaml-cpp   : [0.6.3,  0.7.0 ] -> [0.8.0,  0.8.0 ]
ZLib       : [1.2.8,  1.2.13] -> [1.2.13, 1.3.1 ]
Imath      : [3.1.1,  3.1.12] -> [3.1.1,  3.2.1 ]
pystring   : [1.1.3,  1.1.3 ] -> [1.1.3,  1.1.4 ]
minizip-ng : [3.0.6,  3.0.7 ] -> [4.0.0,  4.0.10]
lcms2      : [2.2,    2.2   ] -> [2.2,    2.17  ]
openfx     : [1.4,    1.4   ] -> [1.4,    1.5   ]
pybind11   : [2.9.2,  2.12.1] -> [2.9.2,  3.0.1 ]
OIIO       : [2.2.14, 2.4   ] -> [2.2.14, 3     ]
OpenEXR    : [3.0.5,  3.1.5 ] -> [3.2.0,  3.4   ]
OSL        : [1.11,   1.11  ] -> [1.13,   1.14  ]

[ minimum, recommended] 
```
See this [spreadsheet](https://docs.google.com/spreadsheets/d/1N5r94M1yryCKy5MYmrYCCHkloS6c_XTrGtpANlQZq1M/edit?gid=0#gid=0) for the details.

### DIFF
--- a/include/OpenColorIO/OpenColorIO.h
+++ b/include/OpenColorIO/OpenColorIO.h
@@ -2044,7 +2044,7 @@ public:
      *
      * Currently supported attribute names are "amf_transform_ids" and
      * "icc_profile_name". Using any other name will throw. If the attribute is
-     * not defined, it'll return an empty string. Setting the value to an empty
+     * not defined, it will return an empty string. Setting the value to an empty
      * string will effectively delete the attribute.
      *
      * The AMF transform IDs are used to identify specific transforms in the
@@ -2407,6 +2407,22 @@ public:
     const char * getDescription() const;
     void setDescription(const char * description);
 
+    /**
+    * Get/Set the interchange attributes.
+    *
+    * Currently the only supported attribute name is "amf_transform_ids". Using
+    * any other name will throw. If the attribute is not defined, it will return
+    * an empty string. Setting the value to an empty string will effectively
+    * delete the attribute.
+    *
+    * The AMF transform IDs are used to identify specific transforms in the ACES
+    * Metadata File. Multiple transform IDs can be specified in a
+    * newline-separated string.
+    */
+    const char *getInterchangeAttribute(const char *attrName) const;
+    void setInterchangeAttribute(const char* attrName, const char *value);
+    std::map<std::string, std::string> getInterchangeAttributes() const noexcept;
+
     Look(const Look &) = delete;
     Look& operator= (const Look &) = delete;
     /// Do not use (needed only for pybind11).
@@ -2541,6 +2557,22 @@ public:
 
     const char * getDescription() const noexcept;
     void setDescription(const char * description);
+
+    /**
+    * Get/Set the interchange attributes.
+    *
+    * Currently the only supported attribute name is "amf_transform_ids". Using
+    * any other name will throw. If the attribute is not defined, it will return
+    * an empty string. Setting the value to an empty string will effectively
+    * delete the attribute.
+    *
+    * The AMF transform IDs are used to identify specific transforms in the ACES
+    * Metadata File. Multiple transform IDs can be specified in a
+    * newline-separated string.
+    */
+    const char *getInterchangeAttribute(const char *attrName) const;
+    void setInterchangeAttribute(const char* attrName, const char *value);
+    std::map<std::string, std::string> getInterchangeAttributes() const noexcept;
 
     /// \see ColorSpace::hasCategory
     bool hasCategory(const char * category) const;
@@ -3351,6 +3383,23 @@ public:
     ///  Set a prefix to the resource name
     void setResourcePrefix(const char * prefix) noexcept;
 
+    /**
+    * \brief Set the descriptor set index and texture binding start index to use for the shader program.
+    * 
+    * \note Only supported for shading languages, such as Vulkan, that use descriptor sets and texture bindings.
+    * 
+    * \param index The descriptor set index to use.
+    * \param textureBindingStart The texture binding start index to use. The default index starts at 1
+    *                            and is incremented by 1 for each texture. Otherwise, the texture binding starts
+    *                            at textureBindingStart and is incremented by 1 for each texture. 
+    *                            The binding of a texture is equal to the texture index + textureBindingStart.
+    *                            The texture binding start index must be greater than 0, as binding 0 is reserved
+    *                            for the uniform buffer binding
+    * */
+    void setDescriptorSetIndex(unsigned index, unsigned textureBindingStart = 1);
+    unsigned getDescriptorSetIndex() const noexcept;
+    unsigned getTextureBindingStart() const noexcept;
+
     virtual const char * getCacheID() const noexcept;
 
     /// Start to collect the shader data.
@@ -3394,13 +3443,23 @@ public:
     virtual bool addUniform(const char * name,
                             const Float3Getter & getFloat3) = 0;
 
+    /// The size of the vector can be smaller than the size of the corresponding 
+    /// array that is declared in the shader. The parameter maxSize must be used 
+    /// to pass the size of the array declared in the shader. This is important for
+    /// being able to calculate the correct uniform buffer offset for subsequent uniforms
     virtual bool addUniform(const char * name,
                             const SizeGetter & getSize,
-                            const VectorFloatGetter & getVectorFloat) = 0;
+                            const VectorFloatGetter & getVectorFloat,
+                            const unsigned maxSize) = 0;
 
+    /// The size of the vector can be smaller than the size of the corresponding 
+    /// array that is declared in the shader. The parameter maxSize must be used 
+    /// to pass the size of the array declared in the shader. This is important for
+    /// being able to calculate the correct uniform buffer offset for subsequent uniforms
     virtual bool addUniform(const char * name,
                             const SizeGetter & getSize,
-                            const VectorIntGetter & getVectorInt) = 0;
+                            const VectorIntGetter & getVectorInt,
+                            const unsigned maxSize) = 0;
 
     /// Adds the property (used internally).
     void addDynamicProperty(DynamicPropertyRcPtr & prop);
@@ -3436,14 +3495,17 @@ public:
      * \note
      *   The 'values' parameter contains the LUT data which must be used as-is as the dimensions and
      *   origin are hard-coded in the fragment shader program. So, it means one GPU texture per entry.
+     * 
+     * \return Index of the texture. For shading languages using explicit texture bindings, the return
+     *         value is the same as the texture binding index in the generated shader program.
      **/
-    virtual void addTexture(const char * textureName,
-                            const char * samplerName,
-                            unsigned width, unsigned height,
-                            TextureType channel,
-                            TextureDimensions dimensions,
-                            Interpolation interpolation,
-                            const float * values) = 0;
+    virtual unsigned addTexture(const char * textureName,
+                                const char * samplerName,
+                                unsigned width, unsigned height,
+                                TextureType channel,
+                                TextureDimensions dimensions,
+                                Interpolation interpolation,
+                                const float * values) = 0;
 
     /**
      *  Add a 3D texture with RGB channel type.
@@ -3452,15 +3514,19 @@ public:
      *   The 'values' parameter contains the 3D LUT data which must be used as-is as the dimension
      *   and origin are hard-coded in the fragment shader program. So, it means one GPU 3D texture
      *   per entry.
+     * 
+     * \return Index of the texture. For shading languages using explicit texture bindings, the return
+     *         value is the same as the texture binding index in the generated shader program.
      **/
-    virtual void add3DTexture(const char * textureName,
+    virtual unsigned add3DTexture(const char * textureName,
                               const char * samplerName,
                               unsigned edgelen,
                               Interpolation interpolation,
                               const float * values) = 0;
 
     // Methods to specialize parts of a OCIO shader program
-    virtual void addToDeclareShaderCode(const char * shaderCode);
+    virtual void addToParameterDeclareShaderCode(const char * shaderCode);
+    virtual void addToTextureDeclareShaderCode(const char* shaderCode);
     virtual void addToHelperShaderCode(const char * shaderCode);
     virtual void addToFunctionHeaderShaderCode(const char * shaderCode);
     virtual void addToFunctionShaderCode(const char * shaderCode);
@@ -3474,7 +3540,8 @@ public:
      *   to change some parts. Some product integrations add the color processing
      *   within a client shader program, imposing constraints requiring this flexibility.
      */
-    virtual void createShaderText(const char * shaderDeclarations,
+    virtual void createShaderText(const char * shaderParameterDeclarations,
+                                  const char * shaderTextureDeclarations,
                                   const char * shaderHelperMethods,
                                   const char * shaderFunctionHeader,
                                   const char * shaderFunctionBody,
@@ -3663,10 +3730,16 @@ public:
      * * UNIFORM_FLOAT3: m_getFloat3.
      * * UNIFORM_VECTOR_FLOAT: m_vectorFloat.
      * * UNIFORM_VECTOR_INT: m_vectorInt.
+     * 
+     * The m_bufferOffset is the offset in bytes from the start of the uniform buffer.
+     * For shading languages that use uniform buffers, the offset can be used to
+     * determine the location of the uniform in the buffer and fill it with the 
+     * corresponding data.
      */
     struct UniformData
     {
         UniformDataType m_type{ UNIFORM_UNKNOWN };
+        std::size_t m_bufferOffset{};
         DoubleGetter m_getDouble{};
         BoolGetter m_getBool{};
         Float3Getter m_getFloat3{};
@@ -3684,6 +3757,16 @@ public:
     virtual unsigned getNumUniforms() const noexcept = 0;
     /// Returns name of uniform and data as parameter.
     virtual const char * getUniform(unsigned index, UniformData & data) const = 0;
+
+    /**
+    * For shading languages that use uniform buffers, a uniform buffer
+    * containing all uniforms is generated in the shader code. This method can
+    * be used to create a buffer of the same size in the client code that can
+    * be filled with the corresponding data.
+    * 
+    * \return Size of the uniform buffer in bytes
+    **/
+    virtual std::size_t getUniformBufferSize() const noexcept = 0;
 
     // 1D lut related methods
     virtual unsigned getNumTextures() const noexcept = 0;
--- a/include/OpenColorIO/OpenColorTypes.h
+++ b/include/OpenColorIO/OpenColorTypes.h
@@ -449,6 +449,7 @@ enum GpuLanguage
     GPU_LANGUAGE_GLSL_1_2,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_1_3,          ///< OpenGL Shading Language
     GPU_LANGUAGE_GLSL_4_0,          ///< OpenGL Shading Language
+    GPU_LANGUAGE_GLSL_VK_4_6,       ///< OpenGL Shading Language for Vulkan
     GPU_LANGUAGE_HLSL_SM_5_0,       ///< DirectX High Level Shading Language
     LANGUAGE_OSL_1,                 ///< Open Shading Language
     GPU_LANGUAGE_GLSL_ES_1_0,       ///< OpenGL ES Shading Language
--- a/src/OpenColorIO/Config.cpp
+++ b/src/OpenColorIO/Config.cpp
@@ -5904,6 +5904,38 @@ void Config::Impl::checkVersionConsistency() const
         throw Exception("Only version 2 (or higher) can have ViewTransforms.");
     }
 
+    // Check for new ViewTransform properties.
+
+    if (hexVersion < 0x02050000) 
+    {
+        for (const auto& vt : m_viewTransforms)
+        {
+            if (vt->getInterchangeAttributes().size()>0)
+            {
+                std::ostringstream os;
+                os << "Config failed validation. The view transform '" << vt->getName() << "' ";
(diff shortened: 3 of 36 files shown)


src/OpenColorIO/SystemMonitor_macos.cpp
KevinJW, line 1: Looks like a stray update, was this required as part of the update? or a new warning from static analysis or something?

remia: My bad for merging all commits, it's a warning you get from recent AppleClang, seems like we were using a non-standard VLA the whole time.

```
In file included from OpenColorIO/src/OpenColorIO/SystemMonitor.cpp:30:
OpenColorIO/src/OpenColorIO/SystemMonitor_macos.cpp:193:29: warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]
  193 |                 char buffer[bufferSize];
      |                             ^~~~~~~~~~
OpenColorIO/src/OpenColorIO/SystemMonitor_macos.cpp:193:29: note: initializer of 'bufferSize' is not a constant expression
OpenColorIO/src/OpenColorIO/SystemMonitor_macos.cpp:192:31: note: declared here
  192 |                 const CFIndex bufferSize = CFStringGetLength(values[0]) + 1; // +1 for null termination
      |                               ^
1 warning generated.
```

KevinJW: Indeed, sounds like we should fix it anyway.

share/cmake/utils/CompilerFlags.cmake
```diff
@@ -60,12 +60,13 @@ if(USE_MSVC)
...
-            "${PLATFORM_COMPILE_OPTIONS};/D_SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING"
-        )
-    endif()
+	# seems no longer needed, there were c++ modernization on 0.7.0. /coz
+#    if(${CMAKE_CXX_STANDARD} GREATER_EQUAL 17)
+#        # Inheriting from std::iterator is deprecated starting with C++17 and Yaml 0.6.3 does that.
+#        set(PLATFORM_COMPILE_OPTIONS
+#            "${PLATFORM_COMPILE_OPTIONS};/D_SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING"
+#        )
+#    endif()
```
cozdas: I think we can remove this part now as the min version of yaml 0.8.0 looks happy without it.

share/cmake/macros/GetPythonPreCommand.cmake
```diff
@@ -42,15 +43,20 @@ macro(get_python_pre_command)
...
         foreach(_PATH ${ARGN})
             file(TO_NATIVE_PATH ${_PATH} _WIN_PATH)
             list(APPEND _WIN_PATHS ${_WIN_PATH})
         endforeach()
-        list(APPEND _WIN_PATHS "%PYTHONPATH%")
+
+        # Double % to escape as the cmd.exe will eat one level of %. That
+        # results in an empty path in the ENV variable and Python fails to
+        # convert that to absolute path. Possibly due to
+        # https://www.cve.news/cve-2023-41105/
```
cozdas: My comment was not accurate here. Please update
```suggestion
        # Double % to escape as an intermediate cmd.exe step causes early
        # expansion. That results in an empty path in the ENV variable and
        # Python 3.11+ fails to convert that to absolute path. Possibly due to
        # https://www.cve.news/cve-2023-41105/
```

share/ci/scripts/multi/install_yaml-cpp.sh
```diff
@@ -23,8 +23,10 @@ if [ "$YAMLCPP_VERSION" == "latest" ]; then
 else
     if [[ "$YAMLCPP_MINOR" -lt 6 && "$YAMLCPP_PATCH" -lt 3 ]]; then
         git checkout tags/release-${YAMLCPP_VERSION} -b release-${YAMLCPP_VERSION}
-    else
+    elif [[ "$YAMLCPP_MINOR" -lt 8 ]]; then
         git checkout tags/yaml-cpp-${YAMLCPP_VERSION} -b yaml-cpp-${YAMLCPP_VERSION}
+    else
+        git checkout tags/${YAMLCPP_VERSION} -b ${YAMLCPP_VERSION}
```
cozdas: Do we still need to check the version<0.8.0 cases?

.github/workflows/ci_workflow.yml
```diff
@@ -61,9 +61,51 @@ jobs:
     strategy:
```
doug-walker, line 61: Are you able to remove the ".1" at the end of the image line now (to use the latest image)?

share/cmake/modules/FindExtPackages.cmake
```diff
@@ -259,8 +258,8 @@ if(OCIO_BUILD_APPS)
         ###############################################################################
         # Calling find_package in CONFIG mode using PREFER_CONFIG option.
         ocio_handle_dependency(  OpenEXR PREFER_CONFIG ALLOW_INSTALL
-                                 MIN_VERSION ${OpenEXR_MININUM_VERSION}
-                                 RECOMMENDED_VERSION 3.1.6
+                                 MIN_VERSION 3.2.0
+                                 RECOMMENDED_VERSION 3.3.5
```
doug-walker: VFX Platform says OpenEXR 3.4 for CY2026, so I think that should be our recommended version.

share/cmake/modules/FindExtPackages.cmake
```diff
@@ -179,19 +179,18 @@ if(OCIO_BUILD_PYTHON OR OCIO_BUILD_DOCS)
         # pybind11 2.9 fixes issues with MS Visual Studio 2022 (Debug).
         ocio_handle_dependency(  pybind11 REQUIRED ALLOW_INSTALL
                                  MIN_VERSION 2.9.2
-                                 RECOMMENDED_VERSION 2.12.1)
+                                 RECOMMENDED_VERSION 3.0.1)
     endif()
 endif()
 
 # Set OpenEXR Minimum version as a variable since it it used at multiple places.
```
cozdas: Leftover comment. (my bad)

share/cmake/utils/CompilerFlags.cmake
remia: Yes, done now, plus addressing the other points.

share/cmake/modules/FindExtPackages.cmake
remia: What I found when trying OpenEXR 3.4.0 were OpenJPH compilation issues, given that the release is very new I figured we might stick to 3.3.5 for now. I'll give it another go here once the ASWF docker test is complete from your other point.

remia:
Hey Doug,

> What change was needed to the test file?
> tests/data/files/configs/context_test1/context_test1_windows.ocioz

I built OCIO on Windows (with the new minizip-ng version), then used ocioarchive to regenerate the file, the issue is described in more details [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/2157).

.github/workflows/ci_workflow.yml
remia: It looks like our CI workflow no longer need it, I'm however seeing an issue in [dependencies latest](https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/17876981722/job/50839666824) that showed up when I removed the forced .1 tag I had there.

remia: I found that ci-base 2025.2 is causing issues because we install Imath in /usr/local/lib64 (the default) but LD_LIBRARY_PATH is missing that path, ci-base 2025.1, however has it. For example see these build with [2025.1](https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/17857251726/job/50778983096) and [2025.2](https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/17877187824/job/50840130035?pr=2192). In the (raw) logs we can see the difference in LD_LIBRARY_PATH. @jfpanisset is that something that rings a bell for you? For now the workaround is to add the missing library path before running the CMake consumer example.

jfpanisset: TL;DR : as of the 2024..3 / 2025.2 / 2026.0 images everything is installed in `/usr/local/lib` instead of `/usr/local/lib64` (with some leakage in `/usr/local/lib`), and you should find a 2025 build of Imath 3.1.12 in `/usr/local/{lib,include}.

In more details:

Until the 2024.2 / 2025.1 releases, I had been fighting a running battle with all the packages included in aswf-docker to land the builds in `lib64`, but that was a constant source of busy work / toil:

- a majority of open source packages expect to land in `lib` rather than lib64, and often have that hardcoded in their build / install / CMake code, so I had to maintain patches against those packages which would often break with every new package
- the Conan recipes from the Conan Center Index explicitly build in `lib`, so I had to maintain patches against the Conan recipes themselves
- no matter what I did, there would always be some "leakage" of installs in `lib` rather than `lib64`

So for the 2024.3 / 2025.2 / 2026.1 releases I gave up and decided to switch to `lib`, and thus installation in `/usr/local/lib`. This is saving a huge amount of time, and helps to minimize the number of differences / patches I need to maintain against upstream packages and Conan recipes. This is mentioned in the [CHANGELOG.md](https://github.com/AcademySoftwareFoundation/aswf-docker/blob/main/CHANGELOG.md#2025-08-14) :

```
These releases no longer build / install libraries in /usr/local/lib64, they revert back to the default /usr/local/lib supported by most packages and Conan recipes.
```

Note that the aim of the `ci-ocio` image is to provide all the pre-built dependencies required by OCIO (same for all the other `ci-aswfproject` images) so that the project can:

- minimize its build time by not having to download and build any dependencies
- exercise the build script path where dependencies aer found locally

Specifically the Conan build script [conanfile.py](https://github.com/AcademySoftwareFoundation/aswf-docker/blob/d9431fc8777257b5a17adcad900c7ad6b8d8baab/packages/conan/recipes/ocio/conanfile.py#L61) specifies explicit dependencies on OpenEXR and Imath, so these should be already in `/usr/local/{include,lib}` in the `ci-ocio:2025.2` image.

Note that the versions of OpenEXR / Imath and others found in the `conanfile.py` recipe get overridden by the [Conan profile for the VFX Platform year](https://github.com/AcademySoftwareFoundation/aswf-docker/blob/main/packages/conan/settings/profiles_aswf/vfx2025), the hard coded values in the Conan recipe are just the ones found in the [Conan Center Index upstream recipe](https://github.com/conan-io/conan-center-index/blob/master/recipes/opencolorio/all/conanfile.py), using Conan profiles allows the same recipe to be used for multiple VFX Platform year and minimize differences from the upstream recipe.

Let me know if this makes sense, and if it will work for OCIO.

remia: Thanks @jfpanisset, that sounds good and the `ci-ocio:2025.2` seem to work as expected now.

I should have provided more context, here we use the `ci-base` (2024 and 2025) image in a nightly workflow "Dependencies latest" which is testing the bleeding edge (latest release tag) of ASWF projects we depend on. We install Imath with default cmake workflow and get the binary installed in /usr/local/lib64, however this path is not LD_LIBRARY_PATH (which I thought would be there by default), but we have a simple work around in any case so this is not a blocking issue.

jfpanisset: Good to know, not always simple to know all the ways in which the image are being used!

I'd have to dig a bit deeper as to where `LD_LIBRARY_PATH` is set explicitly, but yes I believe it now only points to `/usr/local/lib` since the new images don't even have a `/usr/local/lib64` anymore.

`LD_LIBRARY_PATH` is typically not set by default in a Linux distribution, `ld.so` only looks in the "standard" system directories as well as what is specified in `/etc/ld.so.conf` and `/etc/ld.so.conf.d/` (applications which add to the runtime search path with those mechanisms can cause unexpected behavior when they provide DSOs which globally override system installed DSOs).

doug-walker:
@remia , thanks for the detail on the minizip issue.

Regarding OpenEXR 3.4 support, maybe it would be simpler to build it without HTJ2K support, I don't think that is essential for our OCIO tools?  In any case, I think 3.4 support is something we could add for 2.5.1.

remia:
> Regarding OpenEXR 3.4 support, maybe it would be simpler to build it without HTJ2K support, I don't think that is essential for our OCIO tools? In any case, I think 3.4 support is something we could add for 2.5.1.

I think HTJ2K is mandatory in OpenEXR 3.4.0, the latest commits have OpenEXR 3.4.0 working, it actually seems fine in all our CI workflows.

remia:
I'm wondering if it's fine to publish Python 3.14 wheels even if Python 3.14 will release after OCIO 2.5.0 (on October 7)? I believe cibuildwheel currently use a Python 3.14 release candidate version. Maybe we should disable the 3.14 wheel here and add the Python 3.14 wheel in 2.5.1?

share/cmake/modules/install/InstallOpenEXR.cmake
```diff
@@ -79,6 +79,20 @@ if(NOT OpenEXR_FOUND AND OCIO_INSTALL_EXT_PACKAGES AND NOT OCIO_INSTALL_EXT_PACK
 
     set(_OpenEXR_LIB_VER "${_OpenEXR_VERSION_MAJOR}_${_OpenEXR_VERSION_MINOR}")
 
+    # Specify which OpenJPH version to use as we need to know the library name
+    # for the Windows library, and OpenEXR would otherwise pull the latest master
+    # branch commit which may result in less stable builds.
```
remia: I was actually looking at the wrong branch, OpenEXR 3.4.0 would pull OpenJPH 0.22.0, might still be nice to use 0.23.1 here as I believe it comes with some build fixes.

cozdas:
As we discussed in the OCIO working group meeting today, I double checked that the older versions of OCIO will be able to load the ocioz files that this PR generates on the Windows platform. 

### Executive Summary
My findings are positive, OCIO v2.4 on linux and and v2.3 on macos can load the windows v2.5-generated ocioz files. So we're good.👍 

### Details
Here are the details if anyone is interested in:
Zip specification says that the paths should use forward slash in every platform:

> 4.4.17.1 The name of the file, with optional relative path.
       The path stored MUST NOT contain a drive or
       device letter, or a leading slash.  All slashes
       MUST be forward slashes '/' as opposed to
       backwards slashes '\' for compatibility with Amiga
       and UNIX file systems etc.  If input came from standard
       input, there is no file name field.  

But unfortunately, minizip-ng was writing backslashes on Windows. These non-standard paths were corrected while opening the archive file but converting the backslashes to forward slashes was causing other issues. So finally, in v4.0.10 they stopped treating backslashes as forward slashes and thus those ill-formed ocioz archives started to fail loading.

As of minizip-ng v4.0.10, they are now writing standard zip archives according to the spec I posted above thus they should work with every zip unpacking tool on any platform.

I checked the various ocioz archives that both v2.4 and this PR created to make sure that the issue is indeed fixed.  Here are the results

**Created by v2.4 on Linux**
```
Archive:  ./context_test1_linux.ocioz
Zip file size: 3290 bytes, number of entries: 10
?---------  4.5 unx     3100 bl defX 22-Sep-26 09:22 config.ocio
-rwxr-xr-x  4.5 unx      224 bl defX 80-000-00 00:00 shot2/lut1.clf
-rwxr-xr-x  4.5 unx      221 bl defX 80-000-00 00:00 shot2/lut2.clf
-rwxr-xr-x  4.5 unx      221 bl defX 80-000-00 00:00 shot4/lut4.clf
-rwxr-xr-x  4.5 unx      225 bl defX 80-000-00 00:00 shot4/lut1.clf
-rwxr-xr-x  4.5 unx      221 bl defX 80-000-00 00:00 shot3/subdir/lut3.clf
-rwxr-xr-x  4.5 unx      224 bl defX 80-000-00 00:00 shot3/lut1.clf
-rwxr-xr-x  4.5 unx      925 bl defX 80-000-00 00:00 looks.cdl
-rwxr-xr-x  4.5 unx      221 bl defX 80-000-00 00:00 lut1.clf
-rwxr-xr-x  4.5 unx      224 bl defX 80-000-00 00:00 shot1/lut1.clf
10 files, 5806 bytes uncompressed, 2082 bytes compressed:  64.1%
```
**Created by v2.4 on Windows (note the backslashes)**
```
Archive:  ./context_test1_windows.ocioz
Zip file size: 3938 bytes, number of entries: 10
?---------  4.5 t20     3096 bl defX 22-Sep-09 11:32 config.ocio
?---------  4.5 t20      925 bX defX 22-Sep-03 16:30 looks.cdl
?---------  4.5 t20      221 bX defX 22-Sep-03 16:39 lut1.clf
?---------  4.5 t20      224 bX defX 22-Sep-02 19:30 shot1\lut1.clf
?---------  4.5 t20      224 bX defX 22-Sep-02 19:31 shot2\lut1.clf
?---------  4.5 t20      221 bX defX 22-Sep-02 19:32 shot2\lut2.clf
?---------  4.5 t20      224 bX defX 22-Sep-02 19:31 shot3\lut1.clf
?---------  4.5 t20      221 bX defX 22-Sep-02 19:32 shot3\subdir\lut3.clf
?---------  4.5 t20      225 bX defX 22-Sep-03 16:27 shot4\lut1.clf
?---------  4.5 t20      221 bX defX 22-Sep-03 16:25 shot4\lut4.clf
10 files, 5802 bytes uncompressed, 2082 bytes compressed:  64.1%
```

**Created by v2.5 (this PR) on Windows**
```
Archive:  ./context_test1_windows_remi.ocioz
Zip file size: 3966 bytes, number of entries: 10
?---------  4.5 t20     3100 bl defX 25-Sep-16 15:19 config.ocio
?---------  4.5 t20      967 bX defX 24-Sep-09 16:57 looks.cdl
?---------  4.5 t20      230 bX defX 24-Sep-09 16:57 lut1.clf
?---------  4.5 t20      233 bX defX 24-Sep-09 16:57 shot1/lut1.clf
?---------  4.5 t20      233 bX defX 24-Sep-09 16:57 shot2/lut1.clf
?---------  4.5 t20      230 bX defX 24-Sep-09 16:57 shot2/lut2.clf
?---------  4.5 t20      233 bX defX 24-Sep-09 16:57 shot3/lut1.clf
?---------  4.5 t20      230 bX defX 24-Sep-09 16:57 shot3/subdir/lut3.clf
?---------  4.5 t20      235 bX defX 24-Sep-09 16:57 shot4/lut1.clf
?---------  4.5 t20      230 bX defX 24-Sep-09 16:57 shot4/lut4.clf
10 files, 5921 bytes uncompressed, 2110 bytes compressed:  64.4%
```

Here are the details of a file in the last archive verifying that it's indeed created on Windows
```
Central directory entry #10:
---------------------------

  There are an extra 16 bytes preceding this file.

  shot4/lut4.clf

  offset of local header from start of archive:   2774
                                                  (0000000000000AD6h) bytes
  file system or operating system of origin:      TOPS-20
  version of encoding software:                   4.5
  minimum file system compatibility required:     MS-DOS, OS/2 or NT FAT
  minimum software version required to extract:   2.0
  compression method:                             deflated
  compression sub-type (deflation):               maximum
  file security status:                           not encrypted
  extended local header:                          yes
  file last modified on (DOS date/time):          2024 Sep 9 16:57:14
  32-bit CRC value (hex):                         0f9aa390
  compressed size:                                157 bytes
  uncompressed size:                              230 bytes
  length of filename:                             14 characters
  length of extra field:                          36 bytes
  length of file comment:                         0 characters
  disk number on which file begins:               disk 1
  apparent file type:                             binary
  Unix file attributes (000000 octal):            ?---------
  MS-DOS file attributes (20 hex):                arc 

  The central-directory extra field contains:
  - A subfield with ID 0x000a (PKWARE Win32) and 32 data bytes.  The first
    20 are:   00 00 00 00 01 00 18 00 00 a9 f7 ef d0 02 db 01 80 b8 5e 0f.

  There is no file comment.
```

cozdas:
Ok since the zip spec mentioned Amiga, I had to test it on Amiga emulator too. The ocioz file generated by OCIO v2.4 on windows unzips to files with backslashes in their names (Amiga allows that), 
image

while the ocioz file generated by OCIO v2.5 on Windows unzips properly with the correct folder structure.
image
 OCIO v2.5 is Amiga ready :P

remia:
> I'm wondering if it's fine to publish Python 3.14 wheels even if Python 3.14 will release after OCIO 2.5.0 (on October 7)? I believe cibuildwheel currently use a Python 3.14 release candidate version. Maybe we should disable the 3.14 wheel here and add the Python 3.14 wheel in 2.5.1?

I'm disabling these as per @JeanChristopheMorinPerso feedback, we should deploy them in a follow up minor / patch release once Python 3.14 is officially out.

doug-walker (approved):

doug-walker:
As discussed in the TSC meeting, am merging this now.

PR #2194, opened 2025-09-22
Adsk Contrib - Adding interchange attributes to ViewTransforms and Looks

cozdas:
- Currently only the amf_transform_ids key is supported. That key is needed for ACES 2 built-in configs.


KelSolaar (approved):
LGTM!

PR #2197, opened 2025-09-25
Alpha channel should be 1 by default, rather than 0 

pennelee:
Related to issue https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/1781.

For CPU processor, when source has no alpha (RGB) but dest does (RGBA) make it default to 1.
GPU Processors (new and legacy) did not seem to have this issue, alpha was already set to 1. Let me know if this is not the case.

Tested changes with ocioconvert and changed the output to force RGBA, since the default is to use inplace conversion (did not check in those changes).
Tested with all the ocioconvert test cases (basic, --lut, --view, --invertview --namedtransform, --invnamed transform), and various EXR file sizes.

Before the changes alpha would be 0, so image was not viewable. After the changes then images can be viewed and from inspection the alpha channel is now 1.0.

Also updated related CPU tests to reflect non zero alpha values.

### DIFF
--- a/src/OpenColorIO/ImagePacking.cpp
+++ b/src/OpenColorIO/ImagePacking.cpp
@@ -70,7 +70,7 @@ void Generic<Type>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
         inBitDepthBuffer[4*pixelsCopied+0] = *rPtr;
         inBitDepthBuffer[4*pixelsCopied+1] = *gPtr;
         inBitDepthBuffer[4*pixelsCopied+2] = *bPtr;
-        inBitDepthBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (Type)0.0f;
+        inBitDepthBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (Type)1.0f;
 
         pixelsCopied++;
         xIndex++;
@@ -140,7 +140,7 @@ void Generic<float>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
         outputBuffer[4*pixelsCopied+0] = *rPtr;
         outputBuffer[4*pixelsCopied+1] = *gPtr;
         outputBuffer[4*pixelsCopied+2] = *bPtr;
-        outputBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : 0.0f;
+        outputBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : 1.0f;
 
         pixelsCopied++;
         xIndex++;
--- a/tests/cpu/CPUProcessor_tests.cpp
+++ b/tests/cpu/CPUProcessor_tests.cpp
@@ -276,9 +276,9 @@ OCIO_ADD_TEST(CPUProcessor, with_one_matrix)
                 1.0000f,  1.2500f,  1.9900f  };
 
         const std::vector<float> resImg
-            = { -0.01929f,  -0.3995f,  0.4002f, 0.5f,
-                 1.58960f,   0.9050f,  1.5025f, 0.5f,
-                 2.070699f,  1.6505f,  2.4002f, 0.5f   };
+            = { -0.01929f,  -0.3995f,  0.4002f, 1.5f,
+                 1.58960f,   0.9050f,  1.5025f, 1.5f,
+                 2.070699f,  1.6505f,  2.4002f, 1.5f   };
 
         ComputeValues<OCIO::BIT_DEPTH_F32, OCIO::BIT_DEPTH_F32>( 
             __LINE__, processor,
@@ -620,10 +620,10 @@ OCIO_ADD_TEST(CPUProcessor, with_one_1d_lut)
               5120,  20480, 65535  };
 
         const std::vector<uint16_t> resImg
-            = {    95,    24,     0,     0,
-                  268,   178,   123,     0,
-                  955,   598,   394,     0,
-                65535,  8589,  1986,     0 };
+            = {    95,    24,     0,     1,
+                  268,   178,   123,     1,
+                  955,   598,   394,     1,
+                65535,  8589,  1986,     1 };
 
         ComputeValues<OCIO::BIT_DEPTH_UINT16, OCIO::BIT_DEPTH_UINT16>( 
             __LINE__, processor,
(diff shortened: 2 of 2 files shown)


doug-walker:
Thanks for the contribution @pennelee !  

Added the devdays25 label.

src/OpenColorIO/ImagePacking.cpp
```diff
@@ -70,7 +70,7 @@ void Generic<Type>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
         inBitDepthBuffer[4*pixelsCopied+0] = *rPtr;
         inBitDepthBuffer[4*pixelsCopied+1] = *gPtr;
         inBitDepthBuffer[4*pixelsCopied+2] = *bPtr;
-        inBitDepthBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (Type)0.0f;
+        inBitDepthBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (Type)1.0f;
```
cozdas: Thanks for your contribution.

The idea is if the input image doesn't have an alpha channel, we need to consider the pixels as fully opaque instead of fully transparent (alpha=0). For floating point types fully opaque is represented as 1.0f but for integer types it needs to be the max value the type can hold (e.g. for 8-bit types it's 255)

Please check BithDepthUtils.h, the templated struct BitDepthInfo will be useful.

pennelee: Ahh, I got you, thanks @cozdas! Let me take a look at that file and fix it, will comment if any questions.

pennelee:
Hi @cozdas ,
Finally I have made some update, could you please review again and let me know if you have any feedback.  Thank you.

src/OpenColorIO/ImagePacking.cpp
```diff
@@ -23,7 +23,8 @@ void Generic<Type>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
                                           Type * inBitDepthBuffer,
                                           float * outputBuffer,
                                           int outputBufferSize,
-                                          long imagePixelStartIndex)
+                                          long imagePixelStartIndex,
+                                          BitDepth outputBitDepth)
```
cozdas: This function Packs the input image (3 or 4 channel, of type 'Type') into 4-channel f32 buffer, and thus the output bit depth should not be part of the conversion. What needs to be done is to fill the missing input alpha value as the max value the "input" type can represent.

Since the GenericImageDesc class is a simplified version of ImageDesc, it doesn't carry the bithDepth information in it like the ImageDesc does, so you are right to pass it as a separate parameter, but it needs to be the input image bit depth, not the output.

src/OpenColorIO/ImagePacking.cpp
```diff
@@ -62,15 +63,21 @@ void Generic<Type>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
...
     int pixelsCopied = 0;
     while(pixelsCopied < outputBufferSize)
-    {
+    {     
         // Reorder channels from arbitrary channel ordering to RGBA 32-bit float.
         inBitDepthBuffer[4*pixelsCopied+0] = *rPtr;
         inBitDepthBuffer[4*pixelsCopied+1] = *gPtr;
         inBitDepthBuffer[4*pixelsCopied+2] = *bPtr;
-        inBitDepthBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (Type)0.0f;
+        inBitDepthBuffer[4 * pixelsCopied + 3] = aPtr ? *aPtr : (Type)(maxValue);
```
cozdas: please keep the formatting consistent with the above lines

src/OpenColorIO/ImagePacking.cpp
```diff
@@ -93,7 +100,8 @@ void Generic<float>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
                                            float * /*inBitDepthBuffer*/,
                                            float * outputBuffer,
                                            int outputBufferSize,
-                                           long imagePixelStartIndex)
+                                           long imagePixelStartIndex,
+                                           BitDepth outputBitDepth)
```
cozdas: this is a specialization of the template function for float input type, so you can simplify this by using hard-coded 1.0f.

src/OpenColorIO/ScanlineHelper.cpp
```diff
@@ -143,7 +143,8 @@ void GenericScanlineHelper<InType, OutType>::prepRGBAScanline(float** buffer, lo
                                                &m_inBitDepthBuffer[0],
                                                *buffer,
                                                m_dstImg.m_width,
-                                               m_yIndex * m_dstImg.m_width);
+                                               m_yIndex * m_dstImg.m_width, 
+                                               m_outputBitDepth);
```
cozdas: as explained above, needs to be m_inputBitDepth

cozdas (changes requested):
Hi @pennelee,
The unit test expected values are now making much more sense, thanks for the fix but the value that needs to be used should be derived from the input bit depth, not output, I added comments to some of the relevant lines. 

I believe the tests will still pass when you use the input depth instead of the output.

src/OpenColorIO/ImagePacking.cpp
```diff
@@ -140,7 +148,7 @@ void Generic<float>::PackRGBAFromImageDesc(const GenericImageDesc & srcImg,
         outputBuffer[4*pixelsCopied+0] = *rPtr;
         outputBuffer[4*pixelsCopied+1] = *gPtr;
         outputBuffer[4*pixelsCopied+2] = *bPtr;
-        outputBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : 0.0f;
+        outputBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : (float)1.0f;
```
cozdas:
```suggestion
        outputBuffer[4*pixelsCopied+3] = aPtr ? *aPtr : 1.0f;
```

pennelee: Wow thanks for the quick review @cozdas .  Sure let me fix that also.

src/OpenColorIO/ImagePacking.cpp
pennelee: Thank you @cozdas for the helpful explanations!

pennelee:
Hi @doug-walker - Could you also review the changes?  Thank you!

PR #2198, opened 2025-09-26
Add setStringVar to Context Object Python Bindings

mdecaria:
`setStringVar` is missing from the python bindings. Its present in the [Java Bindings](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/bindings/java/org/OpenColorIO/Context.java#L20)

Without this method the user is required to use dictionary-style in Python. 
`context["foo"] = ["bar"]`

I've seen this current way catch quite a few people off guard, as they are expecting a method to set arbitrary Context Variables.

With this proposed change, a user can now also do:
`context.setStringVar("foo", "bar")`

### DIFF
--- a/src/bindings/python/PyContext.cpp
+++ b/src/bindings/python/PyContext.cpp
@@ -168,6 +168,8 @@ void bindPyContext(py::module & m)
             { 
                 return StringVarIterator(self); 
             })
+        .def("setStringVar", &Context::setStringVar, "name"_a, "value"_a, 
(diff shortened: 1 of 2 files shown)


doug-walker (approved):
Thank you for the enhancement, Michael!

Shall we label this as a Dev Days contribution?

mdecaria:
> Shall we label this as a Dev Days contribution?

Sure thing, this PR was also logged to encourage others to participate!

PR #2202, opened 2025-09-30
Update documentation for 2.5 release

doug-walker:
Added a What's New page for the 2.5 release and made updates to existing documentation where necessary based on the new features.

A preliminary description has been provided, but more work is needed for the config merge and hue curve transforms. To be done in a follow-up.

I added a "Breaking Changes" section at the top for people that aren't going to read the whole What's New, but maybe that is too dramatic?  Open to suggestions for how to do that better.


docs/guides/using_ocio/tool_overview.rst
```diff
@@ -43,6 +43,10 @@ This command will expand it back out::
 
 The --list option may be used to see the contents of a .ocioz file.
 
+Note: Archive files generated on Windows machines in OCIO 2.4 or earlier should
```
cozdas, line 46: Some clarification: archives generated with earlier versions of OCIO on Windows will still open fine with OCIO v2.5 on Windows. Only the non-Windows OCIO v2.5 builds will give error. In other words, only the (Windows <v2.5) -> (non-Windows v2.5+) path is broken.

To be on the safer side we can still recommend re-generating the archives, but you may want to include "otherwise they may not open properly in Linux/MacOS systems" to clarify what's broken. Up to you.

docs/releases/ocio_2_5.rst
```diff
@@ -0,0 +1,322 @@
...
+
+Breaking Changes
+================
+
+Please be aware of the following changes when upgrading to OCIO 2.5.
+
+For Users
++++++++++
+
+* OCIOZ archive files created on Windows in prior releases are incompatible with OCIO 2.5
```
cozdas: See the comment above.

docs/releases/ocio_2_5.rst
```diff
@@ -0,0 +1,322 @@
...
+* The Python functions getActiveDisplays and getActiveViews now return iterators, please 
+see below for details.
+
+* Please see the section below about version updates to required third-party dependencies.
+
+
+New Feature Guide
+=================
+
+Config Merging (PREVIEW RELEASE)
```
cozdas, line 46: I think it'd be better if the new feature list is sorted by how wide an audience this feature will serve. Not sure if the config merging will be the most widely used feature amongst all. Built-in ACES 2.0 configs, Vulkan support, hue curve transform might be interesting for more people.

doug-walker: It was more widely requested than any of the other features you mention. And it's the one that we are most interested in feedback about.

docs/releases/ocio_2_5.rst
```diff
@@ -0,0 +1,322 @@
...
+
+
+Built-in ACES 2.0 Configs
+*************************
+
+The ACES 2.0 library support was finalized in the OCIO 2.4.2 release. In the 2.5.0 release,
+built-in ACES 2.0 versions of the Studio and CG config are now provided. This allows use of
+the ACES 2.0 Output Transforms as OCIO views.
+
+This `ASWF Open Source Days presentation <https://youtu.be/VZti3UztRE4?feature=shared>`_
```
cozdas, line 96: That's a great explanation indeed, I'm glad that you included this in the documentation. 👍

docs/releases/ocio_2_5.rst
cozdas: Fair enough. I stand corrected :)

cozdas:
Do we want to mention that 

- ociocheck will now check the interop_id field against the known CIF ids.
- ocioconvert will write the colorInteropID attribute to the output image if available.

cozdas (approved):
Look great!

remia (approved):
Looks great thanks Doug.

docs/releases/ocio_2_5.rst
```diff
@@ -0,0 +1,321 @@
...
+built-in ACES 2.0 versions of the Studio and CG config are now provided. This allows use of
+the ACES 2.0 Output Transforms as OCIO views.
+
+This `ASWF Open Source Days presentation <https://youtu.be/VZti3UztRE4?feature=shared>`_ 
+from ILM's Alex Fry provides a great overview of the benefits of the new configs.
+
+In addition, the new configs make expanded use of OCIO v2 features such as File Rules and
+Viewing Rules and add a virtual display to support use with ICC monitor profiles.
+
+Please take note that the display color spaces in the new configs pass through negative
```
carolalynn, line 102: I'm wondering if this should be added to the "breaking changes" section somehow as well?

carolalynn (approved):
I made one small comment re: negative mirroring but it's minor, happy to approve as is. Great work and thank you Doug :)

docs/releases/ocio_2_5.rst
doug-walker: I had originally added something there, but then removed it because it is not really a breaking change. In other words, the ACES 1 configs still clamp negs as before. The ACES 2 config mirrors, but I felt that it was really a new config and so this wasn't a change to existing behavior and that it was therefore sufficient to describe mirroring in the section on the new ACES 2 configs.  (Yes, external ACES 2 configs were released in the spring, but we did bump the major version number of those.)  

Anyway, that was the logic, but am open to adding something, if you want.

carolalynn: Yeah I agree with you, it's not really *breaking*... let's leave it for now, and if we get a lot of feedback/questions around it, we can revisit making it more prominent.

PR #2206, opened 2025-10-06
Issue 2075 : Apply RGB on transpose matrices.

Asheikhm:
Fix : 

Added C-contiguity detection to the `applyRGB` buffer method in `PyCPUProcessor.cpp`. The fix checks if the input NumPy array is stored in row-major order by verifying that the memory strides match the expected C-contiguous layout. If the array isn't C-contiguous (like when someone passes a transposed column-major array without copying it), the method now throws a runtime error saying "applyRGB only supports C-contiguous (row-major) arrays". This prevents the color processor from silently processing data in the wrong order and forces users to ensure their arrays are properly laid out in memory before calling `applyRGB`.


remia (approved):
I ran into the issue while using colour.LUT3D() utility, this was correctly caught by this PR, thank you @Asheikhm !

src/bindings/python/PyUtils.cpp
```diff
@@ -179,6 +179,27 @@ void checkBufferType(const py::buffer_info & info, BitDepth bitDepth)
...
+    ptrdiff_t expected = itemsize;
+    for (py::ssize_t i = ndim - 1; i >= 0; --i)
+    {
+        if (strides[i] != expected) { isC = false; break; }
+        expected *= shape[i];
+    }
+
+    if (!isC)
+    {
+        throw std::runtime_error("function only supports C-contiguous (row-major) arrays");
```
zachlewis, line 199: Should we throw a PyOpenColorIO.Exception here instead of std::runtime_error?

doug-walker: Zach, it's a fair question, but this new code is just following what the other code in this module does. It should stay consistent, in the context of this PR.

I would have no objection for you or someone else to change the way exceptions are raised in this module, but the goal of these Dev Days tasks is to try and keep the focus quite narrow, so I think that would be out of scope here.

zachlewis: Right on!

PR #2214, opened 2025-10-28
Findminizip-ng.cmake now fully handles library names without trailing…

JGoldstone:
There was what looked to be an incomplete handoff between `Findminizip-ng.cmake` and the library `CMakeLists.txt` file, where there was mechanism to handle the case where a third-party installer had installed `minizip-ng` in such a way that neither headers nor library has a `-ng` suffix (and indeed, if you pull down the base `minizip-ng` from GitHub, it installs its products without suffixes).

Anyway, there was provision in the latter file to handle this case if a certain communication variable was set, but the `Findminizip-ng.cmake` module wasn't setting that module.

I'd very much welcome criticism from those who understand both `cmake` in general and the OCIO build system in particular as to whether this is a good way to handle the situation (which ... sometimes I wonder if I'm the only MacPorts user on the planet who uses ASWF tools), or whether some other approach would be more tasteful.

### DIFF
--- a/share/cmake/modules/Findminizip-ng.cmake
+++ b/share/cmake/modules/Findminizip-ng.cmake
@@ -150,6 +150,13 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
                 lib64 lib 
         )
 
+        if (minizip-ng_LIBRARY)
+            cmake_path(GET minizip-ng_LIBRARY STEM _lib_stem)
+            if (_lib_stem STREQUAL "libminizip")
+                set(minizip_LIBRARY ${minizip-ng_LIBRARY})
+            endif()
+        endif()
+
         # Get version from header or pkg-config
         if(minizip-ng_INCLUDE_DIR)
             list(GET minizip-ng_INCLUDE_DIR 0 _minizip-ng_INCLUDE_DIR)


zachlewis:
Something similar came up with [OpenImageIO and "auto-building" zlib-ng](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4639)

> So, for example, an app that itself uses zlib and OIIO might end up itself using zlib-ng inadvertently. Is that always ok? Maybe it's fine, I just think we need to be confident that this is the behavior we want.

Is minizip_LIBRARY is already defined, is it safe to redefine it to point to the un-suffixed minizip-ng library? 
I don't know much about minizip-ng; but it _seems_ to me like this is the intended behavior, although I imagine it's possible to end up building software with rpath references to dynamic vanilla minizip libraries in some parts, and dynamic minizip-ng libraries in others, which seems like it might cause trouble...

I feel like maybe the safest thing to do is to only try to link static un-suffixed minizip.a (if minizip_LIBRARY is already set), but I could be overthinking this.  

Does macports provide static minizip-ng builds, by any chance?

remia:
@JGoldstone thanks for the PR, like @zachlewis I'm not really happy updating `minizip_LIBRARY` like this. If I'm understanding your original issue correctly, we have an issue when trying to hide the minizip-ng symbols, looking at the [conditional](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/CMakeLists.txt#L409) I see that we check for `minizip-ng_COMPAT` which the current Find module we use should set in your case. Can you check if and why this would not be set, maybe there is a bug / edge case we didn't cover.

Issue #2225, opened 2025-12-21
Vulkan GLSL output can emit duplicate layout(set,binding) for multiple textures (2D/1D + 3D), producing invalid descriptor bindings


Sedles:
When generating GPU shaders using GPU_LANGUAGE_GLSL_VK_4_6, OpenColorIO can emit GLSL where multiple distinct resources share the same layout(set = X, binding = Y) within the same shader stage. This is not a valid Vulkan descriptor binding layout: a (set,binding) pair must uniquely identify a single descriptor binding.

This is observed in pipelines that require both a 3D LUT and 1D/2D LUT(s) (e.g., when a display/view transform chain uses both types of LUT textures).

## Affected Versions

OpenColorIO: v2.5.x

## Steps to Reproduce

1. Use an OCIO config + transform chain whose GPU representation requires:
   - at least one 3D LUT texture, and
   - at least one 1D/2D LUT texture (often represented as 2D textures in OCIO GPU output).

2. Create a shader description for Vulkan GLSL:
```
auto desc = OCIO::GpuShaderDesc::CreateShaderDesc();
desc->setLanguage(OCIO::GPU_LANGUAGE_GLSL_VK_4_6);
desc->setDescriptorSetIndex(0, 1);
gpuProc->extractGpuShaderInfo(desc);
const char* shaderText = desc->getShaderText();
```
3. Inspect shaderText and locate the texture/sampler uniform declarations.

## Actual Result

The generated shader may contain duplicate Vulkan binding declarations, e.g.:
```
layout(set = 0, binding = 1) uniform sampler3D ocio_lut3d_0Sampler;
layout(set = 0, binding = 1) uniform sampler2D ocio_lut1d_1Sampler;
```

The core issue is multiple uniforms sharing the same (set,binding).

## Expected Result

For GPU_LANGUAGE_GLSL_VK_4_6, OCIO should emit Vulkan-valid GLSL where:
- each (set,binding) pair is unique for a given shader stage
- 1D/2D and 3D LUT resources do not overlap in binding assignment

doug-walker:
Thanks for the detailed bug report. We will try to get this fixed in OCIO 2.5.1 in early January.

doug-walker:
This issue was fixed in PR #2226 and will be included in OCIO 2.5.1.

PR #2226, opened 2025-12-22
Fix Vulkan texture binding index issue

doug-walker:
Addresses defect #2225.

The binding index was only taking into account the number of textures of a given kind (1D or 3D) rather than the total number of textures.

### DIFF
--- a/share/ci/scripts/windows/install_doxygen.sh
+++ b/share/ci/scripts/windows/install_doxygen.sh
@@ -10,7 +10,7 @@ DOXYGEN_LOCATION="$1"
 choco install jq
 
 # Get the URL of the latest zip package for Doxygen.
-url=$(curl -s 'https://api.github.com/repos/doxygen/doxygen/releases/latest' | jq -r '.assets[] | select(.name | test("doxygen-.*windows.x64.bin.zip")) | .browser_download_url')
+url=$(curl -s 'https://api.github.com/repos/doxygen/doxygen/releases/latest' | jq -r '.assets[] | select(.name | test("doxygen-.*\\.x64\\.bin\\.zip")) | .browser_download_url')
 
 # Download the zip.
 mkdir $DOXYGEN_LOCATION
--- a/src/OpenColorIO/FileRules.cpp
+++ b/src/OpenColorIO/FileRules.cpp
@@ -117,7 +117,7 @@ std::string ConvertToRegularExpression(const char * globPattern, bool ignoreCase
 
         if (globString[idx] == ']')
         {
-            ThrowInvalidRegex(globPattern, globPattern + idx);
+            ThrowInvalidRegex(globPattern, &globString[idx]);
         }
 
         // Full processing from '[' to ']'.
--- a/src/OpenColorIO/fileformats/FileFormatICC.cpp
+++ b/src/OpenColorIO/fileformats/FileFormatICC.cpp
@@ -15,7 +15,6 @@
 #include "ops/gamma/GammaOp.h"
 #include "ops/lut1d/Lut1DOp.h"
 #include "ops/matrix/MatrixOp.h"
-#include "ops/range/RangeOp.h"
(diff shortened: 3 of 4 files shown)


Shootfast:
LGTM!

michaelHADSK:
Looks good but this way the index that is returned by addTexture() is a different one than the one that is passed to getTexture() and doesn't correspond to the array index of the texture anymore, this could be a problem. If you just get the number of textures and loop over them to pass them to Vulkan you might get wrong indices. For example, if you do something like this:

> for(int i=0; i < shaderDesc->getNumTextures(); ++i)
> {
>     vulkan->setTextureIndex(i);
> }
> for(int i=0; i < shaderDesc->getNum3DTextures(); ++i)
> {
>     vulkan->setTexture3DIndex(shaderDesc->getNumTextures() + i);
> }

If the 3D and 1D/2D texture indices are mixed in the generated shader code, the above code will produce wrong values. You would need another function that returns for each texture the index that was returned when calling the addTexture() function. Something like `getTextureIndexInShader( textureIndexInTextureArray, out_textureIndexInShader);`. Clients that have code like the one above would have to update their code then to make sure the right index is passed.
The original issue might also be solved by using separate processors for each LUT so you can give each of them a unique descriptor set index.

src/OpenColorIO/GpuShader.cpp
```diff
@@ -214,7 +214,8 @@ class PrivateImpl
                 << width << " > " << get1dLutMaxWidth();
             throw Exception(ss.str().c_str());
         }
-        unsigned textureIndex = static_cast<unsigned>(m_textures.size());
+        unsigned textureIndex = static_cast<unsigned>(m_textures.size())
+                              + static_cast<unsigned>(m_textures3D.size());
         unsigned numDimensions = static_cast<unsigned>(dimensions);
         Texture t(textureName, samplerName, width, height, 1, channel, numDimensions, interpolation, values);
```
michaelHADSK: To clarify my other comment a bit more:
First maybe rename textureIndex to something that fits better like textureShaderBindingIndex.
Extend `Texture` struct to also store the binding index `Texture t(textureName, ..., textureShaderBindingIndex);`
Then add methods to the public API to retrieve it `uint getTexture[3D]ShaderBindingIndex(textureIndex) { return m_textures[textureIndex].textureShaderBindingIndex;}`
This way users have all the information to properly bind textures to the correct indices.

doug-walker: Thank you @michaelHADSK, that was very helpful. Please check if my latest commit does what you have in mind. 

Note that the new getTextureShaderBindingIndex methods include the textureBindingStart. I'm considering having the index returned by the addTexture methods include that too, so that the code in the Ops doesn't have to worry about that.  What do you think?

And I will add the new methods to the Python binding, once I have the C++ doing what you'd like.

include/OpenColorIO/OpenColorIO.h
```diff
@@ -3503,8 +3503,9 @@ class OCIOEXPORT GpuShaderCreator
      *   The 'values' parameter contains the LUT data which must be used as-is as the dimensions and
      *   origin are hard-coded in the fragment shader program. So, it means one GPU texture per entry.
      * 
-     * \return Index of the texture. For shading languages using explicit texture bindings, the return
-     *         value is the same as the texture binding index in the generated shader program.
+     * \return Shader binding ndex of the texture. For shading languages using explicit texture bindings,
```
michaelHADSK: typo "ndex"

src/OpenColorIO/GpuShader.cpp
michaelHADSK: Yes, thanks, that's what I had in mind, approved it. Including the textureBindingStart sounds good to me. I'm not sure when I will be able to test the changes in our software but as far as I can see everything should work this way.

Issue #2220, opened 2025-11-25
"Array Subscript Out Of Range" Exception Can Be Thrown Under A Floating Point Rounding Edge Case.


SimonCSmith:
I'm getting an "array subscript out of range" exception (Windows, Debug Build, OCIO 2.5.0) inside of **[reach_m_from_table](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/b59cd64368992aec3dfb4901f3de52b8f7878862/src/OpenColorIO/ops/fixedfunction/ACES2/Transform.cpp#L91)** due to a hue of 360.0 being passed in. This causes a i_lo=361, and i_hi=362. Given the ACES2::Table1D rt is only 362 in size, you get the overrun access scenario. 

The reason you can get a hue of 360.0 is due to floating point rounding issues. These step from the calculations in the [fwd](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/b59cd64368992aec3dfb4901f3de52b8f7878862/src/OpenColorIO/ops/fixedfunction/FixedFunctionOpCPU.cpp#L1101) call.

The RGB values coming into Renderer_ACES_OutputTransform20::fwd
```
0.742242277
0.0931933373
0.321542144
```
ACES2::RGB_to_Aab converts to
```
0.659631252
103.849297
-2.28881836e-05
```
Then ACES2::Aab_to_JMh converts it on to
```
62.3068
103.850
360.000
```
The atan2() in the Aab_to_JMh returns `-2.20398064e-07`, which then results in `-1.26278792e-05` going into the [_wrap_to_hue_limit](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/b59cd64368992aec3dfb4901f3de52b8f7878862/src/OpenColorIO/ops/fixedfunction/ACES2/Common.h#L21) call, and adding that to 360.000000 leaves it at a value of 360.000000.

Note that the random data obtained from the buffer over-read is technically immaterial (as it's multiplied by 0 in the [lerp](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/OpenColorIO/MathUtils.h#L53) call) but it does mean you get an exception firing off due to Windows _STL_VERIFY macro which halts execution in the debugger.

[...]

PR #2227, opened 2025-12-24
Bug/2220 array subscript out of range exception can be thrown under a floating point rounding edge case

KevinJW:
Reproduced #2220

Using Debug build on Windows 11 using VS 2022 Community edition:

1. Open VS 2022 Developer Console
2. Clone OpenColorIO main branch
3. Add new **aces_ot_20_edge_cases** test from this branch
4. Configure for Debug build
```
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX=c:\Temp -DCMAKE_BUILD_TYPE=Debug ..
cmake --build . --config Debug --parallel
```
5. Run ctest

Fix is to add an additional entry at the end of the tables, to handle the boundary condition due to atan2() returning the interval [-pi, +pi] each new entry filled with the corresponding entry from the lower entries to maintain continuity


KevinJW:
Just to note whatever the solution it should probably be back ported to 2.4.x branch. Something to discuss in next TSC

KevinJW:
> Just to note whatever the solution it should probably be back ported to 2.4.x branch. Something to discuss in next TSC

Decision was to not proceed with a back port for now. If any vendors need the fix then this patch should be easy to backport.

PR #2229, opened 2025-12-26
Add release signing workflow using Sigstore

### LINKED ISSUE #2034: Consider signing release artifacts
The OpenSSF Best Practices Badge suggests [signing release artifacts](https://www.bestpractices.dev/en/criteria), using OpenEXR's [release-sign.yml](https://github.com/AcademySoftwareFoundation/openexr/blob/main/.github/workflows/release-sign.yml) workflow as a template. It's triggered on release creation and does these steps:
1. Runs `get archive` to generate a `<release>.tar.gz` artifact
2. Signs the `<release>.tar.gz` via [sigstore](https://www.sigstore.dev/) 
3. Uploads the resulting sigstore signature file along with the tarball.

pmady:
## Summary

This PR adds a GitHub Actions workflow that signs release artifacts using Sigstore, following the OpenSSF Best Practices Badge recommendations.

## Changes

Added `.github/workflows/release-sign.yml` that:
1. Triggers on release publication
2. Creates a `.tar.gz` archive of the source tree at the release tag
3. Signs the archive using `sigstore/gh-action-sigstore-python@v3.2.0`
4. Uploads both the tarball and `.sigstore.json` credential bundle to the release

## Verification

Users can verify a downloaded release using:
```bash
pip install sigstore
sigstore verify github --cert-identity https://github.com/AcademySoftwareFoundation/OpenColorIO/.github/workflows/release-sign.yml@refs/tags/<tag> OpenColorIO-<tag>.tar.gz
```

## Reference

Based on OpenEXR's [release-sign.yml](https://github.com/AcademySoftwareFoundation/openexr/blob/main/.github/workflows/release-sign.yml) workflow as suggested in the issue.

Closes #2034


.github/workflows/release-sign.yml
```diff
@@ -0,0 +1,67 @@
...
+        # should omit the 'v', so the tarball "OpenColorIO-2.4.0.tar.gz"
+        # extracts files into "OpenColorIO-2.4.0/...". This matches
+        # the GitHub release page autogenerated artifact conventions.
+        run: |
+          echo OCIO_PREFIX=OpenColorIO-${TAG//v}/ >> $GITHUB_ENV
+          echo OCIO_TARBALL=OpenColorIO-${TAG//v}.tar.gz >> $GITHUB_ENV
+        shell: bash
+
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
doug-walker, line 52: I notice that the OpenEXR version of this is using a more recent checkout action (6.0.1 vs. 4.2.2). Using 4.2.2 is consistent with what we do in our other Actions scripts, so it's fine for this PR. Just wanted to flag it as something that we may want to update and see what breaks if we try to use something more recent on our more recent vfx platform releases that are not subject to the glibc problem.

doug-walker (approved):
Thanks for the contribution, @pmady !

Issue #2228, opened 2025-12-25
[Bug] Heap-use-after-free in OpenColorIO::ThrowInvalidRegex when parsing malformed FileRules


oneafter:
### Description

We discovered a Heap-use-after-free (UAF) vulnerability in OpenColorIO (OCIO). The crash occurs specifically when processing invalid regular expressions within the FileRules section of a YAML configuration.

The issue manifests in OpenColorIO::ThrowInvalidRegex, where the application attempts to read a string (via strlen) from a memory region that was just freed by the internal state management of std::regex.

### Environment
- OS: Linux x86_64
- Complier: Clang with -fsanitize=address
- Build Configure: Release
- Affected Version: `master branch`

### Vulnerability Details
- Target: OpenColorIO (v2.5 dev)
- Vulnerability Type: Heap-use-after-free (READ of size 1)
- Severity: High (Denial of Service / Potential Memory Corruption)
- Source File: src/OpenColorIO/FileRules.cpp
- Function: OpenColorIO_v2_5dev::(anonymous namespace)::ThrowInvalidRegex
- Line Number: 55
- Root Cause Analysis: 
1. The application parses a YAML config containing a FileRule with a pattern.
2. ValidateRegularExpression calls BuildRegularExpression (Frame #-5), which eventually invokes the std::regex compiler.
3. During regex compilation (std::basic_regex::_M_compile), the internal NFA state vector reallocates memory (_M_realloc_insert), freeing the old buffer (as seen in the "freed by thread T0" stack trace).
4. Crucially, when an error is detected or an exception is thrown during this process, the code enters ThrowInvalidRegex (Frame #-3).
5. Inside ThrowInvalidRegex (Line 55), the code attempts to construct a std::string (Frame #-2) from a pointer. The ASAN report indicates this pointer points to the memory region that was just freed by the regex engine's reallocation.
6. This suggests that ThrowInvalidRegex is accessing a dangling pointer, possibly referring to a part of the regex pattern or an internal error message string that became invalid after the vector resizing.

### Reproduce
1. Compile the ocio test [harness](https://github.com/oneafter/1225/blob/main/harness.cpp) with AddressSanitizer enabled (-fsanitize=address -g)
2. Run the harness with the attached [repro](https://github.com/oneafter/1225/blob/main/uaf) input:
```
./harness < repro
```

ASAN report
```
[...]

PR #2231, opened 2026-01-05
Issue #2228 - Out-of-bounds issue in ThrowInvalidRegex

cozdas:
Addressing Issue #2228 - Heap-use-after-free in OpenColorIO::ThrowInvalidRegex

This turned out to be not a UAF issue but an out-of-bounds-access issue where the throw was accessing the incorrect (and potentially shorter) string with an index derived from another (and potentially longer) string.

### DIFF
--- a/share/ci/scripts/windows/install_doxygen.sh
+++ b/share/ci/scripts/windows/install_doxygen.sh
@@ -10,7 +10,7 @@ DOXYGEN_LOCATION="$1"
 choco install jq
 
 # Get the URL of the latest zip package for Doxygen.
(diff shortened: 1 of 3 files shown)


src/OpenColorIO/FileRules.cpp
```diff
@@ -117,7 +117,7 @@ std::string ConvertToRegularExpression(const char * globPattern, bool ignoreCase
 
         if (globString[idx] == ']')
         {
-            ThrowInvalidRegex(globPattern, globPattern + idx);
+            ThrowInvalidRegex(globPattern, &globString[idx]);
```
cozdas, line 120: idx is looping through the **globString** content which is a potentially longer string (for example when ignore space is true, which expands every alpha character to [xX] string). So it was trying to access pass end of **globPattern**.

PR #2234, opened 2026-01-09
Fix Windows Doxygen install script 

cozdas:
as Doxygen v1.16.0 changed the released zip name.


doug-walker:
Merging this immediately as it is just a fix for the CI.

Issue #1993, opened 2024-07-19
Update Common LUT Format to v4 (SMPTE version)


doug-walker:
SMPTE is working on a standard for the Academy/ASC Common LUT Format. This will include some minor changes and the OCIO implementation will need to be adjusted accordingly.

zachlewis:
Just out of curiosity, what kind of changes are we anticipating? And how would this affect ownership of further development? 
(Like, would an Academy working group be in charge of making changes for v5? Would SMPTE have to make further changes for v6?)

doug-walker:
Closing as completed via PR #2265.

PR #2265, opened 2026-02-11
Adsk Contrib - Add Support for SMPTE ST 2036-1 compliant CLF files

cozdas:
Adding reading and writing support for **SMPTE ST 2036-1** compliant CLF files. Addressing the issue #1993 

- Added "Id" element support per the spec.

- Extended the CTFVersion class to handle non-numeric formats per the spec (which uses xmlns as the version)

- CLF xml parser now strips the namespaces from the elements by default. This makes it possible to parse files with simple name spaces but complex name-spaced clf files may still fail. There is an internal switch to turn on/off the name space stripping. This way elements which need to retain the name spaces (such as the Info element) can still get the un-stripped names.

- Extended the Input and Output Descriptor collection to hold multiple entries (similar to Description field)

- CLF writer now writes the SMPTE xmlns version as well as the CompCLFVersion attribute. The resulting files can be read both by the Academy CLF parsers and SMPTE CLF parsers.

- Generated cache hash IDs are now in the 8-4-4-4-12 UUID format to help complying with SMPTE id requirements

- ociomakeclf tool now takes "--generateid" switch for inserting newly added "Id element" into the target clf file.

- Description, Input and Output Descriptor elements now collect the language attribute too.

- non-default namespace attributes (e.g. xmlns:*) are now collected as root level attributes.

- Some re-factoring in CLF xml parsing code was needed to be able to add extra attribute collection.

- Added OCIO_CHECK_EQUAL_STR unit test macro to ease some pain with "char *" comparisons.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -36,19 +36,52 @@ best leverage them.  We have an OCIO User Experience (UX) working group that is
 gradually working on more coverage.
 
 The documentation is a work-in-progress and we would love to have your help to 
-improve it!  An easy way to get involved is to join the #docs or #ux channel on 
+improve it!  An easy way to get involved is to join the #opencolorio channel on 
 :ref:`slack`.
+
+
+Roadmap
+=======
+
+The project attempts to maintain a roadmap to give the community insight into
+upcoming development. The items are grouped into "Now", "Next", and "Later" categories.
+"Now" indicates work that is currently in-progress. "Next" indicates work that is being
+planned for imminent development (meaning now would be a good time to provide your input).
+The roadmap is available `here. <https://roadmap.opencolorio.org>`_
 
 
 Community
 =========
+
+.. _meetings:
+
+Meetings
+********
+
+The OpenColorIO project has a Technical Steering Committee meeting on Zoom every two weeks
+at noon LA time. The Zoom link is available from the `OCIO Calendar. <https://calendar.opencolorio.org>`_
+
+There is a general working group and "office hours" meeting once a month in the same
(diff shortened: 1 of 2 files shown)


tests/data/files/reference_nested_2.ctf
zachlewis, line 1: Does this mean CTF Reference nodes do not support SMPTE-flavored CLF?

(I suppose I'm okay with this, since Reference nodes aren't publicly documented or explicitly supported)

doug-walker: Yes, the smpte_all_metadata test in FileFormatCTF_tests.cpp checks multiple OutputDescriptors (line 1464).

References will support any loadable LUT format. The one the test uses happens to be pre-SMPTE, but I agree with you that it could give the wrong impression. We will update that to reference one of the other files.

Thanks for the quick review Zach!

include/OpenColorIO/OpenColorTypes.h
```diff
@@ -977,6 +977,13 @@ extern OCIOEXPORT const char * METADATA_NAME;
  */
 extern OCIOEXPORT const char * METADATA_ID;
 
+/**
+ * An ID when stored as an XML element rather than as an attribute. This is the
+ * preferred mechanism in the SMPTE ST 2036-1 version of the CLF format. If
+ * present, it is only available from the top-level FormatMetadata.
+ */
+ extern OCIOEXPORT const char * METADATA_ID_ELEMENT;
```
remia, line 985: Minor, should this be added to PyTypes.cpp?

cozdas: Added in the second commit. Please check.

Issue #526, opened 2018-03-20
GPU Unit Tests - Add Dx11 support


hodoulp:
Ideally, the same unit test code should be used for OpenGL, OpenGL Core Profile and Dx11.

hodoulp:
Dependency to #525

remia:
Note this should probably also include the support of DirectX backend for ociodisplay, similar to what was done for Metal.

doug-walker:
@carolalynn, I think we should keep this open. This is a significant issue - there is currently no way to test that the HLSL shaders are working.

I believe @num3ric has already done some preliminary work on this.

carolalynn:
Sounds good! I think I meant to add the label instead of actually closing
it, so thanks! :)

num3ric:
I should indeed come back to this. My test app was DX12 and quite verbose, but a DX11 one integrated into the library is certainly desirable and would help use maintain stable integrations every release.

PR #2271, opened 2026-03-07
Add DirectX 12 GPU backend for automated unit testing on Windows

num3ric:
Introduce a DirectX 12 / HLSL rendering backend alongside the existing OpenGL / GLSL and Metal / MSL backends, enabling the GPU unit test suite to run natively on Windows without requiring an OpenGL context.

Key changes:

- GraphicalApp abstract interface (graphicalapp.h/cpp)
  - Backend-agnostic base class extracted from OglApp.  OglApp and MetalApp now inherit from it.
- DxApp (dxapp.h/cpp) -- DirectX 12 backend
  - Off-screen RGBA32F render target, full-screen triangle via SV_VertexID, staging readback, SM 6.0 DXC shader compilation.
- HLSLBuilder (hlsl.h/cpp) -- HLSL shader generation
  - Translates GpuShaderDesc into HLSL pixel shaders with 1D and 3D LUT texture uploads in RGBA32F format. 
- CMake integration
  - OCIO_DIRECTX_ENABLED option, FetchContent for DirectX-Headers, auto-copy of DXC runtime DLLs to the test output directory.
- Test tolerance adjustments
  - Minor epsilon increases for 4 tests due to DX12/SM6.0 FMA and pow() precision differences.
  - All 263 GPU tests pass on the DirectX 12 backend.

Build and run:
```
# Configure (OCIO_DIRECTX_ENABLED defaults to ON on Windows)
  cmake -S . -B build -DCMAKE_BUILD_TYPE=Release

  # Build the GPU test binary
  cmake --build build --target test_gpu_exec --config Release

  # Run GPU tests with the DX12 backend
  ctest --test-dir build -C Release -R test_dx
```

Further Notes:
* For a few releases now, we have had to manually test HLSL shader generation. This integration would significantly facilitate this process moving forward.
* While I maintain ownership, this code change was partly AI-assisted. I used Claude Code to help me finish the DirectX implementation (which I had started more than a year ago), debug failing tests during implementation, review the changes, etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,6 +287,14 @@ message(STATUS "")
 message(STATUS "Checking for GPU configuration...")
 include(CheckSupportGL)
 
+# DirectX 12 is only available on Windows.
+if(WIN32)
+    option(OCIO_DIRECTX_ENABLED "Enable DirectX 12 GPU rendering support" ON)
+else()
+    set(OCIO_DIRECTX_ENABLED OFF CACHE BOOL "Enable DirectX 12 GPU rendering support" FORCE)
+endif()
+mark_as_advanced(OCIO_DIRECTX_ENABLED)
+
 
 ###############################################################################
 # Check for ARM neon here because we need to know if ARM NEON is supported
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,6 +295,11 @@ else()
 endif()
 mark_as_advanced(OCIO_DIRECTX_ENABLED)
 
+# Vulkan is cross-platform but requires an external SDK, so it is off by
+# default; enable explicitly with -DOCIO_VULKAN_ENABLED=ON.
+option(OCIO_VULKAN_ENABLED "Enable Vulkan GPU rendering support" OFF)
+mark_as_advanced(OCIO_VULKAN_ENABLED)
+
 
 ###############################################################################
 # Check for ARM neon here because we need to know if ARM NEON is supported
--- a/docs/quick_start/installation.rst
+++ b/docs/quick_start/installation.rst
@@ -287,6 +287,8 @@ Here are the most common OCIO-specific CMake options (the default values are sho
 - ``-DOCIO_BUILD_TESTS=ON`` (Set to OFF to not build the unit tests)
 - ``-DOCIO_BUILD_GPU_TESTS=ON`` (Set to OFF to not build the GPU unit tests)
 - ``-DOCIO_USE_HEADLESS=OFF`` (Set to ON to do headless GPU rendering)
+- ``-DOCIO_DIRECTX_ENABLED=ON`` (Windows only; set to OFF to disable the DirectX 12 GPU rendering backend used by ``test_dx``. Forced OFF on non-Windows platforms.)
+- ``-DOCIO_VULKAN_ENABLED=OFF`` (Set to ON to enable the Vulkan GPU rendering backend used by ``test_vulkan``. Requires the Vulkan SDK to be installed and findable by CMake.)
 - ``-DOCIO_WARNING_AS_ERROR=ON`` (Set to OFF to turn off warnings as errors)
 - ``-DOCIO_BUILD_DOCS=OFF`` (Set to ON to build the documentation)
 
--- a/share/cmake/modules/install/InstallDirectX-Headers.cmake
+++ b/share/cmake/modules/install/InstallDirectX-Headers.cmake
@@ -8,12 +8,18 @@
 
 include(FetchContent)
 
+if(OCIO_DirectX-Headers_RECOMMENDED_VERSION)
+    set(DirectX-Headers_VERSION ${OCIO_DirectX-Headers_RECOMMENDED_VERSION})
+else()
+    set(DirectX-Headers_VERSION ${DirectX-Headers_FIND_VERSION})
+endif()
+
 set(FETCHCONTENT_BASE_DIR "${CMAKE_BINARY_DIR}/ext/build/DirectX-Headers")
 set(DIRECTX_HEADERS_BUILD_TEST OFF CACHE BOOL "" FORCE)
 
 FetchContent_Declare(DirectX-Headers
     GIT_REPOSITORY https://github.com/microsoft/DirectX-Headers.git
-    GIT_TAG        v1.619.1
+    GIT_TAG        v${DirectX-Headers_VERSION}
 )
 
 FetchContent_MakeAvailable(DirectX-Headers)
@@ -23,5 +29,4 @@ FetchContent_MakeAvailable(DirectX-Headers)
 # the Microsoft::DirectX-Headers target via the upstream CMakeLists.
 if(TARGET Microsoft::DirectX-Headers)
     set(DirectX-Headers_FOUND TRUE)
-    set(DirectX-Headers_VERSION "1.619.1")
 endif()
--- a/share/cmake/modules/install/InstallDirectX-Headers.cmake
+++ b/share/cmake/modules/install/InstallDirectX-Headers.cmake
@@ -17,3 +17,11 @@ FetchContent_Declare(DirectX-Headers
 )
 
 FetchContent_MakeAvailable(DirectX-Headers)
+
+# Signal success to ocio_install_dependency so ocio_handle_dependency does not
+# abort at the next required-check. FetchContent_MakeAvailable has just created
+# the Microsoft::DirectX-Headers target via the upstream CMakeLists.
+if(TARGET Microsoft::DirectX-Headers)
+    set(DirectX-Headers_FOUND TRUE)
+    set(DirectX-Headers_VERSION "1.619.1")
+endif()
--- a/src/apps/ociodisplay/main.cpp
+++ b/src/apps/ociodisplay/main.cpp
@@ -64,7 +64,7 @@ float g_display_gamma{1.0f};
 int g_channelHot[4]{1, 1, 1, 1}; // show rgb
 int g_viewsMenuID;
 
-OCIO::OglAppRcPtr g_oglApp;
+OCIO::GraphicalAppRcPtr g_oglApp;
 
 void UpdateOCIOGLState();
 
@@ -115,14 +115,14 @@ static void InitImageTexture(const char * filename)
         }
     }
 
-    OCIO::OglApp::Components comp = OCIO::OglApp::COMPONENTS_RGBA;
+    OCIO::GraphicalApp::Components comp = OCIO::GraphicalApp::COMPONENTS_RGBA;
     if (img.getNumChannels() == 4)
     {
-        comp = OCIO::OglApp::COMPONENTS_RGBA;
+        comp = OCIO::GraphicalApp::COMPONENTS_RGBA;
     }
     else if (img.getNumChannels() == 3)
     {
-        comp = OCIO::OglApp::COMPONENTS_RGB;
+        comp = OCIO::GraphicalApp::COMPONENTS_RGB;
     }
     else
     {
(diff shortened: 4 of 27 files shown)


num3ric:
Links with issue #526. Note that I chose DX12 over DX11 to be forward-looking. It remains backwards compatible with the SM5 shader generation.

_EDIT: Would appreciate help with testing other GPU apps to ensure they all continue to function properly._

tests/gpu/GPUUnitTest.cpp
```diff
@@ -631,6 +570,7 @@ int main(int argc, const char ** argv)
     bool printHelp = false;
     bool useMetalRenderer = false;
     bool useVulkanRenderer = false;
+    bool useDxRenderer = false;
```
doug-walker, line 576: When compiling on Mac, I get unused variable warnings on lines 572 and 573.   I'm surprised that the CI passed given that it is supposed to handle warnings as errors. But please either put these in ifdef or remove the ifdefs for the command-line arguments.

tests/gpu/CMakeLists.txt
```diff
@@ -58,6 +106,11 @@ if(MSVC AND BUILD_SHARED_LIBS)
     set(NEW_PATH "${NEW_PATH}\\\;${GLUT_INCLUDE_DIR}/../bin")
     set(NEW_PATH "${NEW_PATH}\\\;${GLEW_INCLUDE_DIRS}/../bin")
 
-    set_tests_properties(test_gpu PROPERTIES ENVIRONMENT PATH=${NEW_PATH})
+    if(OCIO_GL_ENABLED)
+        set_tests_properties(test_gpu PROPERTIES ENVIRONMENT PATH=${NEW_PATH})
+    endif()
+    if(WIN32 AND OCIO_DIRECTX_ENABLED)
+        set_tests_properties(test_dx PROPERTIES ENVIRONMENT PATH=${NEW_PATH})
+    endif()
```
doug-walker, line 94: I had to add the following for the Vulkan GPU tests to run successfully:

```
    if(OCIO_VULKAN_ENABLED)
        set_tests_properties(test_vulkan PROPERTIES ENVIRONMENT PATH=${NEW_PATH})
    endif()
```

src/libutils/oglapphelpers/CMakeLists.txt
```diff
@@ -1,20 +1,41 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright Contributors to the OpenColorIO Project.
 
-if(NOT OCIO_GL_ENABLED)
-    message(WARNING "GL component missing. Skipping oglapphelpers.")
+if(NOT OCIO_GL_ENABLED AND NOT (WIN32 AND OCIO_DIRECTX_ENABLED) AND NOT OCIO_VULKAN_ENABLED)
+    message(WARNING "GL component missing, DirectX disabled, and Vulkan disabled. Skipping oglapphelpers.")
```
doug-walker, line 5: Please add a description of OCIO_DIRECTX_ENABLED to the list of CMake variables in:
OpenColorIO/docs/quick_start/installation.rst

And, if you could do me a favor, please add OCIO_VULKAN_ENABLED at the same time.

doug-walker (changes requested):
This is awesome @num3ric !

I was able to compile this on Windows and run all three GPU test suites: GL, Vulkan, and DX successfully!

One thing that surprised me is that the DX tests seem to take a lot longer to run than the GL or Vulkan tests, any thoughts about that? Here's what I see on my machine:
```
      Start 10: test_gpu
10/13 Test #10: test_gpu .........................   Passed   12.12 sec
      Start 11: test_dx
11/13 Test #11: test_dx ..........................   Passed   19.55 sec
      Start 12: test_vulkan
12/13 Test #12: test_vulkan ......................   Passed    9.75 sec
```
Here's the GPU info for my machine:
```
GL Vendor:    NVIDIA Corporation
GL Renderer:  Tesla T4/PCIe/SSE2
GL Version:   4.6.0 NVIDIA 528.24
GLSL Version: 4.60 NVIDIA
```

I tested that ociochecklut, ocioconvert, and ociodisplay still run. These still use the GL path, but at least nothing broke.

I compiled and ran the tests and apps on macOS as well, just to check nothing broke.

Great work!

tests/gpu/CMakeLists.txt
```diff
@@ -41,10 +41,58 @@ target_link_libraries(test_gpu_exec
...
+	add_test(NAME test_dx COMMAND test_gpu_exec --dx)
+endif()
+if(OCIO_VULKAN_ENABLED)
+	add_test(NAME test_vulkan COMMAND test_gpu_exec --vulkan)
+endif()
+
+# Copy dxcompiler.dll and dxil.dll to the test output directory.
+# These are required at runtime when DXC (IDxcCompiler3) is used for SM6.0 shader compilation.
+# The Redist/D3D path is the stable, version-independent redistribution location.
+if(WIN32 AND OCIO_DIRECTX_ENABLED)
```
remia, line 59: On my test workstation the unit tests crash, I checked dxdiag and it advertise DX 12_1 support but the dxcompiler.dll found here seems to be around 5y old. When I manually copy the DLL from [dxcompiler](https://github.com/microsoft/DirectXShaderCompiler) latest release, the tests run successfully. The workstation is stuck on Win10 and not updated so probably not expecting it to work flawlessly but maybe there is a way to improve the error. Just a note for now.

tests/gpu/CMakeLists.txt
```diff
@@ -41,10 +41,58 @@ target_link_libraries(test_gpu_exec
 		testutils
 )
 
-add_test(NAME test_gpu COMMAND test_gpu_exec)
+if(OCIO_GL_ENABLED)
+    add_test(NAME test_gpu COMMAND test_gpu_exec)
```
remia: At this point could we rename `test_gpu` to `test_opengl` to make it less confusing?

CMakeLists.txt
```diff
@@ -287,6 +287,14 @@ message(STATUS "")
 message(STATUS "Checking for GPU configuration...")
 include(CheckSupportGL)
 
+# DirectX 12 is only available on Windows.
```
remia, line 290: Not from this PR but maybe we should use this opportunity to also add the same option for `OCIO_VULKAN_ENABLED`.

src/apps/ociodisplay/main.cpp
remia, line 43: Is it correct to say adding support for DirectX and/or Vulkan here would be more work than for Metal due to the use of GLUT (which we should probably move away from)? And for DX12 specifically, it currently only support headless rendering and more work would be needed for windowing support?

It would be nice to have these modes available to use ociodisplay with debugger/profilers like Nvidia NSight / Renderdoc (like you can already do with XCode Metal debugger).

share/cmake/modules/install/InstallDirectX-Headers.cmake
```diff
@@ -0,0 +1,19 @@
...
+#
+###############################################################################
+
+include(FetchContent)
+
+set(FETCHCONTENT_BASE_DIR "${CMAKE_BINARY_DIR}/ext/build/DirectX-Headers")
+set(DIRECTX_HEADERS_BUILD_TEST OFF CACHE BOOL "" FORCE)
+
+FetchContent_Declare(DirectX-Headers
+    GIT_REPOSITORY https://github.com/microsoft/DirectX-Headers.git
```
remia, line 15: Not a blocker I think but may be nice to offer a way to specify a path manually for Windows systems without internet access (unless you already did and I missed it in the PR).

remia (approved):
Great work @num3ric , I like the cleanup you did adding `GraphicalApp`. I cannot comment on the DX12 implementation details but this otherwise looks good to me (left a couple of minor comments).

Tangent but maybe time to remove the `GPU_LANGUAGE_HLSL_DX11` duplicated enum for the next version?

num3ric:
Addressed all notes thus far. Please refer to each commit message for detailed descriptions regarding improvements, fixes, and/or findings.

src/apps/ociodisplay/main.cpp
num3ric: Re-GLUT & metal, that's correct. An alternative framework would be something like GLFW, which apparently can also support DX12 (though I've never tried).

The DX12 implementation is not headless currently, since my starting point was the separate app I'd made for testing purposes (similar to ociodisplay). With this PR, we still create a full window (see `CreateWindowA`), swapchain, present loop, but we render the OCIO transforms to an offscreen RT for test readback. I'd initially picked this approach since it's closer to practical use cases (what most users are going to be implementing), and mirrors the OpenGL implementation. Though as I learned through recent updates, the new Vulkan path is indeed different being fully headless/compute.

We could add an headless option here too, but I'd suggest that being a separate line of work/PR if needed.

docs/quick_start/installation.rst
```diff
@@ -287,6 +287,8 @@ Here are the most common OCIO-specific CMake options (the default values are sho
 - ``-DOCIO_BUILD_TESTS=ON`` (Set to OFF to not build the unit tests)
 - ``-DOCIO_BUILD_GPU_TESTS=ON`` (Set to OFF to not build the GPU unit tests)
 - ``-DOCIO_USE_HEADLESS=OFF`` (Set to ON to do headless GPU rendering)
+- ``-DOCIO_DIRECTX_ENABLED=ON`` (Windows only; set to OFF to disable the DirectX 12 GPU rendering backend used by ``test_dx``. Forced OFF on non-Windows platforms.)
```
num3ric, line 290: Default-enabled for windows platforms, but let me know if you prefer leaving it off by default for now.

share/cmake/modules/install/InstallDirectX-Headers.cmake
remia: I think `ocio_handle_dependency` create the PACKAGE_VERSION variable for you here so you don't have to hard code it.

tests/gpu/CMakeLists.txt
remia: Thanks for the related updates @num3ric ! I'll give this a go next week and report back here.

src/apps/ociodisplay/main.cpp
remia: Ok, so it looks like if/when me move away from GLUT (GLFW, SDL or other candidates), we could add DX12 support for `ociodisplay` relatively easily that's good to know thanks.

share/cmake/modules/install/InstallDirectX-Headers.cmake
num3ric: Fixed, and rebased to address merge conflicts. (Probably should have merged-resolved avoid a force push, but too late now.😅)

num3ric:
By the way, if I delete my local GL shader cache (`%LOCALAPPDATA%\NVIDIA\GLCache`), I get the following results:

- OpenGL Tests: 13.35s  (warm cache only takes 7.67s)
- DX12 Tests: 14.48s

This is what we'd expect, given that DX12 has an extra DXIL signing pass. The way to enable caching on the DX12 would be to add a PSO cache with ID3D12PipelineLibrary. I don't think it's worth it at the moment though, since it's ok to run tests without a cache anyway.

tests/gpu/CMakeLists.txt
remia: Tested with my default environment, it found dxcompiler (presumably very old version), then failed on every test with the same error as before (HRESULT of 0x80004002):
`Found dxcompiler.dll (version 10.0.19041.685 (WinBuild.160101.0800)): C:/Program Files (x86)/Windows Kits/10/Redist/D3D/x64/dxcompiler.dll`

After I used OCIO_DXCOMPILER_DLL to point to the release I download from github, it works fine and every test passed:
`Found dxcompiler.dll (version 1.9.2602.17): ..../dxc_2026_02_20/bin/x64/dxcompiler.dll`

Thanks @num3ric

num3ric:
Minor tweaks on the last commit from a Claude Opus 4.7 `/review` run.

doug-walker (approved):
@num3ric , thanks for your updates, they addressed all my comments. I've tested again on Windows and the DX, GL, and Vulkan tests all succeed. I tested on Mac again to ensure GL and Metal still work. Everything seems good to merge!

Unfortunately the Linux CI is failing at the moment, unrelated to your PR. Once we fix that, I will rerun the CI on the PR and, assuming it passes, which I think it will, go ahead and merge.  

Thanks again, this is a huge benefit to the project!

PR #2273, opened 2026-03-10
Fix linking to self-built deps on Windows + Clang

dimula73:
OCIO fails to build on Windows when llvm-mingw environment is used. It happens because of incorrect conditions are used to mimic the static library names. This patch makes these names generation consistent with the CMakeLists.txt files of the corresponding deps.

1) Expat: 'MD' suffix is added on MSVC only, 'd' suffix is added on WIN32, i.e. on clang as well.
2) ZLib: 'd' suffix is added only on MSVC, not on clang


dimula73:
What is DCO? Isn't the author field enough? I always thought it is reviewers, who sign-off commits...

dimula73:
Anyway, I'll rebase and add the lines a bit later :)

cozdas:
Thanks.

Please check the contribution process doc at https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/CONTRIBUTING.md for the details.

dimula73:
> but the DCO check should pass

Okay, I think it should be fixed now.

cozdas (approved):

remia (approved):

PR #2276, opened 2026-03-19
Adsk Contrib - Hue curve python binding was not copying all parameters

doug-walker:
When setting a GradingBSplineCurve on a GradingHueCurve using the Python binding, the spline type and slopes were not being copied correctly.  This issue does not affect the C++ interface.

In addition, fixed a minor problem in CTFReaderHelper.cpp that was causing the Platform Latest Action to fail.

### DIFF
--- a/share/cmake/modules/install/InstallZLIB.cmake
+++ b/share/cmake/modules/install/InstallZLIB.cmake
@@ -53,8 +53,8 @@ if(NOT ZLIB_FOUND AND OCIO_INSTALL_EXT_PACKAGES AND NOT OCIO_INSTALL_EXT_PACKAGE
 
     set(ZLIB_INCLUDE_DIRS "${_EXT_DIST_ROOT}/${CMAKE_INSTALL_INCLUDEDIR}")
 
-    # Windows need the "d" suffix at the end.
-    if(WIN32 AND BUILD_TYPE_DEBUG)
+    # Windows need the "d" suffix at the end (only for MSVC).
+    if(MSVC AND BUILD_TYPE_DEBUG)
         set(_ZLIB_LIB_SUFFIX "d")
     endif()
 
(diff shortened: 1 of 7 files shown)


src/OpenColorIO/fileformats/ctf/CTFReaderHelper.cpp
```diff
@@ -862,11 +862,11 @@ void CTFReaderDescElt::start(const char **  atttributes )
     m_language = {};
 
     const char ** attr = atttributes;
-    while (*attr)
+    while (attr && *attr)
     {
         if (0 == Platform::Strcasecmp(ATTR_LANGUAGE, *attr))
         {
-            if (!attr || !(attr + 1))
+            if (!(attr + 1) || !*(attr + 1))
```
remia: Not sure if this is going to fix the warning as I seem to recall `!(attr + 1)` was the culprit as adding 1 to a pointer would never test false?

cozdas: This is my bad.

We don't need to check if the passed attributes param is null or not, as in the rest of the system it's assumed non-null. So what this code needs to do is 
- checking if the pointer in the array is null
- check if the string that pointer points to is empty

I guess this is what we need to do
`
while (*attr && **attr)
{
...
    if (!*(attr + 1) || !**(attr + 1))
...
}
`

Does it make sense to you too?

src/OpenColorIO/fileformats/ctf/CTFReaderHelper.cpp
```diff
@@ -862,11 +862,11 @@ void CTFReaderDescElt::start(const char **  atttributes )
     m_language = {};
 
     const char ** attr = atttributes;
-    while (*attr)
+    while (attr && *attr)
```
cozdas:
```suggestion
    while (*attr && **attr)
```

src/OpenColorIO/fileformats/ctf/CTFReaderHelper.cpp
cozdas:
```suggestion
            if (!*(attr + 1) || !**(attr + 1))
```

doug-walker: I changed it to better follow the conventions of the surrounding code.  Please see latest commit.

PR #2281, opened 2026-04-07
Fix OpenGL ES type issues in ACES2 FixedFunction Ops

### LINKED ISSUE #2269: Bug: Type Mismatch in GLES 3.0 Shader Code of _Add_Reach_table() and _Add_Cusp_table()
This is the code I generated for ACES 2.0 ODT from ACEScg to SDR 100 nit Rec.709 on Android platform. The compile complains that there are int/float type mismatch:
```
float _ocio_res_reach_m_table_0_sample(float h)
{
  float i_base = floor(h);
  float i_lo = i_base + 1;
  float i_hi = i_lo + 1;
  float lo = texture(_ocio_res_reach_m_table_0Sampler, vec2((i_lo + 0.5) / 363, 0.0)).r;
  float hi = texture(_ocio_res_reach_m_table_0Sampler, vec2((i_hi + 0.5) / 363, 0.5)).r;
  float t = h - i_base;
  return mix(lo, hi, t);
}
```
Here, the compiler complains around `+1` and `/363`.

```
vec3 _ocio_res_gamut_cusp_table_0_sample(float h)
{
  int i = int(h) + 1;
  int i_lo = int(max(float(0), float(i + 0)));
  int i_hi = int(min(float(361), float(i + 2)));
  while (i_lo + 1 < i_hi)
  {
    float hcur = _ocio_res_gamut_cusp_table_0_hues_array[i];
    if (h > hcur)
    {
      i_lo = i;
    }
    else
    {
      i_hi = i;
    }
    i = (i_lo + i_hi) / 2;
  }
  vec3 lo = texture(_ocio_res_gamut_cusp_table_0Sampler, vec2((i_hi - 1 + 0.5) / 363, 0.5)).rgb;
  vec3 hi = texture(_ocio_res_gamut_cusp_table_0Sampler, vec2((i_hi + 0.5) / 363, 0.5)).rgb;
  float t = (h - _ocio_res_gamut_cusp_table_0_hues_array[i_hi - 1]) / (_ocio_res_gamut_cusp_table_0_hues_array[i_hi] - _ocio_res_gamut_cusp_table_0_hues_array[i_hi - 1]);
  return mix(lo, hi, t);
}
```
Here, the compiler complains around `+ 0.5`.

For now, I can make some simple change to the code to make it work, but I have two concerns:
1. Is there a GLSL ES 3.0 test for this codebase's CI?
2. Is it a correct fix to simply change the constants to float type?

remia:
Address #2269

### DIFF
--- a/src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
+++ b/src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
@@ -901,7 +901,7 @@ std::string _Add_Cusp_table(
 
     if (dimensions == GpuShaderDesc::TEXTURE_1D)
     {
-        ss.newLine() << ss.float3Decl("lo") << " = " << ss.sampleTex1D(name, std::string("(" + ss.floatKeyword() + "(i_hi) - 1 + 0.5) / ") + ss.floatKeyword() + "(" + std::to_string(g.gamut_cusp_table.total_size) + ")") << ".rgb;";
+        ss.newLine() << ss.float3Decl("lo") << " = " << ss.sampleTex1D(name, std::string("(" + ss.floatKeyword() + "(i_hi) - 1.0 + 0.5) / ") + ss.floatKeyword() + "(" + std::to_string(g.gamut_cusp_table.total_size) + ")") << ".rgb;";
         ss.newLine() << ss.float3Decl("hi") << " = " << ss.sampleTex1D(name, std::string("(" + ss.floatKeyword() + "(i_hi) + 0.5) / ") + ss.floatKeyword() + "(" + std::to_string(g.gamut_cusp_table.total_size) + ")") << ".rgb;";
     }
     else
--- a/src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
+++ b/src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
@@ -563,18 +563,18 @@ std::string _Add_Reach_table(
     ss.indent();
 
     ss.newLine() << ss.floatDecl("i_base") << " = floor(h);";
-    ss.newLine() << ss.floatDecl("i_lo") << " = i_base + " << table.base_index << ";";
-    ss.newLine() << ss.floatDecl("i_hi") << " = i_lo + 1;";
+    ss.newLine() << ss.floatDecl("i_lo") << " = i_base + " << ss.floatKeyword() << "(" << table.base_index << ");";
+    ss.newLine() << ss.floatDecl("i_hi") << " = i_lo + 1.0;";
 
     if (dimensions == GpuShaderDesc::TEXTURE_1D)
     {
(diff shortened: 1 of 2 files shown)


src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
```diff
@@ -901,13 +901,13 @@ std::string _Add_Cusp_table(
 
     if (dimensions == GpuShaderDesc::TEXTURE_1D)
     {
-        ss.newLine() << ss.float3Decl("lo") << " = " << ss.sampleTex1D(name, std::string("(i_hi - 1 + 0.5) / ") + std::to_string(g.gamut_cusp_table.total_size)) << ".rgb;";
-        ss.newLine() << ss.float3Decl("hi") << " = " << ss.sampleTex1D(name, std::string("(i_hi + 0.5) / ") + std::to_string(g.gamut_cusp_table.total_size)) << ".rgb;";
+        ss.newLine() << ss.float3Decl("lo") << " = " << ss.sampleTex1D(name, std::string("(" + ss.floatKeyword() + "(i_hi) - 1 + 0.5) / ") + ss.floatKeyword() + "(" + std::to_string(g.gamut_cusp_table.total_size) + ")") << ".rgb;";
```
doug-walker:
```suggestion
        ss.newLine() << ss.float3Decl("lo") << " = " << ss.sampleTex1D(name, std::string("(" + ss.floatKeyword() + "(i_hi) - 1.0 + 0.5) / ") + ss.floatKeyword() + "(" + std::to_string(g.gamut_cusp_table.total_size) + ")") << ".rgb;";
```

doug-walker (approved):

src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp
remia: Good catch, thanks!

remia: Added a unit test to check the 2D Texture path for OpenGL as it was not yet done (maybe for Vulkan and Metal), for OpenGL ES there is no 1D Texture support so nothing more to be done.

KevinJW:
Have not looked at the exact code this now emits, but when I looked at the original issue, I did wonder about the code in the cusp table ...

```
  int i_lo = int(max(float(0), float(i + 0)));
  int i_hi = int(min(float(361), float(i + 2)));
```

The original blink/c++/etc assumed integer division properties when doing the binary search so used integer types. I guess we'll need to check carefully about all these type conversions and the available function support under the various shader languages.

remia:
Just double checked with the OpenGL ES tests I'm running, all tests pass successfully provided I'm using the following defaults precision qualifier (without this almost everything fails).

```
precision highp float;
precision highp int;
precision highp sampler2D;
precision highp sampler3D;
```

doug-walker:
@KevinJW , could we get your review on this please? We need two approvals in order to merge. Thx!

cozdas (approved):

KevinJW (approved):

PR #2282, opened 2026-04-11
Adsk Contrib - Update CI to VFX 2026

doug-walker:
- Added Linux VFX CY2026
- Removed Linux VFX CY2022
- Updated actions commands, based on dependabot warnings
- Locked actions commands to fixed hashes, like OIIO
- Removed/modified the macOS runners based on TSC discussion, this reduces total runners to 20
- Removed support for Python 3.9 in both CI and Wheel generation, based on TSC discussion

- Included a very minor bug fix in GradingTone.cpp

The new build matrix may be found here (scroll to the bottom table):
https://docs.google.com/spreadsheets/d/1Iw5uRw6voB54wFpkSSOGrcAXkciZfx9nlBSVHz6N4P8/edit?usp=sharing

At the TSC we discussed what runner image to use for macOS and Windows. I left the images as they were because they are already the options that are best aligned with the VFX Platform for CY2026.

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -41,11 +41,11 @@ jobs:
       CC: gcc
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 50
-      - name:  Install sonar-scanner and build-wrapper
-        uses: sonarsource/sonarcloud-github-c-cpp@v2
+      - name:  Install build-wrapper
+        uses: SonarSource/sonarqube-scan-action/install-build-wrapper@v4
       - name: Install docs env
         run: share/ci/scripts/linux/dnf/install_docs_env.sh
       - name: Install tests env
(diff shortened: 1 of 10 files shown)


remia (approved):

remia:
Thanks Doug looks good to me! It could be nice to update the actions/checkout and actions/setup-python similarly for Platform Latest / Dependencies Latest as we currently have related warnings in these 2 workflows.

doug-walker:
Yes, totally agree Remi!  I should have marked this as a draft as I wanted to test that the CI passed before updating the other workflows. And I was going to ask at the TSC if we should remove some of the Intel macOS runs.

remia:
We should also remove Python 3.9 from the package metadata in [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/setup.cfg#L10) and [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/setup.cfg#L24) plus I'm seeing Platform and Dependencies latest still uses `python-version: '3.9'`, but other wise looks good to me, thanks @doug-walker !

remia (approved):

doug-walker:
Note: I believe it is expected that the analysis CI will fail on PRs from a fork. In any case, it was already failing, so this should at least get us closer.

cozdas (approved):

doug-walker:
The CI failures are expected here, so will go ahead and merge this.

PR #2285, opened 2026-04-14
Adsk Contrib - Update Python documentation requirements

doug-walker:
Fixed the issues with the outdated installations requested by the documentation system, as mentioned by dependabot.

This fixes three "Dependabot alerts" vulnerabilities in the project's Security and Quality tab:
"setuptools vulnerable to Command Injection via package URL"
"setuptools has a path traversal vulnerability in PackageIndex.download that leads to Arbitrary File Write"
"urllib3 redirects are not disabled when retries are disabled on PoolManager instantiation"

I was unable to update Sphinx because it seems our documentation template is not compatible.

Note: It is expected that the Linux 2022 CI will fail, however, we are moving that out of the CI matrix in a separate PR.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,7 @@
 urllib3<3
 # The builds for documentation fails with <0.18.0
 docutils>=0.18.1
-sphinx<=9.1.0
+sphinx<=7.1.2
 six
 testresources
 recommonmark
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,6 +1,3 @@
-# Fix an issue with the OCIO's Linux container images that have OpenSSL under 1.1.1.
-# If the container images are updated with OpenSSL 1.1.1+, the restriction on
-# urllib3 version <2 can be removed.
 urllib3<3
 # The builds for documentation fails with <0.18.0
 docutils>=0.18.1
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,14 +1,14 @@
 # Fix an issue with the OCIO's Linux container images that have OpenSSL under 1.1.1.
 # If the container images are updated with OpenSSL 1.1.1+, the restriction on
 # urllib3 version <2 can be removed.
-urllib3<2
+urllib3<3
 # The builds for documentation fails with <0.18.0
 docutils>=0.18.1
-sphinx<=7.1.2
+sphinx<=9.1.0
 six
 testresources
 recommonmark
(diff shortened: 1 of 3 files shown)


zachlewis (approved):

zachlewis:
Apparently,  urllib3 v2 is causing problems on the CY2022 containers.

Copilot says....

> 
> The failure is coming from the CY2022 (“linux-old”) container’s OpenSSL (`OpenSSL 1.0.2k-fips`), which is incompatible with `urllib3 v2` (it requires OpenSSL 1.1.1+). Sphinx fails while importing `sphinx.builders.linkcheck` because that import pulls in urllib3.
> 
> Rather than pinning `urllib3<2` globally (which blocks updating doc deps for newer CYs), we can **pin only on old-OpenSSL containers** by adding a small conditional in `share/ci/scripts/linux/dnf/install_docs_env.sh`:
> 
> ```bash
> pip3 install -r $HERE/../../../../../docs/requirements.txt
> 
> OPENSSL_VERSION="$(python3 -c 'import ssl; print(ssl.OPENSSL_VERSION)' || true)"
> echo "Detected: ${OPENSSL_VERSION}"
> 
> if echo "${OPENSSL_VERSION}" | grep -qE 'OpenSSL 1\.0\.|OpenSSL 1\.1\.0'; then
>   echo "Old OpenSSL detected; forcing urllib3<2 for compatibility."
>   pip3 install --upgrade --force-reinstall "urllib3<2"
> fi
> ```
> 
> This keeps CY2022 green while allowing CY2023+ to move to urllib3 v2 and newer doc tooling.
> 
> Ref: failing job log in Actions run `24375992243`, job `71189536261`.
>

doug-walker:
Thanks for the review Zach! Please note that we are removing CY2022 from the CI in PR #2282, so that doesn't need to fixed.

zachlewis:
Ohhh yes. And now I remember, I think you mentioned this in the TSC meeting. Forget I said anything! LGTM.

docs/requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Fix an issue with the OCIO's Linux container images that have OpenSSL under 1.1.1.
 # If the container images are updated with OpenSSL 1.1.1+, the restriction on
 # urllib3 version <2 can be removed.
```
cozdas: Is this comment still valid? At least the "<2" part needs to be updated it seems.

cozdas (approved):

PR #2302, opened 2026-04-30
Improve CMake and Actions settings

doug-walker:
Several minor build fixes:
- Fix a CMake error in how OCIO_NAMESPACE was being used. This was discussed at the TSC meeting and the preference was for this to continue to be the full string that is used. Therefore, cleaning up some code that was incorrectly using it as part of a longer string.
- Reducing the frequency of Dependabot PRs and requesting only security fixes. Currently it seems to be proposing fixes for minor version increases that have no security aspect.
- Fix documentation comment that was causing a build failure on Windows.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,13 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "quarterly"
+    # Only do security updates rather than all version updates.
+    open-pull-requests-limit: 0
 
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "quarterly"
+    # Only do security updates rather than all version updates.
+    open-pull-requests-limit: 0
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,13 +8,13 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "quarterly"
+      interval: "monthly"
     # Only do security updates rather than all version updates.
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
```diff
@@ -8,9 +8,13 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "quarterly"
+    # Only do security updates rather than all version updates.
+    open-pull-requests-limit: 0
```
remia, line 13: Maybe I didn't find the documentation section related to this but it seems this would simply disable dependabot completely instead of doing what the comment mention (I'm also fine with it but maybe there is a middle ground)?

remia (approved):

.github/dependabot.yml
doug-walker: Good question Remi, but apparently this is how one is supposed to request only security updates. See item three here:
https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/customizing-dependabot-security-prs#preparing-to-customize-pull-requests

Security updates are not affected by this limit, so it essentially just turns of version number updates.
https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#open-pull-requests-limit-

remia: That's my bad for missing that thanks for the links Doug.

carolalynn (approved):

remia:
Thanks for fixing the Windows documentation issue Doug, I saw this in my PR too.

doug-walker:
Will go ahead and merge this since Analysis runs never succeed on PRs.

PR #2308, opened 2026-05-05
Adsk Contrib - Miscellaneous improvements suggested by Claude

doug-walker:
I used Claude Code (Sonnet 4.6) to review most of the core library. It found a number of bugs and minor issues such as pointers not being checked before use or memory allocations not being thoroughly validated. The only item that I would describe as a major bug was an issue that caused the inverse video shader for the GradingPrimaryTransform to not compile, which happened to be hidden by a flaw in one of the unit tests. The rest of it is a large collection of minor fixes to improve library robustness.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,20 +14,16 @@ would be naive to say our code is immune to every exploit.
 
 ## Reporting Vulnerabilities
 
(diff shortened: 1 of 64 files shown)


remia (approved):

KevinJW (approved):
Based

Issue #2256, opened 2026-01-26
Proposal to remove pystring dependency


KevinJW:
Our current usage of pystring falls into 2 camps

String manipulation e.g. split, startswith, endswith, splitlines, lstrip etc
Filesystem interaction pystring::os::*

I would propose that this be carried out in 3 stages:

1) Replace string functionality with code based on our own StringUtils (add functionality where needed). We already have some tests for replacing the split() behaviour for example.
2) Replace filesystem interaction by calls to std::filesystem (C++17)
3) Remove all pystring from CMake

meimchu:
Based on some work I did in #2310 I think there is value in adding some filesystem functionality to PathUtils and am interested in exploring this further if still desired.

Some of the name may need to be different in order to avoid confusion. FileExists() is one where its current use case seems quite specific with caching hashes. If we want some basic filesystem usages such as filesystem::exists, it will probably need a different name than FileExists() to avoid confusion?

I can take a look at the first stage and replace pystring::split() with existing StringUtils::Split() and go from there.

meimchu:
I have created a PR that should address the first phase and would love to get some feedback on it. Thank you! https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/2313

PR #2310, opened 2026-05-13
ociocheck to check OpenColorIO library version first

### LINKED ISSUE #2303: ociocheck to check OpenColorIO library version first
I brought this up as a question recently in ASWF OpenColorIO's slack channel. If I were to use `ociocheck` built from an earlier version of OCIO (v2.2 for example) and try to load in a OCIO config file built with a later version of OCIO (v2.4 for example), the operation will stop immediately with a generic message `ERROR: Error could not read...`. I thought it would be nice to have a more deterministic error message by checking for `ocio_profile_version:` value first. This is especially so when the various DCCs can have different OCIO versions loaded. This would be just a nice little quality of life improvement for debugging.

`ociocheck` already prints out `OpenColorIO Library Version:` at the beginning so my plan is to just expand on that by checking it against `ocio_profile_version` prior to `OCIO::CreateFromFile()`.

meimchu:
The idea behind this PR originated with me using `ociocheck` and getting stuck with config read error (#2303). So I just wanted clearer error message when the error was due to the file path provided being non-existent. Upon looking at the existing issues, I noticed there is an issue about removing pystring dependency (#2256). So I took this opportunity to get started by adding `PathUtils::IsPathAbs()`, `PathUtils::NormalizePath()`... etc with standard library `filesystem` and using them in `Config`. Also a change to `ExceptionMissingFile` raised if file does not exist. (Can revert back to just raise Exception if it's causing too much API change)

I would appreciate if @KevinJW has any thought about this as a starting point and the right track. Thank you! Fix #2303.

### DIFF
--- a/src/OpenColorIO/Config.cpp
+++ b/src/OpenColorIO/Config.cpp
@@ -11,6 +11,7 @@
 #include <vector>
 #include <regex>
 #include <functional>
+#include <filesystem>
 
 #include <pystring.h>
 
@@ -1167,10 +1168,10 @@ ConstConfigRcPtr Config::CreateFromFile(const char * filename)
         return CreateFromBuiltinConfig(uri.c_str());
     }
 
-    if (!FileExists(filename))
+    if (!std::filesystem::exists(filename))
     {
         std::ostringstream oss;
-        oss << "'" << filename << "' does not exist.";
+        oss << "'" << filename << "' file does not exist.";
         throw ExceptionMissingFile(oss.str().c_str());
     }
 
@@ -6019,7 +6020,7 @@ bool Config::isArchivable() const
     // Current archive implementation needs a working directory to look for LUT files and 
     // working directory must be an absolute path.
     const char * workingDirectory = getWorkingDir();
-    if ((workingDirectory && !workingDirectory[0]) || !IsPathAbs(workingDirectory))
+    if ((workingDirectory && !workingDirectory[0]) || !pystring::os::path::isabs(workingDirectory))
     {
         return false;
     }
@@ -6028,10 +6029,10 @@ bool Config::isArchivable() const
     auto validatePathForArchiving = [](const std::string & path) 
     {
         // Using the normalized path.
-        const std::string normPath = NormalizePath(path, false);
+        const std::string normPath = pystring::os::path::normpath(path);
         if (    
                 // 1) Path may not be absolute.
-                IsPathAbs(normPath)  || 
+                pystring::os::path::isabs(normPath)  || 
                 // 2) Path may not start with double dot ".." (going above working directory).
                 pystring::startswith(normPath, "..") ||
                 // 3) A context variable may not be located at the start of the path.
--- a/src/OpenColorIO/Config.cpp
+++ b/src/OpenColorIO/Config.cpp
@@ -1155,7 +1155,7 @@ ConstConfigRcPtr Config::CreateFromFile(const char * filename)
 {
     if (!filename || !*filename)
     {
-        throw ExceptionMissingFile ("The config filepath is missing.");
+        throw ExceptionMissingFile("The config filepath is missing.");
     }
 
     // Check for URI Pattern: ocio://<config name>
@@ -1165,6 +1165,13 @@ ConstConfigRcPtr Config::CreateFromFile(const char * filename)
     if (std::regex_search(uri, match, uriPattern))
     {
         return CreateFromBuiltinConfig(uri.c_str());
+    }
+
+    if (!FileExists(filename))
+    {
+        std::ostringstream oss;
+        oss << filename << " does not exists.";
+        throw ExceptionMissingFile(oss.str().c_str());
     }
 
     std::ifstream ifstream = Platform::CreateInputFileStream(
@@ -6012,7 +6019,7 @@ bool Config::isArchivable() const
     // Current archive implementation needs a working directory to look for LUT files and 
     // working directory must be an absolute path.
     const char * workingDirectory = getWorkingDir();
-    if ((workingDirectory && !workingDirectory[0]) || !pystring::os::path::isabs(workingDirectory))
+    if ((workingDirectory && !workingDirectory[0]) || !IsPathAbs(workingDirectory))
     {
         return false;
     }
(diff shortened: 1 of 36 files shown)


src/OpenColorIO/Config.cpp
```diff
@@ -1155,7 +1155,7 @@ ConstConfigRcPtr Config::CreateFromFile(const char * filename)
 {
     if (!filename || !*filename)
     {
-        throw ExceptionMissingFile ("The config filepath is missing.");
+        throw ExceptionMissingFile("The config filepath is missing.");
```
KevinJW: Probably OK for now, but it is worth noting that as you highlighted in the issue this condition isn't quite the same as a missing file. ideally we'd have a different type for this.

src/OpenColorIO/Config.cpp
```diff
@@ -6012,7 +6019,7 @@ bool Config::isArchivable() const
     // Current archive implementation needs a working directory to look for LUT files and 
     // working directory must be an absolute path.
     const char * workingDirectory = getWorkingDir();
-    if ((workingDirectory && !workingDirectory[0]) || !pystring::os::path::isabs(workingDirectory))
+    if ((workingDirectory && !workingDirectory[0]) || !IsPathAbs(workingDirectory))
```
KevinJW: My gut suggests we should deal with the switch away from the pystring implementation as a separate issue to cleaning up the error messages. Changing the implementation to use std::filesystem could have more subtle issues

src/OpenColorIO/PathUtils.cpp
```diff
@@ -120,6 +121,11 @@ bool FileExists(const std::string & filename, const Context & context)
     return (!hash.empty());
 }
 
+bool FileExists(const std::string & filename)
```
KevinJW: As an example of the subtlety When comparing this function with the other overload

```
bool FileExists(const std::string & filename, const Context & context)
```
Shows a much more complicated implementation which immediately makes me wonder why. Should we be dealing with the IOProxy here or not, if not then should the function have a different name? or at least a clear documentation stating why etc.

At this point I have not reasoned this through as I don't have the understanding of the IOProxy workings without studying it.

meimchu: That is all good points. I cannot say for sure myself about that other particular FileExists overload. I see it being used for trying to find lut files with `search_path` context. I have no issue with it being a different name when we want to replace pystring::path usages!

Perhaps we can create our own internal pystring-like module that basically does some of those pystring tasks that we use then? That was sort of where I was heading by adding these path-related stuff to PathUtil.

src/OpenColorIO/Config.cpp
meimchu: No issue with that. I can reduce the scope of this specific change to just simply a better message if config file does not exists.

src/OpenColorIO/Config.cpp
meimchu: Agreed! Though should I switch this to just OCIO::Exception for now then? Ultimately it's a runtime error so it kind of fits!

src/OpenColorIO/PathUtils.cpp
meimchu: Reverted changes done to PathUtil which can be put in a separate PR.

src/OpenColorIO/Config.cpp
meimchu: Reverted changes done to PathUtil which can be put in a separate PR.

meimchu:
For now, I've reverted the more ambitious changes and opted for a targeted solution. I looked a little deeper into the existing PathUtils::FileExists function which seems caching file hash within a context. My shallow understanding is that it helps speed up the various lut files found via search_path. I also tried creating an empty context to try to utilize that same FileExists() with the config file path but it doesn't feel right to me.

The primary impact of this though is that ExceptionMissingFile() is raised if input file path does not exist. Relying on ifstream.fail() for other reasons why reading that file could fail.

src/OpenColorIO/Config.cpp
KevinJW: Maybe just add a comment at the site of the 2 throws indicating that ideally they should be distinguishable?

src/OpenColorIO/Config.cpp
```diff
@@ -1153,9 +1154,11 @@ ConstConfigRcPtr Config::CreateFromEnv()
 
 ConstConfigRcPtr Config::CreateFromFile(const char * filename)
 {
+    // Specifically check if a config filepath is provided.
+    // TODO: Consider raising a different exception for this?
```
doug-walker: I don't think we need a TODO here, what you did is fine.

doug-walker (approved):
Thanks Mei!

doug-walker:
@meimchu , if you merge in the latest changes from the main branch it should fix the CI problem for you.

meimchu:
> @meimchu , if you merge in the latest changes from the main branch it should fix the CI problem for you.

Just did! Thank you for fixing that. Though I noticed my DCO check failed and maybe that's due to the fact that I did not `git rebase` with `--signoff`. Can that be rectified by simply doing another `git commit` with `--signoff`? I didn't realize I needed to rebase with signoff.

meimchu:
Did `git rebase HEAD~10 --signoff` and then `git push --force-with-lease origin feature/config_exists_check` per suggestions from the failed DCO job. All jobs appear to have passed now.

doug-walker (approved):

KevinJW (approved):
Based on visual check of the code changes, this looks fine to me.

I'll add it to my growing OCIO backlog to verify with running code tests.

remia:
Looks good to me, might be good to check if the OCIO library size increases with the use of std::filesystem, as we have ongoing aim to reduce the binary size.

meimchu:
> Looks good to me, might be good to check if the OCIO library size increases with the use of std::filesystem, as we have ongoing aim to reduce the binary size.

Thanks! I don't have a very good way to check the binary size for the other OSs. On my mac, the binary size change for the dylib is negligible at best. I am also working on another issue to reduce pystring dependency so that should hopefully help towards that goal.

remia:
Thanks for checking @meimchu, sounds good!

meimchu:
Looks like this little PR is ready for merge!

PR #2244, opened 2026-01-12
Feature/optimiser debugging improvements

KevinJW:
This is a couple of changes related to improving the debug output of the optimiser.

Basically it outputs the results of each stage within the passes, that allows for better reasoning about what changes though the process, I've tried to ensure we don't suffer too much from the overhead from the extra logging during non-debug conditions.

I have not done any performance evaluation to confirm this so far.

As a side effect there is also a slight change to the includes to remove the dependency on pystring in the op/optimiser code.

PR #2313, opened 2026-05-19
Reduce pystring usages

meimchu:
First step of pystring removal by improving StringUtils (https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/2256)
StringUtils have been updated to include some pystring functionalities. Focused on replacing:
- pystring::split()
- pystring::strip()
- pystring::replace()
- pystring::mul()
- pystring::startswith()
- pystring::endswith()

These usages have also been replaced throughout the modules that use these as well. Only pystring::os usages should be left now.

Additional changes of note:
- Temporarily changed ci_workflow.yml. Additional explanation below.
- Added an additional include path for OpenFX build.

@KevinJW please take a look when you have a moment! Thank you.


meimchu:
I have temporarily removed a couple of combinations of our CICD Linux tests here to at least allow more tests to run. Since one tests from the OS category can stop the other jobs from running.

There is an on-going discussions in Slack regarding some Linux combinations and the recent `aswf/ci-ocio:2023` docker image update. That is not the focus of this PR so I have temporarily removed it as a distraction.

I will not merge this request without a conclusion is reached in that regards.

The initial message from Jean-Francois Panisset:
```
See my comments in #wg-ci, it does seem like the update of the aswf/ci-ocio:2023 image from 2023.2 to 2023.3 has "changed something", and now the self build of OpenEXR 3.4 (the version from 2025) is breaking since that makes use of C++ 20 features.

Quick workaround would be to pin this build to use the aswf/ci-ocio:2023.2 image which should restore the previous behavior, but more generally it would be interesting to figure out "what does building latest OCIO with VFX 2025 OpenEXR in a VFX 2023 environment" really mean, and is that something we need / want to support.
```

src/utils/StringUtils.h
```diff
@@ -304,6 +340,18 @@ inline bool Remove(StringVec & list, const std::string & entry)
...
 }
 
+// Repeat the 'str' by the 'n' value.
+inline std::string Multiply(const std::string & str, int n)
+{
+    // Early exits
+    if (n <= 0) return "";
+    if (n == 1) return str;
+
+    std::ostringstream os;
```
KevinJW, line 352: I wonder if something like the following might be 'better'/faster

```
std::string repeat_fast(const std::string& input, std::size_t n) {
    if (n == 0) {
        return {};
    }
    if (n == 1) {
        return input;
    }
    std::string result;
    const auto input_size = input.size();
    result.reserve(input_size * n);
    result = input;
    std::size_t current = 1;
    while (current * 2 <= n) {
        result += result;
        current *= 2;
    }
    const auto remaining = n - current;
    if (remaining > 0) {
        result += std::string_view(result.data(), input_size * remaining);
    }
    return result;
}

```
Edited because I nerd sniped myself and ran some performance tests
And again avoid an extra allocation

src/utils/StringUtils.h
```diff
@@ -249,39 +270,54 @@ inline std::string::size_type ReverseFind(const std::string & subject, const std
     return subject.rfind(search);
 }
 
-// In place replace the 'search' substring by the 'replace' string in 'str'.
-inline bool ReplaceInPlace(std::string & subject, const std::string & search, const std::string & replace)
+// In place replace the 'search' substring by the 'replace' string in 'subject'. Limited by 'count'.
+inline bool ReplaceInPlace(std::string & subject, const std::string & search, const std::string & replace, int count)
```
KevinJW: count should possibly be a size_t ? if we really need signed values then C++20 has std::ssize_t or  we could use std::ptrdiff_t

src/utils/StringUtils.h
```diff
@@ -249,39 +270,54 @@ inline std::string::size_type ReverseFind(const std::string & subject, const std
...
 {
     if (search.empty()) return false;
 
     bool changed = false;
 
-    size_t pos =  0;
+    size_t pos = 0;
+    int iter = 0;
```
KevinJW: ditto

src/utils/StringUtils.h
```diff
@@ -249,39 +270,54 @@ inline std::string::size_type ReverseFind(const std::string & subject, const std
...
-// Replace the 'search' substring by the 'replace' string in 'str'.
-inline std::string Replace(const std::string & subject, const std::string & search, const std::string & replace)
+// In place replace the 'search' substring by the 'replace' string in 'subject'.
+inline bool ReplaceInPlace(std::string & subject, const std::string & search, const std::string & replace)
+{
+    return ReplaceInPlace(subject, search, replace, -1);
+}
+
+// Replace the 'search' substring by the 'replace' string in 'subject'. Limited by 'count'.
+inline std::string Replace(const std::string & subject, const std::string & search, const std::string & replace, int count)
```
KevinJW: and again?

src/utils/StringUtils.h
KevinJW: might want to call the function Repeat or Repeat_n rather than Multiply. 

Note that reserve() can raise an exception if we would go over std::string::max_size() but likely we would run out of memory before that happens.

src/utils/StringUtils.h
meimchu: Thank you for this suggestion. I've replaced it with `size_t` as you've suggested to make it consistent with the other `size_t` usages. As well, I've removed a signed integer comparison `ReplaceInPlace()` accordingly.

src/utils/StringUtils.h
meimchu: More `size_t` replacement as suggested!

src/utils/StringUtils.h
meimchu: More `size_t` replacement as suggested!

src/utils/StringUtils.h
meimchu: What a wonderful suggestion. I just want to speak briefly on what I've done so far to get your eyes on it.
1) Implemented your function above as `Repeat()`. Made some small changes for syntax consistency.
2) I attempted to put a max_limit check as you've pointed out prior to `reserve()`. I chose not to throw an exception but I'm open to suggestions. It is a new behaviour compared to the old `pystring::mul`.
3) I have replaced the single `StringUtils::Multiply()` usage found in the code base (which was using `pystring::mul()` before).
4) I did do a quick benchmark testing between Multiply() and Repeat(). I got a string of 23 chars in length and repeated it 1000 times. The difference is significant and I like the pointer arithmetic in this implementation. I'm happy to remove `Multiply()` completely in favour of `Repeat()` if that is the desired decision.
```
Multiply() Total Time: 388.173 ms
Multiply() Avg Time:   0.0388174 ms/iter

Repeat() Total Time:   4.99192 ms
Repeat() Avg Time:     0.000499192 ms/iter
```

src/utils/StringUtils.h
```diff
@@ -340,18 +340,46 @@ inline bool Remove(StringVec & list, const std::string & entry)
...
+{
+    // Early exit and match pystring::mul behaviour.
+    if (n == 0) { return {}; }
+    if (n == 1) { return str; }
+    const auto str_size = str.size();
+
+    // Check for overflow if n is greater than maximum repeatable amount via string max_size.
+    // New behaviour compared to pystring::mul and StringUtil::Multiply.
+    // limits.h says size_t max size is 18446744073709551615.
+    if (n > str.max_size() / str_size) { return {}; }
```
KevinJW: Ah I didn't mean you needed to detect this case, I don't think you need this. if you remove it then the reserve() call below will deal with it. With this check you could think of it as silently "failing".

src/utils/StringUtils.h
```diff
@@ -340,18 +340,46 @@ inline bool Remove(StringVec & list, const std::string & entry)
...
     std::ostringstream os;
     for(size_t i = 0; i < n; ++i) { os << str; }
     return os.str();
 }
 
+// Repeat the 'str' by the 'n' value.
+inline std::string Repeat(const std::string & str, size_t n)
+{
+    // Early exit and match pystring::mul behaviour.
+    if (n == 0) { return {}; }
```
KevinJW, line 361: I've not benchmarked it but it is possible that under some circumstances we might want to check `|| str.empty()` to catch the case of appending an empty string many times. though the extra test probably slows down the function when `n` is small.

meimchu: I agree with you that doing an early check on `str.empty()` for early return makes sense. I ran a benchmark on both Multiply() and Repeat() with an empty string input and 10,000 repeats. Both Multiply() and Repeat() will return itself and Repeat() just barely beat out Multiply(). I think at this point, I feel pretty good about just removing Multiply() in favour of Repeat() if you agree with it.

```
Multiply() Total Time: 0.129125 ms
Multiply() Avg Time:   1.29125e-05 ms/iter

Repeat() Total Time:   0.119125 ms
Repeat() Avg Time:     1.19125e-05 ms/iter
```

src/utils/StringUtils.h
meimchu: Removed and let it YOLO! When I was testing this, I did mistakenly let it ran and it filled up my swap pretty fast. Also of note is that this is a private function only used in Op's `std::string SerializeOpVec(const OpRcPtrVec & ops, int indent=0);`. I have reservations about that function's `indent` parameter as `int`. As `SerializeOpVec` is internal and infrequently used, I have switched this to `size_t` to remove negative integer possibility. If a contributor decided to use Repeat() or SerializeOpVec() with a negative number though, then they're going to have a bad time!

cozdas:
Looks like commits 
57a94a80 (DX12)
907a643a (Linux container)
f705d2e6 (CVE)
are replayed (cherry picked?) into the PR branch with different commit ID's and that confuses the PR diff.

I think rebasing your branch to main and force-pushing your branch should clear the duplicate changes and remove the extra changes listed here.

meimchu:
> Looks like commits [57a94a8](https://github.com/AcademySoftwareFoundation/OpenColorIO/commit/57a94a8003c00be506a41f54a080d9fefaba6d22) (DX12) [907a643](https://github.com/AcademySoftwareFoundation/OpenColorIO/commit/907a643afed5df80121788815af972ddc4879f64) (Linux container) [f705d2e](https://github.com/AcademySoftwareFoundation/OpenColorIO/commit/f705d2e6f057884eca3c98858137c57bca4ea5aa) (CVE) are replayed (cherry picked?) into the PR branch with different commit ID's and that confuses the PR diff.
> 
> I think rebasing your branch to main and force-pushing your branch should clear the duplicate changes and remove the extra changes listed here.

Thanks for your feedback! I have done the rebase as you've suggested. I did pull in a bunch of changes from other commits the last time I rebased to get an updated CICD workflow (it was busted for a while). I wonder if there has been recent changes again to CICD that's causing different issues again?
- Docs building seems to be failing in the last few actions/jobs the last couple of days.
- I'm not entirely sure why, but C++20 builds are mostly failing now for linux platform. I see errors in there about linker with `opencolorio_v2_3` namespace... while linking `ociolutimage`? I am not sure why this is happening and don't believe I've done anything that would impact `ociolutimage` like this. Any help regarding this would be much appreciated!

meimchu:
https://github.com/AcademySoftwareFoundation/OpenColorIO/actions/runs/28687774279/job/85083494812?pr=2313#step:2:258

In the initializing containers step, I see `ASWF_OPENCOLORIO_VERSION=2.3.2`. I wonder if that is a hint of what's causing this issue? I also see it later doing a git sparse-checkout, so maybe there is some clobbering going on there?

cozdas:
@meimchu 
There were several CI changes and fixes last week, including an OIIO related fix. Please update the main branch in your fork and rebase your branch to it. This may resolve the CI errors.

meimchu:
> @meimchu There were several CI changes and fixes last week, including an OIIO related fix. Please update the main branch in your fork and rebase your branch to it. This may resolve the CI errors.

Looks working now. Thank you!

src/OpenColorIO/Op.cpp
```diff
@@ -470,15 +469,15 @@ std::ostream& operator<< (std::ostream & os, const Op & op)
...
+std::string SerializeOpVec(const OpRcPtrVec & ops, size_t indent)
 {
     std::ostringstream oss;
 
     for (OpRcPtrVec::size_type idx = 0, size = ops.size(); idx < size; ++idx)
     {
         const OpRcPtr & op = ops[idx];
 
-        oss << pystring::mul(" ", indent);
+        oss << StringUtils::Repeat(" ", indent);
```
KevinJW, line 480: in parallel I've proposed something more direct using std::string only see [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/2244/changes#diff-737cb30b311ff4c8bd9b5218af68d149ee7cd73432339ef8ac05705fc61ad26bR324)

but either way could probably extract the function call outside of the loop and assign to a const string

vendor/openfx/OCIOUtils.h
```diff
@@ -9,6 +9,8 @@
 
 #include "ofxsImageEffect.h"
 
+#include "utils/StringUtils.h"
```
KevinJW, line 12: If there are no direct uses in the header, it is probably better to include the StringUtils header in the .cpp files directly.

src/OpenColorIO/Config.cpp
```diff
@@ -2571,7 +2571,8 @@ const char * Config::getInactiveColorSpaces() const
 bool Config::isInactiveColorSpace(const char * colorspace) const noexcept
 {
     StringUtils::StringVec svec;
-    pystring::split(getImpl()->m_inactiveColorSpaceNamesConf.c_str(), svec, ", ");
+    svec = StringUtils::Split(getImpl()->m_inactiveColorSpaceNamesConf.c_str(), ',');
```
KevinJW, line 2574: Disclaimer - I have not tested this myself

I'm wondering if we have enough existing test coverage to make sure this change to the split sequence results in the same behaviour. Was it ever possible for m_inactiveColorSpaceNamesConf
to have a comma without a space to the "right" of it, which this new version might split on but the previous one did not?

i.e. `"some thing,like this, results-in"` how many entries in the string vector? I think previously we'd end up with 2 `["some thing,like this", "results-in"]` and now we might have 3 `["some thing", "like this", "results-in"]`

src/utils/StringUtils.h
```diff
@@ -110,6 +110,14 @@ inline bool StartsWith(const std::string& str, char prefix)
 }
 
 // Starting from the left, trim the character.
+inline std::string LeftTrim(std::string str, const std::string & prefix)
+{
+    size_t first_good_char = str.find_first_not_of(prefix);
+    if (first_good_char == std::string::npos) { return str; }
+    str.erase(0, first_good_char);
+    return str;
+}
```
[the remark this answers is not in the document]
cozdas: @meimchu I think this is a valid point, it should return empty.

src/utils/StringUtils.h
```diff
@@ -126,6 +134,14 @@ inline std::string LeftTrim(std::string str)
 }
 
 // Starting from the right, trim the character.
+inline std::string RightTrim(std::string str, const std::string & suffix)
+{
+    size_t last_good_char = str.find_last_not_of(suffix);
+    if (last_good_char == std::string::npos) { return str; }
+    str.erase(last_good_char + 1);
+    return str;
+}
```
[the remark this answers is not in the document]
cozdas: @meimchu this too.

cozdas (changes requested):
LeftTrim and RightTrim should return empty when the input contains only trim characters.

PR #2318, opened 2026-05-24
Adsk Contrib - Avoid possible cycles in transforms

doug-walker:
This PR prevents cycles in a config from causing a crash by creating an infinite recursion while building a ColorSpaceTransform, DisplayViewTransform, or LookTransform.


remia (approved):
Thanks @doug-walker, looks good to me. If this pattern continue to repeat maybe we can at some point extract the DepthGuard into a utility header and use it consistently (`ResolveContextVariablesImpl` don't yet) but probably not important with this few cases.

cozdas (approved):
Looks good and clean.

PR #2319, opened 2026-05-29
Fix GPU shader correctness and warnings surfaced by DXC warning flags

num3ric:
* GpuShaderUtils: add intCast() helper that emits a C-style cast on HLSL (silences -Wfloat-conversion) and a functional cast on GLSL/MSL/OSL.
  * FixedFunctionOpGPU: use intCast() for the three implicit float->int conversions in the ACES2 hue-weight setup and cusp-table search.
  * FixedFunctionOpGPU: drop the unused _Add_SinCos_Shader calls in Gamut_Compress_Fwd/Inv_Shader (gamut-compress stays in JMh space, so cos_hr/sin_hr were dead).
  * GradingBSplineCurve: in the HUE-FX inverse path (curveIdx == 7), offset knStartY by knStart to match the CPU evalCurveRevHue.
  * GradingRGBCurveOpGPU: inverse eval no longer declares identity_x (it's never used in the inverse), and the in/out qualifier branch is folded into a single emitter — matches the CPU evalCurveRev signature.
  * DxApp: when verbose (test_gpu_exec --dx -v), pass -Wall -Wextra -Wconversion to DXC and print DXC_OUT_ERRORS even on success, so generated-shader warnings are visible during debugging.

### DIFF
--- a/src/OpenColorIO/GpuShaderUtils.cpp
+++ b/src/OpenColorIO/GpuShaderUtils.cpp
@@ -404,6 +404,17 @@ std::string GpuShaderText::intKeyword() const
     return "int";
 }
 
+std::string GpuShaderText::intCast(const std::string & expr) const
+{
+    // HLSL/DXC flags the functional cast int(x) under -Wfloat-conversion; a C-style cast suppresses it (GLSL forbids C-style casts).
+    if (m_lang == GPU_LANGUAGE_HLSL_SM_5_0)
+    {
+        return "(int)(" + expr + ")";
+    }
+
+    return intKeyword() + "(" + expr + ")";
+}
+
 std::string GpuShaderText::intKeywordConst() const
 {
     std::string str;
--- a/src/OpenColorIO/GpuShaderUtils.h
+++ b/src/OpenColorIO/GpuShaderUtils.h
@@ -76,6 +76,8 @@ public:
     std::string intKeyword() const;
     std::string intKeywordConst() const;
     std::string intDecl(const std::string& name) const;
+    // Cast a float expression to int using language-appropriate syntax.
+    std::string intCast(const std::string& expr) const;
 
     std::string colorDecl(const std::string& name) const;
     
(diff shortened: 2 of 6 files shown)


num3ric:
See attached file for warning logs (without fixes included here): [dxc_warnings_full.log](https://github.com/user-attachments/files/28405652/dxc_warnings_full.log)

fallenmi (approved):
Reviewed exact head `1d41e95d8738fdccc3d4bac4104c665c5bcbc519` against exact PR base `57a94a8003c00be506a41f54a080d9fefaba6d22` and current `main` `5a808fb57a94c7229640a97835c420c9a1fbd1fe`. Although GitHub marks the branch behind, it merges cleanly; I tested the resulting current-main tree `a28f55e4e14a070f2a82f779bb73c0de038a9be9`.

The shader changes are internally consistent. The HLSL-only C-style integer cast preserves existing truncation while avoiding DXC's functional-cast warning; other shader languages retain their required functional syntax. The HUE-FX inverse lower bound now matches the CPU path's knot offset, the inverse RGB-curve helper safely drops an argument that was redundant because its empty-curve path returns `x`, and the removed sine/cosine setup was dead in the standalone JMh gamut-compress paths.

Exact base, head, and current merge all build successfully. The same 50 shader/curve/fixed-function focused tests pass on all three. The full CPU runs are also baseline-equivalent: base and head each have 1144 passes / 4 failures, and current merge has 1149 passes / the same 4 failures (two unavailable-monitor failures and two pre-existing AppleClang floating-point tolerance failures). Live exact-head CI is fully green: 21 check-runs plus EasyCLA and ReadTheDocs, with no pending, failed, or `action_required` workflow. I found no blocking issue.

AI disclosure: I used OpenAI Codex to inspect the exact revisions and repository policy, build the exact base/head/current-main merge oracle, run and compare the focused and full CPU suites, verify live checks and review history, and draft this review. I verified the evidence and conclusion.

doug-walker (approved):
@num3ric, this all looks very good, thanks for the fixes!  

I ran your branch locally and verified that the GLSL, HLSL, and Vulkan GPU tests all pass on Windows.

Apologies for the delayed review. We will get this into 2.6.0.

PR #2322, opened 2026-06-15
Remove ocioview

michdolan:
This PR removes `ocioview` from the main OpenColorIO project repository, since it has a new home at https://github.com/AcademySoftwareFoundation/ocioview . Notes have been added to existing app references in the docs pointing to this new location.

I wonder if we should also remove `pyociodisplay`? It's useful on its own as an example of a GPU viewport written in Python, but its code is largely duplicated in `ocioview`.


doug-walker (approved):
LGTM!

Agreed that pyociodisplay may make more sense in the other repo too.

PR #2323, opened 2026-06-28
Fix ocioconvert invalid bidepth handling

### LINKED ISSUE #2110: ocioconvert silently errors out when given invalid bitdepth running on Windows
When calling ocioconvert from a terminal on windows and an invalid bitdepth is given, it does not output the intended error message:

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/src/apps/ocioconvert/main.cpp#L142

This is because the throw is not within a try/catch block and exceptions produce no output unless you do something with them.

Suggest turning this into a simple std:cerr message and exit() with an appropriate error code.

zlrs1013:
Fixes #2110

This updates ocioconvert so an invalid --bitdepth value reports a normal CLI error and usage text instead of throwing an uncaught exception.

Verified with:

cmake --build build --target ocioconvert --config Release
.\build\src\apps\ocioconvert\Release\ocioconvert.exe --bitdepth 32 --view input.exr "ACES2065-1" output.exr "sRGB - Display" "ACES 2.0 - SDR 100 nits (Rec.709)"


zlrs1013:
I checked the failing Read the Docs build. It appears unrelated to this PR. The build failure appears to be related to a .readthedocs.yml config validation error for build.os, while this PR only changes src/apps/ocioconvert/main.cpp.

src/apps/ocioconvert/main.cpp
doug-walker, line 541: @zlrs1013 , please remove this throw too. The "else if" at line 535 could be replaced by an "else".

doug-walker (changes requested):
Thanks for the contribution Zhen! 

Yes, the ReadTheDocs issue is unrelated to your work.

src/apps/ocioconvert/main.cpp
doug-walker: Sorry, my mistake, there is actually a catch statement for this one.

doug-walker (approved):
Thank you Zhen!

zachlewis (approved):

PR #2232, opened 2026-01-06
pyocioamf: Add AMF v2.0 and OCIO 2.5+ support

giardiello:
## Summary

This PR enhances pyocioamf with support for AMF v2.0 and OCIO 2.5+ configurations:

- **AMF v2.0 support** with automatic version detection from namespace URI or version attribute
- **OCIO 2.5+ config support** using the `amf_transform_ids` interchange attribute for transform lookup
- **New CLI options** for flexible conversion control
- **Split CTF generation** for AMF v2.0 `workingLocation` marker
- **New example file** demonstrating v2.0 format

### AMF Version Support
- Auto-detects v1.0 vs v2.0 from namespace URI (`urn:ampas:aces:amf:v1.0` / `v2.0`) or version attribute
- Handles both v1.0 element names (`SOPNode`/`SatNode`) and v2.0 names (`ASC_SOP`/`ASC_SAT`)

### OCIO Config Compatibility
- For OCIO 2.5+ configs: Uses dedicated `amf_transform_ids` interchange attribute
- For OCIO 2.1-2.4 configs: Falls back to description-based transform ID search
- Dynamic ACES2065-1 colorspace detection via role lookup, common names, and aliases

### CLI Enhancements
- `--config`: Specify a custom OCIO config file (supports OCIO 2.1+)
- `--no-idt`: Exclude input transform from conversion
- `--no-lmt`: Exclude look transforms from conversion  
- `--no-odt`: Exclude output transform from conversion
- `--split-by-working-location`: Generate split CTFs at workingLocation marker

### New Files
- `example_v2.amf`: AMF v2.0 example with `ASC_SOP`/`ASC_SAT` elements

## Test Plan

- [x] Convert `example.amf` (v1.0) - backwards compatible
- [x] Convert `example_v2.amf` (v2.0) - new format support
- [x] Test with OCIO 2.4 config (description-based lookup)
- [x] Test with OCIO 2.5 config (amf_transform_ids lookup)
- [x] Test `--no-idt`, `--no-lmt`, `--no-odt` exclusion options
- [x] Test `--split-by-working-location` with workingLocation marker

## Backwards Compatibility

All existing functionality is preserved. The script will:
- Auto-detect AMF version and use appropriate element parsers
- Auto-detect OCIO config version and use appropriate lookup method
- Default to existing behavior when no CLI options are specified

PR #2325, opened 2026-07-05
pyocioamf: Add AMF v2.0 and OCIO 2.5+ support

giardiello:
## Summary

This PR enhances `pyocioamf` with support for AMF v2.0 and OCIO 2.5+ configurations. It supersedes #2232 (closed to move the branch onto a fresh public fork and add the refinements below).

- **AMF v2.0 support** with automatic version detection from namespace URI or version attribute
- **OCIO 2.5+ config support** using the `amf_transform_ids` interchange attribute for transform lookup
- **New CLI options** for flexible conversion control
- **Split CTF generation** for the AMF v2.0 `workingLocation` marker
- **New example file** demonstrating the v2.0 format

### AMF Version Support
- Auto-detects v1.0 vs v2.0 from namespace URI (`urn:ampas:aces:amf:v1.0` / `v2.0`) or version attribute
- Handles both v1.0 element names (`SOPNode`/`SatNode`) and v2.0 names (`ASC_SOP`/`ASC_SAT`)

### OCIO Config Compatibility
- For OCIO 2.5+ configs: uses the dedicated `amf_transform_ids` interchange attribute
- For OCIO 2.1-2.4 configs: falls back to description-based transform ID search
- Multi-strategy ACES2065-1 colorspace resolution via `aces_interchange` role, then common names, then alias matching (works across config naming variants)

### CLI Enhancements
- `--config`: specify a custom OCIO config file (supports OCIO 2.1+)
- `--no-idt` / `--no-lmt` / `--no-odt`: exclude input / look / output transforms
- `--split-by-working-location`: generate split CTFs at the `workingLocation` marker
- `--ignore-applied-tag`: process transforms even when `applied="true"`
- `--output-prefix` / `--output-dir`: collision-free output naming for batch/concurrent use

### Robustness
- `check_lut_path` now always returns an absolute path so OCIO's CTF `write()` resolves referenced LUTs correctly regardless of the current working directory

### New Files
- `example_v2.amf`: AMF v2.0 example with `ASC_SOP`/`ASC_SAT` elements

## Backwards Compatibility

All existing functionality is preserved: the script auto-detects the AMF version and OCIO config version, and defaults to prior behavior when no CLI options are given.

### DIFF
--- a/src/apps/pyocioamf/README.md
+++ b/src/apps/pyocioamf/README.md
@@ -4,20 +4,48 @@
 pyocioamf
 =========
 
-**Work in progress**. Script to convert an ACES Metadata File (AMF) into OCIO's 
-native CTF file format.  The CTF may then be used with any of the other OCIO tools 
-to process pixels.
+Script to convert an ACES Metadata File (AMF) into OCIO's native CTF file format.
+The CTF may then be used with any of the other OCIO tools to process pixels.
 
-An example AMF file named ``example.amf`` is provided.  It references the LUT file
-``example_referenced_lut.clf``, which is in the Academy/ASC Common LUT Format (CLF).
+**Supports both AMF v1.0 and v2.0** with automatic version detection.
+
+Example files are provided:
+- ``example.amf`` - AMF v1.0 format
+- ``example_v2.amf`` - AMF v2.0 format (uses ``ASC_SOP``/``ASC_SAT`` elements)
+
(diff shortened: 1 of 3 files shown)


giardiello:
@doug-walker Sorry if this was messy, I've updated the script and re-issued a PR.

doug-walker:
No worries, thanks for the extra context @giardiello !  I've added this PR to the OCIO 2.6 milestone.

Issue #2275, opened 2026-03-19
Finalize ACES 2 FixedFunctions and remove warnings


doug-walker:
When ACES 2 was implemented, there was some uncertainty in how to decompose the overall Output Transforms into FixedFunctions, and so the YAML reader and writer issue warnings when any of these are used:

FIXED_FUNCTION_ACES_OUTPUT_TRANSFORM_20
FIXED_FUNCTION_ACES_RGB_TO_JMH_20
FIXED_FUNCTION_ACES_TONESCALE_COMPRESS_20
FIXED_FUNCTION_ACES_GAMUT_COMPRESS_20

At this point, we should try to finalize what the fixed functions are and remove the warnings.

I like having a function to convert to JMh, but there are two things I find inconvenient about the current one:
1. The results are in the hundreds (e.g., J goes up above 800, h goes up to 360) and this makes it inconvenient to use in other tools that would prefer more of a 0 to roughly 1 normalization.
2. The more conventional channel layout is for hue to be first and lightness to be last. For example, the existing conversions to HSV and HSY use that channel layout. 

Currently, in order to use the existing RGB_TO_JMH with the GradingHueCurveTransform, I need to put a matrix on either side to rescale and reorder the channels. My proposal would be to create a new function that is ordered hMJ and has h divided by 360 and J divided by roughly 500.

Similarly, my only concern about the others is whether the inputs/outputs are normalized in a way that is convenient.

As we are currently developing LMTs that may work in JMh, it would be helpful if the optimizer was able to decompose an ACES_OUTPUT_TRANFORM into one or more fixed functions that expects JMh (from an LMT) as input so that a pair of RGB_TO_JMH and JMH_to_RGB could be dropped. Currently, is it possible to replace an ACES_OUTPUT_TRANSFORM with RGB_to_JMH + TONESCALE_COMPRESS + GAMUT_COMPRESS + JMH_to_RGB?

Tagging @KevinJW and @remia for input.

Note: When implementing this, search Config_tests.cpp for "experimental" and clean up accordingly. Likewise remove "experimental" from OpenColorTypes.h fixed function descriptions.

remia:
[...]

PR #2336, opened 2026-08-31
Adsk Contrib - Add RGB-to-HMJ FixedFunction

doug-walker:
This PR adds a new FixedFunction option FIXED_FUNCTION_ACES_RGB_TO_HMJ_20. In comparison to the experimental FIXED_FUNCTION_ACES_RGB_TO_JMH_20, this one reorders the channels to be more consistent with the existing HSV and HSY FixedFunctions and scales them as {h/360, M/200, J/100} so that images in this space are easier to work with in compositing applications.

Note that unlike the experimental ACES 2 FixedFunctions, this one is intended to be final and therefore, there is no warning logged in OCIOYaml when reading a config containing this transform.

In addition, it fixes a bug in the CPU renderer for the GradingHueCurveTransform which affected how the SAT_LUM curve was applied when in video mode.

The config version is bumped to 2.6, in preparation for the upcoming release. Version checks are added for the new FixedFunction and BuiltinTransform.

Some prototype documentation is added in preparation for the 2.6 release.

I removed the look transform from the first version of this PR. The decision was made to collect additional user feedback before moving forward with that.

In addition, this adds the missing Config::Impl::checkVersionConsistency check for the new Apple Log 2 builtin.

Implements issue #2275.

Assisted by: Claude Code / Sonnet 5.0


remia (approved):
Pending the parameters adjustments as mentioned but looks good to me.

src/OpenColorIO/Config.cpp
```diff
@@ -5674,6 +5674,27 @@ void Config::Impl::checkVersionConsistency(ConstTransformRcPtr & transform) cons
...
+                throw Exception(os.str().c_str());
+            }
+            if (m_majorVersion == 2 && m_minorVersion < 6
+                    && 0 == Platform::Strcasecmp(blt->getStyle(), "ACES-LMT - ACES 2.0 DCC Look 1"))
+            {
+                std::ostringstream os;
+                os << "Only config version 2.6 (or higher) can have BuiltinTransform style '"
+                   << blt->getStyle() << "'.";
+                throw Exception(os.str().c_str());
+            }
```
cozdas, line 5697: If I'm not missing it, I think this throw is not exercised in the tests (new fixed function throws are tested though). Do we want to add that too?

cozdas (approved):
Looks good to me. 
I added a comment for a missing version check for the new LMT.

doug-walker:
I removed the Look Transform part of this PR and the rest was approved, so I'll go ahead and merge this one now. It has the 2.6 version update which is needed for other PRs.

PR #2339, opened 2026-09-07
tests/cpu/Config_tests.cpp: Fix compat with yaml-cpp 0.9.0

OPNA2608:
Array contents that are continued on subsequent lines must be indented, otherwise:

```
FAILED: exception thrown from config = OCIO::Config::CreateFromStream(is): "Error: Loading the OCIO profile failed. yaml-cpp: error at line 25, column 1: plain value cannot start with flow indicator character"
```

---

https://github.com/jbeder/yaml-cpp/commit/1f2b841949de53366508d44e59a6d5bbac562fa3


OPNA2608:
I find it unreasonable to provide you my full mailing address in the CLA for just a minor fix to one of your YAML expressions. Closing & opening issue instead…

Issue #2337, opened 2026-08-31
DisplayViewTransform should fall back to display color space aliases


doug-walker:
Currently, OCIO displays do not have aliases. This makes it difficult to change the name of a display in a config without breaking backwards compatibility (and using the inactive list is not ideal). Regular color spaces, including display color spaces already have this feature via the aliases attribute of the color space.

For displays that use views that define their display_colorspace as `<USE_DISPLAY_NAME>` (which is most views in modern config files), I propose that if a DisplayViewTransform is unable to find the requested display that it fall back to looking at the aliases for all display color spaces.

One possible unintended consequence is that it would potentially allow some DisplayViewTransforms that legitimately should have failed to now succeed. I'm unable to think of a likely example of where that would be a problem, but we could potentially add a config-level attribute such as `allow_display_aliases: true` to require config authors to opt-in to the behavior.

Similar to the display scenario, I propose that if a DisplayViewTransform is unable to find the view, that it search in a new aliases attribute of the ViewTransforms.

Issue #2210, opened 2025-10-22
Add Display description to OCIO configs


remia:
With OCIO v2 we now have a description field in the View specification, maybe it would be interesting to have a corresponding field for the display itself. An example use case would be a DCC with separated list to pick Display and View, each having a tooltip. Sometime the display name is not enough to really know what is the actual calibration target (example being ACES configs with the `P3-D65 - Display`, we also sometime use display names in our internal configs that are meant to be easy to use for non colour savvy artists but more advanced users usually want more information).

With the current OCIO YAML structure it looks like this would not be straightforward to add.

doug-walker:
@remia, rather than adding a description to displays, what do you think about taking the description from the corresponding display color space? I just added a related issue, please see #2337.

remia:
That sounds reasonable Doug, assuming we want to add an API call for this to hide the logic. If I follow the reasoning from the linked issue, it would have to find at least one view using `<USE_DISPLAY_NAME>` and then query the description from the associated `display_colorspace`.

PR #2341, opened 2026-09-10
Adsk Contrib - Implement display and view aliases

doug-walker:
This PR allows config authors to define aliases for displays and views, similar to aliases on color spaces. This allows display and view names to evolve over time while allowing the earlier names to continue to work in DisplayViewTransforms. In addition, it will now be possible to define aliases that simplify working with command-line tools such as oiiotool to apply DisplayViewTransforms. 

This new flexibility in naming should be very useful with the upcoming built-in ACES configs.

For displays, this works by treating the aliases to the display's corresponding display color space as display aliases.

For views, this works by treating the name and aliases of the view's view_transform as aliases to the view. ViewTransforms did not already have aliases, so this PR adds those as well.

One design decision was to add the aliases to the ViewTransform class rather than the views themselves. There were a number of reasons for this choice, including:
- There is no class for views, so it would require adding at least six more functions to the already long Config class.
- This is consistent with how we handled the addition of interchange attributes, and is likely how we would handle any additional view-related functionality.

This functionality is off by default and requires config authors to opt-in by adding the new use_display_view_aliases attribute to their config file. This avoids any unintended consequences related to existing configs that were not designed with this in mind. The new functionality requires the config file version to be at least 2.6 or higher.

While working on this, I wound up noticing and fixing a few bugs involving either aliases or the USE_DISPLAY_NAME token. 

If application developers are directly using Config::getDisplayViewColorSpaceName, they will probably need to switch to the new Config::getResolvedDisplayViewColorSpaceName, in order to properly handle configs that use display/view aliases. I will mention that in the release notes.

The majority of the code is unit tests, as I was trying to anticipate all scenarios, though a lot of them are edge cases.

Addresses issues #2337 and #2210.

Assisted by: Claude Code / Sonnet 5.


include/OpenColorIO/OpenColorIO.h
```diff
@@ -656,10 +656,9 @@ class OCIOEXPORT Config
     void removeColorSpace(const char * name);
 
     /**
-     * Return true if the color space is used by a transform, a role, or a look.
-     *
-     * \note
-     *    Name must be the canonical name.
+     * Return true if the color space is used by a transform, a role, a look, a (display, view)
+     * pair, or a file rule. The argument may be either an alias or the canonical name. While
+     * searching the config, aliases are always resolve to their canonical names for comparison.
```
cozdas:
```suggestion
     * searching the config, aliases are always resolved to their canonical names for comparison.
```

include/OpenColorIO/OpenColorIO.h
```diff
@@ -2555,14 +2635,35 @@ class OCIOEXPORT ViewTransform
     ViewTransformRcPtr createEditableCopy() const;
 
     const char * getName() const noexcept;
+    /// \see ColorSpace::setName
     void setName(const char * name) noexcept;
 
+    /**
+     * \note
+     * ViewTransform aliases are available in config file versions 2.6 or higher. They
+     * are availaable regardless of how \ref Config::getUseDisplayViewAliases is set.
```
cozdas:
```suggestion
     * are available regardless of how \ref Config::getUseDisplayViewAliases is set.
```

docs/guides/authoring/displays_views.rst
```diff
@@ -132,6 +132,25 @@ A View Transform may use the following keys:
 
 .. TODO: Good spot for an example in a future revision.
 
+
+``use_display_view_aliases``
```
remia: Terminology: should we drop the use_ prefix? Comparing to `strictparsing: true` for example.

src/OpenColorIO/Config.cpp
```diff
@@ -3823,6 +3853,275 @@ void Config::clearDisplays()
...
+        return iter->first.c_str();
+    }
+
+    const size_t numAliases = cs->getNumAliases();
+    for (size_t i = 0; i < numAliases; ++i)
+    {
+        iter = FindDisplay(getImpl()->m_displays, cs->getAlias(i));
+        if (iter != getImpl()->m_displays.end())
+        {
+            // A display exists with the name of an alias of the color space. In this case,
```
remia, line 3923: I think this edge case is a bit convoluted, could it help linking to the corresponding (if any) unit test in the comment to help understanding with an example?

src/OpenColorIO/Config.cpp
```diff
@@ -3823,6 +3853,275 @@ void Config::clearDisplays()
...
+
+        if (!resolvedDisplayCs)
+        {
+            // There is no display color space in the config corresponding to this display,
+            // so regardless of what the candidate's display_colorspace is, there is 
+            // nothing to check it against. Accept the match.
+            return candidate->m_name.c_str();
+        }
+
+        // There is a display color space for this display, so only accept this candidate
```
remia, line 4110: Same as the above, could help linking to relevant unit test example? Including the case where a view using  a matched view/named transform is not matching due to it using a color space name not matching the display name, what's the reasoning behind this (obviously it seem like something very rare to encounter)?

I haven't look thoroughly in the new unit tests, but what happen if multiple views only differ by the looks they use? The resolution is going to pick the first matching one?

docs/guides/authoring/displays_views.rst
```diff
@@ -132,6 +132,25 @@ A View Transform may use the following keys:
...
+Optional.  Activates aliases for display and view names.
+
+By default, the arguments to DisplayViewTransform must be the exact strings found
+in the display / view section of the config. However, if ``ocio_profile_version`` 
+is 2.6 or higher, ``use_display_view_aliases`` may be set to true. This allows a
+display to be referred to by the name or aliases of its corresponding display
+ColorSpace and the view to be referred to by the name or aliases of its
+corresponding ViewTransform. This config-level attribute defaults to false and 
+must be omitted from the config file if its value is not "true".
+
```
remia, line 148: It's already indirectly said, but should we make it explicit that this aliasing mechanism only applies to display/view using the v2 view transform / display color space workflow and will not work for older style configs (from my understanding at least)?

remia (commented):
I think display / view aliases will be a useful feature and I already see case where it could help us transition display names more smoothly (though in our case it's almost as if you would want a simple / advanced name for each display depending on the target audience), provided DCCs are updated to use the new API to handle this gracefully. Agree with leaving it off by default for now, as I find it a bit hard to wrap my head around all the potential edge cases / consequences involved.

Some of the edge cases handling are a bit convoluted, which may be just me, but could lead to unexpected resolving so may need extra clarity?

Overall looks good, thank you for addressing the side issues like resolving <USE_DISPLAY_NAME> and adding the display description API.

Issue #2116, opened 2025-02-06
Metal implementation is very slow with ACES 2.0 view transform


cozdas:
We hit a performance issue with ACES 2.0 output transforms using a metal shader. Metal implementation is very slow especially on Apple silicon and the performance hit increases with input size. After the preliminary analysis, the main culprit is the 362-element long const float array that the transform uses. Since the OCIO metal shader implementation uses a "wrapper" struct to encapsulate the functions and data, this array ends up being a member of the encapsulating struct and looks like the array is created in the run-time many times, trashing the L1 cache (90% miss rate). See **ocio_gamut_cusp_table_0_hues_array** defined in line 88 of the attached shader code which is generated with the v2.4.1 tagged branch.

OpenGL and HLSL generators don't use a wrapper and thus don't suffer from the same issue even on the same hardware. Also pulling the offending array outside of the struct and marking it "constant float" fixes the performance problem. (see the attached shader which has the array outside of the struct).

[shader generated with v2.4.1.txt](https://github.com/user-attachments/files/18680675/v2.4.1.txt)
[shader where array pulled outside of the struct](https://github.com/user-attachments/files/18680703/constant_float_array.txt)

### Repro steps

- checkout OCIO **v2.4.1 branch** and compile
- download the test config that's provided here: https://github.com/AcademySoftwareFoundation/OpenColorIO/wiki/ACES-2.0-optimization
- set **OCIO** environment variable to point to that config.
- run ociodisplay utility with "-metal" and "-gpuinfo" flags and pass a relatively large image (4k) as the input image 
   **ociodisplay -metal -gpuinfo large_image.exr**
- with the right click in the UI set 
   - the **image color space** to "**ACES / ACES2065-1**"
   - the **display** to "**sRGB - Display**"
   - set the **view** to "**ACES 2.0 - SDR 100 nits (Rec.709)**"

You'll see the generated code in the console and if you do a metal capture you'll see that in the draw call performance report, buffer L1 cache miss rate is very high.

PR #2342, opened 2026-09-10
Bugfix: RGB/Hue Curve Performance with MSL

carolalynn:
While implementing RGB and Hue Curves, noticed a severe performance drop in playback on edited curves vs. other OCIO ops. This fix is specific to Metal on MacOS, so I have already asked @Morteeza to take a look on my fork - fix confirmed. 

The MSL class wrapper gave the generated struct owning copies of every array uniform, with a constructor that copied the used range out of constant memory into per-thread private memory and then zero-filled the remainder up to the full declared capacity. The struct is instantiated once per shader invocation, so a compute shader dispatched one thread per pixel paid that copy per pixel: for a dynamic GradingRGBCurve that is 8 + 120 + 8 + 360 = 496 private-memory writes for every pixel, regardless of how many knots the curve actually has.

Point at the constant memory instead. The array declarations arrive from the ops shared with the other GPU languages, so the wrapper rewrites them to constant pointers rather than changing the ops; it already knows which parameters are arrays. Indexing syntax is unchanged, so every read site in the generated shader is untouched, and the generated function signature is unchanged so existing MSL consumers keep working.

Dropping the zero-fill does not change any result. Reads of knots/coefs are bounded by the offset taken from the offsets array, and the offsets arrays are always fully populated -- GetNumOffsetValues() is a compile-time constant equal to the declared capacity (8 for RGB, 16 for hue) -- so no read past the used range was ever possible.

Measured with a Metal compute kernel over an RGBA16F 4K UHD frame, one thread per pixel, output texels bit-identical before and after:

  rgb curve only        33.45 ms -> at the read+write bandwidth floor
  luma curve only       35.65 ms -> at the floor
  hue curve only        35.46 ms -> at the floor
  rgb + luma + hue     115.28 ms -> at the floor

GradingRGBCurve and GradingHueCurve are the only ops that declare array uniforms; everything else passes scalars or textures and is unaffected.

MetalSupport9 pins the generated text, so its expected shader is updated. Three assertions are added alongside it stating the no-copy invariant on its own, so a future refresh of the expected text cannot quietly restore the per-pixel copy.

Assisted by: Claude / Opus 5

### DIFF
--- a/src/OpenColorIO/GpuShaderClassWrapper.cpp
+++ b/src/OpenColorIO/GpuShaderClassWrapper.cpp
@@ -181,36 +181,10 @@ std::string MetalShaderClassWrapper::generateClassWrapperHeader(GpuShaderText& k
     kw.indent();
     for(const auto& param : m_functionParameters)
     {
-        size_t openAngledBracketPos = param.m_name.find('[');
-        if(!param.m_isArray)
-        {
-            kw.newLine()    << "this->" << param.m_name  << " = " << param.m_name  << ";";
-        }
-        else
-        {
-            size_t closeAngledBracketPos = param.m_name.find(']');
-            std::string variableName = param.m_name.substr(0, openAngledBracketPos);
-            
-            kw.newLine()    << "for(int i = 0; i < "
-                            << GetArrayLengthVariableName(variableName)
-                            << "; ++i)";
-            kw.newLine()    << "{";
-            kw.indent();
-            kw.newLine()    << "this->" << variableName << "[i] = " << variableName << "[i];";
-            kw.dedent();
-            kw.newLine()    << "}";
-            
-            kw.newLine()    << "for(int i = "
-                            << GetArrayLengthVariableName(variableName)
-                            << "; i < "
-                            << param.m_name.substr(openAngledBracketPos+1, closeAngledBracketPos-openAngledBracketPos-1)
-                            << "; ++i)";
-            kw.newLine()    << "{";
-            kw.indent();
-            kw.newLine()    << "this->" << variableName << "[i] = 0;";
-            kw.dedent();
-            kw.newLine()    << "}";
-        }
+        // Array members are pointers into constant memory (see rewriteArrayDeclarations) so they
+        // are bound rather than copied, using the parameter name without its array size.
+        std::string variableName = param.m_name.substr(0, param.m_name.find('['));
+        kw.newLine() << "this->" << variableName << " = " << variableName << ";";
     }
     kw.dedent();
     kw.newLine() <<"}";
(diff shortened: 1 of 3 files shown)


src/OpenColorIO/GpuShaderClassWrapper.cpp
```diff
@@ -378,6 +352,36 @@ void MetalShaderClassWrapper::prepareClassWrapper(const std::string& resourcePre
     extractFunctionParameters(originalHeader);
 }
 
+std::string MetalShaderClassWrapper::rewriteArrayDeclarations(const std::string& declarations) const
+{
+    // The uniform declarations are shared with the other GPU languages so arrays arrive here as
+    // fixed-size members i.e. 'float name[120];'. Owning them would make the constructor copy the
+    // whole array out of constant memory into per-thread memory, which is prohibitive as the
+    // struct is typically instantiated once per pixel. Hold a pointer to the constant memory
+    // instead; every read site is unchanged as the indexing syntax is the same.
```
doug-walker: Is it true that "the struct is typically instantiated once per pixel"?  Isn't it more typical that a group of pixels is processed in a batch, to amortize cost like this?  

I'm all in favor of the change to improve performance for the single-pixel case, but I suspect it's not the "typical" situation.

tests/cpu/GpuShader_tests.cpp
```diff
@@ -1361,8 +1333,21 @@ float4 OCIOMain(
...
+    // uniforms rather than pointing at them therefore copies the whole declared
+    // capacity into per-thread memory for every pixel, which dominated the cost
+    // of any curve op. These checks state that invariant on its own so a future
+    // update of the expected text above cannot quietly restore the copy.
+    OCIO_CHECK_ASSERT(text.find("constant float* ocio_grading_rgbcurve_knots;")
+                      != std::string::npos);
+    OCIO_CHECK_ASSERT(text.find("this->ocio_grading_rgbcurve_knots = "
+                                "ocio_grading_rgbcurve_knots;") != std::string::npos);
+    OCIO_CHECK_ASSERT(text.find("this->ocio_grading_rgbcurve_knots[i]")
+                      == std::string::npos);
```
doug-walker: Maybe this is a nit pick but it seems like kind of an odd test. The first two asserts are redundant after the comparison of the shader program above. And for the third, if we don't want people to change the code back in the future, perhaps just a comment where the original code is should suffice? If someone did try to return to the original approach, it would have to be written exactly the same in order to fail this test (e.g., maybe someone uses j rather than i for their loop)?

doug-walker (commented):
Looks like a nice improvement!

src/OpenColorIO/GpuShaderClassWrapper.cpp
carolalynn: Fair — "typically" was sloppy. It's not really a claim about how you dispatch: the struct is a temporary constructed inside the generated OCIOMain(), which takes and returns one float4. No batch entry point, nowhere to hoist it to, so it's once per call.

The numbers are from the normal case rather than a synthetic one — the 30.01 → 0.45 ms row is a fragment shader built the way metalapp.mm generates one, and compute (29.30 → 0.44) lands in the same place, so nothing is amortizing it either way.

You're right that a kernel doing a tile per thread could hoist it. That's just not what the wrapper emits today, and even there the copy is still per-thread and a 496-float private array costs occupancy anyway.

Reworded the comment to give the mechanism instead.

tests/cpu/GpuShader_tests.cpp
carolalynn: Agreed, dropped all three.

They were aimed at someone regenerating expected wholesale from actual output, where the equality check passes and a restored copy slips through. But the negative one only catches a character-identical reintroduction, so it doesn't really do that job — and the first two are redundant, as you say.

The explanation lives in the comment on rewriteArrayDeclarations now.

src/OpenColorIO/GpuShaderClassWrapper.cpp
doug-walker: Thanks for the update. Yes, all of the fragment shaders are written to work on one pixel at a time, though on the GPU these essentially get executed inside massive for-loops on large numbers of pixels. The problem here was that the shader was doing initialization work that it should not have been doing. It now references existing uniforms that have been pre-allocated rather than copying them all into its local memory. So the uniforms now work more like the LUT samplers work. The comment essentially said that, but I didn't quite follow what it was saying when I first read it.

doug-walker (approved):

cozdas (approved):
This is a similar fix to what we have done for issue #2116 that we faced and fixed for ACES 2.0 view transform back in the day. 👍 

Approved.

doug-walker:
Merging, thank you Carol!

PR #2343, opened 2026-09-11
Add Apple Log 2 built-in transform

carolalynn:
Apple Log 2 uses the same transfer function as Apple Log but encodes in the Apple Wide Gamut primaries rather than Rec.2020.  Adds the APPLE_LOG-APPLEWG_to_ACES2065-1 built-in and its unit test values.

Assisted by: Claude / Opus 5


doug-walker (approved):
Looks good!  

I verified the chromaticities are correct.

We need to update Config::Impl::checkVersionConsistency, but we could merge this first and I'll fix that in a follow-up, since it requires bumping the config version.

carolalynn:
Yeah sounds good, thanks Doug! I'll follow with the PR to the configs repo, too.

remia (approved):

doug-walker:
@carolalynn, I'm unable to merge main into your branch (Allow edits from maintainers is off?). If you update you branch with main, I'll merge it.

carolalynn:
Weird - done!

PR #2346, opened 2026-09-16
Fix ocio.bat: create build/install directories if missing

### LINKED ISSUE #2031: Should ocio.bat create the required build and install location directories?
I am VERY new to compiling software, and I'm sure all the subtleties that go along with it.

I'm trying to build OpenColourIO on Windows 10. I found the instructions on what we needed to install for a full build, and followed them as best I could. I think the required applications have been installed correctly.

I used the instructions here.

https://github.com/AcademySoftwareFoundation/OpenColorIO/tree/main/share/dev/windows

Although I would note that I downloaded VS 2022 Community edition, since I do not need either of the paid versions. This caused a glitch where ocio.bat insisted VS wasn't installed, even though I had supplied the path to the relevant directory using the flag --msvs.

I had to go into the batch script and change the VS path in there. So it seems the --msvs flag isn't working.

Anyway, I got through these initial difficulties, and the batch file started to run. Then it bailed with an error:

C:/TEMP/OCIO/build/Release is not a directory

C:/TEMP is of course my temp path.

And its right, those directories don't exist. Surely the batch file should create them?

hikmetba-bit:
## Summary
Fixes #2031.

Without `--configure`, `cmake -B` never runs, so `BUILD_PATH` and `INSTALL_PATH` are never created on disk. The subsequent `cmake --build`, `cmake --install`, and `ctest` steps then fail with an error like `C:/TEMP/OCIO/build/Release is not a directory`, as reported in the issue.

## Fix
Explicitly create both directories with `mkdir` right after they are resolved, regardless of whether `--configure` was passed on that run.

## Test plan
- Ran `ocio.bat --help` before and after the change to confirm the script still parses correctly.
- Verified in isolation (`cmd /c "if NOT EXIST ... mkdir ..."`) that the two added lines create nested build/install directories that don't yet exist, and are a no-op when they already exist.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


carolalynn:
Hi @hikmetba-bit , thanks for the contribution. We'll need you to follow our contribution guidelines in order to keep moving on this - which include signing your PRs (DCO), signing our contribution license agreement (CLA), and following our guidelines for use of AI tool assistance in your work (please state Assisted by: Tool / Model in your PR description) thanks!

Issue #2340, opened 2026-09-07
Invalid YAML in test_cpu suite rejected by yaml-cpp 0.9.0


OPNA2608:
https://github.com/NixOS/nixpkgs/pull/560590#issuecomment-5564165433

When built against yaml-cpp version 0.9.0, `test_cpu_exec` fails:

```
[ 129/1169] [Config / inactive_color_space_precedence                    ] - PASSED
/build/source/tests/cpu/Config_tests.cpp:6502:
FAILED: exception thrown from config = OCIO::Config::CreateFromStream(is): "Error: Loading the OCIO profile failed. yaml-cpp: error at line 25, column 1: plain value cannot start with flow indicator character"
Segmentation fault         (core dumped) ./test_cpu_exec
```

https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/c52966a6677723d5bd2dbef0ccec3fed9cbc3790/tests/cpu/Config_tests.cpp#L6492-L6502

If you want to continue using funky YAML in this test, then the continued lines of the array with more than just whitespace content must be indented. Example fix (without CLA signature): #2339

Issue #2314, opened 2026-05-20
Should the 2023 aswf-docker container images include minizip-ng 4.0.10?


jfpanisset:
The 2023.3 aswf-docker container images have minizip-ng 3.0.10, where mz.h contains an invalid hex constant:

```
#define MZ_VERSION_BUILD                (03000a)
```

which breaks code in OCIOZArchive.cpp starting with OCIO 2.3.x:

```
#if MZ_VERSION_BUILD >= 040000
```

minizip-ng 4.0.10, included in the 2024 / 2025 / 2026 images fixes that:

```
#define MZ_VERSION_BUILD (0x04000A)
```

Some CI permutations in other projects want to build OCIO 2.3.x in the 2023.3 images. Would it be an issue for OCIO if 2023 images moved to minizip-ng 4.0.10?

doug-walker:
@jfpanisset , I don't think it would be an issue for OCIO if you raise the minizip-ng version in that container.

lgritz:
I'm seeing this issue in broken 2023 containers for OSL as well -- because OSL has to build OIIO, and OIIO has to build OCIO, and OCIO breaks with the minizip-ng in the container.

jfpanisset:
I was hoping that a simple hack like:

```
sed -i 's/03000a/0x03000A/' /usr/local/include/mz.h
```

would be enough to allow OCIO 2.3.2 to build against minizip-ng 3.0.10, but it seems like this code in OCIOZArchive.cpp:

```
#if MZ_VERSION_BUILD >= 040000
    write_mem_stream = mz_stream_mem_create();
#else
    mz_stream_mem_create(&write_mem_stream);
#endif
```

isn't really doing what it thinks it is doing, it would need to be something like:

```
#if MZ_VERSION_BUILD >= 0x040000
    write_mem_stream = mz_stream_mem_create();
#else
    mz_stream_mem_create(&write_mem_stream);
#endif
```

The following patch kinda makes this work, but of course it requires a patched version of mz.h 3.0.10 with a correct hex constant (the `sed` line above):

```
cat << 'EOF' | patch -p1
diff --git a/src/OpenColorIO/OCIOZArchive.cpp b/src/OpenColorIO/OCIOZArchive.cpp
index d99dd79ac..bfda2778a 100644
--- a/src/OpenColorIO/OCIOZArchive.cpp
+++ b/src/OpenColorIO/OCIOZArchive.cpp
[...]

Issue #2294, opened 2026-04-23
ocioarchive symbolic link handling broken (upstream bug/feature)


KevinJW:
We were trying to use OCIO archive on a tree that happened to contain symbolic links to another location. The archive runs with no errors indicated but neither `unzip` not `ocioarchive` can extract the files. It appears to be related to https://github.com/zlib-ng/minizip-ng/issues/797

In our case the symbolic link was to a LUT file, which when trying to extract resulted in an attempt to create a symbolic link pointing to the contents of the file... example contents from `zipinfo`

```
Central directory entry #24:
---------------------------

  There are an extra 16 bytes preceding this file.

  colour/look/aces_1.2/luts/Log2_48_nits_Shaper_to_linear.spi1d

  offset of local header from start of archive:   16072021
                                                  (0000000000F53D55h) bytes
  file system or operating system of origin:      Unix
  version of encoding software:                   4.5
  minimum file system compatibility required:     MS-DOS, OS/2 or NT FAT
  minimum software version required to extract:   2.0
  compression method:                             deflated
  compression sub-type (deflation):               maximum
  file security status:                           not encrypted
  extended local header:                          yes
  file last modified on (DOS date/time):          2023 Aug 1 08:30:30
  32-bit CRC value (hex):                         83a5291c
  compressed size:                                27242 bytes
  uncompressed size:                              73776 bytes
  length of filename:                             61 characters
  length of extra field:                          0 bytes
  length of file comment:                         0 characters
  disk number on which file begins:               disk 1
  apparent file type:                             binary
  Unix file attributes (120777 octal):            lrwxrwxrwx
  MS-DOS file attributes (00 hex):                none
```

So we see the file contents are stored compressed but the file attributes are a for a symbolic link.

KevinJW:
Upstream looks like it may have fixed it https://github.com/zlib-ng/minizip-ng/releases/tag/4.2.0 needs testing. If it works we should consider updating our minimum version check.

KevinJW:
Currently testing with a minimal patch to OCIO something like this:

```
diff --git a/share/cmake/modules/install/Installminizip-ng.cmake b/share/cmake/modules/install/Installminizip-ng.cmake
index 28b6e6a..3a5e64e 100644
--- a/share/cmake/modules/install/Installminizip-ng.cmake
+++ b/share/cmake/modules/install/Installminizip-ng.cmake
[...]

PR #2350, opened 2026-09-21
Adsk Contrib - Support Yaml 0.9.0 and other build fixes

doug-walker:
As requested by @lgritz, this PR adds support for Yaml 0.9.0 and makes it the recommended version, which gets installed when OCIO_INSTALL_EXT_PACKAGES is set to MISSING or ALL. Addresses issue #2340.

Support for Yaml 0.9.0 simplifies building OCIO with gcc-15.

In addition, it fixes a build problem on Windows where recent vcpkg installs of glew no longer seem to have glu. This only seems to be necessary for gluErrorString, and it required pulling in windows.h. So another potential fix would just be to not use gluErrorString, as is done on macOS. Any opinions on that?

In addition, it fixes a bug in how the version of minizip_ng was being detected, as pointed out by @jfpanisset in issue #2314.


lgritz (approved):

share/cmake/modules/FindExtPackages.cmake
remia: I don't know if this is really necessary, the version hex fix was introduced in 4.0.5, so maybe we should bump the min version to that (or just 4.0.10) to officially move past these issues?

remia (approved):

share/cmake/modules/FindExtPackages.cmake
KevinJW: I think it is reasonable to jump to 4.0.10 as both minimum and recommended version as a fairly minimal change at this point in the 2.6 release.

Personally I'd like to jump the recommended to a more recent version to include the symbolic link handling bug fixed in 4.2.0+(#2294) but that is probably not a an OCIO 2.6 change.

jfpanisset:
For context here's what I currently do in aswf-docker ci-ocio images. The version of yaml-cpp:

- 2023: 0.8.0
- 2024: 0.8.0
- 2025: 0.8.0
- 2026: 0.8.0
- 2027: 0.9.0

Since for now I'm building 2027 images with OCIO 2.5.2, that demonstrates that you can build / link at least 2.5.2 against 0.9.0. Since OCIO wants to be able to build its latest version against older VFX Platform images, I can update the 2023-2026 images to yaml-cpp 0.9.0 to make that easier.

As for minizip-ng, I'm already including minizip-ng 4.0.10 in 2023-2026 (and rev'ed up to 4.2.1 for 2027), and for OCIO 2.2.1 builds for vfx2023 I apply this patch to support building against minizip-ng 4:

https://github.com/AcademySoftwareFoundation/aswf-docker/blob/d14a1c2c94ca88fbbabb9c966f6e6fe371b6a4cb/scripts/vfx/build_ocio.sh#L46

doug-walker:
Following up on the TSC discussion today, I've made the following updates to the dependencies. I'm listing the minimum and recommended versions (the latter is what OCIO installs when requested). Arrows indicate where I've increased the version. And I list the current latest version of each package.

**Library external dependencies**

expat
- min: 2.6.0
- rec: 2.7.2 --> 2.8.4 (current 2.8.4, has CVE fixes)

yaml-cpp
- min: 0.8.0
- rec: 0.8.0 --> 0.9.0 (current 0.9.0)

pystring
- min: 1.1.3
- rec: 1.1.4 (current 1.2.0)

Imath
- min: 3.1.1
- rec: 3.2.1 --> 3.2.3 (current 3.2.3)

zlib
- min: 1.2.13
- rec: 1.3.1 --> 1.3.2 (current 1.3.2, has CVE fixes)

minizip-ng
- min: 4.0.0 --> 4.0.5
- rec: 4.0.10 --> 4.2.2 (current 4.2.2)

**Library vendored**

sampleicc (not maintained)

xxhash 0.8.1 (current 0.8.4)

In addition to the Actions runs, I've gotten the local build/tests to pass on macOS (arm64, x86_64), Windows, and Linux. 

The only actual library code change was to modify how exceptions are handled due to changes in the most recent expat. We have a lot of unit tests around the XML parsing, so hopefully this is solid.

cozdas (approved):
Looks great.

Regarding the glu issue, the current fix looks correct but considering that multiple new developers hit glew/glut/glu related build problems in the past, and glu being deprecated many years ago I'm ok with removing the glu entirely. We can either return generic error string like mac does or add utility function to convert error enums to string.

There's another glu function usage in oglapp.cpp (gluOrtho2D) but that one seems ok as in that case  glu is included via freeglut. Looks like it can be replaced with glOrtho() easily too if needed.

I'm ok with keeping the current glu fix as is and creating a ticket to remove glu in glsl.cpp in a later version.

doug-walker:
@remia, my latest commit makes a fix to get the Wheel workflow back online, please take a look.

remia:
That seem reasonable, thanks @doug-walker, maybe we should consider dropping manylinux2014 but that can be done somewhere else.

According to Claude:

> Who would lose wheels if you drop it: only Linux users with glibc older than 2.27. Your manylinux_2_28 wheels are also tagged manylinux_2_27, so Ubuntu 18.04 is already covered. What's left:
> 
> RHEL/CentOS 7 and its clones: studios still on CY2022 or older pipelines, and paid extended-support RHEL 7.
> Amazon Linux 2, which reached end of support on 30 June 2026.
> Very old distros such as Debian 9, Ubuntu 16.04 and SLES 12.
> 
> Those users would fall back to building from the sdist, which is painful for a C++ project, so say so in the release notes.

